Issue #3, opened 2020-02-18
Use the ILogger rather than Console.Writeline


jennyf19:
Why?

Microsoft.Identity.Web's code uses Console.Writeline, whereas ASP.NET Core provides a ILogger interface which can be injected (by DI), and offers great logging, including for apps that would be deployed in App services

What?

Inject ILogger in the constructor of the classes using Console.WriteLine
Use ILogger instead of Console.Writeline

Copies from https://github.com/Azure-Samples/active-directory-aspnetcore-webapp-openidconnect-v2/issues/289

jennyf19:
Included in [release]https://github.com/AzureAD/microsoft-identity-web/releases/tag/0.1.0-preview)

PR #36, opened 2020-02-25
implement ILogger<>

jennyf19:
https://github.com/AzureAD/microsoft-identity-web/issues/3

### DIFF
--- a/src/Microsoft.Identity.Web/Resource/JwtBearerMiddlewareDiagnostics.cs
+++ b/src/Microsoft.Identity.Web/Resource/JwtBearerMiddlewareDiagnostics.cs
@@ -45,8 +45,7 @@ namespace Microsoft.Identity.Web.Resource
         /// preserving the previous handlers (which are called)
         /// </summary>
         /// <param name="events">Events to subscribe to</param>
-        public JwtBearerEvents Subscribe(
-            JwtBearerEvents events)
+        public JwtBearerEvents Subscribe(JwtBearerEvents events)
         {
             if (events == null)
             {
--- a/src/Microsoft.Identity.Web/Resource/JwtBearerMiddlewareDiagnostics.cs
+++ b/src/Microsoft.Identity.Web/Resource/JwtBearerMiddlewareDiagnostics.cs
@@ -2,43 +2,51 @@
 // Licensed under the MIT License.
 
 using System;
-using System.Diagnostics;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Authentication.JwtBearer;
+using Microsoft.Extensions.Logging;
 
 namespace Microsoft.Identity.Web.Resource
 {
     /// <summary>
     /// Diagnostics for the JwtBearer middleware (used in Web APIs)
     /// </summary>
-    public class JwtBearerMiddlewareDiagnostics
+    public class JwtBearerMiddlewareDiagnostics : IJwtBearerMiddlewareDiagnostics
     {
+        private readonly ILogger _logger;
+
+        public JwtBearerMiddlewareDiagnostics(ILogger<JwtBearerMiddlewareDiagnostics> logger)
+        {
+            _logger = logger;
+        }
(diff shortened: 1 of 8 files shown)


src/Microsoft.Identity.Web/Resource/JwtBearerMiddlewareDiagnostics.cs
```diff
@@ -2,43 +2,51 @@
...
+        private Func<JwtBearerChallengeContext, Task> s_onChallenge;
 
         /// <summary>
         /// Subscribes to all the JwtBearer events, to help debugging, while
         /// preserving the previous handlers (which are called)
         /// </summary>
         /// <param name="events">Events to subscribe to</param>
-        public static JwtBearerEvents Subscribe(JwtBearerEvents events)
+        public JwtBearerEvents Subscribe(
+            JwtBearerEvents events)
```
jmprieur: you might want to remove the extra line?

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -135,8 +138,7 @@ public class TokenAcquisition : ITokenAcquisition
             }
             catch (MsalException ex)
             {
-                // brentsch - todo, write to a log
-                Debug.WriteLine(ex.Message);
+                _logger.LogInformation(ex.Message);
```
jmprieur: LogInformation? or LogDebug?

jennyf19: @jmprieur i'm not sure. See [#8 here](https://nblumhardt.com/2017/07/library-logging/) this is why i put the others in Debug, but this is an exception.

src/Microsoft.Identity.Web/Resource/JwtBearerMiddlewareDiagnostics.cs
jennyf19: yes. thanks.

jmprieur:
yay!

src/Microsoft.Identity.Web/Resource/JwtBearerMiddlewareDiagnostics.cs
```diff
@@ -2,43 +2,50 @@
...
+        public JwtBearerMiddlewareDiagnostics(ILogger<JwtBearerMiddlewareDiagnostics> logger)
+        {
+            _logger = logger;
+        }
+
         /// <summary>
         /// Invoked if exceptions are thrown during request processing. The exceptions will be re-thrown after this event unless suppressed.
         /// </summary>
-        private static Func<AuthenticationFailedContext, Task> s_onAuthenticationFailed;
+        private Func<AuthenticationFailedContext, Task> s_onAuthenticationFailed;
```
pmaytak: Proposal to not use 's_' prefix for statics. Also these should use proper casing.

src/Microsoft.Identity.Web/WebApiServiceCollectionExtensions.cs
```diff
@@ -209,7 +210,9 @@ public static class WebApiServiceCollectionExtensions
 
                 if (subscribeToJwtBearerMiddlewareDiagnosticsEvents)
                 {
-                    options.Events = JwtBearerMiddlewareDiagnostics.Subscribe(options.Events);
+                    var diags = builder.Services.BuildServiceProvider().GetRequiredService<IJwtBearerMiddlewareDiagnostics>();
```
pmaytak: I would spell the name out to make it more readable.

src/Microsoft.Identity.Web/Resource/JwtBearerMiddlewareDiagnostics.cs
jennyf19: that's something to bring up to the team. we have this as a rule in MSAL, so all the statics should shave this format.

PR #50, opened 2020-03-11
Added unit tests and updated dependencies

pmaytak:
Added unit tests for AadIssuerValidator, AccountExtensions, AuthorityHelpers, Base64Helpers, ClaimsPrincipalFactory, ClientInfo, Extensions, ServiceCollectionExtensions. Added NSubstitute Nuget to Web.Test project. Updated xUnit Nuget from 2.4.0 to 2.4.1.

Arrange-Act-Assert pattern is followed to the extent applicable (personal opinion, of course) trying to reduce code duplication while maintaining readability.

The general pattern for test naming is UnitOfWork_Scenario_ExpectedResult. UnitOfWork is a method. ExpectedResult can be whether an exception is thrown or certain return object is expected. Scenario is the behavior we're testing, like different parameter values, call order. At times I group multiple similar test cases into one test method under one scenario to prevent creating almost duplicate test methods (ex. test validating that exception is thrown if at least one of the parameters is null or empty).


tests/Microsoft.Identity.Web.Test/Resource/AadIssuerValidatorTests.cs
```diff
@@ -0,0 +1,340 @@
...
+                    {
+                        ValidIssuers = new[] {
+                            "https://host1/{tenantid}/v2.0",
+                            "https://host2/{tenantid}/v2.0"
+                        }
+                    }));
+            Assert.Equal(expectedErrorMessage, exception.Message);
+        }
+
+        // Regression test for https://github.com/Azure-Samples/active-directory-dotnet-native-aspnetcore-v2/issues/68
```
pmaytak: I left the tests below here from previous version. The main differences is that they pass B2C related info and call GetIssuerValidator instead of the constructor. I think these are duplicates and can be removed. Also I think AadIssuerValidator class itself needs to be refactored; so these tests might change then too.

tests/Microsoft.Identity.Web.Test/Microsoft.Identity.Web.Test.csproj
```diff
@@ -8,8 +8,12 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="16.0.1" />
-    <PackageReference Include="xunit" Version="2.4.0" />
-    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.0" />
+    <PackageReference Include="NSubstitute" Version="4.2.1" />
```
henrik-me: ><PackageReference Include="NSubstitute" Version="4.2.1" /> [](start = 4, length = 58)

Did you consider if we can make code changes instead? (I know we use it in MSAL .NET and it's serving the purpose well).

pmaytak: Yes. Keeping track of changes.

PR #67, opened 2020-03-30
Simplification of the token cache serialization providers

jmprieur:
Simplification of the token cache serialization providers, which don't need a reference to the options (the ClientId, which was extracted from these options, can be provided by the TokenCacheArguments, or in the call ClearAsync

### DIFF
--- a/src/Microsoft.Identity.Web/TokenCacheProviders/IMsalTokenCacheProvider .cs
+++ b/src/Microsoft.Identity.Web/TokenCacheProviders/IMsalTokenCacheProvider .cs
@@ -22,6 +22,6 @@ namespace Microsoft.Identity.Web.TokenCacheProviders
         /// Clear the cache
         /// </summary>
         /// <returns></returns>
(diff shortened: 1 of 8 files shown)


src/Microsoft.Identity.Web/TokenCacheProviders/IMsalTokenCacheProvider .cs
```diff
@@ -22,6 +22,6 @@ public interface IMsalTokenCacheProvider
         /// Clear the cache
         /// </summary>
         /// <returns></returns>
-        Task ClearAsync();
+        Task ClearAsync(string clientId);
```
pmaytak: Should we also update the comments or will there be a task to go over all of the comments in the project?

jmprieur:
@jennyf19 @pmaytak 
This would not be needed if we solve https://github.com/AzureAD/microsoft-identity-web/issues/235

Proposing to close this

PR #68, opened 2020-03-30
Enables developers to use the domain instead of the TenantId

jmprieur:
Enables developers to use the domain instead of the TenantId for the authority (in the case of AAD)

### DIFF
--- a/tests/Microsoft.Identity.Web.Test/AuthorityHelpersTests.cs
+++ b/tests/Microsoft.Identity.Web.Test/AuthorityHelpersTests.cs
@@ -125,6 +125,22 @@ namespace Microsoft.Identity.Web.Test
         }
 
         [Fact]
+        public void BuildAuthority_AadValidOptionsWithDomain_ReturnsValidAadAuthority()
+        {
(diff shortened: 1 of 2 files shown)


tests/Microsoft.Identity.Web.Test/AuthorityHelpersTests.cs
```diff
@@ -124,6 +124,22 @@ public void BuildAuthority_AadValidOptions_ReturnsValidAadAuthority()
...
+            MicrosoftIdentityOptions options = new MicrosoftIdentityOptions
+            {
+                Domain = TestConstants.Domain,
+                Instance = TestConstants.AadInstance
+            };
+            string expectedResult = $"{options.Instance}/{options.Domain}/v2.0";
+
+            string result = AuthorityHelpers.BuildAuthority(options);
+
+            Assert.NotNull(result);
```
pmaytak: I think this is not needed, since nullability will be checked in the `.Equal`. (I realize other test have the same pattern, so maybe those should be updated too then.)

PR #69, opened 2020-03-30
Proposal for the parameters of AddProtectedApi ... methods

jmprieur:
I propose to get rid of hybrid methods, that is provide only:
- configuration and section name
- or delegates to set the option (several delegates)

@jennyf19 : let's discuss this when/if you have time.

### DIFF
--- a/src/Microsoft.Identity.Web/WebApiServiceCollectionExtensions.cs
+++ b/src/Microsoft.Identity.Web/WebApiServiceCollectionExtensions.cs
@@ -22,112 +22,6 @@ namespace Microsoft.Identity.Web
     /// </summary>
     public static class WebApiServiceCollectionExtensions
     {
-        #region Compatibility
-        /// <summary>
-        /// Protects the Web API with Microsoft identity platform (formerly Azure AD v2.0)
-        /// This supposes that the configuration files have a section named configSectionName (typically "AzureAD")
-        /// </summary>
-        /// <param name="services">Service collection to which to add authentication</param>
-        /// <param name="configuration">Configuration</param>
-        /// <returns></returns>
-        public static IServiceCollection AddProtectedApiCallsWebApis(
-            this IServiceCollection services,
-            IConfiguration configuration,
-            string configSectionName = "AzureAd")
-        {
-            return AddProtectedWebApiCallsProtectedWebApi(services,
-                                                          configuration,
-                                                          configSectionName);
-        }
-        #endregion
-
-        /// <summary>
-        /// Protects the Web API with Microsoft identity platform (formerly Azure AD v2.0)
-        /// This method expects the configuration file will have a section named "AzureAd" with the necessary settings to initialize authentication options.
-        /// </summary>
-        /// <param name="services">Service collection to which to add this authentication scheme</param>
-        /// <param name="configuration">The Configuration object</param>
-        /// <param name="subscribeToJwtBearerMiddlewareDiagnosticsEvents">
-        /// Set to true if you want to debug, or just understand the JwtBearer events.
-        /// </param>
-        /// <returns></returns>
-        public static IServiceCollection AddProtectedWebApi(
-            this IServiceCollection services,
-            IConfiguration configuration,
-            string configSectionName = "AzureAd",
-            X509Certificate2 tokenDecryptionCertificate = null,
-            bool subscribeToJwtBearerMiddlewareDiagnosticsEvents = false)
-        {
-            services.AddAuthentication(JwtBearerDefaults.AuthenticationScheme)
-                .AddProtectedWebApi(configuration,
-                    configSectionName,
-                    JwtBearerDefaults.AuthenticationScheme,
-                    tokenDecryptionCertificate,
-                    subscribeToJwtBearerMiddlewareDiagnosticsEvents);
-
-            return services;
-        }
-
-        /// <summary>
-        /// Protects the Web API with Microsoft identity platform (formerly Azure AD v2.0)
-        /// This method expects the configuration file will have a section, named "AzureAd" as default, with the necessary settings to initialize authentication options.
-        /// </summary>
-        /// <param name="builder">AuthenticationBuilder to which to add this configuration</param>
-        /// <param name="configuration">The Configuration object</param>
-        /// <param name="configureOptions">An action to configure JwtBearerOptions</param>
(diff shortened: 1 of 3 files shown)


src/Microsoft.Identity.Web/WebApiServiceCollectionExtensions.cs
```diff
@@ -225,17 +253,37 @@ public static class WebApiServiceCollectionExtensions
...
+        /// This supposes that the configuration files have a section named configSectionName (typically "AzureAD")
+        /// </summary>
+        /// <param name="services">Service collection to which to add authentication</param>
+        /// <param name="configuration">Configuration</param>
+        /// <returns></returns>
+        public static IServiceCollection AddProtectedWebApiCallsProtectedWebApi(
+            this IServiceCollection services,
+            string jwtBearerScheme,
+            Action<ConfidentialClientApplicationOptions> msalAppOptionsMapper,
+            Action<MicrosoftIdentityOptions> microsoftIdentityOptionsMapper)
```
pmaytak: Should these two Action parameters be renamed? The user can pass any action, so I think `*Mapper` is not generic enough suffix. The parameter name for the ASP.NET `Configure` method below is `configureOptions`. We could rename these to `configureMsalAppOptions` and `configureMicrosoftIdentityOptions` for consistency.

src/Microsoft.Identity.Web/WebApiServiceCollectionExtensions.cs
```diff
@@ -225,17 +253,37 @@ public static class WebApiServiceCollectionExtensions
...
             string jwtBearerScheme = JwtBearerDefaults.AuthenticationScheme)
+        {
+            return services.AddProtectedWebApiCallsProtectedWebApi(
+                jwtBearerScheme,
+                options => configuration.Bind(configSectionName, options),
+                options => configuration.Bind(configSectionName, options));
+        }
+        /// <summary>
+        /// Protects the Web API with Microsoft identity platform (formerly Azure AD v2.0)
+        /// This supposes that the configuration files have a section named configSectionName (typically "AzureAD")
```
pmaytak: This comment is no longer applicable.

src/Microsoft.Identity.Web/WebApiServiceCollectionExtensions.cs
```diff
@@ -135,34 +133,64 @@ public static class WebApiServiceCollectionExtensions
...
 
             builder.Services.AddHttpContextAccessor();
             builder.Services.AddSingleton<IJwtBearerMiddlewareDiagnostics, JwtBearerMiddlewareDiagnostics>();
 
             // Change the authentication configuration to accommodate the Microsoft identity platform endpoint (v2.0).
             builder.AddJwtBearer(jwtBearerScheme, options =>
             {
-                var microsoftIdentityOptions = configuration.GetSection(configSectionName).Get<MicrosoftIdentityOptions>();
+                // TODO:
+                // Suspect. Why not get the IOption<MicrosoftIdentityOptions>?
```
pmaytak: Makes sense.

src/Microsoft.Identity.Web/WebApiServiceCollectionExtensions.cs
```diff
@@ -135,34 +133,64 @@ public static class WebApiServiceCollectionExtensions
...
+        public static AuthenticationBuilder AddProtectedWebApi(
+        this AuthenticationBuilder builder,
+        string jwtBearerScheme,
+        Action<JwtBearerOptions> jwtOptionsMapper,
+        Action<MicrosoftIdentityOptions> microsoftIdentityOptionsMapper,
+        X509Certificate2 tokenDecryptionCertificate = null,
+        bool subscribeToJwtBearerMiddlewareDiagnosticsEvents = false)
+        {
+            builder.Services.Configure(jwtBearerScheme, jwtOptionsMapper);
+            builder.Services.Configure<MicrosoftIdentityOptions>(microsoftIdentityOptionsMapper);
```
pmaytak: I think the generic type can be removed, since it can be inferred or we can add one in the line above for consistency and if we want to be explicit or not. (Same applied to similar methods below).

src/Microsoft.Identity.Web/WebApiServiceCollectionExtensions.cs
```diff
@@ -135,34 +133,64 @@ public static class WebApiServiceCollectionExtensions
...
+        /// <param name="tokenDecryptionCertificate">Token decryption certificate</param>
+        /// <param name="subscribeToJwtBearerMiddlewareDiagnosticsEvents">
+        /// Set to true if you want to debug, or just understand the JwtBearer events.
+        /// </param>
+        /// <returns></returns>
+        public static AuthenticationBuilder AddProtectedWebApi(
+        this AuthenticationBuilder builder,
+        string jwtBearerScheme,
+        Action<JwtBearerOptions> jwtOptionsMapper,
+        Action<MicrosoftIdentityOptions> microsoftIdentityOptionsMapper,
```
pmaytak: Possibly rename `Mapper` variables (see comments below).

src/Microsoft.Identity.Web/WebApiServiceCollectionExtensions.cs
```diff
@@ -82,6 +80,7 @@ public static class WebApiServiceCollectionExtensions
         /// Set to true if you want to debug, or just understand the JwtBearer events.
         /// </param>
         /// <returns></returns>
+        [Obsolete("rather use the override with the configSectionName, or the configure option actions")]
```
pmaytak: Update the message and the one below to be consistent with what we have in the WebApp extensions class.

jennyf19:
Looks good. I'll take a look at the related web app work. thanks.

---
In reply to: [383584388](https://github.com/AzureAD/microsoft-identity-web/pull/69#issuecomment-383584388) [](ancestors = 383584388)

src/Microsoft.Identity.Web/WebApiServiceCollectionExtensions.cs
jmprieur: Done!

src/Microsoft.Identity.Web/WebApiServiceCollectionExtensions.cs
jmprieur: I removed the obsolete methods.

PR #70, opened 2020-03-30
Added new unit tests, cleaned up older ones

pmaytak:
Added unit tests for
- AzureADB2COpenIDConnectEventHandlersTests
- CookiePolicyOptionsExtensionsTests

Cleaned up some tests by using InlineData and changing some local variables to instance.

### DIFF
--- a/tests/Microsoft.Identity.Web.Test/AzureADB2COpenIDConnectEventHandlersTests.cs
+++ b/tests/Microsoft.Identity.Web.Test/AzureADB2COpenIDConnectEventHandlersTests.cs
@@ -1,0 +1,108 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Microsoft.AspNetCore.Authentication;
+using Microsoft.AspNetCore.Authentication.OpenIdConnect;
+using Microsoft.AspNetCore.Http;
+using Microsoft.Identity.Web.Test.Common;
+using Microsoft.Identity.Web.Test.Common.TestHelpers;
+using Microsoft.IdentityModel.Protocols.OpenIdConnect;
+using NSubstitute;
+using Xunit;
+
+namespace Microsoft.Identity.Web.Test
+{
+    public class AzureADB2COpenIDConnectEventHandlersTests
+    {
+        private const string PathBase = "/PathBase";
+        private const string DefaultUserFlow = TestConstants.B2CSignUpSignInUserFlow;
+        private const string CustomUserFlow = TestConstants.B2CResetPasswordUserFlow;
+        private string _defaultIssuer = $"IssuerAddress/{DefaultUserFlow}/";
+        private string _customIssuer = $"IssuerAddress/{CustomUserFlow}/";
+        private AuthenticationScheme _authScheme;
+
+        public AzureADB2COpenIDConnectEventHandlersTests()
+        {
+            _authScheme = new AuthenticationScheme(OpenIdConnectDefaults.AuthenticationScheme, OpenIdConnectDefaults.AuthenticationScheme, typeof(OpenIdConnectHandler));
+        }
+
+        [Fact]
+        public async void OnRedirectToIdentityProvider_CustomUserFlow_UpdatesContext()
+        {
+            var options = new MicrosoftIdentityOptions() { SignUpSignInPolicyId = DefaultUserFlow };
+            var handler = new AzureADB2COpenIDConnectEventHandlers(OpenIdConnectDefaults.AuthenticationScheme, options);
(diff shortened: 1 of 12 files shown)


tests/Microsoft.Identity.Web.Test/AzureADB2COpenIDConnectEventHandlersTests.cs
```diff
@@ -0,0 +1,107 @@
...
+using Microsoft.IdentityModel.Protocols.OpenIdConnect;
+using NSubstitute;
+using Xunit;
+
+namespace Microsoft.Identity.Web.Test
+{
+    public class AzureADB2COpenIDConnectEventHandlersTests
+    {
+        private const string PathBase = "/PathBase";
+        private const string DefaultUserFlow = "DefaultUserFlow";
```
jennyf19: >DefaultUserFlow [](start = 48, length = 15)

could you use something like "b2c_1_susi", which is closer to what the value would be? makes the test easier to understand for someone who doesn't know b2c

tests/Microsoft.Identity.Web.Test/AzureADB2COpenIDConnectEventHandlersTests.cs
```diff
@@ -0,0 +1,107 @@
...
+        private const string NonDefaultUserFlow = "NonDefaultUserFlow";
```
jennyf19: maybe call this "CustomUserFlow", which would be the non-default, and the value similar to something above, at least which starts w/"b2c_1_*"

tests/Microsoft.Identity.Web.Test/AzureADB2COpenIDConnectEventHandlersTests.cs
```diff
@@ -0,0 +1,107 @@
...
+        private string _defaultIssuer = $"IssuerAddress/{DefaultUserFlow}/";
+        private string _nonDefaultIssuer = $"IssuerAddress/{NonDefaultUserFlow}/";
```
jennyf19: not sure what you mean by default and non-default issuer and how that's impacted by the user flow? the user flow is independent of the issuer. maybe we can sync later, i think confused. :)

tests/Microsoft.Identity.Web.Test/Base64UrlHelpersTests.cs
```diff
@@ -188,48 +88,18 @@ public void CreateString_NonUTF8Bytes_ReturnsInvalidString()
...
-            actualDecodedByteArray = Base64UrlHelpers.DecodeToBytes(stringToDecodeWithBase64Slash);
-
-            Assert.Equal(expectedDecodedByteArray, actualDecodedByteArray);
-
-            var emptyStringToDecode = string.Empty;
-            expectedDecodedByteArray = Encoding.UTF8.GetBytes(emptyStringToDecode);
+            var expectedDecodedByteArray = Encoding.UTF8.GetBytes(expectedDecodedString);
 
-            actualDecodedByteArray = Base64UrlHelpers.DecodeToBytes(emptyStringToDecode);
+            var actualDecodedByteArray = Base64UrlHelpers.DecodeToBytes(stringToDecode);
```
jennyf19: yay to removing code!

tests/Microsoft.Identity.Web.Test/AzureADB2COpenIDConnectEventHandlersTests.cs
pmaytak: If the user flow in the `RedirectContext` is different from the default user flow (which is susi), then `IssuerAddress` is updated to replace the default user flow with the one from context.

https://github.com/AzureAD/microsoft-identity-web/blob/41798cebfe28a75f2984beca8407018fc9e77b72/src/Microsoft.Identity.Web/AzureADB2COpenIDConnectEventHandlers.cs#L28-L53

tests/Microsoft.Identity.Web.Test/Base64UrlHelpersTests.cs
pmaytak: 👍

tests/Microsoft.Identity.Web.Test/AzureADB2COpenIDConnectEventHandlersTests.cs
pmaytak: Agree. While the code under test doesn't specifically look for these values, updated to make it easier to understand.

tests/Microsoft.Identity.Web.Test/AzureADB2COpenIDConnectEventHandlersTests.cs
pmaytak: Updated.

PR #75, opened 2020-04-02
clean up public api and...

jennyf19:
- move to internal some classes
- add internalsvisibleto to properties
- add msal key for internals
- fix description for nuget package


src/Microsoft.Identity.Web/Microsoft.Identity.Web.csproj
```diff
@@ -8,14 +8,20 @@
...
+    <Description>
+      This package enables ASP.NET Core Web apps and Web APIs to use the Microsoft identity platform (formerly Azure AD v2.0). 
+      This package is specifically used for web applications, which sign-in users, and protected web APIs, which optionally call downstream web APIs.
+    </Description>
     <Copyright>© Microsoft Corporation. All rights reserved.</Copyright>
     <PackageLicenseExpression>MIT</PackageLicenseExpression>
     <PackageProjectUrl>https://github.com/AzureAD/microsoft-identity-web</PackageProjectUrl>
     <RepositoryUrl>https://github.com/AzureAD/microsoft-identity-web</RepositoryUrl>
-    <PackageTags>Microsoft identity web .NET ASP.NET Core</PackageTags>
+    <PackageTags>Microsoft identity web .NET ASP.NET Core Web App Web API B2C</PackageTags>
```
pmaytak, line 24: Nit: Identity Web capitalized?

src/Microsoft.Identity.Web/Microsoft.Identity.Web.csproj
```diff
@@ -40,13 +46,18 @@
   
   <PropertyGroup>
     <TargetFramework>netcoreapp3.1</TargetFramework>
+    <SignAssembly>true</SignAssembly>
+    <AssemblyOriginatorKeyFile>../../build/MSAL.snk</AssemblyOriginatorKeyFile>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <!-- The MSAL.snk has both private and public keys -->
```
pmaytak, line 54: Should we check in only public though?

tests/Microsoft.Identity.Web.Test.Integration/Microsoft.Identity.Web.Test.Integration.csproj
```diff
@@ -4,6 +4,10 @@
     <TargetFramework>netcoreapp3.1</TargetFramework>
 
     <IsPackable>false</IsPackable>
+
+    <SignAssembly>true</SignAssembly>
```
pmaytak, line 8: Do we need to sign test assemblies?

tests/Microsoft.Identity.Web.Test/Microsoft.Identity.Web.Test.csproj
```diff
@@ -4,6 +4,10 @@
     <TargetFramework>netcoreapp3.1</TargetFramework>
 
     <IsPackable>false</IsPackable>
+
+    <SignAssembly>true</SignAssembly>
+
+    <AssemblyOriginatorKeyFile>../../build/MSAL.snk</AssemblyOriginatorKeyFile>
```
pmaytak, line 10: Same as above.

src/Microsoft.Identity.Web/Microsoft.Identity.Web.csproj
bgavrilMS: No, it's fine. Do what MSAL does. And you don't need this property either, delaySign is false by default I think.

Assembly signing is NOT a security mechanism. Our Release build will add a proper digital signature. The only reason we do this "sign assembly" is to make sure the DLL is `strongly named`. I believe that in the past .net folks wanted this to be a security measure as well, but it doesn't hold up. The strong naming is good though. So current guidance is to commit an snk file that contains both public and private keys. The goal is to have contributors that are able to create strongly-signing DLLs, otherwise they can't test their changes.

tests/B2CWebAppCallsWebApi/TodoListService/AuthorizationPolicies/ScopesRequirement.cs
```diff
@@ -28,17 +28,19 @@ public ScopesRequirement(params string[] acceptedScopes)
         protected override Task HandleRequirementAsync(AuthorizationHandlerContext context,
                                                         ScopesRequirement requirement)
         {
+            string scope = "http://schemas.microsoft.com/identity/claims/scope";
+            string scp = "scp";
             // If there are no scopes, do not process
-            if (!context.User.Claims.Any(x => x.Type == ClaimConstants.Scope)
-               && !context.User.Claims.Any(y => y.Type == ClaimConstants.Scp))
+            if (!context.User.Claims.Any(x => x.Type == scope)
```
bgavrilMS, line 34: Please use string.equals
Also consider adding static analysis that will prevent you from making this mistake in the future. I recommend you copy the static analysis rules from asp.net core.

tests/Microsoft.Identity.Web.Test.Integration/Microsoft.Identity.Web.Test.Integration.csproj
bgavrilMS: As far as I know you need to sign all assemblies (or I should call it "strongly name"). AFAIK you cannot mix strongly-named and non-strongly-named assemblies. I may be wrong, there might be some subtleties around this. In any case, this is what is done in MSAL.

tests/B2CWebAppCallsWebApi/TodoListService/AuthorizationPolicies/ScopesRequirement.cs
pmaytak: As far as I remember equality operator for strings calls string.Equals with OrdinalIgnoreCase behind the scenes. Is your suggestion just for convention?

Which static analyzers? Is it the FxCop package or something else?

src/Microsoft.Identity.Web/Microsoft.Identity.Web.csproj
pmaytak: Makes sense. Thanks. 👍

tests/Microsoft.Identity.Web.Test.Integration/Microsoft.Identity.Web.Test.Integration.csproj
pmaytak: Makes sense. Thanks. 👍

src/Microsoft.Identity.Web/Microsoft.Identity.Web.csproj
jennyf19: fixed! nice catch

---
In reply to: [402046519](https://github.com/AzureAD/microsoft-identity-web/pull/75#discussion_r402046519) [](ancestors = 402046519)

src/Microsoft.Identity.Web/Microsoft.Identity.Web.csproj
jennyf19: thanks bogdan. :)

---
In reply to: [402375802](https://github.com/AzureAD/microsoft-identity-web/pull/75#discussion_r402375802) [](ancestors = 402375802)

tests/B2CWebAppCallsWebApi/TodoListService/AuthorizationPolicies/ScopesRequirement.cs
jennyf19: yes. and added an issue for this: [https://github.com/AzureAD/microsoft-identity-web/issues/76](https://github.com/AzureAD/microsoft-identity-web/issues/76)

---
In reply to: [402376930](https://github.com/AzureAD/microsoft-identity-web/pull/75#discussion_r402376930) [](ancestors = 402376930)

Issue #77, opened 2020-04-03
[Bug] ITokenAcquisition.GetAcquireTokenForApp should be named GetAccessTokenForAppAsync


jmprieur:
[Bug] ITokenAcquisition.GetAcquireTokenForApp should be named GetAccessTokenForAppAsync

jennyf19:
Included in [release]https://github.com/AzureAD/microsoft-identity-web/releases/tag/0.1.0-preview)

PR #79, opened 2020-04-06
move some ITokenAcquisition methods to internal interface

jennyf19:
rename to GetATForAppAsync

#77

### DIFF
--- a/src/Microsoft.Identity.Web/ITokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web/ITokenAcquisition.cs
@@ -16,49 +16,6 @@ namespace Microsoft.Identity.Web
     public interface ITokenAcquisition
     {
         /// <summary>
-        /// In a Web App, adds, to the MSAL.NET cache, the account of the user authenticating to the Web App, when the authorization code is received (after the user
-        /// signed-in and consented)
-        /// An On-behalf-of token contained in the <see cref="AuthorizationCodeReceivedContext"/> is added to the cache, so that it can then be used to acquire another token on-behalf-of the 
-        /// same user in order to call to downstream APIs.
-        /// </summary>
-        /// <param name="context">The context used when an 'AuthorizationCode' is received over the OpenIdConnect protocol.</param>
-        /// <param name="scopes">Scopes to request</param>
-        /// <example>
-        /// From the configuration of the Authentication of the ASP.NET Core Web API: 
-        /// <code>OpenIdConnectOptions options;</code>
-        /// 
-        /// Subscribe to the authorization code received event:
-        /// <code>
-        ///  options.Events = new OpenIdConnectEvents();
-        ///  options.Events.OnAuthorizationCodeReceived = OnAuthorizationCodeReceived;
-        /// }
-        /// </code>
-        /// 
-        /// And then in the OnAuthorizationCodeRecieved method, call <see cref="AddAccountToCacheFromAuthorizationCode"/>:
-        /// <code>
-        /// private async Task OnAuthorizationCodeReceived(AuthorizationCodeReceivedContext context)
-        /// {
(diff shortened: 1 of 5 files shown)


jmprieur:
methods of an interface cannot be internal ? can they?

pmaytak:
Yes. They do need to be explicitly implemented in the implementing class. For example: 
```CSharp
internal Task RemoveAccountAsync(RedirectContext context);
```
and 
```CSharp
async Task ITokenAcquisition.RemoveAccountAsync(RedirectContext context)...
```

In this case we just have to update `TokenAcquisition` and `ITokenAcquisition`. If we want to expose a method, we just remove the internal modifier.

Also, as it is currently, I think this line might look funky. We are asking to resolve ITokenAcquisition, but then it's actually ITokenAcquisitionInternal...
```CSharp
var tokenAcquisition = context.HttpContext.RequestServices.GetRequiredService<ITokenAcquisition>() as ITokenAcquisitionInternal;
```

Now, I am not sure from a software design perspective, which of the strategies is more adaptable long-term. I do prefer the one interface way because it seems cleaner and easier.

bgavrilMS:
I think the decision is about choosing the solution with least complexity so that more junior developers understand how this works. In MSAL we used both approaches but generally I think having 2 interfaces, one internal and one public, is somewhat easier to comprehend. So think about: 

P0: public API should be as simple as possible
P1: complexity is getting hold of internal API and working with it
P2: complexity in setting all of this up (since you only do it once)


Btw, can't you do `context.HttpContext.RequestServices.GetRequiredService<ITokenAcquisitionInternal>();` directly?

jmprieur:
Thanks @bgavrilMS : we need to check, but I doubt it as the DI services in .NET Core associate one interface with one implementation (not 2 interfaces)

pmaytak:
- Yea I think its 1:1 mapping for the dependencies.
- Hmm, well to me having one interface is simpler to understand and maintain. All the token acquisition methods are in one logical place. The same public methods are exposed with both strategies. 
  - If down the road we decide to expose additional methods to public:
    - With one interface, we (1) change the method's access modifer on the interface; (2) optionally change the implemented method's access modifer to public (if we want it to be visible from the concrete implementation rather than only the interface.)
    - With two interfaces, we (1) we move the method into public interface; (2) update all the code where we explicitly cast to the internal interface to use the public interface.
- @bgavrilMS, in your experience using both approaches in MSAL, does one have clear advantages over the other?
- After writing this out, it seems one interface approach is more adaptable to change. However if everyone sees this as just a personal style preference, then I'm fine with either choice. I just want to make sure we use it throughout and make our code base consistent, and to your point, Bogdan, easier to understand.

@jennyf19, thoughts, experiences?

jmprieur:
@pmaytak : I still don't understand how you would do with one interface. I don't want consumers of ITokenAcquisition to be able to redeem a code.

bgavrilMS:
@jmprieur - I think Peter is talking about "explicit interfaces". This is discussed in more detail for example [here](https://alexfranchuk.com/blog/internal-interface-classes-in-csharp/). 

@pmaytak - all of those are excellent points, especially making setters public. Just note that explicit interfaces are a somewhat obscure feature of C# that has no equivalent in Java etc. You'll also not be able to use the object to call your methods, you need to always cast to your interface (I guess that's alright if you're doing DI properly). 

I don't have strong feelings either way, if I were to choose I would go with the explicit interface just because you need to keep 1 object in your head instead of 2 when debugging etc.

pmaytak:
@jmprieur Keep `ITokenAcquisition` interface public, mark certain methods internal. Explicitly implement this interface in `TokenAcquisition` class. However, we also must add default implementation to these internal interface methods (C# 8 feature). (Without this, when an external library tries to implement this interface, the compiler will complain that the external class doesn't implement the internal interface method, which it can't.)
> You'll also not be able to use the object to call your methods, you need to always cast to your interface

@bgavrilMS Yea, I think that adds a bit to the complexity.

Issue #78, opened 2020-04-05
[Bug] Missing method: AddProtectedWebApi


SaqibS:
**Which Version of Microsoft Identity Web are you using ?**
Note that to get help, you need to run the latest version. 
Latest from master branch

**Where is the issue?**
Web API

**Is this a new or existing app?**
<!-- Ex:
This is a new app or experiment

**Repro**

```csharp
  public void ConfigureServices(IServiceCollection services)
  {
   services.AddProtectedWebApi(Configuration);
  }
```

**Expected behavior**
The `AddProtectedWebApi` method should be there.

**Actual behavior**
There is a method called `AddProtectedWebApi` on `AuthenticationBuilder`, which suggests that this may need to go in `Configure`', but that returns `IApplicationBuilder` not `ApplicationBuilder`.

**Possible Solution**
Updated the readme.

**Additional context/ Logs / Screenshots**
None.

SaqibS:
Very strange - I filled in the bug template but not all content shows up here.

Basically, in `ConfigureServices`, I can't find the `AddProtectedWebApi` method in latest master. There is a method with that name on `AuthenticationBuilder`, but how to create one of those?

jmprieur:
Hello @SaqibS : you need to call .AddAuthentication first.
Would you want to AddProtectedWebApi on the services collection?
 
 ...
  public void ConfigureServices(IServiceCollection services)
  {
   ...
   services.AddAuthentication(AzureADDefaults.AuthenticationScheme)
           .AddProtectedWebApi(Configuration);
   ...
  }
  ...

jmprieur:
BTW, we discsussed this bug and we've decided to bring back the `services.AddProtectedWebApi()`.  
See https://github.com/AzureAD/microsoft-identity-web/pull/81

jennyf19:
Included in [release]https://github.com/AzureAD/microsoft-identity-web/releases/tag/0.1.0-preview)

PR #81, opened 2020-04-06
split the web api/web app service collection extention methods and cr…

jennyf19:
…eated ones for the builder in a separate class to simplify the api

update unit tests

fix for #78

### DIFF
--- a/src/Microsoft.Identity.Web/WebApiServiceCollectionExtensions.cs
+++ b/src/Microsoft.Identity.Web/WebApiServiceCollectionExtensions.cs
@@ -6,12 +6,8 @@ using Microsoft.AspNetCore.Authentication.JwtBearer;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Identity.Client;
-using Microsoft.Identity.Web.Resource;
-using Microsoft.IdentityModel.Tokens;
 using System;
-using System.Collections.Generic;
 using System.IdentityModel.Tokens.Jwt;
-using System.Linq;
 using System.Security.Cryptography.X509Certificates;
 using System.Threading.Tasks;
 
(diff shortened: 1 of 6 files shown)


jennyf19:
New Api diagrams:
image

image

image

image

src/Microsoft.Identity.Web/WebApiServiceCollectionExtensions.cs
```diff
@@ -36,111 +32,21 @@ public static class WebApiServiceCollectionExtensions
...
         {
+            AuthenticationBuilder builder = services.AddAuthentication(jwtBearerScheme);
             return builder.AddProtectedWebApi(options => configuration.Bind(configSectionName, options),
                 options => configuration.Bind(configSectionName, options),
                 tokenDecryptionCertificate,
                 jwtBearerScheme,
                 subscribeToJwtBearerMiddlewareDiagnosticsEvents);
         }
 
-        /// <summary>
```
jmprieur, line 53: (Moved to WebApiAuthenticationBuidlerExtensions.cs)

jmprieur:
@pmaytak : actually, Jenny asked me and I though we'd only add them if customers ask for them

jennyf19:
Thanks for the feedback...I'll merge this and we can discuss the above. :)

pmaytak:
My reasoning:
- Overloads with action parameters give more flexibility and power to the end user. (That's why we added them, as I understand.)
- Methods extending `ServiceCollection` simplify and allow the user to not call `AddAuthentication` methods every time.
- Other extension methods already have the two types of overloads. Adding these two will contribute to the consistency of the API.
- We can simplify _Startup_ classes in our samples to use the `ServiceCollection` extensions, like so:
```CSharp
...
// Adds Microsoft Identity platform (AAD v2.0) support to protect this Api
services.AddProtectedWebApi(Configuration, "AzureAd");
...
```
- Lastly, I think we don't even have to call `AddAuthentication(JwtBearerDefaults.AuthenticationScheme)` and `AddAuthentication(OpenIdConnectDefaults.AuthenticationScheme)` methods at all because in the configuration methods we eventually call `builder.AddJwtBearer(jwtBearerScheme, options => ...);` and `builder.AddOpenIdConnect(openIdConnectScheme, options => ... );`, which set the same authentication scheme anyway. I am not a 100% sure on this; it's just a hunch based on the Core source code I saw. So we would need to create a separate issue for this.

jmprieur:
Thanks @pmaytak 
- For the last bullet point it's worth asking George. Do you want to do it?
- For the bullet point before last, this is already the case?  they are simple already for the simple scenario.

I prefer that we work for now on making sure the public API has the shape we want. We can add the other methods later.

pmaytak:
- I know where the code for that is, but I'll verify with George. I created #83 so I won't lose track of this.
- Below is how it is right now. I agree that it's simple already, but there's always room for improvement.
```CSharp
// Adds Microsoft Identity platform (AAD v2.0) support to protect this Api
       .AddProtectedWebApi(Configuration, "AzureAd");
```
- That's fine, as long as we had a discussion on this.

Issue #84, opened 2020-04-09
[Bug] Add overrides with delegates of AddSignIn, and AddProtectedWebAPI taking Service collection (in addition to AuthenticationBuilder)


jmprieur:
**Where is the issue?**
* Web App
    * [ x] Sign-in users
    * [ ] Sign-in users and call web APIs
* Web API
    * [ x] Protected web APIs (Validating tokens)
    * [ ] Protected web APIs (Validating scopes)
    * [ ] Protected web APIs call downstream web APIs
* Token cache serialization
     * [ ] In Memory caches
     * [ ] Session caches
     * [ ] Distributed caches

**Repro**

Today, we need to write
```csharp
services.AddAuthentication(OpenIdConnectAuthentication.DefaultScheme)
             .AddSignIn(options => ..., options => ....);
```
but we cannot do directly :
services..AddSignIn(options => ..., options => ....);

whereas the overrides with a configuration section exist.

**Expected behavior**
Have the overrides with delegates for both AddSignIn and AddProtectedWebAPI, from a service collection in addition to from an AuthenticationBuilder

jennyf19:
Included in [release]https://github.com/AzureAD/microsoft-identity-web/releases/tag/0.1.0-preview)

PR #85, opened 2020-04-09
fix xml comments

jennyf19:
add overrides with delegates
#84 

@pmaytak after updating samples, decided to address the suggestion you had earlier this week and just do it now. makes the api much simpler.

### DIFF
--- a/src/Microsoft.Identity.Web/WebAppServiceCollectionExtensions.cs
+++ b/src/Microsoft.Identity.Web/WebAppServiceCollectionExtensions.cs
@@ -22,16 +22,23 @@ namespace Microsoft.Identity.Web
     {
         /// <summary>
         /// Add authentication with Microsoft identity platform.
-        /// This method expects the configuration file will have a section, named "AzureAd" as default, with the necessary settings to initialize authentication options.
+        /// This method expects the configuration file will have a section, (by default named "AzureAd"), with the necessary settings to 
+        /// initialize the authentication options.
         /// </summary>
         /// <param name="services">Service collection to which to add authentication</param>
         /// <param name="configuration">The IConfiguration object</param>
-        /// <param name="configSectionName">The configuration section with the necessary settings to initialize authentication options</param>
+        /// <param name="configSectionName">The name of the configuration section with the necessary
+        /// settings to initialize authentication options</param>
+        /// <param name="openIdConnectScheme">Optional name for the open id connect authentication scheme 
+        /// (by default OpenIdConnectDefaults.AuthenticationScheme). This can be specified when you want to support
+        /// several OpenIdConnect identity providers.</param>
+        /// <param name="cookieScheme">Optional name for the acookie uthentication scheme 
+        /// (by default OpenIdConnectDefaults.AuthenticationScheme) </param>
         /// <param name="subscribeToOpenIdConnectMiddlewareDiagnosticsEvents">
         /// Set to true if you want to debug, or just understand the OpenIdConnect events.
         /// </param>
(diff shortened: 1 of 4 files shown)


src/Microsoft.Identity.Web/WebAppServiceCollectionExtensions.cs
```diff
@@ -40,20 +47,60 @@ public static class WebAppServiceCollectionExtensions
...
+            bool subscribeToOpenIdConnectMiddlewareDiagnosticsEvents = false)
+        {
+            AuthenticationBuilder builder = services.AddAuthentication(openIdConnectScheme);
+            builder.AddSignIn(
+                 configureOpenIdConnectOptions,
+                 configureMicrosoftIdentityOptions,
+                 openIdConnectScheme,
+                 cookieScheme,
+                 subscribeToOpenIdConnectMiddlewareDiagnosticsEvents);
+            return services;
```
pmaytak, line 90: I think we need to return `builder` object here, which actually has the `Services` property in itself. From what I've seen in the ASP.NET Core `IServiceCollection` extensions classes, if the extension method only adds some services to the collection, then it's fine to return `services`, but if there's more setup using a builder, then builder is returned. Also `AuthenticaitonBuilder` has some specific auth related methods that can be chained. For example:
```CSharp
services.AddAuthentication(X).AddSignIn(Y).AddCookie(Z).Services.AddLogging();
```

jennyf19: like the idea. let's dicuss

jmprieur: I get your point, @pmaytak. Thanks for this feedback.

Today, by returning services, we enable (simple) code like:

```csharp
services.AddSignIn(Configuration)
             .AddProtectedWebApiCallsWebAPi(Configuration)
             .AddTokenCacheXXX();
```

If we returned a builder, the code would have to be

```csharp
services.AddSignIn(Configuration);
services.AddProtectedWebApiCallsWebAPi(Configuration)
             .AddTokenCacheXXX();
```

which is ok, but I like a bit less, and given we are already taking care of cookies ...

Finally, it's always possible for customers who want to get a builder to use the following override:

```csharp
services.AddAuthentication(OpenIdConnectAuthentication.Default)
             .AddSignIn(Configuration)
             .AddCookie(Y);

services.AddProtectedWebApiCallsWebAPi(Configuration)
             .AddTokenCacheXXX();
```

pmaytak: If we returned a builder, your second code example could be this:
```CSharp
services.AddSignIn(Configuration);
        .Services
        .AddProtectedWebApiCallsWebAPi(Configuration)
        .AddTokenCacheXXX();
```
Similar to code in [ServiceCollectionTests.cs](https://github.com/dotnet/aspnetcore/blob/43c2c9ea97dfacce87f6bd578c1863ee0bf677c4/src/DataProtection/DataProtection/test/ServiceCollectionTests.cs#L15)

I agree with your point of making the API simple, and both of the ways are pretty close in that regard. The builder does provide more authentication setup functionality while still providing access to the `Services` property for chaining. I guess my perspective is a bit more conceptual - do we return a more specific `AuthenticationBuilder` or a more general `IServicesCollection`.

jmprieur: Proposing to do a user study on that
(ask users what they prefer)?

PR #105, opened 2020-04-16
validate v2.0 authority with web app, like we do w/web api.

### LINKED ISSUE #19: AddSignIn should be more robust to authority having or not a trailing slash.
`WebApiAuthenticationBuilderExtensions.EnsureAuthorityIsV2_0` was extracted from the ` WebApiAuthenticationBuilderExtensions.AddProtectedWebAPi` method so that we can unit test it. I think it should stay there as this method takes a `JwtBearerOptions` parameter and It's unit tested. 

It provides a more robust processing of authorities accepting them to end in / or not. 

We want to do somthing similar for `WebAppAuthenticationBuilderExtensions.AddSignIn(
            this AuthenticationBuilder builder,
            Action<OpenIdConnectOptions> configureOpenIdConnectOptions,
            Action<MicrosoftIdentityOptions> configureMicrosoftIdentityOptions,
            string openIdConnectScheme = OpenIdConnectDefaults.AuthenticationScheme,
            string cookieScheme = CookieAuthenticationDefaults.AuthenticationScheme,
            bool subscribeToOpenIdConnectMiddlewareDiagnosticsEvents = false)`, but with a method enabling an `OpenIdConnectOptions`.

jennyf19:
Fixes #19

### DIFF
--- a/src/Microsoft.Identity.Web/WebAppAuthenticationBuilderExtensions.cs
+++ b/src/Microsoft.Identity.Web/WebAppAuthenticationBuilderExtensions.cs
@@ -82,8 +82,8 @@ namespace Microsoft.Identity.Web
                 if (string.IsNullOrWhiteSpace(options.Authority))
                     options.Authority = AuthorityHelpers.BuildAuthority(microsoftIdentityOptions);
 
-                if (!AuthorityHelpers.IsV2Authority(options.Authority))
-                    options.Authority += "/v2.0";
+                // This is a Microsoft identity platform Web APP
+                EnsureAuthorityIsV2_0(options);
 
                 // B2C doesn't have preferred_username claims
                 if (microsoftIdentityOptions.IsB2C)
@@ -162,5 +162,18 @@ namespace Microsoft.Identity.Web
             return builder;
         }
 
+        /// <summary>
+        /// Ensures that the authority is a v2.0 authority
+        /// </summary>
+        /// <param name="options">OpenIdConnect options read from the config file
+        /// or set by the developer, for which we want to ensure the authority
+        /// is a v2.0 authority</param>
+        internal static void EnsureAuthorityIsV2_0(OpenIdConnectOptions options)
+        {
+            var authority = options.Authority.Trim().TrimEnd('/');
+            if (!authority.EndsWith("v2.0"))
+                authority += "/v2.0";
+            options.Authority = authority;
+        }
     }
 }
--- a/tests/Microsoft.Identity.Web.Test/WebAppExtentionsTests.cs
+++ b/tests/Microsoft.Identity.Web.Test/WebAppExtentionsTests.cs
@@ -1,0 +1,32 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Microsoft.AspNetCore.Authentication.OpenIdConnect;
+using Microsoft.Identity.Web.Test.Common;
+using Xunit;
+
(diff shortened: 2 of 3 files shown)


tests/Microsoft.Identity.Web.Test/WebAppExtentionsTests.cs
```diff
@@ -0,0 +1,32 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Microsoft.AspNetCore.Authentication.OpenIdConnect;
+using Microsoft.Identity.Web.Test.Common;
+using Xunit;
+
+namespace Microsoft.Identity.Web.Test
+{
+    public class WebAppExtentionsTests
```
pmaytak, line 10: typo: > `WebAppExtensionsTests` (+ filename).

src/Microsoft.Identity.Web/WebAppAuthenticationBuilderExtensions.cs
```diff
@@ -82,8 +82,8 @@ public static class WebAppAuthenticationBuilderExtensions
                 if (string.IsNullOrWhiteSpace(options.Authority))
                     options.Authority = AuthorityHelpers.BuildAuthority(microsoftIdentityOptions);
 
-                if (!AuthorityHelpers.IsV2Authority(options.Authority))
-                    options.Authority += "/v2.0";
+                // This is a Microsoft identity platform Web APP
```
pmaytak, line 85: typo: > `app`

src/Microsoft.Identity.Web/WebAppAuthenticationBuilderExtensions.cs
```diff
@@ -162,5 +162,18 @@ public static class WebAppAuthenticationBuilderExtensions
             return builder;
         }
 
+        /// <summary>
+        /// Ensures that the authority is a v2.0 authority
+        /// </summary>
+        /// <param name="options">OpenIdConnect options read from the config file
+        /// or set by the developer, for which we want to ensure the authority
+        /// is a v2.0 authority</param>
+        internal static void EnsureAuthorityIsV2_0(OpenIdConnectOptions options)
```
pmaytak, line 171: Maybe the main part of this method and the one in `WebApiAuthenticationBuilderExtensions` can be extracted into `AuthorityHelpers` class.
Something like so:
```csharp
internal static bool EnsureAuthorityIsV2_0(string authority)
{
    authority = authority.Trim().TrimEnd('/');
    if (!authority.EndsWith("v2.0"))
        authority += "/v2.0";
    return authority;
}
```

Then just called like so from this method:
```csharp
...
options.Authority = AuthorityHelpers.EnsureAuthorityIsV2_0(options.Authority);
...
```

pmaytak, line 171: We should also decide on how the naming convention for the version in the name. AuthorityHelpers.Is**V2**Authority is one way, EnsureAuthorityIs**V2_0** is another.

jmprieur: with EnsureAuthorityIsV2_0 returning a string, I guess? not a bool?

pmaytak: Good catch, @jmprieur 👍. I am not perfect. @jennyf19 got it right, though. 😃

Issue #100, opened 2020-04-13
[Feature Request] Make the ClaimConstants class public again


o1oo11oo:
**Is your feature request related to a problem? Please describe.**
As part of a web API, we're using integration tests with a custom authentication handler like the one suggested [on the Microsoft Docs](https://docs.microsoft.com/en-us/aspnet/core/test/integration-tests#mock-authentication). To build a similar ClaimsPrincipal and have the extension functions like `User.GetObjectId()` working, I'm adding a bunch of claims from ClaimConstants to the principal, namely `Name`, `ObjectId` and `PreferredUserName`. This stopped working after #75 was merged, because the ClaimConstants class was set to internal.

**Describe the solution you'd like**
To me it seems like the easiest solution is to set the ClaimConstants class to public again, or provide a limited subset of claims, that can be public, if not everything should be publically available.

**Describe alternatives you've considered**
I could copy the constants I need from ClaimConstants or just define those as strings where I need them, but that might break in case those claims get changed at some point. Having them as constants that can be used prevents that.

jennyf19:
Thanks for the feedback @o1oo11oo 
cc: @jmprieur

jmprieur:
@jennyf19 @pmaytak : let's prioritize this to unblock the customers.

jennyf19:
@jmprieur sounds good. will get this in the next release.

jennyf19:
@o1oo11oo @chris5287 included in [0.1.1-preview release](https://github.com/AzureAD/microsoft-identity-web/releases/tag/0.1.1-preview

PR #106, opened 2020-04-16
make claimConstant class public

jennyf19:
Issue #100


src/Microsoft.Identity.Web/ClaimConstants.cs
```diff
@@ -6,7 +6,7 @@ namespace Microsoft.Identity.Web
     /// <summary>
     /// Constants for claim types.
     /// </summary>
-    internal static class ClaimConstants
+    public static class ClaimConstants
```
pmaytak, line 9: Is there a class in the .NET/Core/ASP.NET that we can just use for these constants? Wondering since if we expose this, we allow users to take dependency on this and maybe we can reuse something existing. But if it's a small enough class that will stay constant, maybe it's fine.

PR #109, opened 2020-04-17
Added Roslyn analyzers to the projects

pmaytak:
- Added Roslyn analyzers to all the projects. (If they turn out to be not useful for test projects, we can remove later).
- Added editorconfig files at whole solution and tests levels.
- Removed old legacy ruleset file. (The rules were already in Roslyn by default, some are deprecated.)


pmaytak:
@jmprieur Yes, there are actually more of them. I was thinking about fixing them in the next PRs and keep this one just for the Nuget package addition.

jmprieur:
Sounds good, @pmaytak !
Thanks for the update!

PR #114, opened 2020-04-20
- Updating the Web API template and adding the Web app template

jmprieur:
- Updating the Web API template
- Adding a Web App template


ProjectTemplates/templates/StarterWeb-CSharp/appsettings.json
```diff
@@ -0,0 +1,42 @@
+{
+////#if (IndividualB2CAuth)
```
jennyf19, line 2: can this be deleted?

ProjectTemplates/templates/StarterWeb-CSharp/Program.cs
```diff
@@ -0,0 +1,26 @@
+using System;
```
jennyf19, line 1: do we need license info on these files?

jennyf19:
I didn't have time to try it out, but looks fine. thx for adding this. we'll need to update the wiki page too.

ProjectTemplates/templates/StarterWeb-CSharp/appsettings.json
jmprieur: No we should not remove them, and they won't appear in the generated project. 
Those are special comments which are about conditionnally generating code based on the parameters. 

What happens is that when the user writes:
`dotnet new mvc -auth IndividualB2C`, this defines the `IndividualB2CAuth` variable, and then the dotnet new template generation engine will generate the files (here the appsettings.json) conditionnally based on these variables. This section will appear in the appsettings.json if the user wants a B2C application

This is pretty awesome technology, BTW. See https://github.com/dotnet/templating/wiki for details

ProjectTemplates/templates/StarterWeb-CSharp/Program.cs
jmprieur: That's a good question
however these are the projects that customers will create, and they won't necessarilly want to have open source code with the MIT license :)
so, no!

jmprieur:
@pmaytak : I'm going to merge this PR. I'll address your PR comments, if any, in a next batch: I'd like to ahve these templates available to be able to update the docs.ms ASAP.

jmprieur:
@jennyf19 @pmaytak 
- The following PR adds tests to unfold and build the templates: https://github.com/AzureAD/microsoft-identity-web/pull/116/
- Thanks @jennyf19 for the remainder about the wiki. This is updated: https://github.com/AzureAD/microsoft-identity-web/wiki/web-app-template. Please review!

PR #116, opened 2020-04-21
Jmprieur/project template tests

jmprieur:
Adding a .bat file that tests the project templates:
- Web app (SingleOrg, MultiOrg and B2C)
- Web api (SingleOrg, B2C)

Fixes the templates based on issues discoverered while running these tests

### DIFF
--- a/ProjectTemplates/templates/WebApi-CSharp/Startup.cs
+++ b/ProjectTemplates/templates/WebApi-CSharp/Startup.cs
@@ -48,7 +48,7 @@ namespace Company.WebApplication1
             //         .AddInMemoryTokenCaches();
 #elif (IndividualB2CAuth)
             services.AddAuthentication(JwtBearerDefaults.AuthenticationScheme)
-                    .AddProtectedWebApi("AzureAdB2C", Configuration, options => Configuration.Bind("AzureAdB2C", options));
+                    .AddProtectedWebApi(Configuration, "AzureAdB2C");
             // Uncomment the following lines if you want your Web API to call a downstream API
             // services.AddProtectedWebApiCallsProtectedWebApi(Configuration, "AzureAdB2C")
             //         .AddInMemoryTokenCaches();
--- a/ProjectTemplates/test-templates.bat
+++ b/ProjectTemplates/test-templates.bat
@@ -1,0 +1,58 @@
+REM Build and Install templates
+msbuild /t:pack AspNetCoreMicrosoftIdentityPlatformTemplates.csproj
+cd bin
+cd Debug
+dotnet new -i Microsoft.IdentityPlatform.Templates.0.1.0.nupkg
+
+REM Test templates
+mkdir tests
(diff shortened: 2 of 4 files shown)


ProjectTemplates/templates/WebApi-CSharp/Startup.cs
```diff
@@ -48,7 +48,7 @@ public void ConfigureServices(IServiceCollection services)
             //         .AddInMemoryTokenCaches();
 #elif (IndividualB2CAuth)
             services.AddAuthentication(JwtBearerDefaults.AuthenticationScheme)
-                    .AddProtectedWebApi("AzureAdB2C", Configuration, options => Configuration.Bind("AzureAdB2C", options));
+                    .AddProtectedWebApi(Configuration, "AzureAdB2C");
```
jmprieur, line 51: This was a bug in the template;

ProjectTemplates/test-templates.bat
```diff
@@ -0,0 +1,51 @@
+REM Build and Install templates
+msbuild /t:pack AspNetCoreMicrosoftIdentityPlatformTemplates.csproj
```
bgavrilMS, line 2: Very nice. Are you going to copy this to the CI or just run it manuallly everytime a template changes?

jmprieur: I was thinking of adding a task in the CI, but of course it needs to pickup the nuget package produced by the source tree (not from NuGet). I guess a NuGet config somewhere should do the trick @bgavrilMS  ?

I would also think we could run it part of the github action, but that means I'd need to write a shell script (as it's running on ubuntu). I guess this is almost the same? `mkdir` instead of `md` ?

jmprieur: Oh, and @bgavrilMS : it would not be only when templates change, but also when the public API changes .... this would help catching templates becoming out of sync ..
(not that we want to change the public API ... but ...)

bgavrilMS: There are several ways of doing this. You could publish the package to ADO / GitHub package manager and then have a step that verifies the templates. Or do what this script is doing (you can even copy the script in a build step, perhaps use variable so that you don't hardcode the version)

For Linux, yes, `mkdir` will work. I am not sure if `msbuild` is installed on a linux box by default, so you can replace `msbuild` with `dotnet build` / `dotnet pack` if you have problems with that.

jmprieur: @bgavrilMS : msbuild is installed in the context of a GitHub action: See https://github.com/AzureAD/microsoft-identity-web/blob/4cfedc9771ff75ec611f48f504dd0eea546e1cdc/.github/workflows/dotnetcore.yml#L17. So that should work. Thanks for confirmining!

jennyf19: Nice

PR #118, opened 2020-04-22
Fixing the Web app template:

jmprieur:
- added references to Microsoft.Identity.Web.UI in the **Startup.cs** (using Microsoft.Identity.Web.UI and call to AddMicrosoftIdentityUI())
- added package reference to  Microsoft.Identity.Web.UI in the web app project
- changed the metadata of the template (mvc2) to mention Microsof identity platform and explicit Microsoft.Identity.Web

Fixing the Web API (reference to Microsoft.Identity.Web)

Renaming the generated NuGet package for the templates so that it starts with Microsoft.Identity.Web (Microsoft.Identity.Web.ProjectTemplates)


jmprieur:
note that you might want to hide the white space changes for the cshtml files (there are less changes than it appears)

PR #122, opened 2020-04-23
fix polycheck violation

jennyf19:
😹


jmprieur:
Thanks for fixing, @jennyf19 
I wonder how ASP.NET core can ship these templates. The violation is in an open source dependency, which almost everybody has ... a mistery

PR #127, opened 2020-04-24
leverage msal.net's httpclient factory

jennyf19:
Fix for #6

### DIFF
--- a/src/Microsoft.Identity.Web/TokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web/TokenAcquisition.cs
@@ -15,6 +15,7 @@ using System.Collections.Generic;
 using System.IdentityModel.Tokens.Jwt;
 using System.Linq;
 using System.Net;
+using System.Net.Http;
 using System.Security.Claims;
 using System.Threading.Tasks;
 
@@ -33,6 +34,7 @@ namespace Microsoft.Identity.Web
         private IConfidentialClientApplication _application;
         private readonly IHttpContextAccessor _httpContextAccessor;
         private HttpContext CurrentHttpContext => _httpContextAccessor.HttpContext;
+        private IMsalHttpClientFactory _httpClientFactory;
         private readonly ILogger _logger;
 
         /// <summary>
@@ -48,12 +50,14 @@ namespace Microsoft.Identity.Web
             IHttpContextAccessor httpContextAccessor,
             IOptions<MicrosoftIdentityOptions> microsoftIdentityOptions,
             IOptions<ConfidentialClientApplicationOptions> applicationOptions,
+            IHttpClientFactory httpClientFactory,
             ILogger<TokenAcquisition> logger)
         {
             _httpContextAccessor = httpContextAccessor;
             _microsoftIdentityOptions = microsoftIdentityOptions.Value;
             _applicationOptions = applicationOptions.Value;
             _tokenCacheProvider = tokenCacheProvider;
+            _httpClientFactory = new MsalAspNetCoreHttpClientFactory(httpClientFactory);
             _logger = logger;
         }
 
@@ -337,7 +341,7 @@ namespace Microsoft.Identity.Web
                 request.PathBase,
                 _microsoftIdentityOptions.CallbackPath.Value ?? string.Empty);
 
-            if (!_applicationOptions.Instance.EndsWith("/"))
+            if (!_applicationOptions.Instance.EndsWith("/", StringComparison.InvariantCulture))
                 _applicationOptions.Instance += "/";
 
             string authority;
@@ -352,6 +356,7 @@ namespace Microsoft.Identity.Web
                         .CreateWithApplicationOptions(_applicationOptions)
                         .WithRedirectUri(currentUri)
                         .WithB2CAuthority(authority)
+                        .WithHttpClientFactory(_httpClientFactory)
                         .Build();
                 }
                 else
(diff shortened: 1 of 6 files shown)


src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -48,12 +50,14 @@ internal class TokenAcquisition : ITokenAcquisition, ITokenAcquisitionInternal
             IHttpContextAccessor httpContextAccessor,
             IOptions<MicrosoftIdentityOptions> microsoftIdentityOptions,
             IOptions<ConfidentialClientApplicationOptions> applicationOptions,
+            IHttpClientFactory httpClientFactory,
```
bgavrilMS, line 53: Please update comments.

src/Microsoft.Identity.Web/WebApiAuthenticationBuilderExtensions.cs
```diff
@@ -74,6 +74,7 @@ public static class WebApiAuthenticationBuilderExtensions
 
             builder.Services.AddHttpContextAccessor();
             builder.Services.AddSingleton<IJwtBearerMiddlewareDiagnostics, JwtBearerMiddlewareDiagnostics>();
+            builder.Services.AddHttpClient();
```
bgavrilMS, line 77: Just for my understanding - this will ensure that the high performance HTTP client factory is used? Is there any scenario where people might not want to use this, for example for testing?

src/Microsoft.Identity.Web/TokenAcquisition.cs
bgavrilMS, line 53: Can this ever be null? I don't understand the DI that well, so apologies for asking silly questions.

tests/Microsoft.Identity.Web.Test.Integration/AcquireTokenForAppIntegrationTests.cs
```diff
@@ -91,6 +92,7 @@ private void InitializeTokenAcquisitionObjects()
             IOptions<MsalMemoryTokenCacheOptions> tokenOptions = _provider.GetService<IOptions<MsalMemoryTokenCacheOptions>>();
             IOptions<ConfidentialClientApplicationOptions> ccOptions = _provider.GetService<IOptions<ConfidentialClientApplicationOptions>>();
             ILogger<TokenAcquisition> logger = _provider.GetService<ILogger<TokenAcquisition>>();
+            IHttpClientFactory httpClientFactory = _provider.GetService<IHttpClientFactory>();
```
bgavrilMS, line 95: I wonder if we can add a better unit test here, as this doesn't really assert anything? At the very we should assert the httpClientFactory is not null. 

Ideally you would inject a mock IHttpClientFactory and make sure that MSAL actually uses it for token acquisition.

src/Microsoft.Identity.Web/TokenAcquisition.cs
jmprieur: No it cannot, @bgavrilMS:  The [ASP] .NET Core framework will throw if an `IHttpClientFactory` was not present as a service. It's presence is ensured in Microsoft.Identity.Web by the call to `services.AddHttpClient()`which was added in the `WebApiAuthenticationBuilderExtensions` and `WebAppAuthenticationBuilderExtensions extension methods`

src/Microsoft.Identity.Web/WebApiAuthenticationBuilderExtensions.cs
jmprieur: That's a good question. 
Would it be ok to wait for customer feedback ?

bgavrilMS: Absolutely.

tests/Microsoft.Identity.Web.Test.Integration/AcquireTokenForAppIntegrationTests.cs
jennyf19: When you do `BuildTheRequiredServices()` in the test, you do ` services.AddHttpClient();` which adds the httpClient, which will do the DI for the IHttpClientFactory. The implementation of the IHttpClientFactory is tightly tied to the DI implementation, so i'm not sure how we would do this without doing something outside of what actually happens with the library.

jennyf19:
thanks for reviewing @pmaytak ...thought i had added you originally as well. :(

tests/Microsoft.Identity.Web.Test.Integration/AcquireTokenForAppIntegrationTests.cs
bgavrilMS: Fair enough.

src/Microsoft.Identity.Web/WebApiAuthenticationBuilderExtensions.cs
pmaytak: [AddHttpClient()](https://github.com/dotnet/runtime/blob/master/src/libraries/Microsoft.Extensions.Http/src/DependencyInjection/HttpClientFactoryServiceCollectionExtensions.cs#L23) adds a few required services. Adding my own factory should still work. I can either remove these default services and add my own before the service provider is built. I can also just add my own factory after I call AddHttpClient, it should overwrite the same service types that were added earlier. AddHttpClient also uses TryAdd* methods to add the services, so if I add my own factory to the collection first, calling AddHttpClient should not overwrite what I've added.

I haven't read the docs, so there might be some behavior that I'm missing; but I think this should work.

src/Microsoft.Identity.Web/TokenAcquisition.cs
pmaytak: The framework resolves these references recursively. So if a service DI mapping was not added for some deep dependency, it will throw an error.

PR #132, opened 2020-04-28
Added unit tests for Web app extensions plus minor fixes.

pmaytak:
- Added unit tests for
  - WebAppAuthenticationBuilderExtensions
  - WebAppServiceCollectionExtensions
- Minor fixes.

_Might need to refactor tests later to improve readability, etc._


tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
```diff
@@ -1,32 +1,493 @@
...
+            var configuredOidcOptions = provider.GetServices<IConfigureOptions<OpenIdConnectOptions>>().Cast<ConfigureNamedOptions<OpenIdConnectOptions>>();
+            var configuredMsOptions = provider.GetServices<IConfigureOptions<MicrosoftIdentityOptions>>().Cast<ConfigureNamedOptions<MicrosoftIdentityOptions>>();
+
+            Assert.Contains(configuredOidcOptions, o => o.Action == _configureOidcOptions);
+            Assert.Contains(configuredMsOptions, o => o.Action == _configureMsOptions);
+
+            AddSignIn_TestCommon(services, provider);
+        }
+
+        private void AddSignIn_TestCommon(IServiceCollection services, ServiceProvider provider)
```
jennyf19, line 104: Please put all public methods before the private ones.

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
```diff
@@ -1,32 +1,493 @@
...
+        public void AddSignIn_WithConfigName(bool useServiceCollectionExtension)
+        {
+            var configMock = Substitute.For<IConfiguration>();
+            configMock.Configure().GetSection(_configSectionName).Returns(_configSection);
+              
+            var services = new ServiceCollection();
+            services.AddDataProtection();
+
+            if (useServiceCollectionExtension)
+                services.AddSignIn(configMock, _configSectionName, _oidcScheme, _cookieScheme, false);
```
jennyf19, line 63: can you add the { }...with the analyzers you added, this becomes a warning, which I've fixed in the branch i pushed to your other branch. thx.

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
```diff
@@ -1,32 +1,493 @@
...
+            else
+                new AuthenticationBuilder(services)
+                    .AddSignIn(configMock, _configSectionName, _oidcScheme, _cookieScheme, false);
+
+            var provider = services.BuildServiceProvider();
+
+            // Assert config bind actions added correctly
+            provider.GetRequiredService<IOptionsFactory<OpenIdConnectOptions>>().Create(_oidcScheme);
+            provider.GetRequiredService<IOptionsFactory<MicrosoftIdentityOptions>>().Create("");
```
jennyf19, line 72: >"" [](start = 92, length = 2)

use string.Empty()

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
jmprieur: why not merge the other branch?

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
```diff
@@ -1,32 +1,493 @@
...
+            var configuredOidcOptions = provider.GetService<IConfigureOptions<OpenIdConnectOptions>>() as ConfigureNamedOptions<OpenIdConnectOptions>;
+
+            Assert.Equal(_oidcScheme, configuredOidcOptions.Name);
+
+            // Assert properties set
+            var oidcOptions = provider.GetRequiredService<IOptionsFactory<OpenIdConnectOptions>>().Create(_oidcScheme);
+
+            Assert.Equal(_cookieScheme, oidcOptions.SignInScheme);
+            Assert.NotNull(oidcOptions.Authority);
+            Assert.NotNull(oidcOptions.TokenValidationParameters.IssuerValidator);
```
jennyf19, line 123: Is there a reason to not check the values here, as you do for preferredUserName? Might as well check the authority is correct, client_id, callback path, scopes, issuer address, etc...

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
```diff
@@ -1,32 +1,493 @@
...
+            var configMock = Substitute.For<IConfiguration>();
+            configMock.Configure().GetSection(_configSectionName).Returns(_configSection);
+
+            var redirectFunc = Substitute.For<Func<RedirectContext, Task>>();
+            var services = new ServiceCollection()
+                .Configure<OpenIdConnectOptions>(_oidcScheme, (options) => {
+                    options.Events ??= new OpenIdConnectEvents();
+                    options.Events.OnRedirectToIdentityProvider += redirectFunc;
+                });
+            services.AddDataProtection();
```
jennyf19, line 141: what does `.AddDataProtection()` do?

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
```diff
@@ -1,32 +1,493 @@
...
+            else
+                new AuthenticationBuilder(services)
+                    .AddSignIn(_configureOidcOptions, _configureMsOptions, _oidcScheme, _cookieScheme, false);
+
+            await AddSignIn_TestOnRedirectToIdentityProviderEvent(services, redirectFunc).ConfigureAwait(false);
+        }
+
+        private async Task AddSignIn_TestOnRedirectToIdentityProviderEvent(IServiceCollection services, Func<RedirectContext, Task> redirectFunc)
+        {
+            var provider = services.BuildServiceProvider();
```
jennyf19, line 176: >var provider = services.BuildServiceProvider(); [](start = 12, length = 47)

in some tests you build the service provider in the test and pass it along, others you build it in the private method. can you move them all to the private method and reduce the amount of parameters?

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
```diff
@@ -1,32 +1,493 @@
...
+
+            var oidcOptions = provider.GetRequiredService<IOptionsFactory<OpenIdConnectOptions>>().Create(_oidcScheme);
+
+            var httpContext = HttpContextUtilities.CreateHttpContext();
+            var authScheme = new AuthenticationScheme(OpenIdConnectDefaults.AuthenticationScheme, OpenIdConnectDefaults.AuthenticationScheme, typeof(OpenIdConnectHandler));
+            var authProperties = new AuthenticationProperties();
+            authProperties.Items[OidcConstants.AdditionalClaims] = "additional_claims";
```
jennyf19, line 183: >"additional_claims" [](start = 67, length = 19)

we have constant values for these right?

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
```diff
@@ -1,32 +1,493 @@
...
+                ProtocolMessage = new OpenIdConnectMessage(),
+            };
+
+            await oidcOptions.Events.RedirectToIdentityProvider(redirectContext).ConfigureAwait(false);
+
+            // Assert properties set, events called
+            await redirectFunc.ReceivedWithAnyArgs().Invoke(Arg.Any<RedirectContext>()).ConfigureAwait(false);
+            Assert.NotNull(redirectContext.ProtocolMessage.LoginHint);
+            Assert.NotNull(redirectContext.ProtocolMessage.DomainHint);
+            Assert.NotNull(redirectContext.ProtocolMessage.Parameters[OidcConstants.AdditionalClaims]);
```
jennyf19, line 197: what about checking the `redirectContext.ProtocolMessage.RequestType` and `Claims.IsAuthenticated` values as well? which is beyond something that you've intentionally set.

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
```diff
@@ -1,32 +1,493 @@
...
+        }
+
+        [Theory]
+        [InlineData(true, true)]
+        [InlineData(true, false)]
+        [InlineData(false, true)]
+        [InlineData(false, false)]
+        public void AddSignIn_WithConfigName_SubscribesToDiagnostics(bool useServiceCollectionExtension, bool subscribeToDiagnostics)
+        {
```
jennyf19, line 206: can this be merged w/the AddSignIn_WithConfigActions(bool userServiceCollectionExtension) test, as it's just adding logging, but lots more code to these tests. I think you could streamline it more and do less but more in each test. just a thought. things seem to be repeating with slight changes.

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
```diff
@@ -1,32 +1,493 @@
...
+            else
+                diagnosticsMock.DidNotReceiveWithAnyArgs().Subscribe(Arg.Any<OpenIdConnectEvents>());
+        }
+
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public async Task AddSignIn_WithConfigName_B2cSpecificSetup(bool useServiceCollectionExtension)
+        {
+            var configMock = Substitute.For<IConfiguration>();
```
jennyf19, line 266: there shouldn't be too many differences between b2c and AAD set up. could you have an additional bool, like `isB2C` and then route accordingly? This is basically the same as AddSIgnIn_WithConfigName()

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
```diff
@@ -1,32 +1,493 @@
...
-        [InlineData(TestConstants.B2CCustomDomainAuthority, TestConstants.B2CCustomDomainAuthorityWithV2)]
-        public void EnsureAuthorityIsV2_0(string initialAuthority, string expectedAuthority)
-        {
-            OpenIdConnectOptions options = new OpenIdConnectOptions
+        [InlineData(true)]
+        [InlineData(false)]
+        public async Task AddSignIn_WithConfigActions_B2cSpecificSetup(bool useServiceCollectionExtension)
+        {
+            _configureMsOptions = (options) => {
+                options.Instance = TestConstants.AadInstance;
```
jennyf19, line 288: B2C shouldn't use an AadInstance...use something like `"[https://fabrikamb2c.b2clogin.com](https://fabrikamb2c.b2clogin.com)"`

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
```diff
@@ -1,32 +1,493 @@
...
+                options.TenantId = TestConstants.TenantIdAsGuid;
+                options.ClientId = TestConstants.ClientId;
+                options.SignUpSignInPolicyId = TestConstants.B2CSignUpSignInUserFlow;
+                options.Domain = TestConstants.Domain;
```
jennyf19, line 292: domain should be b2c specific -> `"fabrikamb2c.onmicrosoft.com"`

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
```diff
@@ -1,32 +1,493 @@
...
+            var redirectContext = new RedirectContext(httpContext, authScheme, oidcOptions, authProperties)
+            {
+                ProtocolMessage = new OpenIdConnectMessage() { IssuerAddress = $"IssuerAddress/{TestConstants.B2CSignUpSignInUserFlow}/" },
+            };
+
+            await oidcOptions.Events.RedirectToIdentityProvider(redirectContext).ConfigureAwait(false);
+
+            // Assert issuer is updated to non-default user flow
+            Assert.Contains(TestConstants.B2CResetPasswordUserFlow, redirectContext.ProtocolMessage.IssuerAddress);
+        }
```
jennyf19, line 329: I wouldn't use password reset here, as that policy will move under sign-up/sign-in soon, so maybe use edit profile or something else.
This is similar to the tests, expect you're setting the issuer address and the claim name...can you combine TestCommon and this? And then we will be asserting more values w/b2c as well.

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
```diff
@@ -1,32 +1,493 @@
...
+                { $"{configSectionName}:Instance", TestConstants.AadInstance },
+                { $"{configSectionName}:TenantId", TestConstants.TenantIdAsGuid },
+                { $"{configSectionName}:ClientId", TestConstants.TenantIdAsGuid },
+                { $"{configSectionName}:Domain", TestConstants.Domain },
             };
 
-            options.Authority = AuthorityHelpers.EnsureAuthorityIsV2(options.Authority);
-            Assert.Equal(expectedAuthority, options.Authority);
+            if (includeB2cConfig)
+            {
```
jennyf19, line 343: needs to include b2c specific instance and domain

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
```diff
@@ -1,32 +1,493 @@
...
+            Assert.Contains(services, s => s.ServiceType == typeof(IConfigureOptions<MicrosoftIdentityOptions>));
+            Assert.Contains(services, s => s.ServiceType == typeof(IConfigureOptions<OpenIdConnectOptions>));
+
+            // Assert OIDC options added correctly
+            var configuredOidcOptions = provider.GetService<IConfigureOptions<OpenIdConnectOptions>>() as ConfigureNamedOptions<OpenIdConnectOptions>;
+
+            Assert.Equal(_oidcScheme, configuredOidcOptions.Name);
+        }
+
+        private async Task AddWebAppCallsProtectedWebApi_TestAuthorizationCodeReceivedEvent(IServiceProvider provider, OpenIdConnectOptions oidcOptions, Func<AuthorizationCodeReceivedContext, Task> authCodeReceivedFuncMock, ITokenAcquisition tokenAcquisitionMock)
```
jennyf19, line 445: nit: can you move each parameter to a new line for the longer method signatures? thx.

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
```diff
@@ -1,32 +1,493 @@
...
+            // Assert event called
+            await authCodeReceivedFuncMock.ReceivedWithAnyArgs().Invoke(Arg.Any<AuthorizationCodeReceivedContext>()).ConfigureAwait(false);
+            await ((ITokenAcquisitionInternal)tokenAcquisitionMock).ReceivedWithAnyArgs().AddAccountToCacheFromAuthorizationCodeAsync(Arg.Any<AuthorizationCodeReceivedContext>(), Arg.Any<IEnumerable<string>>()).ConfigureAwait(false);
+        }
+
+        private async Task AddWebAppCallsProtectedWebApi_TestTokenValidatedEvent(OpenIdConnectOptions oidcOptions, Func<TokenValidatedContext, Task> tokenValidatedFuncMock)
+        {
+            var httpContext = HttpContextUtilities.CreateHttpContext();
+            httpContext.Request.Form = new FormCollection(
+                new Dictionary<string, StringValues>() { { ClaimConstants.ClientInfo, Base64UrlHelpers.Encode($"{{\"uid\":\"{TestConstants.TenantIdAsGuid}\",\"utid\":\"{TestConstants.TenantIdAsGuid}\"}}") } });
```
jennyf19, line 464: nit: make a specific uid constant value and utid constant value...i don't think they are normally the tenantId. but @jmprieur would probably know. And then let's assert those specific values are there below.

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
```diff
@@ -1,32 +1,493 @@
...
+
+            await oidcOptions.Events.TokenValidated(tokenValidatedContext).ConfigureAwait(false);
+
+            // Assert event called, properties set
+            await tokenValidatedFuncMock.ReceivedWithAnyArgs().Invoke(Arg.Any<TokenValidatedContext>()).ConfigureAwait(false);
+            Assert.True(tokenValidatedContext.Principal.HasClaim(c => c.Type == ClaimConstants.Tid));
+            Assert.True(tokenValidatedContext.Principal.HasClaim(c => c.Type == ClaimConstants.UniqueObjectIdentifier));
+        }
+
+        private async Task AddWebAppCallsProtectedWebApi_TestRedirectToIdentityProviderForSignOutEvent(IServiceProvider provider, OpenIdConnectOptions oidcOptions, Func<RedirectContext, Task> redirectFuncMock, ITokenAcquisition tokenAcquisitionMock)
```
jennyf19, line 478: nit: These last three have similar code set up, could you move some of it to another private method to reduce the code duplication? then there is less noise in the differences between the methods.

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
jennyf19: I don't know. It's Peter's branch. I need you to look at the doc warnings though. I'm not sure what we want to do w/some of the xml stuff. there are 400+ warnings still in that branch. I think majority are docs now...we can discuss.

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
pmaytak: I grouped these private methods right under the public ones that use them. I think it makes it more readable, instead of jumping to the end of the class each time. But I can add them to the end to be consistent with the rest of the project.

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
pmaytak: Will do.

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
pmaytak: Done.

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
pmaytak: I haven't looked exactly at what it does but the data protection services are required when creating an instance of OIDC options.

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
pmaytak: I was only testing the specific things to that method. Valid authority format is tested in AuthorityHelpers and other properties aren't touched here.

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
pmaytak: The reason why I build the provider in the public methods is because they have specific tests to that method that need a provider, like testing config by name vs by actions, and then I also need the provider in the private method to test the common things.

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
pmaytak: Well unit tests should test one thing, have no logic, be simple, act as documentation. This way if this one unit test breaks but the rest are fine, I know something related to B2C broke. Writing tests this way does create a ton of duplicated quote (partially because setup code can't be abstracted away because of small differences). On the other extreme I can write just one 250+ line method to test everything. This reduces the code duplication but adds a potential for errors and, if it breaks, I have to debug it, basically.

I haven't strictly followed the [best practices](AddSignIn_TestOnRedirectToIdentityProviderEvent) but tried to strike a balance between the two extremes whenever it made sense. You can look at [AadIsserValidatorTests](https://github.com/AzureAD/microsoft-identity-web/blob/master/tests/Microsoft.Identity.Web.Test/Resource/AadIssuerValidatorTests.cs) where I took the opposite approach of writing smaller tests with more duplication.

So in a sense, since both extremes are not perfect, I'm okay with writing tests that more or less makes sense and work and we can refactor them later, if needed. (The 80-20 rule).

@jmprieur @henrik-me thoughts?

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
pmaytak: Reply in https://github.com/AzureAD/microsoft-identity-web/pull/132#discussion_r417813473

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
pmaytak: Nice to know. The specific values don't matter in this case but I updated it to Edit Profile for readability.

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
pmaytak: Specific value doesn't affect the outcome of the test but I've updated it for readability.

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
pmaytak: Specific value doesn't affect the outcome of the test but I've updated it for readability.

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
pmaytak: Specific value doesn't affect the outcome of the test but I've updated it for readability.

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
pmaytak: Done. Do we follow a parameter limit or just when it wraps around or arbitrary?

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
jennyf19: i moved these around in the analyzer branch, so mentioned it for consistency. it came up as a warning, so i just implemented the suggestion instead of suppressing it.

---
In reply to: [417791959](https://github.com/AzureAD/microsoft-identity-web/pull/132#discussion_r417791959) [](ancestors = 417791959)

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
jennyf19: I understand that, but if someone were to read the code in the tests, it doesn't make sense.

---
In reply to: [417818463](https://github.com/AzureAD/microsoft-identity-web/pull/132#discussion_r417818463) [](ancestors = 417818463)

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
jennyf19: we don't have a specific limit. @bgavrilMS started this for us in .NET. Usually when we have 3+, we start making new lines.

---
In reply to: [417819635](https://github.com/AzureAD/microsoft-identity-web/pull/132#discussion_r417819635) [](ancestors = 417819635)

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
jennyf19: I get that for the individual test, I guess I mean more for the private methods that seem to have very little difference between them. Or just move the configuration that's the same in these public methods somewhere else to remove "noise". just a suggestion.

---
In reply to: [417813473](https://github.com/AzureAD/microsoft-identity-web/pull/132#discussion_r417813473) [](ancestors = 417813473)

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
jmprieur: I think that this is the DPAPI?

jmprieur: Should we have it in the product code as well? @pmaytak ?

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
jmprieur: I"ll let @bgavrilMS provide his expertise (#not-a-dev)

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
jmprieur: I second @jennyf19 that we should avoid duplication.
@pmaytak there is a took in Visual Studio named "Code Clone" which will tell you which code is copied/paste. you might want to use it.

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
pmaytak: In prod code [AddAuthentication](https://github.com/dotnet/aspnetcore/blob/c9ce0664ea2d9ff842af8f6242d1609c93e8b683/src/Security/Authentication/Core/src/AuthenticationServiceCollectionExtensions.cs#L16) method adds the data protection services.

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
pmaytak: Yea, I'm only testing for the output caused by the current method. That other stuff holistically should be covered by integration tests.

PR #134, opened 2020-04-29
Updated AadIssuerValidator class to cache authority aliases under correct cache key.

pmaytak:


src/Microsoft.Identity.Web/Resource/AadIssuerValidator.cs
```diff
@@ -47,35 +47,36 @@ public static AadIssuerValidator GetIssuerValidator(string aadAuthority)
...
-                    .SelectMany(m => m.Aliases)
-                    .Append(authority) // For b2c scenarios, the alias will be the authorityHost itself
-                    .Distinct();
-                s_issuerValidators[authority] = new AadIssuerValidator(aliases);
-                return s_issuerValidators[authority];
+                return aadIssuerValidator;
             }
+
+            // In the constructor, we hit the Azure AD issuer metadata endpoint and cache the aliases. The data is cached for 24 hrs.
+            var issuerMetadata = s_configManager.GetConfigurationAsync().ConfigureAwait(false).GetAwaiter().GetResult();
```
bgavrilMS, line 69: Doesn't this potentially make an http request? A blocking call here would take pretty long. What is the impact of making this method async?

pmaytak: It does make a request. Making it an async is a breaking change, but it's an internal class... We should refactor this to use DI in general. I just wanted this PR/bug fix to be small and related to the cache issue.

bgavrilMS: Block-Waiting for an http request to complete is kind of bad :) Good things there is a static cache...

src/Microsoft.Identity.Web/Resource/AadIssuerValidator.cs
```diff
@@ -47,35 +47,36 @@ public static AadIssuerValidator GetIssuerValidator(string aadAuthority)
             if (string.IsNullOrEmpty(aadAuthority))
                 throw new ArgumentNullException(nameof(aadAuthority));
 
-            if (s_issuerValidators.TryGetValue(aadAuthority, out AadIssuerValidator aadIssuerValidator))
+            string authority;
+            try
```
jmprieur, line 51: Instead of the try/catch to create a Uri, could we use Uri.TryParse() ?

src/Microsoft.Identity.Web/Resource/AadIssuerValidator.cs
```diff
@@ -47,35 +47,36 @@ public static AadIssuerValidator GetIssuerValidator(string aadAuthority)
...
             {
-                return aadIssuerValidator;
+                authority = new Uri(aadAuthority).Authority;
```
jmprieur, line 53: Let's be careful. the OIDC authority contains paths (the instance does not), whereas the Uri.Authority is really the instance. I'm not a big fan of naming the instance `authority`. @jennyf19, @bgavrilMS what do you think?

pmaytak: Hmm, yea it's an overloaded term. Uri.Authority is host + port. Could rename it to authorityHost?

src/Microsoft.Identity.Web/Resource/AadIssuerValidator.cs
pmaytak: Hm, yes, I'll try.

Issue #66, opened 2019-10-27
[Bug] Null reference exception calling AcquireTokenByAuthorizationCode when client secret is not specified (need to improve error message)


winterlimelight:
**Which Version of MSAL are you using ?**
4.5.1

**Platform**
.NET Core 3

**What authentication flow has the issue?**
* Web App
    * [ ] Authorization code
 

**Is this a new or existing app?**
c. This is a new app or experiment

**Repro**
This is based on https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/wiki/Acquiring-tokens-with-authorization-codes-on-web-apps but using WithB2CAuthority rather than WithAuthority.

```csharp
IConfidentialClientApplication cca = ConfidentialClientApplicationBuilder
	.Create(AzureAdB2COptions.ClientId)
	.WithRedirectUri(AzureAdB2COptions.RedirectUri)
	.WithB2CAuthority(AzureAdB2COptions.Authority)
	.Build();

var builder = cca.AcquireTokenByAuthorizationCode(new[] { "openid" }, context.ProtocolMessage.Code);
AuthenticationResult result = await builder.ExecuteAsync();
```

**Expected behavior**
I was expecting to retrieve an access token in line with https://docs.microsoft.com/en-us/azure/active-directory-b2c/active-directory-b2c-reference-oauth-code#2-get-a-token

**Actual behavior**
Exception is thrown by ExecuteAsync(). Stack-trace as follows (awaits removed):
```
at Microsoft.Identity.Client.Internal.ClientCredentialWrapper.get_Thumbprint()
at Microsoft.Identity.Client.Internal.JsonWebToken.JWTHeaderWithCertificate..ctor(ClientCredentialWrapper credential, Boolean sendCertificate)
at Microsoft.Identity.Client.Internal.JsonWebToken.EncodeHeaderToJson(ClientCredentialWrapper credential, Boolean sendCertificate)
at Microsoft.Identity.Client.Internal.JsonWebToken.Encode(ClientCredentialWrapper credential, Boolean sendCertificate)
at Microsoft.Identity.Client.Internal.JsonWebToken.Sign(ClientCredentialWrapper credential, Boolean sendCertificate)
[...]

Issue #12, opened 2020-02-18
[Feature Request] Support certificates for app authentication


jennyf19:
see [this issue for more info](https://github.com/Azure-Samples/active-directory-aspnetcore-webapp-openidconnect-v2/pull/189)

See also See also https://github.com/AzureAD/microsoft-identity-web/wiki/Spec-certificates

jmprieur:
Closing as duplicate of #165

PR #139, opened 2020-04-30
Jennyf/null ref

jennyf19:
#66 
and
https://github.com/AzureAD/microsoft-identity-web/issues/19#issuecomment-621239808

### DIFF
--- a/src/Microsoft.Identity.Web/AuthorityHelpers.cs
+++ b/src/Microsoft.Identity.Web/AuthorityHelpers.cs
@@ -8,44 +8,20 @@ namespace Microsoft.Identity.Web
 {
     internal static class AuthorityHelpers
     {
-        internal static bool IsV2Authority(string authority)
-        {
-            if (string.IsNullOrEmpty(authority))
-                return false;
-
-            return authority.EndsWith("/v2.0");
-        }
-
         internal static string BuildAuthority(MicrosoftIdentityOptions options)
         {
-            if (options == null)
-                return null;
-
-            // Cannot build authority without AAD Instance
-            if (string.IsNullOrWhiteSpace(options.Instance))
-                return null;
-
             var baseUri = new Uri(options.Instance);
             var pathBase = baseUri.PathAndQuery.TrimEnd('/');
             var domain = options.Domain;
             var tenantId = options.TenantId;
 
-            // If there are user flows, then it must build a B2C authority 
-            if (!string.IsNullOrWhiteSpace(options.DefaultUserFlow))
+            if (options.IsB2C)
             {
-                // Cannot build B2C authority without domain
-                if (string.IsNullOrWhiteSpace(domain))
-                    return null;
-
                 var userFlow = options.DefaultUserFlow;
                 return new Uri(baseUri, new PathString($"{pathBase}/{domain}/{userFlow}/v2.0")).ToString();
             }
-            else
+            else // Create an AAD authority
             {
-                // Cannot build AAD authority without tenant id
-                if (string.IsNullOrWhiteSpace(tenantId))
-                    return null;
-
                 return new Uri(baseUri, new PathString($"{pathBase}/{tenantId}/v2.0")).ToString();
             }
         }
@@ -53,7 +29,7 @@ namespace Microsoft.Identity.Web
         internal static string EnsureAuthorityIsV2(string authority)
         {
             authority = authority.Trim().TrimEnd('/');
-            if (!authority.EndsWith("v2.0"))
+            if (!authority.EndsWith("v2.0", StringComparison.InvariantCulture))
                 authority += "/v2.0";
             return authority;
         }
--- a/src/Microsoft.Identity.Web/TokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web/TokenAcquisition.cs
@@ -12,6 +12,7 @@ using Microsoft.Identity.Web.TokenCacheProviders;
 using Microsoft.Net.Http.Headers;
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.IdentityModel.Tokens.Jwt;
 using System.Linq;
 using System.Net;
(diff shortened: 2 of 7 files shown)


jennyf19:
@bgavrilMS this is related to the MSAL null ref we discussed this morning. I'll do a minor fix there and open an issue. FYI.

src/Microsoft.Identity.Web/AuthorityHelpers.cs
```diff
@@ -8,52 +8,28 @@ namespace Microsoft.Identity.Web
...
             {
-                // Cannot build B2C authority without domain
-                if (string.IsNullOrWhiteSpace(domain))
-                    return null;
-
                 var userFlow = options.DefaultUserFlow;
                 return new Uri(baseUri, new PathString($"{pathBase}/{domain}/{userFlow}/v2.0")).ToString();
             }
-            else
+            else // Create an AAD authority
```
bgavrilMS, line 23: If it's not B2C, it can be ADFS.

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -378,11 +384,29 @@ private async Task<IConfidentialClientApplication> BuildConfidentialClientApplic
...
-                _logger.LogInformation(ex, "Exception acquiring token for a confidential client. ");
+                _logger.LogInformation(
+                    ex,
+                    string.Format(CultureInfo.InvariantCulture, 
+                    "Exception acquiring token for a confidential client. "));
                 throw;
             }
         }
 
+        internal void CheckApplicationOptionsContainClientSecret(ConfidentialClientApplicationOptions applicationOptions)
```
bgavrilMS, line 395: I think that this fix is not necessary in Identity.Web, and that it should be only in MSAL. But I can understand having a tactical fix in Identity.Web if you cannot wait for the next MSAL release (or you can hotfix MSAL).

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -349,6 +353,8 @@ private async Task<IConfidentialClientApplication> BuildConfidentialClientApplic
 
             try
             {
+                CheckApplicationOptionsContainClientSecret(_applicationOptions);
```
bgavrilMS, line 356: TLDR: What about certificates?



There are 4 ways of specifying the secure contract between the server and AAD in a confidential client: 

- client secret (should not really be used in production)
- cert (recommended) 
- client assertion (which is essentially a string with cert details - we use this when MSAL can't handle wierd certs)
- cert + extra claims 

ConfidentialClientApplicationOptions only allows users to pass in a secret. But you must provide mechanism for user to configure certificates.

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -378,11 +384,29 @@ private async Task<IConfidentialClientApplication> BuildConfidentialClientApplic
...
                 throw;
             }
         }
 
+        internal void CheckApplicationOptionsContainClientSecret(ConfidentialClientApplicationOptions applicationOptions)
+        {
+            if (string.IsNullOrEmpty(applicationOptions.ClientSecret))
+            {
+                string msg = string.Format(CultureInfo.InvariantCulture, "Client secret cannot be null or whitespace, " +
+                    "and must be included in the configuration of the web app when calling a web API. " +
```
bgavrilMS, line 400: A certificate can be used instead of a secret.

src/Microsoft.Identity.Web/WebAppAuthenticationBuilderExtensions.cs
```diff
@@ -162,5 +165,41 @@ public static class WebAppAuthenticationBuilderExtensions
...
+            {
+                if (string.IsNullOrEmpty(microsoftIdentityOptions.Domain))
+                {
+                    throw new ArgumentNullException(nameof(microsoftIdentityOptions.Domain),
+                        string.Format(CultureInfo.InvariantCulture, "The Domain " + msg));
+                }
+            }
+            else
+            {
+                if (string.IsNullOrEmpty(microsoftIdentityOptions.TenantId))
```
bgavrilMS, line 197: What about ADFS?

jmprieur: That's a good question @bgavrilMS 
For the moment, I did not even considered that Microsoft.Identity.Web coud support ADFS, but maybe it should.

src/Microsoft.Identity.Web/TokenAcquisition.cs
jmprieur: @bgavrilMS : yes, and we have an enhancements issue for this: (https://github.com/AzureAD/microsoft-identity-web/issues/12).

For the moment, proposing an incremental approach

src/Microsoft.Identity.Web/TokenAcquisition.cs
jmprieur: @bgavrilMS : yes, and we have an enhancements issue for this: (#12).

For the moment, proposing an incremental approach

src/Microsoft.Identity.Web/TokenAcquisition.cs
jmprieur: We want to check the other parameters anyway.

src/Microsoft.Identity.Web/AuthorityHelpers.cs
jmprieur: @bgavrilMS : do you think that Microsoft.Identity.Web should support ADFS?

bgavrilMS: I think you should at least throw a "not supported" exception or smth.

jennyf19:
> LGTM. @jennyf19
> Thanks for making the configuration more reliable
> 
> for the future, I just discovered that we could implement a IValidateOptions. See for instance https://github.com/dotnet/aspnetcore/blob/master/src/Azure/AzureAD/Authentication.AzureAD.UI/src/AzureADOptionsValidation.cs
> 
> It's then registered like this: https://github.com/dotnet/aspnetcore/blob/49b4f418d5e7351b192023515e33e06c84d25e0e/src/Azure/AzureAD/Authentication.AzureAD.UI/src/AzureADAuthenticationBuilderExtensions.cs#L63
> 
> Maybe we'd want to raise a new issue to use this pattern (which separates concerns even better)?

that's much easier. i just did it on a separate branch and it's easier to check web api options now. thx. will add some tests adn push that branhc.

PR #140, opened 2020-04-30
taking JM"s suggestion of checking the options in a separate class

jennyf19:
- added unit tests
- fix for null ref w/no client secret

Taking @jmprieur 's suggestions. thank you. much better.

@bgavrilMS - if you want to take a quick look. IDK what to do w/ADFS. ignoring for now.

### DIFF
--- a/src/Microsoft.Identity.Web/TokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web/TokenAcquisition.cs
@@ -12,6 +12,7 @@ using Microsoft.Identity.Web.TokenCacheProviders;
 using Microsoft.Net.Http.Headers;
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.IdentityModel.Tokens.Jwt;
 using System.Linq;
 using System.Net;
@@ -145,7 +146,10 @@ namespace Microsoft.Identity.Web
             }
             catch (MsalException ex)
             {
-                _logger.LogInformation(ex, "Exception occured while adding an account to the cache from the auth code. ");
+                _logger.LogInformation(
+                    ex,
+                    string.Format(CultureInfo.InvariantCulture,
+                    "Exception occured while adding an account to the cache from the auth code. "));
                 throw;
             }
         }
@@ -347,6 +351,19 @@ namespace Microsoft.Identity.Web
             string authority;
             IConfidentialClientApplication app;
 
+            MicrosoftIdentityOptionsValidation microsoftIdentityOptionsValidation = new MicrosoftIdentityOptionsValidation();
+            if (microsoftIdentityOptionsValidation.ValidateClientSecret(_applicationOptions).Failed)
+            {
+                string msg = string.Format(CultureInfo.InvariantCulture, "Client secret cannot be null or whitespace, " +
+                   "and must be included in the configuration of the web app when calling a web API. " +
+                   "For instance, in the appsettings.json file. ");
+
+                _logger.LogInformation(msg);
+                throw new MsalClientException(
+                    "missing_client_credentials",
+                    msg);
+            }
+
             try
             {
                 if (_microsoftIdentityOptions.IsB2C)
@@ -378,7 +395,10 @@ namespace Microsoft.Identity.Web
             }
             catch (Exception ex)
             {
-                _logger.LogInformation(ex, "Exception acquiring token for a confidential client. ");
+                _logger.LogInformation(
+                    ex,
+                    string.Format(CultureInfo.InvariantCulture,
+                    "Exception acquiring token for a confidential client. "));
                 throw;
             }
         }
--- a/src/Microsoft.Identity.Web/WebApiAuthenticationBuilderExtensions.cs
+++ b/src/Microsoft.Identity.Web/WebApiAuthenticationBuilderExtensions.cs
@@ -5,6 +5,8 @@ using Microsoft.AspNetCore.Authentication;
 using Microsoft.AspNetCore.Authentication.JwtBearer;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Extensions.Options;
 using Microsoft.Identity.Web.Resource;
 using Microsoft.IdentityModel.Tokens;
 using System;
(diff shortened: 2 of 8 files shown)


src/Microsoft.Identity.Web/WebApiAuthenticationBuilderExtensions.cs
```diff
@@ -76,6 +78,7 @@ public static class WebApiAuthenticationBuilderExtensions
             builder.Services.Configure(jwtBearerScheme, configureJwtBearerOptions);
             builder.Services.Configure<MicrosoftIdentityOptions>(configureMicrosoftIdentityOptions);
 
+            builder.Services.TryAddEnumerable(ServiceDescriptor.Singleton<IValidateOptions<MicrosoftIdentityOptions>, MicrosoftIdentityOptionsValidation>());
```
jennyf19, line 81: checking in the web api now as well.

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -347,6 +351,19 @@ private async Task<IConfidentialClientApplication> BuildConfidentialClientApplic
             string authority;
             IConfidentialClientApplication app;
 
+            MicrosoftIdentityOptionsValidation microsoftIdentityOptionsValidation = new MicrosoftIdentityOptionsValidation();
```
jmprieur, line 354: You could also inject a `IValidateOptions<MicrosoftIdentityOptions>` in the constructor by dependency injection

jennyf19: I tried that, but I only need to check the secret here, not all the options, so i just need the class and that method. it didn't work passing it into the constructor.

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -347,6 +351,19 @@ private async Task<IConfidentialClientApplication> BuildConfidentialClientApplic
...
+            if (microsoftIdentityOptionsValidation.ValidateClientSecret(_applicationOptions).Failed)
```
pmaytak, line 355: Perhaps we can create a CCAOptionsValidation class that validates this. It would also be more extensible and we could get the failure message from the result itself.

tests/Microsoft.Identity.Web.Test.Integration/AcquireTokenForAppIntegrationTests.cs
```diff
@@ -86,6 +87,33 @@ public async Task GetAccessTokenForApp_WithUserScope_MsalServiceExceptionThrownA
...
         }
 
+        [Theory]
+        [InlineData("")]
+        [InlineData(null)]
+        [InlineData("some_secret")]
+        public void ApplicationOptionsIncludeClientSecret(string clientSecret)
+        {
+            // Arrange
+            InitializeTokenAcquisitionObjects();
```
pmaytak, line 97: I don't think we need this line. 
Plus seems like this whole test method belongs to unit tests and not integration.

tests/Microsoft.Identity.Web.Test/MicrosoftIdentityOptionsTests.cs
```diff
@@ -1,13 +1,19 @@
...
+using System.Globalization;
 using Xunit;
 
 namespace Microsoft.Identity.Web.Test
 {
     public class MicrosoftIdentityOptionsTests
     {
+        private MicrosoftIdentityOptions microsoftIdentityOptions;
+        private const string AzureAd = "AzureAd";
+        private const string AzureAdB2C = "AzureAdB2C";
```
pmaytak, line 15: Hmm what's the correct naming convention - `AzureAdB2C` or `AzureAdB2c`?

tests/Microsoft.Identity.Web.Test/MicrosoftIdentityOptionsTests.cs
```diff
@@ -30,5 +36,66 @@ public void IsB2C_NullOrEmptyUserFlow_ReturnsFalse()
...
+
+            MicrosoftIdentityOptionsValidation microsoftIdentityOptionsValidation = new MicrosoftIdentityOptionsValidation();
+            ValidateOptionsResult result = microsoftIdentityOptionsValidation.Validate(optionsName, microsoftIdentityOptions);
+
+            CheckReturnValueAgainstExpectedMissingParam(missingParam, result);
+        }
+
+        private void CheckReturnValueAgainstExpectedMissingParam(MissingParam missingParam, ValidateOptionsResult result)
+        {
+            if (result.Failed == true)
```
pmaytak, line 81: `if (result.Failed)`

tests/Microsoft.Identity.Web.Test/MicrosoftIdentityOptionsTests.cs
```diff
@@ -30,5 +36,66 @@ public void IsB2C_NullOrEmptyUserFlow_ReturnsFalse()
...
+            {
+                string msg = string.Format(CultureInfo.InvariantCulture, "The '{0}' option must be provided.", missingParam);
```
pmaytak, line 83: `message`

tests/Microsoft.Identity.Web.Test/MicrosoftIdentityOptionsTests.cs
```diff
@@ -30,5 +36,66 @@ public void IsB2C_NullOrEmptyUserFlow_ReturnsFalse()
...
+            if (optionsName == AzureAdB2C)
+            {
+                microsoftIdentityOptions.SignUpSignInPolicyId = signUpSignInPolicyId;
+                microsoftIdentityOptions.Domain = domain;
+            }
+
+            MicrosoftIdentityOptionsValidation microsoftIdentityOptionsValidation = new MicrosoftIdentityOptionsValidation();
+            ValidateOptionsResult result = microsoftIdentityOptionsValidation.Validate(optionsName, microsoftIdentityOptions);
+
+            CheckReturnValueAgainstExpectedMissingParam(missingParam, result);
```
pmaytak, line 76: I would just put the asserts in the main method.

tests/Microsoft.Identity.Web.Test/MicrosoftIdentityOptionsTests.cs
```diff
@@ -30,5 +36,66 @@ public void IsB2C_NullOrEmptyUserFlow_ReturnsFalse()
...
+        public void ValidateRequiredMicrosoftIdentityOptions(
+           string clientId,
+           string instance,
+           string tenantid,
+           string signUpSignInPolicyId,
+           string domain,
+           string optionsName,
+           MissingParam missingParam = MissingParam.None)
+        {
+            microsoftIdentityOptions = new MicrosoftIdentityOptions
```
pmaytak, line 60: We can just instantiate the variable here instead of at instance level.

PR #150, opened 2020-05-08
Imported analyzer rules from MSAL and included some warning fixes

pmaytak:
- Included rules from MSAL .editorconfig into root .editorconfig.
- Included rules from MSAL SolutionWideAnalyzerConfig.ruleset into root .editorconfig.
- Added some rule suppressions to the tests .editorconfig.
- Added StyleCop analyzers.
- Moved stylecop.json config into solution level and referenced from the projects.
- Also includes Jenny's fixes.

This PR has a lot of files mostly because some warnings had a lot of occurrences (but it's mostly tiny changes). Following PRs should be smaller. Once this one is approved and merged, I will send a PR for the rest of the changes.

Includes fixes for:
- CA1305. Specify IFormatProvider.
- CA2007. Call ConfigureAwait on the awaited Task.
- SA1001. The spacing around a comma is incorrect, within a C# code file.
- SA1003. The spacing around an operator symbol is incorrect, within a C# code file.
- SA1004. A line within a documentation header above a C# element does not begin with a single space.
- SA1005. A single-line comment within a C# code file does not begin with a single space.
- SA1028. A line of code ends with a space, tab, or other whitespace characters before the end of line character(s).
- SA1208. A using directive which declares a member of the System namespace appears after a using directive which declares a member of a different namespace, within a C# code file.
- SA1413. A multi-line initializer should use a comma on the last item.
- SA1503. The opening and closing braces for a C# statement have been omitted.
- SA1600. A C# code element is missing a documentation header.
- SA1629. A section of the Xml header documentation for a C# element does not end with a period (also known as a full stop).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,5 +1,479 @@
-# NOTE: Requires **VS2019 16.3** or later
+# Note: MSAL rules used as a starting point
 
-# Code files
-[*.{cs,vb}]
+# editorconfig.org
+# Copied from the .NET Core repo https://github.com/dotnet/corefx/blob/master/.editorconfig
 
+# top-most EditorConfig file
+root = true
+
+# Default settings:
+# A newline ending every file
+# Use 4 spaces as indentation
+[*]
+insert_final_newline = true
+indent_style = space
+indent_size = 4
+
+[project.json]
+indent_size = 2
+
+# C# files
+[*.cs]
+# New line preferences
+csharp_new_line_before_open_brace = all
+csharp_new_line_before_else = true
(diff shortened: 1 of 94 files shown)


pmaytak:
@jmprieur Yes, I'll try to reduce the noise. And then perhaps we can fix as we go.

jmprieur:
yes, @pmaytak : I think that fixing the style as we go is a good technics (less stressful, and does not require work that could lead to regression if we don't fully pay attention)
Clean-up as you go, is BTW what is recommended by Microsoft technical debt team.

PR #153, opened 2020-05-10
Minor fixes to samples home page

pmaytak:
In both sample client projects:
- moved the cookie consent bar to bottom. (Currently it completely hides the menu bar, which is a bit confusing.)
**Before:**
image

- updated images on home pages.
**Before:** 
image
**After:**
image


pmaytak:
@jennyf19 Perhaps. I can take a look. Although I don't have any samples set up, so can't really test. But I can send a PR.

Issue #154, opened 2020-05-10
[Bug] AddProtectedWebApiCallsProtectedWebApi overwrites existing OnTokenValidated event handler


jg11jg:
The AddProtectedWebApiCallsProtectedWebApi method registers an event handler for OnTokenValidated, but without preserving any existing registered event handler. As a side effect, any existing event registration gets lost, for example, the OnTokenValidated event handler registered in AddProtectedWebApi(), and indeed the OnTokenValidated event handler in JwtBearerMiddlewareDiagnostics (if subscribed).

I believe this is a bug.

https://github.com/AzureAD/microsoft-identity-web/blob/3151e1ca8e5bb9348cf439d999b27fb1e1afee3c/src/Microsoft.Identity.Web/WebApiServiceCollectionExtensions.cs#L127

jennyf19:
automatically closed, will reopen until it's in the release

jennyf19:
Included in [0.1.3-preview release](https://github.com/AzureAD/microsoft-identity-web/releases/tag/0.1.3-preview)

PR #155, opened 2020-05-11
fix: #154 AddProtectedWebApiCallProtectedWebApi overwrites handler

jg11jg:
Code now preserves existing event handler when registering OnTokenValidated event handler in AddProtectedWebApiCallsProtectedWebApi.  Previous code was overwriting existing event handler meaning that logging via JwtBearerMiddlewareDiagnostics would no longer work and perhaps more seriously the OnTokenValidated event handler registered in AddProtectedWebApi (intended to "ensure that the Web API only accepts tokens from tenants where it has been consented and provisioned.")

https://github.com/AzureAD/microsoft-identity-web/issues/154

### DIFF
--- a/src/Microsoft.Identity.Web/WebApiServiceCollectionExtensions.cs
+++ b/src/Microsoft.Identity.Web/WebApiServiceCollectionExtensions.cs
@@ -124,11 +124,15 @@ namespace Microsoft.Identity.Web
             {
                 options.Events ??= new JwtBearerEvents();
 
+                var onTokenValidatedHandler = options.Events.OnTokenValidated;
+
                 options.Events.OnTokenValidated = async context =>
                 {
                     context.HttpContext.StoreTokenUsedToCallWebAPI(context.SecurityToken as JwtSecurityToken);
                     context.Success();
                     await Task.FromResult(0).ConfigureAwait(false);
+
+                    await onTokenValidatedHandler(context).ConfigureAwait(false);
                 };
             });
 
--- a/src/Microsoft.Identity.Web/WebApiServiceCollectionExtensions.cs
+++ b/src/Microsoft.Identity.Web/WebApiServiceCollectionExtensions.cs
@@ -128,11 +128,10 @@ namespace Microsoft.Identity.Web
 
                 options.Events.OnTokenValidated = async context =>
                 {
+                    await onTokenValidatedHandler(context).ConfigureAwait(false);                
                     context.HttpContext.StoreTokenUsedToCallWebAPI(context.SecurityToken as JwtSecurityToken);
                     context.Success();
                     await Task.FromResult(0).ConfigureAwait(false);
-
-                    await onTokenValidatedHandler(context).ConfigureAwait(false);
                 };
             });
 


jmprieur:
Thanks @jg11jg. Good catch!
I think that the previous event should be called before storing the token

src/Microsoft.Identity.Web/WebApiServiceCollectionExtensions.cs
```diff
@@ -124,15 +124,19 @@ public static class WebApiServiceCollectionExtensions
...
 
+                var onTokenValidatedHandler = options.Events.OnTokenValidated;
+
                 options.Events.OnTokenValidated = async context =>
                 {
                     context.HttpContext.StoreTokenUsedToCallWebAPI(context.SecurityToken as JwtSecurityToken);
                     context.Success();
                     await Task.FromResult(0).ConfigureAwait(false);
+
+                    await onTokenValidatedHandler(context).ConfigureAwait(false);
```
jmprieur, line 135:
```suggestion
                options.Events.OnTokenValidated = async context =>
                {
                    await onTokenValidatedHandler(context).ConfigureAwait(false);                
                    context.HttpContext.StoreTokenUsedToCallWebAPI(context.SecurityToken as JwtSecurityToken);
                    context.Success();
                    await Task.FromResult(0).ConfigureAwait(false);
```

src/Microsoft.Identity.Web/WebApiServiceCollectionExtensions.cs
```diff
@@ -124,15 +124,19 @@ public static class WebApiServiceCollectionExtensions
             {
                 options.Events ??= new JwtBearerEvents();
 
+                var onTokenValidatedHandler = options.Events.OnTokenValidated;
+
                 options.Events.OnTokenValidated = async context =>
```
pmaytak, line 129: @jmprieur Related to your suggestion we can just `+=` our handler to the existing ones, right?
```suggestion
                options.Events ??= new JwtBearerEvents();
                
                options.Events.OnTokenValidated += async context =>
```

jg11jg: possibly but the convention within the rest of the library is to do it as I originally did it (in fact I have used exactly the same variable names to be consistent with the "house style".  e.g: https://github.com/AzureAD/microsoft-identity-web/blob/d8608bb000b0d9415a290f5dde836248b179ae38/src/Microsoft.Identity.Web/WebAppServiceCollectionExtensions.cs#L175)

Possibly the += syntax requires null checks or similar?  Also in the *MiddlewareDiagnostics classes += is not used.  So I (personally) would stick with "house style" (or change it everywhere if += truly better).

src/Microsoft.Identity.Web/WebApiServiceCollectionExtensions.cs
jg11jg: Similar to my comments below I would say that in other examples in the library it seems conventional to "do your own stuff THEN call the other event handler".  Is there a specific reason to do differently in this case?  

Also I suspect the  await Task.FromResult(0).ConfigureAwait(false); was only there to stop Resharper/StyleCop/Rosyln nags about async method with no awaits.  There's an "real" await now so that line can go.

jennyf19: We want it as @jmprieur suggests because if the event does additional token validation, you don't want to store the token if this validation fails.

src/Microsoft.Identity.Web/WebApiServiceCollectionExtensions.cs
jennyf19: I agree with @jg11jg, we should keep things consistent. I would leave as-is and we can update things later if we want.

jennyf19:
Thanks @jg11jg will merge this in. @jmprieur @pmaytak we can look into adding a unit test for this.

Issue #14, opened 2020-02-18
No user scopes leads to error


jennyf19:
from @onovotny and copied from https://github.com/Azure-Samples/active-directory-aspnetcore-webapp-openidconnect-v2/issues/241

If you want to use .AddMsal(), there's no way to do so without any user scopes. It might be needed for Application flows.

If you use `new string[0]` for the scopes parameter, it'll throw in the OnAuthorizationCodeReceived callback since options.Scope will be null:
`await tokenAcquisition.AddAccountToCacheFromAuthorizationCodeAsync(context, options.Scope).ConfigureAwait(false);`

bgavrilMS:
Not sure this is a work item for the extenion. MSAL needs to support this scenario, and it is not trivial (but not super complex either). Tracking work item: 

https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/715

jmprieur:
@jennyf19 : proposing to take it in 0.2 given that MSAL.NET fixed it;

bgavrilMS:
This should work now and the behaviour of passing no scopes I observed is as follows: 

- MSAL will still pass in the required scopes for itself to function - `offline_access`, `openid`, `profile`
- AAD and ADFS do issue an access token (I think this wasn't the case some time ago). I assume it has to do with the fact that profile is transformed to User.Read.
- B2C does not issue an access token. 
- IDTokens are always issued
- `AcquireTokenSilent` with no scopes is not supported on B2C.

jmprieur:
Spec: @jennyf19 @pmaytak 

**Why?**
It should be possible to write:

```CSharp
.AddSignIn(Configuration)
.AddWebAppCallsProtectedWebApi(Configuration)
.AddInMemoryTokenCaches();
```

and then have the scopes requested in the Web app controllers for a complelety dynamic experience (with incremental consent, or with pre-authorized scopes)

**What?**
The proposal (non breaking) is to enable initialScopes to be null or an empty string[] in `AddWebAppCallsProtectedWebApi`.
https://github.com/AzureAD/microsoft-identity-web/blob/4cfedc9771ff75ec611f48f504dd0eea546e1cdc/src/Microsoft.Identity.Web/WebAppServiceCollectionExtensions.cs#L104-L109

pmaytak:
Proposal makes sense.

jennyf19:
[...]

PR #158, opened 2020-05-13
handle no scopes

jennyf19:
#14

### DIFF
--- a/src/Microsoft.Identity.Web/WebAppServiceCollectionExtensions.cs
+++ b/src/Microsoft.Identity.Web/WebAppServiceCollectionExtensions.cs
@@ -101,6 +101,8 @@ namespace Microsoft.Identity.Web
         /// (by default OpenIdConnectDefaults.AuthenticationScheme). This can be specified when you want to support
         /// several OpenIdConnect identity providers.</param>
         /// <returns>The service collection for chaining.</returns>
+        /// <remarks>This method cannot be used with Azure AD B2C as, with B2C an initial scope needs
+        /// to be provided</remarks>
         public static IServiceCollection AddWebAppCallsProtectedWebApi(
             this IServiceCollection services,
             IConfiguration configuration,
--- a/src/Microsoft.Identity.Web/WebAppServiceCollectionExtensions.cs
+++ b/src/Microsoft.Identity.Web/WebAppServiceCollectionExtensions.cs
@@ -95,13 +95,38 @@ namespace Microsoft.Identity.Web
         /// </summary>
         /// <param name="services">Service collection to which to add authentication.</param>
         /// <param name="configuration">Configuration.</param>
+        /// <param name="configSectionName">The name of the configuration section with the necessary
+        /// settings to initialize authentication options.</param>
+        /// <param name="openIdConnectScheme">Optional name for the open id connect authentication scheme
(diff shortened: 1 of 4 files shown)


tests/B2CWebAppCallsWebApi/Client/Startup.cs
```diff
@@ -53,9 +53,7 @@ public void ConfigureServices(IServiceCollection services)
 
             // Token acquisition service based on MSAL.NET
             // and chosen token cache implementation
-            services.AddWebAppCallsProtectedWebApi(Configuration,
-                initialScopes: new string[] { Configuration["TodoList:TodoListScope"] },
-                configSectionName: "AzureAdB2C")
+            services.AddWebAppCallsProtectedWebApi(Configuration, "AzureAdB2C")
```
bgavrilMS, line 56: Just wanted to point out the difference between AAD and B2C: 

- AAD will return an access token if you pass 0 scopes. AcquireTokenSilent works.
- B2C will not return an access token if you pass 0 scopes. This is a known issue that B2C team plans to fix at some point. AcquireTokenSilent does not work (MSAL will thrown an exception). AcquireToken*** will still return an ID token though.

So, long story short, I'm not sure you want to make this change for B2C ?

jennyf19: yes. i just talked about it w/ @jmprieur .I think we will document it for now. I'll let @jmprieur make that decision.

tests/B2CWebAppCallsWebApi/Client/Startup.cs
```diff
@@ -53,7 +53,9 @@ public void ConfigureServices(IServiceCollection services)
 
             // Token acquisition service based on MSAL.NET
             // and chosen token cache implementation
-            services.AddWebAppCallsProtectedWebApi(Configuration, "AzureAdB2C")
+            services.AddWebAppCallsProtectedWebApi(Configuration,
+                     initialScopes: new string[] { Configuration["TodoList:TodoListScope"] },
+                     configSectionName: "AzureAd")
```
jmprieur, line 58:
```suggestion
                     configSectionName: "AzureAdB2C")
```

src/Microsoft.Identity.Web/WebAppServiceCollectionExtensions.cs
```diff
@@ -90,6 +90,31 @@ public static class WebAppServiceCollectionExtensions
...
+        /// Add MSAL support to the Web App or Web API.
+        /// </summary>
+        /// <param name="services">Service collection to which to add authentication.</param>
+        /// <param name="configuration">Configuration.</param>
+        /// <param name="configSectionName">The name of the configuration section with the necessary
+        /// settings to initialize authentication options.</param>
+        /// <param name="openIdConnectScheme">Optional name for the open id connect authentication scheme
+        /// (by default OpenIdConnectDefaults.AuthenticationScheme). This can be specified when you want to support
+        /// several OpenIdConnect identity providers.</param>
+        /// <returns>The service collection for chaining.</returns>
```
jmprieur, line 103:
```suggestion
        /// <returns>The service collection for chaining.</returns>
        /// <remarks>This method cannot be used with Azure AD B2C as, with B2C an initial scope needs
        /// to be provided</remarks>
```

PR #159, opened 2020-05-13
Adding a sample using the Singleton Token acquisition and the GraphServiceClient

jmprieur:
This is a sample using the singleton TokenAcquisition. I think, however, that I did not parent the branch to Jenny when I created it ... bah ... you get the idea.


tests/WebAppCallsMicrosoftGraph/Pages/Index.cshtml.cs
```diff
@@ -0,0 +1,43 @@
...
+using Microsoft.Graph;
+
+namespace WebAppCallsMicrosoftGraph.Pages
+{
+    public class IndexModel : PageModel
+    {
+        private readonly ILogger<IndexModel> _logger;
+        private readonly GraphServiceClient _graphServiceClient;
+
+        public IndexModel(ILogger<IndexModel> logger, GraphServiceClient graphServiceClient)
```
jmprieur, line 18: This is the interesting part. We directly inject the GraphServiceClient

tests/WebAppCallsMicrosoftGraph/Startup.cs
```diff
@@ -0,0 +1,74 @@
...
+        public IConfiguration Configuration { get; }
+
+        // This method gets called by the runtime. Use this method to add services to the container.
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.AddSignIn(Configuration)
+                    .AddWebAppCallsProtectedWebApi(Configuration, new string[] { "user.read" })
+                    .AddInMemoryTokenCaches();
+
+            services.AddMicrosoftGraph(Configuration);
```
jmprieur, line 30: That's also the interesting thing.

jennyf19:
Looks fine. Will you add something similar to the external sample? I guess wait to merge until i merge the singleton.

tests/WebAppCallsMicrosoftGraph/MicrosoftGraphServiceExtensions.cs
```diff
@@ -0,0 +1,19 @@
...
+using Microsoft.Graph;
+using Microsoft.Identity.Web;
+
+namespace WebAppCallsMicrosoftGraph
+{
+    public static class MicrosoftGraphServiceExtensions
+    {
+        public static void AddMicrosoftGraph(this IServiceCollection services, IConfiguration configuration)
+        {
+            services.AddSingleton<GraphServiceClient, GraphServiceClient>(serviceProvider =>
```
pmaytak, line 12: `IGraphServiceClient`?
```suggestion
            services.AddSingleton<IGraphServiceClient, GraphServiceClient>(serviceProvider =>
```

tests/WebAppCallsMicrosoftGraph/Pages/Index.cshtml.cs
pmaytak, line 18: `IGraphServiceClient`?
```suggestion
        public IndexModel(ILogger<IndexModel> logger, IGraphServiceClient graphServiceClient)
```

tests/WebAppCallsMicrosoftGraph/Pages/Index.cshtml.cs
```diff
@@ -0,0 +1,43 @@
...
+        {
+            _logger = logger;
+            _graphServiceClient = graphServiceClient;
+        }
+
+        public async Task OnGet()
+        {
+            var x = await _graphServiceClient.Me.Request().GetAsync();
```
pmaytak, line 26: What does this `x` variable represent?

tests/WebAppCallsMicrosoftGraph/MicrosoftGraphServiceExtensions.cs
jmprieur: @pmaytak. The Graph SDK team does not want customer to use the interface but directly the GraphServiceClient. @darrelmiller can you please explain why?

tests/WebAppCallsMicrosoftGraph/Pages/Index.cshtml.cs
jmprieur: Same thing. @darrelmiller ?

jmprieur:
Yes @jennyf19 we'll update all the samples to use this pattern once we have agreed of the right one. @navyasric is also preparing a user study
@kalyankrishna1 @TiagoBrenck and @Shama-K were asked to do it.

tests/WebAppCallsMicrosoftGraph/MicrosoftGraphServiceExtensions.cs
pmaytak: Hmm. Interesting.

tests/WebAppCallsMicrosoftGraph/Startup.cs
```diff
@@ -0,0 +1,74 @@
...
+            Configuration = configuration;
+        }
+
+        public IConfiguration Configuration { get; }
+
+        // This method gets called by the runtime. Use this method to add services to the container.
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.AddSignIn(Configuration)
+                    .AddWebAppCallsProtectedWebApi(Configuration, new string[] { "user.read" })
```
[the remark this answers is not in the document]
navyasric: Per our conversation, the AddWebAppCallsProtectedWebApi can be used when calling other APIs than Graph. The additional scopes can be passed in the controller at the time of API call.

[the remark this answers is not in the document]
jmprieur: It also needs to be used when calling Graph for the moment (it sets the token acquisition service)

tests/WebAppCallsMicrosoftGraph/WebSignInCredential.cs
```diff
@@ -0,0 +1,23 @@
+﻿using Microsoft.Graph;
+using Microsoft.Identity.Web;
+using System.Net.Http;
+using System.Threading.Tasks;
+
+namespace WebAppCallsMicrosoftGraph
+{
+    public class WebSignInCredential : IAuthenticationProvider
```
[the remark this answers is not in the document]
jmprieur: @navyasric , as discussed
This would require Microsoft.Identity.Web to depend on the Graph SDK.
What do you think, @jennyf19 @pmaytak @henrik-me @bgavrilMS ?

I would think that have another NuGet package Microsoft.Identity.Web.MsGraph that would depend on both Microsoft.Identity.Web and Graph SDK.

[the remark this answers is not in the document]
Shama-K: I have a query regarding the use of National clouds through Graph SDK.
By default, the configured endpoint uses MS graph global service i.e., graph.microsoft.com. To use other national cloud services, do we need to pass the specific URL or do we have any method to handle this?

tests/WebAppCallsMicrosoftGraph/Startup.cs
navyasric: As discussed, we can consider providing initial scopes for MS Graph as follows: AddMicrosoftGraph(Configuration, new string[] { "user.read" }) and avoid the AddWebAppCallsProtectedWebApi.
AddWebAppCallsProtectedWebApi can be used when calling other APIs.

tests/WebAppCallsMicrosoftGraph/WebSignInCredential.cs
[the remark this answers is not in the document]
jmprieur: Actually this is a good point. I had thought of it yesterday, and then forgot. The AddMicrosoftGraph method should (optionally) take a base URL for Microsoft Graph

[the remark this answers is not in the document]
pmaytak: Although creating a whole separate Nuget package seems like too much, I think it makes sense in this case to put this class and also the Graph service extensions into a separate package. Not every user will want to use Graph, so it'll keep the main package clean, plus maybe we might end up putting more Graph related code in there anyway.

tests/WebAppCallsMicrosoftGraph/Startup.cs
```diff
@@ -0,0 +1,73 @@
...
+                    .AddInMemoryTokenCaches();
+
+            services.AddMicrosoftGraph(Configuration);
```
navyasric, line 33: @jmprieur , @jennyf19 We reviewed this proposal with the working team today and I'm channeling some feedback here. Since the AddMicrosoftGraph abstraction is introducing a dependency on the Graph SDK, we feel this implementation can stay out of the Microsoft.Identity.Web and the developer can use the AddSingleton pattern which is pretty standard in ASP.NET.

tests/WebAppCallsMicrosoftGraph/WebSignInCredential.cs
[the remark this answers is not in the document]
navyasric: @jmprieur We reviewed this proposal and discussed with the working team today and I'm channeling some feedback here. We would like to avoid Graph specific dependencies and separate package. We want to design for Microsoft.Identity.Web to work with multiple APIs and have that flexibility.

tests/WebAppCallsMicrosoftGraph/Startup.cs
```diff
@@ -0,0 +1,73 @@
...
+        {
+            Configuration = configuration;
+        }
+
+        public IConfiguration Configuration { get; }
+
+        // This method gets called by the runtime. Use this method to add services to the container.
+        public void ConfigureServices(IServiceCollection services)
+        {
+            services.AddSignIn(Configuration)
```
navyasric, line 29: @jmprieur , @jennyf19 Question about AddSignIn. Would the developer need to do this even when they configure EasyAuth and have signed in the user? 
If they need to configure sign in again with client creds, what is the benefit of EasyAuth? Does the integration still make sense?

tests/WebAppCallsMicrosoftGraph/Startup.cs
jennyf19: thanks @navyasric

tests/WebAppCallsMicrosoftGraph/WebSignInCredential.cs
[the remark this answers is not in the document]
bgavrilMS: I'm under the impression that there exists an URI endpoint that you can query to find out national resources - AAD authority, Graph endpoint and others. Using a web service to discover endpoints is better than a package because it's easier to update values.

tests/WebAppCallsMicrosoftGraph/Startup.cs
peterson4ts: I have a question too, regarding AddSignin and the other service collection extenions, there is a motive to we are not using PKCE? After all, its the actual recomentation from the last [BCP](https://tools.ietf.org/html/draft-ietf-oauth-security-topics-14).

jmprieur: IT's not yet implemented in MSAL.NET, @peterson4ts. When it is, we'll use it.
If you don't need to call a Web API, feel free to use it.

tests/WebAppCallsMicrosoftGraph/WebSignInCredential.cs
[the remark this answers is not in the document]
jmprieur: @bgavrilMS : I don't think it exists yet for 3P.

tests/WebAppCallsMicrosoftGraph/MicrosoftGraphServiceExtensions.cs
darrelmiller: IGraphServiceClient contains an entry point for each top level entityset or singleton.  Adding a new top level on the resource is not considered a breaking change and can happen fairly regularly.  Changing an interface is a breaking change.  This would be very undesirable behavior for a customer to update to a new version of the Graph SDK and find their implementations of IGraphServiceClient keep breaking.

Additionally, it is not a necessary interface.  There is no reason for creating an alternate implementation of GraphServiceClient other than for testing and there are better ways of doing testing than attempting to mock that interface.

tests/WebAppCallsMicrosoftGraph/Pages/Index.cshtml.cs
darrelmiller: Yup.

tests/WebAppCallsMicrosoftGraph/WebSignInCredential.cs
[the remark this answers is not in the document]
jmprieur: @pmaytak : I'm completely confused. I read one thing and its contrary? What is your recommendation? it's too much? or not?
> Although creating a whole separate Nuget package seems like too much, I think it makes sense in this case to put this class and also the Graph service extensions into a separate package. Not every user will want to use Graph, so it'll keep the main package clean, plus maybe we might end up putting more Graph related code in there anyway.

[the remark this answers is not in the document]
darrelmiller: @jmprieur   I wasn't sure of we could convince Azure.Identity to take a dependency on Microsoft.Identity.Web and then implement the WebSignInCredential there.  However, the .net 3.1 dependency is a problem to make that happen.
Alternatively, Microsoft.Identity.Web could take a dependency on Azure.Core for the TokenCredential class and implement the WebSignInCredential there.  Microsoft.Identity.Web would not need to depend on the Graph SDK.

@bgavrilMS You can get the Graph endpoint from the OpenID Connect discovery URL for some sovereign clouds.  If you look here https://login.microsoftonline.com/microsoft.onmicrosoft.com/v2.0/.well-known/openid-configuration  it is a property called msgraph_host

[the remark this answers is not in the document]
jmprieur: @@darrelmiller : What do we need to implement so that the AddScope can work with the provider?

tests/WebAppCallsMicrosoftGraph/Startup.cs
peterson4ts: > IT's not yet implemented in MSAL.NET, @peterson4ts. When it is, we'll use it.
> If you don't need to call a Web API, feel free to use it.

@jmprieur  Okay. I'm already using my implementation of AddSignIn with PKCE. But there is something that I still don't understand, for example, when using the AddWebAppCallsProtectedWebApi method it injects the ResponseType with the "code id_token" that would not allow me to use the PKCE with the challenge and still use the MSAL?

tests/WebAppCallsMicrosoftGraph/WebSignInCredential.cs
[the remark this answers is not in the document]
darrelmiller: @jmprieur Nothing.  The WithScope works on our Request object and it puts the scope information into an object in the HttpRequestMessage Properties bag.  The AuthHandler middleware pulls it out just before calling MSAL.

Also, I see that WebSignInCredential is implementing IAuthenticationProvider.  Now I understand why you thought you needed to take a dependency on the Graph SDK.    As we try and align with the Azure SDK, we are trying to make IAuthenticationProvider go away and use the TokenCredential class as the main interface.  If Microsoft.Identity.Web could take a dependency on Identity.Core then  we can use that as the "base class" for all credentials and we can make IAuthenticationProvider a Graph implementation detail.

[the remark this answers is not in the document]
jmprieur: I think we'd want to depend on Identity.Core indeed. We'll need it anyway to get certs from KeyVaults using managed identity.

tests/WebAppCallsMicrosoftGraph/MicrosoftGraphServiceExtensions.cs
jennyf19: thanks for the explanation @darrelmiller

jmprieur: reverted to use GraphServiceClient per @darrelmiller's guidance

PR #156, opened 2020-05-12
Removing a useless instruction and generating XML comments

jmprieur:
- Removing a useless instruction (as pointed to by @jg11jg in https://github.com/AzureAD/microsoft-identity-web/issues/154)
- Regenerating the XML comments.
- Fixing a couple of static analysis warnings

Issue #117, opened 2020-04-21
[Bug] AccessDenied Routing


mattosaurus:
**Which Version of Microsoft Identity Web are you using ?**
Microsoft.Identity.Web - v0.1.0 Preview
Microsoft.Identity.Web.UI - v0.1.0 Preview

**Where is the issue?**
* Web App
    * [x] Sign-in users
    * [ ] Sign-in users and call web APIs
* Web API
    * [ ] Protected web APIs (Validating tokens)
    * [ ] Protected web APIs (Validating scopes)
    * [ ] Protected web APIs call downstream web APIs
* Token cache serialization
     * [ ] In Memory caches
     * [ ] Session caches
     * [ ] Distributed caches

**Is this a new or existing app?**
c. This is a new app or experiment

**Repro**

Clone and run the below sample project.

https://github.com/Azure-Samples/active-directory-aspnetcore-webapp-openidconnect-v2/blob/master/5-WebApp-AuthZ/5-2-Groups

Log in with an unauthorized account.

**Expected behavior**
Unauthorized account should be redirected to /MicrosoftIdentity/Account/AccessDenied.

**Actual behavior**
Unauthorized account is redirected to /Account/AccessDenied which doesn't exist.

**Possible Solution**
The Microsoft.Identity.Web.UI AccountController exists in the MicrosoftIdentity area but there doesn't seem to be a way to provide the AccessDenied response with this route info.

I'd expect this to be configurable using something like the AccessDenied property like this but it doesn't seem to work.

```
services.Configure<OpenIdConnectOptions>(options =>
{
	// Use the groups claim for populating roles
	options.TokenValidationParameters.RoleClaimType = "groups";
	options.AccessDeniedPath = "/MicrosoftIdentity/Account/AccessDenied";
});
```

Possibly this is a redirect URL that should be set in the portal but it's not obvious where this is.

A workaround is to create an account controller with the required view but this goes against the point of Microsoft.Identity.Web.UI.

jmprieur:
@mattosaurus 

as a work around you could specify your page yourself doing the following:

public void ConfigureServices(IServiceCollection services)
{
    services.AddSignIn(Configuration);
    services.Configure<OpenIdConnectOptions>(OpenIdConnectDefaults.AuthenticationScheme,
        options => {
         	// Use the groups claim for populating roles
	       options.TokenValidationParameters.RoleClaimType = "groups";
[...]

PR #161, opened 2020-05-14
Setting Access Denied path to Identity.Web.UI Access Denied page

pmaytak:
There are two ways to implement the fix. I put related comments in PR. Please give opinion. More info in #117.

Also
- Fixed warning SA1400
- Updated unit tests related to #156, #154.

### DIFF
--- a/src/Microsoft.Identity.Web.UI/ServiceCollectionExtensions.cs
+++ b/src/Microsoft.Identity.Web.UI/ServiceCollectionExtensions.cs
@@ -1,6 +1,9 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System;
+using Microsoft.AspNetCore.Authentication.Cookies;
+using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.DependencyInjection;
 
 namespace Microsoft.Identity.Web.UI
@@ -14,12 +17,25 @@ namespace Microsoft.Identity.Web.UI
         /// <summary>
         /// Adds a controller and Razor pages for the accounts management.
         /// </summary>
-        /// <param name="builder">MVC Builder.</param>
+        /// <param name="builder">MVC builder.</param>
         public static IMvcBuilder AddMicrosoftIdentityUI(this IMvcBuilder builder)
         {
+            if (builder == null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
             builder.ConfigureApplicationPartManager(apm =>
             {
                 apm.FeatureProviders.Add(new MicrosoftIdentityAccountControllerFeatureProvider());
+            });
+
+            builder.Services.ConfigureAll<CookieAuthenticationOptions>(options =>
+            {
+                if (string.IsNullOrEmpty(options.AccessDeniedPath))
+                {
+                    options.AccessDeniedPath = new PathString("/MicrosoftIdentity/Account/AccessDenied");
+                }
             });
 
             return builder;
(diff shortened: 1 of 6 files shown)


src/Microsoft.Identity.Web/WebAppAuthenticationBuilderExtensions.cs
```diff
@@ -79,7 +80,13 @@ public static class WebAppAuthenticationBuilderExtensions
             var b2cOidcHandlers = new AzureADB2COpenIDConnectEventHandlers(openIdConnectScheme, microsoftIdentityOptions);
 
             builder.Services.AddSingleton<IOpenIdConnectMiddlewareDiagnostics, OpenIdConnectMiddlewareDiagnostics>();
-            builder.AddCookie(cookieScheme);
+            builder.AddCookie(cookieScheme, options =>
```
pmaytak, line 83: Option 1.
If the path is not set, we set it to default one of our choice.
**Pro** The setup is all in one place and the scheme we use is abstracted from the user.
**Con** If the user doesn't call `AddMicrosoftIdentityUI` and wants to use the default ASP.NET Core path (/Account/AccessDenied), we'll end up replacing it. 
**Con** We're kind of mixing non-UI logic with UI logic.

src/Microsoft.Identity.Web.UI/ServiceCollectionExtensions.cs
```diff
@@ -17,6 +18,11 @@ public static class ServiceCollectionExtensions
         /// <param name="builder">MVC Builder.</param>
         public static IMvcBuilder AddMicrosoftIdentityUI(this IMvcBuilder builder)
         {
+            if (builder == null)
+            {
+                throw new ArgumentNullException(nameof(builder));
+            }
+
```
pmaytak, line 25: Option 2.
Set the path inside `AddMicrosoftIdentityUI`. We need to set the path on the scheme with the same name that we assigned to `OpenIdConnectOptions.SignInScheme` in `AddSignIn`. That's why I added a scheme parameter. 
**Pro** UI related code is together.
**Con** It looks a bit weird passing a cookie scheme to this UI related method. 
**Con** If a user uses a non-default cookie scheme name, they would have to pass it here. A bit fragile.
```csharp
public static IMvcBuilder AddMicrosoftIdentityUI(this IMvcBuilder builder, string cookieScheme = CookieAuthenticationDefaults.AuthenticationScheme)
{
	if (builder == null)
	{
		throw new ArgumentNullException(nameof(builder));
	}

	builder.ConfigureApplicationPartManager(apm =>
	{
		apm.FeatureProviders.Add(new MicrosoftIdentityAccountControllerFeatureProvider());
	});

	builder.Services.Configure<CookieAuthenticationOptions>(cookieScheme, options =>
	{
		options.AccessDeniedPath = new PathString("/MicrosoftIdentity/Account/AccessDenied");
	});

	return builder;
}
```

jmprieur: That's my prefered option (even if I understand the fragility), because /MicrosoftIdentity/Account/AccessDenied is defined in Microsoft.Identity.Web.UI, which customers are not obliged to use.

We'll document the fragility


Is there a way for us to understand which CookieScheme would have been passed into AddSignIn ?

pmaytak: I think we could check the `Services` here if the cookie or OIDC options were added. But that won't work if `AddSignIn` is called after this method.

I also tried configuring without the scheme name, which should be the default, but that didn't work. 

I did have an idea to maybe use PostConfigure somehow, which is trigged after all of the options are added. Will have to explore.

pmaytak:
@jmprieur Agree that option 2 seems better. I'll update the code to use it and see if, perhaps, there's a way to improve it by using something like `PostConfigure`.

src/Microsoft.Identity.Web.UI/ServiceCollectionExtensions.cs
pmaytak: @jmprieur `ConfigureAll` works. No need to specify scheme name for this.

PR #163, opened 2020-05-14
Jennyf/audience

jennyf19:
leverage middleware to validate token audience


src/Microsoft.Identity.Web/Resource/RegisterValidAudience.cs
```diff
@@ -0,0 +1,75 @@
...
+            SecurityToken securityToken,
+            TokenValidationParameters validationParameters)
+        {
+            JwtSecurityToken token = securityToken as JwtSecurityToken;
+            string validAudience;
+
+            if (string.IsNullOrEmpty(validationParameters.ValidAudience) &&
+                validationParameters.ValidAudiences == null)
+            {
+                // handle v2.0 access token
```
jmprieur, line 49:
```suggestion
                // handle v2.0 access token or Azure AD B2C tokens (even if v1.0)
```

tests/Microsoft.Identity.Web.Test/Resource/RegisterValidAudienceTests.cs
```diff
@@ -0,0 +1,116 @@
...
+            JwtSecurityToken token = new JwtSecurityToken(null, null, claims);
+            var validationParams = new TokenValidationParameters();
+            RegisterValidAudience registerValidAudience = new RegisterValidAudience();
+            registerValidAudience.RegisterAudienceValidation(validationParams, options);
+            IEnumerable<string> validAudiences;
+
+            switch (testNumber)
+            {
+                case 1:
+                    validAudiences = new List<string> { expectedAudience };
```
jmprieur, line 74: Would you want to extract this line outside the switch ? (it seems it's the first line of each case?)

jennyf19: yes. forgot to clean that up.

src/Microsoft.Identity.Web/Resource/RegisterValidAudience.cs
```diff
@@ -0,0 +1,75 @@
...
+            validationParameters.AudienceValidator = ValidateAudience;
+        }
+
+        internal /*for test only*/ bool ValidateAudience(
+            IEnumerable<string> audiences,
+            SecurityToken securityToken,
+            TokenValidationParameters validationParameters)
+        {
+            JwtSecurityToken token = securityToken as JwtSecurityToken;
+            string validAudience;
```
pmaytak, line 44: Do we need this? Can we just pass string to `Contains`?

src/Microsoft.Identity.Web/Resource/RegisterValidAudience.cs
```diff
@@ -0,0 +1,75 @@
...
+                }
+
+                // handle v1.0 access token
+                if (token.Claims.Any(c => c.Type == Version && c.Value == V1))
+                {
+                    validAudience = $"api://{ClientId}";
+                    return audiences.Contains(validAudience);
+                }
+
+                throw new SecurityTokenValidationException("Token does not contain a valid audience type. ");
```
pmaytak, line 63:
```suggestion
                throw new SecurityTokenValidationException("Token does not contain a valid audience. ");
```

tests/Microsoft.Identity.Web.Test/Resource/RegisterValidAudienceTests.cs
```diff
@@ -0,0 +1,113 @@
...
+                  new Claim(Version, tokenVersion),
+                  new Claim(Audience, expectedAudience),
+            };
+
+            JwtSecurityToken token = new JwtSecurityToken(null, null, claims);
+            var validationParams = new TokenValidationParameters();
+            RegisterValidAudience registerValidAudience = new RegisterValidAudience();
+            registerValidAudience.RegisterAudienceValidation(validationParams, options);
+            IEnumerable<string> validAudiences;
+            validAudiences = new List<string> { expectedAudience };
```
pmaytak, line 70:
```suggestion
            IEnumerable<string> validAudiences = new List<string> { expectedAudience };
```

tests/Microsoft.Identity.Web.Test/Resource/RegisterValidAudienceTests.cs
```diff
@@ -0,0 +1,113 @@
...
+
+            switch (testNumber)
+            {
+                case 1:
```
pmaytak, line 74: Hmm at a minimum I would replace test number with a descriptive enum. 
And then maybe refactor to either use theory data or into more test methods to simplify, so it's easier to see at a glance what actual and expected conditions are tested.

src/Microsoft.Identity.Web/Resource/RegisterValidAudience.cs
jennyf19: I don't know what you mean.

---
In reply to: [425456868](https://github.com/AzureAD/microsoft-identity-web/pull/163#discussion_r425456868) [](ancestors = 425456868)

pmaytak: Meant like so in the below code:
```csharp
return audiences.Contains($"{ClientId}");

return audiences.Contains($"api://{ClientId}");
```

Issue #165, opened 2020-05-15
[Feature Request] Microsoft.Identity.Web supports certificates


jmprieur:
# Why? 

## Client Certificates 
Web apps and Web APIs are confidential client applications.

They can prove their identity to Azure AD by 3 means:
- client secrets
- client certificates
- client assertions

Today, Microsoft.Identity.Web enables developers to provide client secrets.
In addition to Client secret, we'd want Microsoft.Identity.Web to support client certificates. The constraints are the following:
- enable several ways of getting the certificate. You'd provide a description on how to get the certificate.
  - from the certificate store (Windows) and a thumbprint ("440A5BE6C4BE2FF02A0ADBED1AAA43D6CF12E269")
  - from the certificate store (Windows) and a distinguished name ("CN=TestCert")
  - from a path on the disk (probably only for debugging locally)
  - directly from a base64 representation of the certificate
  - from a KeyVault address.
- getting the certificate just in time, rather than paying the startup cost. For instance for a web app that signs in a user, do not load the certificate until an access token is needed to call a Web API.
- when the certificate is stored in KeyVault, leverage Managed identity (probably though the Azure SDK for .NET)
- help you rotating your certificates but letting you provide several (2) certificates

## Decrypt certificates

It's a different topic, but still touching on certificates: Web APIs can request token encryption (for privacy reasons). This is even compulsory for 1P Web APIs that access MSA identities. We also want to let you pass decryption certificates with the same flexibility as for client certificates.

## Proposal 

The current proposal is to have the following options to specify both client certificates and decrypt certificates from the configuration file.

```Json
{
  "AzureAd": {
    "Instance": "https://login.microsoftonline.com/",
    "Domain": "msidentitysamplestesting.onmicrosoft.com",
    "TenantId": "7f58f645-c190-4ce5-9de4-e2b7acd2a6ab",
    "ClientId": "86699d80-dd21-476a-bcd1-7c1a3d471f75",
    "CallbackPath": "/signin-oidc",
    "SignedOutCallbackPath ": "/signout-callback-oidc",

    // To call an API
    "ClientSecret": "[Copy the client secret added to the app from the Azure portal]",

    // Exclusive with "ClientSecret"
    "ClientCertificates": [
      {
        "CertificateFromThumbprint": {
          "StoreLocation": "CurrentUser", // Optional, default = CurrentUser
          "StoreName": "My",              // Optional, default = My
[...]

PR #176, opened 2020-05-26
Create certificate support proposal.md

bgavrilMS:


bgavrilMS:
Copied from https://github.com/AzureAD/microsoft-identity-web/issues/165 but made into a PR so that we can add comments

docs/certificate support proposal.md
```diff
@@ -0,0 +1,229 @@
...
+      },
+      {
+        "CertificateFromPath": "C:\Users\myname\Documents\TestJWE.pfx",
+        "CertificatePassword": "TestJWE"
+      },
+      {
+        "CertificateFromBase64Encoded": "Base64Encoded"
+      },
+      {
+        "CertificateFromKeyVault": "https://vaultname.vault.azure.net/certificates/certificateid/3fb1c62f74b844b0a2d9f1a3d289648d"
```
bgavrilMS, line 68: I wonder if the naming here is a bit misleading if you expect to use managed identity here? AFAIK, you can only use this if the code is running in an Azure VM or an Azure Function?

docs/certificate support proposal.md
```diff
@@ -0,0 +1,229 @@
+# Why? 
+
+## Client Certificates 
+Web apps and Web APIs are confidential client applications.
+
+They can prove their identity to Azure AD by 3 means:
+- client secrets
+- client certificates
+- client assertions
```
bgavrilMS, line 9: Would you want to integrate with Wilson and expose configuring a `SigningCredential` here? I think this is similar, if not better, than `client assertions` option and would open up more scenarios (not that these 2 are mutually exclusive)

docs/certificate support proposal.md
```diff
@@ -0,0 +1,229 @@
...
+
+Today, Microsoft.Identity.Web enables developers to provide client secrets.
+In addition to Client secret, we'd want Microsoft.Identity.Web to support client certificates. The constraints are the following:
+- enable several ways of getting the certificate. You'd provide a description on how to get the certificate.
```
bgavrilMS, line 13: Can the developers programatically configure an `X509Certificate` object?

docs/certificate support proposal.md
```diff
@@ -0,0 +1,229 @@
...
+  - from the certificate store (Windows) and a thumbprint ("440A5BE6C4BE2FF02A0ADBED1AAA43D6CF12E269")
+  - from the certificate store (Windows) and a distinguished name ("CN=TestCert")
```
bgavrilMS, line 15: These will possibly also work from Linux apparently ... https://devblogs.microsoft.com/aspnet/configuring-https-in-asp-net-core-across-different-platforms/

docs/certificate support proposal.md
```diff
@@ -0,0 +1,229 @@
...
+- we would start by only supporting certificates obtained from KeyVault, as this is the use case that most customers are after. But we'd propose a design that would work with other source of certificates (local file, base64 encoded are the ones I've seen the most)
+
+Proposed sequence:
+1. Experiment with the KeyVault SDK part of the Azure SDK. This has the advantage of getting the secrets from KeyVault both locally, and when the Web app or Web API is deployed, as it leverages Managed identity. See https://docs.microsoft.com/en-us/azure/key-vault/general/tutorial-net-create-vault-azure-web-app#update-the-code (this one is for client secrets, we need to find the equivalent with certificates, for instance by looking at https://docs.microsoft.com/en-us/dotnet/api/overview/azure/security.keyvault.certificates-readme?view=azure-dotnet#retrieve-a-certificate). We could use, for the client the DefaultAzureCredentials (which also work well in the dev environment)
+
+   This means we would now leverage the following NuGet packages: **Azure.Identity** and **Azure.Security.KeyVault.Certificates**. This seems fine but we need to check if there is a negative impact.
```
bgavrilMS, line 125: As part of the experiment, we should also understand if there is significant logic related to MSI in those assemblies. If not, perhaps we can avoid the dependencies.

docs/certificate support proposal.md
```diff
@@ -0,0 +1,229 @@
...
+        /// <item>If <see cref="SourceType"/> equals <see cref="CertificateSource.Base64Encoded"/>, then
+        /// this value is not used</item>
+        /// <item>If <see cref="SourceType"/> equals <see cref="CertificateSource.Path"/>, then
+        /// this value is the path on disk where to find the certificate</item>
+        /// <item>If <see cref="SourceType"/> equals <see cref="CertificateSource.StoreWithDistinguihedName"/>,
+        /// or <see cref="CertificateSource.StoreWithThumbprint"/>, then
+        /// this value is the path to the certificate in the cert store, for instance <c>CurrentUser/My</c></item>
+        /// </list>
+        /// </summary>
+        public string Container { get; set; }
```
bgavrilMS, line 194: Personally I don't like this "weakly typed" API. Why not use a builder pattern that resembles the json document you described

```csharp
Certificate.LoadFromFile(string path);
Certificate.LoadFromKeyVault(...)
```

jmprieur: @bgavrilMS : so you like the initial json? I think it's a bit confusing.

docs/certificate support proposal.md
jmprieur: Actually no: Managed identities also work for App services, plenty of Azure services (increasing) and even locally on the dev box (using the shared token cache)

docs/certificate support proposal.md
jmprieur: Do you know if SigningCredential also works with MSI?

bgavrilMS: It does not as far as I know, but @GeoK would know best

GeoK: SigningCredentials works with any SecurityKey, including [ManagedKeyVaultSecurityKey ](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/Microsoft.IdentityModel.ManagedKeyVaultSecurityKey/ManagedKeyVaultSecurityKey.cs) that supports MSI.
SigningCredentials object has to be constructed with a KeyVaultCryptoProvider as CustomCryptoProvider, but that can be simplified by introducing a new class, e.g. _ManagedKeyVaultSigningCredentials_.

This would be a current setup:
```
var sc = new SigningCredentials(new ManagedKeyVaultSecurityKey([keyId]), [alg])
{
    CryptoProviderFactory = new CryptoProviderFactory
    {
        CustomCryptoProvider = new KeyVaultCryptoProvider()
    }
}
```

jmprieur: Thanks @GeoK : 
- Does it use the new Azure SDK `TokenCredential`?
- Is MSI working with the developer credentials in the shared token cache (between VS, Azure Rm PowerShell and Azure CLI) ?

docs/certificate support proposal.md
jmprieur: Yes, that's a good idea;

docs/certificate support proposal.md
GeoK: I'm not familiar with TokenCredential. _ManagedKeyVaultSecurityKey_ leverages `Microsoft.Azure.KeyVault` and `Microsoft.Azure.Services.AppAuthentication`. Would these libraries have the business logic that you are looking for?

SigningCredentials is usually used in token signing scenarios where algorithm needs to be specified.

jmprieur: @GeoK :  I believe that the new Azure SDK uses Azure.Security.KeyVault (and Azure.Identity). Underneath, it uses MSAL.NET.

Issue #15, opened 2020-02-18
Implement C# 8.0 nullable standard


jennyf19:
from @sujayvsarma and copied from https://github.com/Azure-Samples/active-directory-aspnetcore-webapp-openidconnect-v2/issues/249

Even if the project is not C# 8.0, consider implementing the `nullable` standard by suffixing `?` to all parameters and returns that may be `NULL`. 

This will be a breaking changing internal to the library & to consuming code. Therefore, consider publishing the changes only during a major cycle.

bgavrilMS:
That would be very cool indeed and the best time to do it is before releasing. Happy to help with this one, should not be too hard. 

This blog describes the feature better https://devblogs.microsoft.com/dotnet/embracing-nullable-reference-types/

To quote from the blog: 

`Libraries can accurately describe the null behavior of the APIs, in a way that is checkable in client code at the call site. This raises expressiveness on API boundaries so that everyone can get a handle on the safe propagation and dereferencing of nulls. Nobody likes null reference exceptions or argument-null exceptions! This helps you write the code right the first time, and avoid the sources of those exceptions before you even compile and run the code.`

jennyf19:
@bgavrilMS or @pmaytak can you put this in the correct milestone please? not a breaking change. thx

jennyf19:
Included in [0.2.0-preview release](https://github.com/AzureAD/microsoft-identity-web/releases/tag/0.2.0-preview)

PR #183, opened 2020-06-01
Nullable partial work

bgavrilMS:
This is most of the work needed to conform to the nullable reference types. This makes the API more expressive, by showing which input / output types can be null or not. 

https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references

This is NOT a breaking change. 

Please read through my PR comments and decide if this is an avenue you wish to pursue. To sum up:

PROs: 

- better semantics (you know when to pass null or not)
- will help us discover bugs
- much less chances of NULL refs
- better for customers who also use this feature, as it will help the compiler analyze NULLs as well

CONs: 

- you still have to program defensively
- about ~2h more of work (I am happy to finish it if)
- one more things to understand

### DIFF
--- a/src/Microsoft.Identity.Web/AccountExtensions.cs
+++ b/src/Microsoft.Identity.Web/AccountExtensions.cs
@@ -17,7 +17,7 @@ namespace Microsoft.Identity.Web
         /// </summary>
         /// <param name="account">The IAccount instance.</param>
         /// <returns>A <see cref="ClaimsPrincipal"/> built from IAccount.</returns>
-        public static ClaimsPrincipal ToClaimsPrincipal(this IAccount account)
+        public static ClaimsPrincipal? ToClaimsPrincipal(this IAccount? account)
         {
             if (account != null)
             {
--- a/src/Microsoft.Identity.Web/AuthorizeForScopesAttribute.cs
+++ b/src/Microsoft.Identity.Web/AuthorizeForScopesAttribute.cs
@@ -32,12 +32,12 @@ namespace Microsoft.Identity.Web
         /// <summary>
         /// Scopes to request.
         /// </summary>
-        public string[] Scopes { get; set; }
+        public string[]? Scopes { get; set; }
 
         /// <summary>
         /// Key section on the configuration file that holds the scope value.
         /// </summary>
-        public string ScopeKeySection { get; set; }
+        public string? ScopeKeySection { get; set; }
 
         /// <summary>
         /// Handles the MsalUiRequiredException.
@@ -45,51 +45,47 @@ namespace Microsoft.Identity.Web
         /// <param name="context">Context provided by ASP.NET Core.</param>
         public override void OnException(ExceptionContext context)
         {
-            // Do not re-use the attribute param Scopes. For more info: https://github.com/Azure-Samples/active-directory-aspnetcore-webapp-openidconnect-v2/issues/273
-            string[] incrementalConsentScopes = Array.Empty<string>();
-            MsalUiRequiredException msalUiRequiredException = context.Exception as MsalUiRequiredException;
+            MsalUiRequiredException? msalUiRequiredException =
+                (context.Exception as MsalUiRequiredException) ??
+                (context.Exception?.InnerException as MsalUiRequiredException);
 
-            if (msalUiRequiredException == null)
+            if (msalUiRequiredException != null &&
+                CanBeSolvedByReSignInOfUser(msalUiRequiredException))
             {
-                msalUiRequiredException = context.Exception?.InnerException as MsalUiRequiredException;
-            }
-
-            if (msalUiRequiredException != null)
-            {
-                if (CanBeSolvedByReSignInOfUser(msalUiRequiredException))
+                // the users cannot provide both scopes and ScopeKeySection at the same time
+                if (!string.IsNullOrWhiteSpace(ScopeKeySection) && Scopes != null && Scopes.Length > 0)
                 {
-                    // the users cannot provide both scopes and ScopeKeySection at the same time
-                    if (!string.IsNullOrWhiteSpace(ScopeKeySection) && Scopes != null && Scopes.Length > 0)
-                    {
-                        throw new InvalidOperationException($"Either provide the '{nameof(ScopeKeySection)}' or the '{nameof(Scopes)}' to the 'AuthorizeForScopes'.");
-                    }
-
-                    // If the user wishes us to pick the Scopes from a particular config setting.
-                    if (!string.IsNullOrWhiteSpace(ScopeKeySection))
-                    {
-                        // Load the injected IConfiguration
-                        IConfiguration configuration = context.HttpContext.RequestServices.GetRequiredService<IConfiguration>();
-
-                        if (configuration == null)
-                        {
-                            throw new InvalidOperationException($"The {nameof(ScopeKeySection)} is provided but the IConfiguration instance is not present in the services collection");
-                        }
-
-                        incrementalConsentScopes = new string[] { configuration.GetValue<string>(ScopeKeySection) };
-
-                        if (Scopes != null && Scopes.Length > 0 && incrementalConsentScopes != null && incrementalConsentScopes.Length > 0)
-                        {
-                            throw new InvalidOperationException("no scopes provided in scopes...");
-                        }
-                    }
-                    else
-                    {
-                        incrementalConsentScopes = Scopes;
-                    }
-
-                    var properties = BuildAuthenticationPropertiesForIncrementalConsent(incrementalConsentScopes, msalUiRequiredException, context.HttpContext);
-                    context.Result = new ChallengeResult(properties);
+                    throw new InvalidOperationException($"Either provide the '{nameof(ScopeKeySection)}' or the '{nameof(Scopes)}' to the 'AuthorizeForScopes'.");
                 }
+
+                // Do not re-use the attribute param Scopes. For more info: https://github.com/Azure-Samples/active-directory-aspnetcore-webapp-openidconnect-v2/issues/273
+                string[]? incrementalConsentScopes;
+
+                // If the user wishes us to pick the Scopes from a particular config setting.
+                if (!string.IsNullOrWhiteSpace(ScopeKeySection))
+                {
+                    // Load the injected IConfiguration
+                    IConfiguration configuration = context.HttpContext.RequestServices.GetRequiredService<IConfiguration>();
+
+                    if (configuration == null)
+                    {
+                        throw new InvalidOperationException($"The {nameof(ScopeKeySection)} is provided but the IConfiguration instance is not present in the services collection");
+                    }
+
+                    incrementalConsentScopes = new string[] { configuration.GetValue<string>(ScopeKeySection) };
+
+                    if (Scopes != null && Scopes.Length > 0 && incrementalConsentScopes.Length > 0)
+                    {
+                        throw new InvalidOperationException("no scopes provided in scopes...");
+                    }
+                }
+                else
+                {
+                    incrementalConsentScopes = Scopes;
+                }
+
+                var properties = BuildAuthenticationPropertiesForIncrementalConsent(incrementalConsentScopes, msalUiRequiredException, context.HttpContext);
+                context.Result = new ChallengeResult(properties);
             }
 
             base.OnException(context);
@@ -114,7 +110,7 @@ namespace Microsoft.Identity.Web
         /// <param name="context">current HTTP context in the pipeline.</param>
         /// <returns>AuthenticationProperties.</returns>
         private AuthenticationProperties BuildAuthenticationPropertiesForIncrementalConsent(
-            string[] scopes,
+            string[]? scopes,
             MsalUiRequiredException ex,
             HttpContext context)
         {
@@ -128,9 +124,13 @@ namespace Microsoft.Identity.Web
                  OidcConstants.ScopeProfile,
             };
 
+            // TODO: scopes can actually be null here - how do we treat this case? 
+            // if this is not allowed then we should throw before calling this method 
+            // if this is allowed, we need to avoid the null ref below
+
             properties.SetParameter<ICollection<string>>(
                 OpenIdConnectParameterNames.Scope,
-                scopes.Union(additionalBuiltInScopes).ToList());
+                scopes.Union(additionalBuiltInScopes).ToList()); // potential null ref exception
 
             // Attempts to set the login_hint to avoid the logged-in user to be presented with an account selection dialog
             var loginHint = context.User.GetLoginHint();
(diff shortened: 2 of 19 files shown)


src/Microsoft.Identity.Web/AccountExtensions.cs
```diff
@@ -17,7 +17,7 @@ public static class AccountExtensions
         /// </summary>
         /// <param name="account">The IAccount instance.</param>
         /// <returns>A <see cref="ClaimsPrincipal"/> built from IAccount.</returns>
-        public static ClaimsPrincipal ToClaimsPrincipal(this IAccount account)
+        public static ClaimsPrincipal? ToClaimsPrincipal(this IAccount? account)
```
bgavrilMS, line 20: Public API changes are backwards compatible.

src/Microsoft.Identity.Web/ClaimsPrincipalExtensions.cs
```diff
@@ -16,7 +16,7 @@ public static class ClaimsPrincipalExtensions
         /// </summary>
         /// <param name="claimsPrincipal">Claims principal.</param>
         /// <returns>A string corresponding to an account identifier as defined in <see cref="Microsoft.Identity.Client.AccountId.Identifier"/>.</returns>
-        public static string GetMsalAccountId(this ClaimsPrincipal claimsPrincipal)
+        public static string? GetMsalAccountId(this ClaimsPrincipal claimsPrincipal)
```
bgavrilMS, line 19: If a consuming application also enforces the "nullable reference" feature, then they will get **warnings** when they try to pass in a NULL, for example a null ClaimsPrincinpal here. 

If the consuming application does not enforce this feature, they will not get warnings.

src/Microsoft.Identity.Web/ClaimsPrincipalExtensions.cs
```diff
@@ -16,7 +16,7 @@ public static class ClaimsPrincipalExtensions
...
         {
             if (claimsPrincipal == null)
```
bgavrilMS, line 21: From a code perspective, we still need to make null checks and program defensively, because NULLs can still be passed everywhere.

src/Microsoft.Identity.Web/Microsoft.Identity.Web.csproj
```diff
@@ -1,6 +1,14 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
-  <PropertyGroup>
 
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <SignAssembly>true</SignAssembly>
+    <AssemblyOriginatorKeyFile>../../build/MSAL.snk</AssemblyOriginatorKeyFile>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <Nullable>enable</Nullable>
```
bgavrilMS, line 8: This is how the feature is enabled. It needs C# 8, which is available with .net core 3.1 or with .net class 4.smth (4.7.2?)

src/Microsoft.Identity.Web/Resource/RegisterValidAudience.cs
```diff
@@ -61,7 +61,12 @@ internal class RegisterValidAudience
             SecurityToken securityToken,
             TokenValidationParameters validationParameters)
         {
-            JwtSecurityToken token = securityToken as JwtSecurityToken;
+            JwtSecurityToken? token = securityToken as JwtSecurityToken;
+            if (token == null)
```
bgavrilMS, line 65: This is the kind of bugs you discover when you turn this feature on. THe compiler does NULL analysis and points out areas of the code that are potentially unsafe.

src/Microsoft.Identity.Web/Resource/RegisterValidAudience.cs
```diff
@@ -61,7 +61,12 @@ internal class RegisterValidAudience
...
+            {
+                throw new SecurityTokenValidationException("Token is not JWT token.");
```
bgavrilMS, line 67: Hope this a good exception here.

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -446,9 +444,9 @@ private async Task<IConfidentialClientApplication> BuildConfidentialClientApplic
         /// <param name="tenant"></param>
         private async Task<string> GetAccessTokenOnBehalfOfUserFromCacheAsync(
```
bgavrilMS, line 445: You can see how the methods become more expressive - you now know which input params can be null and which ones can't.

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -446,9 +444,9 @@ private async Task<IConfidentialClientApplication> BuildConfidentialClientApplic
...
             IConfidentialClientApplication application,
-            IAccount account,
+            IAccount? account,
             IEnumerable<string> scopes,
-            string tenant)
+            string? tenant)
         {
             if (scopes == null)
```
bgavrilMS, line 451: But you still have to code defensively...

src/Microsoft.Identity.Web/AuthorizeForScopesAttribute.cs
```diff
@@ -32,64 +32,60 @@ public class AuthorizeForScopesAttribute : ExceptionFilterAttribute
...
         /// Key section on the configuration file that holds the scope value.
         /// </summary>
-        public string ScopeKeySection { get; set; }
+        public string? ScopeKeySection { get; set; }
 
         /// <summary>
         /// Handles the MsalUiRequiredException.
         /// </summary>
         /// <param name="context">Context provided by ASP.NET Core.</param>
         public override void OnException(ExceptionContext context)
```
bgavrilMS, line 46: I rewrote this method a bit, but all changes are non-breaking (done with analyzers)

src/Microsoft.Identity.Web/AuthorizeForScopesAttribute.cs
```diff
@@ -128,9 +124,13 @@ private bool CanBeSolvedByReSignInOfUser(MsalUiRequiredException ex)
                  OidcConstants.ScopeProfile,
             };
 
+            // TODO: scopes can actually be null here - how do we treat this case?
```
bgavrilMS, line 127: This is the kind of bugs that can be discovered. In this particular case I am not sure how to handle it - please advise.

pmaytak:
Thanks @bgavrilMS. Is there an official Microsoft doc that talks about this pattern, plus when to use it or not, etc.?

jennyf19:
Thanks @bgavrilMS I will try to have a look later today. I think this is definitely worth exploring and thanks for you time thus far.

jennyf19:
@pmaytak here's a [doc on it](https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references)

bgavrilMS:
@pmaytak - sorry I didn't add the official guidance - https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references

jmprieur:
Thanks @bgavrilMS for starting this conversation.
This is BTW related to GitHub issue https://github.com/AzureAD/microsoft-identity-web/issues/15 requested by a customer.

I like that the API shows what can be null and not, and that's not a breaking change (except a renaming you've done)
I also like that this enables to find out possible null reference exceptions

Did you notice that the branch does not build?

Do you think customers will understand the subtlety? 
@jennyf19 @pmaytak @henrik-me : how do you feel?

bgavrilMS:
@jmprieur - I broke a unit test related to json handling...

src/Microsoft.Identity.Web/AuthorizeForScopesAttribute.cs
jennyf19: yes...good point. This is for incremental consent, so we technically should only be here if we have scopes, because that's the point, that additional scopes are required. So I think in `OnException()`, we should make sure we have something. @jmprieur thoughts?

---
In reply to: [433215274](https://github.com/AzureAD/microsoft-identity-web/pull/183#discussion_r433215274) [](ancestors = 433215274)

src/Microsoft.Identity.Web/Resource/RegisterValidAudience.cs
jennyf19: LOL.

---
In reply to: [433752648](https://github.com/AzureAD/microsoft-identity-web/pull/183#discussion_r433752648) [](ancestors = 433752648)

jennyf19:
Looks fine to me, bogdan. thanks for all the effort. You still have a few hours of this? thanks again.

---
In reply to: [637603887](https://github.com/AzureAD/microsoft-identity-web/pull/183#issuecomment-637603887) [](ancestors = 637603887)

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -199,7 +197,7 @@ internal class TokenAcquisition : ITokenAcquisition, ITokenAcquisitionInternal
         /// OpenIdConnectOptions.Events.OnAuthorizationCodeReceived.</remarks>
         public async Task<string> GetAccessTokenForUserAsync(
             IEnumerable<string> scopes,
-            string tenant = null)
+            string? tenantId = null)
```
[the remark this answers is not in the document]
bgavrilMS: Ah, you mean the rename? Didn't realize, but you are right. I have an analyzer that flaged it up because the name of the param on the interface != name of the param on the class. That analyzer is SQ btw...

src/Microsoft.Identity.Web/AccountExtensions.cs
pmaytak: Here and in similar methods, we can also use [attributes](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/nullable-analysis) like `[return: NotNullIfNotNull("account")]

src/Microsoft.Identity.Web/ClaimsPrincipalExtensions.cs
pmaytak: How do we decide if we should change the parameter to nullable or keep it non-nullable and throw the ArgumentNullException? 

In this case, a null claimsPrincipal can return a null ID value and non-null claimsPrincipal returns non-null ID value.

src/Microsoft.Identity.Web/ClientInfo.cs
```diff
@@ -25,19 +25,11 @@ public static ClientInfo CreateFromJson(string clientInfo)
                 throw new ArgumentNullException(nameof(clientInfo), $"client info returned from the server is null");
             }
 
-            return DeserializeFromJson<ClientInfo>(Base64UrlHelpers.DecodeToBytes(clientInfo));
-        }
-
-        internal static T DeserializeFromJson<T>(byte[] jsonByteArray)
```
pmaytak, line 31: I would keep `DeserializeFromJson`. I think you can also leave the generic parameter and add `where T : notnull`.

src/Microsoft.Identity.Web/HttpContextExtensions.cs
```diff
@@ -24,7 +24,7 @@ internal static void StoreTokenUsedToCallWebAPI(this HttpContext httpContext, Jw
         /// </summary>
         /// <param name="httpContext">Http context associated with the current request.</param>
         /// <returns><see cref="JwtSecurityToken"/> used to call the Web API.</returns>
-        internal static JwtSecurityToken GetTokenUsedToCallWebAPI(this HttpContext httpContext)
+        internal static JwtSecurityToken? GetTokenUsedToCallWebAPI(this HttpContext httpContext)
```
pmaytak, line 27: In this method (and similar ones), we expect `httpContext` to have a value here. However, if null is passed in, we get an NPE. So we either need to return null or throw.

src/Microsoft.Identity.Web/InstanceDiscovery/Metadata.cs
```diff
@@ -15,19 +15,19 @@ internal class Metadata
         /// Preferred alias.
         /// </summary>
         [JsonProperty(PropertyName = "preferred_network")]
-        public string PreferredNetwork { get; set; }
+        public string? PreferredNetwork { get; set; }
```
pmaytak, line 18: This case is similar to [Late-initialized properties, Data Transfer Objects and nullability](https://docs.microsoft.com/en-us/dotnet/csharp/nullable-migration-strategies#late-initialized-properties-data-transfer-objects-and-nullability). What is our intent - value is required for all these properties or optional?

src/Microsoft.Identity.Web/ClaimsPrincipalExtensions.cs
bgavrilMS: At line 23 we throw an `ArgumentNullEx` if the claimsPrincipal is null, we don't return null.

src/Microsoft.Identity.Web/ClientInfo.cs
```diff
@@ -25,19 +25,11 @@ public static ClientInfo CreateFromJson(string clientInfo)
...
-                return default;
-            }
+            var jsonByteArray = Base64UrlHelpers.DecodeToBytes(clientInfo);
 
-            using var stream = new MemoryStream(jsonByteArray);
-            using var reader = new StreamReader(stream, Encoding.UTF8);
-            return (T)JsonSerializer.Create().Deserialize(reader, typeof(T));
+            using MemoryStream stream = new MemoryStream(jsonByteArray);
+            using StreamReader reader = new StreamReader(stream, Encoding.UTF8);
+            return (ClientInfo)JsonSerializer.Create().Deserialize(reader, typeof(ClientInfo));
```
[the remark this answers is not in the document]
bgavrilMS: JSON deserialization can always return NULL (since you don't know what you deserialize). But a nullable value type is not the same as a nullable reference type. So general purpose serialization helpers that just take a "T" won't work. I think @pmaytak 's suggestions (`where T : notnull`) might do it, I didn't research this enough.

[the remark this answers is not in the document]
bgavrilMS: But from a code perspective, why have a general purpose deserialization method in a class that only deserializes ClientInfo? Makes the code hard to read...

src/Microsoft.Identity.Web/HttpContextExtensions.cs
bgavrilMS: For public APIs, we should always check for NULLs and throw NullArgumentException. 
For internal APIs like this - use your best judgement - I think in most cases we can add a null check, but this is where you know best. If you can't decide, my advice would be throw, i.e. be stingy with returning NULLs.

src/Microsoft.Identity.Web/InstanceDiscovery/Metadata.cs
bgavrilMS: Very good point, I missed this one!

pmaytak: Good thing about this work and PR is that it starts the discussions about the intent of our code. 👍

jennyf19:
@bgavrilMS what's the status on this? just concerned the longer it sits the more out of sync it will get w/master. and the more work ultimately to address merge conflicts.

bgavrilMS:
Folks, I apologise but this is probably more work that I can handle for the next few weeks, especially as @pmaytak makes an excellent point about how to handle JSON + nullables, which requires more investment.

jennyf19:
Thanks @bgavrilMS ... @pmaytak is going to finish the work

PR #188, opened 2020-06-04
Replaced Newtonsoft.Json with System.Text.Json in projects. Plus minor misc fixes.

pmaytak:
Replaced Newtonsoft.Json with System.Text.Json in projects (including samples). Plus minor spelling, warning fixes.

**Note:** Newtonsoft.Json by default does case-**insensitive** deserializaton, while System.Text.Json is case-**sensitive** by default to improve performance. In order to make it case-insensitive, we have to pass it options object to the deserializer.

Related resources:
[Try the new System.Text.Json APIs](https://devblogs.microsoft.com/dotnet/try-the-new-system-text-json-apis/)
[How to serialize and deserialize (marshal and unmarshal) JSON in .NET](https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-how-to)
[How to migrate from Newtonsoft.Json to System.Text.Json](https://docs.microsoft.com/en-us/dotnet/standard/serialization/system-text-json-migrate-from-newtonsoft-how-to)

### DIFF
--- a/tests/B2CWebAppCallsWebApi/TodoListService/Models/TodoItem.cs
+++ b/tests/B2CWebAppCallsWebApi/TodoListService/Models/TodoItem.cs
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System.Text.Json.Serialization;
+
 namespace TodoListService.Models
 {
     public class Todo
--- a/tests/WebAppCallsWebApiCallsGraph/TodoListService/Models/TodoItem.cs
+++ b/tests/WebAppCallsWebApiCallsGraph/TodoListService/Models/TodoItem.cs
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System.Text.Json.Serialization;
+
 namespace TodoListService.Models
 {
     public class Todo
--- a/.editorconfig
+++ b/.editorconfig
@@ -476,9 +476,6 @@ dotnet_diagnostic.CA1303.severity = none
 # SA1004: Documentation lines should begin with single space
 dotnet_diagnostic.SA1004.severity = none
 
-# SA1027: Use tabs correctly
-dotnet_diagnostic.SA1027.severity = none
-
 # SA1118: Parameter should not span multiple lines
 dotnet_diagnostic.SA1118.severity = none
 
@@ -509,17 +506,8 @@ dotnet_diagnostic.SA1308.severity = none
 # SA1311: Static readonly fields should begin with upper-case letter
 dotnet_diagnostic.SA1311.severity = none
 
-# SA1312: Variable names should begin with lower-case letter
-dotnet_diagnostic.SA1312.severity = none
-
-# SA1313: Parameter names should begin with lower-case letter
-dotnet_diagnostic.SA1313.severity = none
-
 # SA1401: Fields should be private
 dotnet_diagnostic.SA1401.severity = none
-
-# SA1502: Element should not be on a single line
-dotnet_diagnostic.SA1502.severity = none
 
 # SA1512: Single-line comments should not be followed by blank line
 dotnet_diagnostic.SA1512.severity = none
--- a/src/Microsoft.Identity.Web/ClientInfo.cs
+++ b/src/Microsoft.Identity.Web/ClientInfo.cs
@@ -2,20 +2,18 @@
 // Licensed under the MIT License.
 
 using System;
-using System.IO;
 using System.Runtime.Serialization;
-using System.Text;
-using Newtonsoft.Json;
+using System.Text.Json;
+using System.Text.Json.Serialization;
 
 namespace Microsoft.Identity.Web
 {
-    [DataContract]
     internal class ClientInfo
     {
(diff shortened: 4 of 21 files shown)


tests/B2CWebAppCallsWebApi/TodoListService/Models/TodoItem.cs
```diff
@@ -1,14 +1,19 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System.Text.Json.Serialization;
+
 namespace TodoListService.Models
 {
     public class Todo
     {
+        [JsonPropertyName("id")]
```
jmprieur, line 10: Do the sample need to be changed?

tests/WebAppCallsWebApiCallsGraph/TodoListService/Models/TodoItem.cs
```diff
@@ -1,14 +1,19 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System.Text.Json.Serialization;
+
 namespace TodoListService.Models
 {
     public class Todo
     {
+        [JsonPropertyName("id")]
```
jmprieur, line 10: @pmaytak
Why do we need to do that in the samples? Has anything changed?
Or is it to be cleaner?

pmaytak: We don't have to. Yea, just so it's cleaner and more consistent.

bgavrilMS: Also more robust, because without the attributes the serializer has to use the name of the properties, which a developer might change inadvertedly.

pmaytak: I misunderstood the question. Let me rephrase. We don't have to replace Newtonsoft. I just did it so it's cleaner and more consistent with Identity Web. I chose to put attributes on properties because System.Text.Json does case-sensitive deserialization by default, so I just specified that JSON propeties in this object are lowercase. Otherwise I would have to pass deserialization options into every deserealize method.

jennyf19: What if a customer doesn't do this? We need to document this and have it in the samples then.

---
In reply to: [436013645](https://github.com/AzureAD/microsoft-identity-web/pull/188#discussion_r436013645) [](ancestors = 436013645)

pmaytak: Hmm. Good point. I could:
1. Revert back to using Newtonsoft - it's more flexible.
2. Instead of the `JsonPropertyName` I can just pass Json options to the deserealize method - it's a bit more straightforward.
3. If I keep it this way, where should I document this? It would probably be mostly just a link to the MS docs, since they explain it well.

jennyf19: @pmaytak I'm not sure. @jmprieur what are your thoughts on this?

jmprieur: How would option 2 look like, @pmaytak. It seems to me I prefer it better, but I've a hard time figuring out what it would mean?

Also, do we need to modify the Project templates?

pmaytak: @jmprieur 
1. Something like:
```csharp
            var options = new JsonSerializerOptions
            {
                PropertyNameCaseInsensitive = true,
            };
            JsonSerializer.Deserialize<Todo>(content, options);
```
2. I didn't find any references to Newtonsoft in the projects.

jmprieur: Indeed, @pmaytak, there are no references to Newtonsoft in the service (Web API) because the serialization is handled by ASP.NET core, hence my surprise (and me being worried) that the behavior would have changed and that we'd need to add these attribute to the samples POCO objects.

ASP.NET core handles the deserialization/serialization automatically in the controller based on the attributes, for instance `[FormBody]`

https://github.com/AzureAD/microsoft-identity-web/blob/68beba144a1b9b7c44e1ebaf34c048b1f69cf838/tests/B2CWebAppCallsWebApi/TodoListService/Controllers/TodoListController.cs#L59

I'm confused. Just to make sure I understand: do we really need to add the attributes or set the `JsonSerializerOptions`? or is it about demonstrating best practices, but not breaking anything.

pmaytak: Yes, to your 3 points.

Interestingly, ASP.NET deserializaton from string body is case-**insensitive**. And ASP.NET serializes Todo to all lowercase property names before sending the result.

https://github.com/AzureAD/microsoft-identity-web/blob/5f70e638989f85d0d0b398eb185e77e301d7de2b/tests/WebAppCallsWebApiCallsGraph/TodoListService/Controllers/TodoListController.cs#L63-L72

So yea, the JsonPropertyName attribute or JSON options are only when JsonSerializer class is used.

PR #205, opened 2020-06-13
fix tests

jennyf19:
move files to migrationaid folder

didn't want to do too much in this one.


jennyf19:
@jmprieur i think we should get rid of it, and then we can re-add it based on customer feedback. We have history of the tests, so shouldn't be a problem to re-add them if needed.

PR #212, opened 2020-06-16
Proposing to also remove the token decrypt certificate from AddMicrosoftWebAPI 

jmprieur:
@jennyf19  

Proposing to also remove the decrypt certificate from the public API of **AddMicrosoftProtectedApi** as now we have the configuration to express it (something like the following PR)

Note that I have not tried to fix the unit tests (this PR is kind of a spec, and this is to discuss)

### DIFF
--- a/src/Microsoft.Identity.Web/MigrationAid/Obsolete-WebApiAuthenticationBuilderExtensions.cs
+++ b/src/Microsoft.Identity.Web/MigrationAid/Obsolete-WebApiAuthenticationBuilderExtensions.cs
@@ -36,14 +36,13 @@ namespace Microsoft.Identity.Web
             X509Certificate2 tokenDecryptionCertificate = null,
             bool subscribeToJwtBearerMiddlewareDiagnosticsEvents = false)
         {
-            if (tokenDecryptionCertificate != null)
-            {
-                throw new ArgumentException("Use the TokenDecryptionCertificates config property. See https://aka.ms/ms-id-web-certificates");
-            }
-
-            return builder.AddMicrosoftWebApi(
-                configuration,
-                configSectionName,
+            // Just call the obsolete method below (which takes delegates).
+            // This method will do the work of taking into account the legacy
+            // parameter for the token decyrption certificate
+            return builder.AddProtectedWebApi(
+                options => configuration.Bind(configSectionName, options),
+                options => configuration.Bind(configSectionName, options),
+                tokenDecryptionCertificate,
                 jwtBearerScheme,
                 subscribeToJwtBearerMiddlewareDiagnosticsEvents);
         }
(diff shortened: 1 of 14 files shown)


src/Microsoft.Identity.Web/MigrationAid/Obsolete-WebApiAuthenticationBuilderExtensions.cs
```diff
@@ -36,11 +36,14 @@ public static partial class WebApiAuthenticationBuilderExtensions
...
             bool subscribeToJwtBearerMiddlewareDiagnosticsEvents = false)
         {
-            return builder.AddMicrosoftWebApi(
-                configuration,
-                configSectionName,
-                jwtBearerScheme,
+            // Just call the obsolete method below (which takes delegates).
+            // This method will do the work of taking into account the legacy
+            // parameter for the token decyrption certificate
+            return builder.AddProtectedWebApi(
```
jmprieur, line 42: The principle is to not call immediately the AddMicrosoftWebApi method but delegate to an obsolete method which will handle the legacy certificate.

src/Microsoft.Identity.Web/WebApiAuthenticationBuilderExtensions.cs
```diff
@@ -28,7 +28,6 @@ public static partial class WebApiAuthenticationBuilderExtensions
         /// <param name="configuration">The Configuration object.</param>
         /// <param name="configSectionName">The configuration section with the necessary settings to initialize authentication options.</param>
         /// <param name="jwtBearerScheme">The JwtBearer scheme name to be used. By default it uses "Bearer".</param>
-        /// <param name="tokenDecryptionCertificate">Token decryption certificate (null by default).</param>
```
jmprieur, line 31: The AddMicrosoftWebApi don't have the tokenDecryptionCertificate any longer. The AddProtectedWebApi still do.

tests/Microsoft.Identity.Web.Test/WebApiExtensionsTests.cs
```diff
@@ -39,6 +45,7 @@ public class WebApiExtensionsTests
             options.Instance = TestConstants.AadInstance;
             options.TenantId = TestConstants.TenantIdAsGuid;
             options.ClientId = TestConstants.ClientId;
+            options.TokenDecryptionCertificates = TokenDecryptionCertificatesDescription;
```
jmprieur, line 48: I hesitated passing directly DecryptTokenCertificateDescriptionJson ?

tests/Microsoft.Identity.Web.Test/WebApiExtensionsTests.cs
```diff
@@ -65,7 +72,11 @@ public void AddProtectedWebApi_WithConfigName()
             provider.GetRequiredService<IOptionsFactory<MicrosoftIdentityOptions>>().Create(string.Empty);
             config.Received(3).GetSection(_configSectionName);
 
-            AddProtectedWebApi_TestCommon(services, provider);
+            // Ideally we'd want to configure the GetValue for "TokenDecryptionCertificates" but that's not possible!
```
jmprieur, line 75: @pmaytak : how can we substitute so that the value of TokenDecryptionCertificates is right? 
I'm even questioning if we want to substitute the config. Why not directly use the configuration methods?

jmprieur:
@jennyf19 , thanks for your perspective: 

> this makes sense. let's take the breaking change with this. but not a fan of the implementation, i think we could do something better. let's discuss

I think I've proposed something way better now. There is no longer a breaking change. Small issue is the mocking? (substitute?) of the options is limited (does not find the collection . I think it's ok for now but we can discuss.

tests/Microsoft.Identity.Web.Test/WebApiExtensionsTests.cs
pmaytak: Well we're mocking config to test that the config was bound (for the overloads with config name). We probably don't need to test that because configuration values are inherently tested in the methods themselves when they are used. So we can just provide test config data without mocking. To get the value for any config property, could just do something like this:
`provider.GetRequiredService<IOptionsFactory<MicrosoftIdentityOptions>>().Create(string.Empty);`

PR #214, opened 2020-06-17
Some changes to build

pmaytak:
- Added task to run project tests.
  - I used `dotnet test` command. Not sure if there's a better way to run these tests.
- Changed VS version parameter to 16.0 to remove warnings.
- Fixed typo related to artifacts packages folder name. (Removed quotes.)

Before:
image

After:
image

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -473,9 +473,6 @@ dotnet_diagnostic.CA1062.severity = none
 # CA1303: Do not pass literals as localized parameters
 dotnet_diagnostic.CA1303.severity = none
 
-# SA1004: Documentation lines should begin with single space
-dotnet_diagnostic.SA1004.severity = none
-
 # SA1118: Parameter should not span multiple lines
 dotnet_diagnostic.SA1118.severity = none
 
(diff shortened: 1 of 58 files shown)


jmprieur:
@pmaytak : given that we have merged the .NET 5.0 support in master you might want to rework the following file: https://github.com/AzureAD/microsoft-identity-web/blob/pmaytak/updateBuild/build/template-restore-build-MSIdentityWeb.yaml to use dotnet build or dotnet test

See:
https://github.com/AzureAD/microsoft-identity-web/blob/master/.github/workflows/dotnetcore.yml

pmaytak:
> Thanks @pmaytak
> 
> I wonder if you don't want to run the **Post-build code analysis (e.g. Roslyn)** (template: template-postbuild-code-analysis.yaml) before running the tests? (to avoid unnecessary warnings during the build)?

- Oh, I don't know. Ideally we wouldn't have any warnings, so task order shouldn't matter. Plus seeing from artifacts analyzers run for all projects anyway.
- Ok. Will update to dotnet core task.

jmprieur:
@pmaytak : please close this one. It's superseeded by #218

PR #220, opened 2020-06-18
update to latest MSAL

jennyf19:
Tested the dev samples.
@jmprieur B2C sample works fine for me, not sure what's up w/yours :(


jmprieur:
Ok. thanks for checking @jennyf19. I'll reclone the folder and will have a look at the user secret files

Issue #211, opened 2020-06-15
[Bug] Custom JwtBearerOptions not applied


CalamityLorenzo:
**Which Version of Microsoft Identity Web are you using ?**
Note that to get help, you need to run the latest version. 

0.14-preview and current souce
**Where is the issue?**
* Web App
    * [ ] Sign-in users
    * [ ] Sign-in users and call web APIs
* Web API
    * [ x] Protected web APIs (Validating tokens)
    * [ ] Protected web APIs (Validating scopes)
    * [ ] Protected web APIs call downstream web APIs
* Token cache serialization
     * [ ] In Memory caches
     * [ ] Session caches
     * [ ] Distributed caches

Other? - please describe;

**Is this a new or existing app?**
New App
**Repro**

```csharp
            services.AddProtectedWebApi(Configuration, subscribeToJwtBearerMiddlewareDiagnosticsEvents: true);
            services.Configure<JwtBearerOptions>(AzureADDefaults.JwtBearerAuthenticationScheme, options =>
                    {
                        // This is an Microsoft identity platform Web API
                        options.Authority += "/v2.0";

                        // The valid audiences are both the Client ID (options.Audience) and api://{ClientID}
                        options.TokenValidationParameters.ValidAudiences = new string[]
                        {
                        options.Audience, $"api://{options.Audience}", $"https://{options.Audience}"
                        };
                        // D-d-d-delegate
                        options.TokenValidationParameters.IssuerValidator = Microsoft.IdentityModel.Tokens.Validators.ValidateIssuer;
                    });
```

**Expected behavior**
A clear and concise description of what you expected to happen (or code).
The **JwtBearerOptions** are never applied. So my settings audience is <<Guid>>, but I expect the valid audenices to be [<<Guid>>, api://<<Guid>>,https://<<Guid>>]
**Actual behavior**
Instead the only valid audiences is only <<Guid>>

**Additional context/ Logs / Screenshots**
The audience is an easy example. The problem looks like the JwtBearerOptions are not being applied by Identity.Web

pmaytak:
Hello.

- What is the value of `AzureADDefaults.JwtBearerAuthenticationScheme`?
- Please instead try 
`services.Configure<JwtBearerOptions>(JwtBearerDefaults.AuthenticationScheme, options =>`.

[...]

PR #227, opened 2020-06-21
fix for honoring valid audiences provided by the developer in JwtBearerOptions

jennyf19:
see [issue](https://github.com/AzureAD/microsoft-identity-web/issues/211)


src/Microsoft.Identity.Web/WebApiAuthenticationBuilderExtensions.cs
```diff
@@ -95,7 +95,8 @@ public static partial class WebApiAuthenticationBuilderExtensions
                 // This is a Microsoft identity platform Web API
                 options.Authority = AuthorityHelpers.EnsureAuthorityIsV2(options.Authority);
 
-                if (options.TokenValidationParameters.AudienceValidator == null)
+                if (options.TokenValidationParameters.AudienceValidator == null
+                && options.TokenValidationParameters.ValidAudiences == null)
```
pmaytak, line 99: Should we check `options.TokenValidationParameters.ValidAudience` too?

PR #231, opened 2020-06-23
Implemented C# 8 Nullable Reference Type standard.

pmaytak:
Issue #15 
Original branch #183

- Implemented C# 8 Nullable Reference Type standard.
- Includes additional null checks for public method parameters.
- Few cases where I had to refactor a little to prevent null exceptions.
- Misc warning and comment fixes.

**Note 1:** 
Created a separate branch because rebasing #183 had some conflicts. Work from #183 is included here.

**Note 2:**
Analyzers had some funky behavior. Some code didn't have squiggly underlines. Error list had some weird warning types. Perhaps something to do with preview .NET build?
image

References:
[Nullable reference types](https://docs.microsoft.com/en-us/dotnet/csharp/nullable-references)
[Reserved attributes contribute to the compiler's null state static analysis](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/attributes/nullable-analysis)
[Try out Nullable Reference Types](https://devblogs.microsoft.com/dotnet/try-out-nullable-reference-types/)
[Update libraries to use nullable reference types and communicate nullable rules to callers](https://docs.microsoft.com/en-us/dotnet/csharp/nullable-migration-strategies)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -467,9 +467,6 @@ dotnet_diagnostic.SA1309.severity = none
 # CA1031: Do not catch general exception types
 dotnet_diagnostic.CA1031.severity = none
 
-# CA1062: Validate arguments of public methods
-dotnet_diagnostic.CA1062.severity = none
-
 # CA1303: Do not pass literals as localized parameters
 dotnet_diagnostic.CA1303.severity = none
 
@@ -527,19 +524,22 @@ dotnet_diagnostic.SA1614.severity = none
 # SA1615: Element return value should be documented
 dotnet_diagnostic.SA1615.severity = none
 
-# SA1616: Element return value documentation should have text
-dotnet_diagnostic.SA1616.severity = none
-
 # SA1623: Property summary documentation should match accessors
 dotnet_diagnostic.SA1623.severity = none
 
 # SA1642: Constructor summary documentation should begin with standard text
 dotnet_diagnostic.SA1642.severity = none
 
-# SA1649: File name should match first type name
+## SA1649: File name should match first type name
 dotnet_diagnostic.SA1649.severity = none
 #########################################################
 #########################################################
+
+# SA1011: Closing square brackets should be spaced correctly
+dotnet_diagnostic.SA1011.severity = none
+
+# SA1009: Closing parenthesis should be spaced correctly
+dotnet_diagnostic.SA1009.severity = none
 
 [*.{cpp,h,in}]
 curly_bracket_next_line = true
--- a/src/Microsoft.Identity.Web/AuthorityHelpers.cs
+++ b/src/Microsoft.Identity.Web/AuthorityHelpers.cs
@@ -10,8 +10,8 @@ namespace Microsoft.Identity.Web
     {
         internal static string BuildAuthority(MicrosoftIdentityOptions options)
         {
-            var baseUri = new Uri(options.Instance);
-            var pathBase = baseUri.PathAndQuery.TrimEnd('/');
+            Uri baseUri = new Uri(options.Instance);
+            string pathBase = baseUri.PathAndQuery.TrimEnd('/');
             var domain = options.Domain;
             var tenantId = options.TenantId;
 
--- a/src/Microsoft.Identity.Web/AuthorizeForScopesAttribute.cs
+++ b/src/Microsoft.Identity.Web/AuthorizeForScopesAttribute.cs
@@ -32,37 +32,36 @@ namespace Microsoft.Identity.Web
         /// <summary>
         /// Scopes to request.
         /// </summary>
-        public string[] Scopes { get; set; }
+        public string[]? Scopes { get; set; }
 
         /// <summary>
         /// Key section on the configuration file that holds the scope value.
         /// </summary>
-        public string ScopeKeySection { get; set; }
+        public string? ScopeKeySection { get; set; }
 
         /// <summary>
-        /// Handles the MsalUiRequiredException.
+        /// Handles the <see cref="MsalUiRequiredException"/>.
         /// </summary>
         /// <param name="context">Context provided by ASP.NET Core.</param>
         public override void OnException(ExceptionContext context)
         {
-            // Do not re-use the attribute param Scopes. For more info: https://github.com/Azure-Samples/active-directory-aspnetcore-webapp-openidconnect-v2/issues/273
-            string[] incrementalConsentScopes = Array.Empty<string>();
-            MsalUiRequiredException msalUiRequiredException = context.Exception as MsalUiRequiredException;
-
-            if (msalUiRequiredException == null)
+            if (context != null)
             {
-                msalUiRequiredException = context.Exception?.InnerException as MsalUiRequiredException;
-            }
+                MsalUiRequiredException? msalUiRequiredException =
+                    (context.Exception as MsalUiRequiredException)
+                    ?? (context.Exception?.InnerException as MsalUiRequiredException);
 
-            if (msalUiRequiredException != null)
-            {
-                if (CanBeSolvedByReSignInOfUser(msalUiRequiredException))
+                if (msalUiRequiredException != null &&
+                    CanBeSolvedByReSignInOfUser(msalUiRequiredException))
                 {
                     // the users cannot provide both scopes and ScopeKeySection at the same time
                     if (!string.IsNullOrWhiteSpace(ScopeKeySection) && Scopes != null && Scopes.Length > 0)
                     {
                         throw new InvalidOperationException($"Either provide the '{nameof(ScopeKeySection)}' or the '{nameof(Scopes)}' to the 'AuthorizeForScopes'.");
                     }
+
+                    // Do not re-use the attribute param Scopes. For more info: https://github.com/Azure-Samples/active-directory-aspnetcore-webapp-openidconnect-v2/issues/273
+                    string[]? incrementalConsentScopes;
 
                     // If the user wishes us to pick the Scopes from a particular config setting.
                     if (!string.IsNullOrWhiteSpace(ScopeKeySection))
@@ -77,7 +76,7 @@ namespace Microsoft.Identity.Web
 
                         incrementalConsentScopes = new string[] { configuration.GetValue<string>(ScopeKeySection) };
 
-                        if (Scopes != null && Scopes.Length > 0 && incrementalConsentScopes != null && incrementalConsentScopes.Length > 0)
+                        if (Scopes != null && Scopes.Length > 0 && incrementalConsentScopes.Length > 0)
                         {
                             throw new InvalidOperationException("no scopes provided in scopes...");
                         }
@@ -87,7 +86,7 @@ namespace Microsoft.Identity.Web
                         incrementalConsentScopes = Scopes;
                     }
 
-                    var properties = BuildAuthenticationPropertiesForIncrementalConsent(incrementalConsentScopes, msalUiRequiredException, context.HttpContext);
+                    AuthenticationProperties properties = BuildAuthenticationPropertiesForIncrementalConsent(incrementalConsentScopes, msalUiRequiredException, context.HttpContext);
                     context.Result = new ChallengeResult(properties);
                 }
             }
@@ -107,17 +106,18 @@ namespace Microsoft.Identity.Web
         }
 
         /// <summary>
-        /// Build Authentication properties needed for incremental consent.
+        /// Build authentication properties needed for incremental consent.
         /// </summary>
         /// <param name="scopes">Scopes to request.</param>
-        /// <param name="ex">MsalUiRequiredException instance.</param>
-        /// <param name="context">current HTTP context in the pipeline.</param>
+        /// <param name="ex"><see cref="MsalUiRequiredException"/> instance.</param>
+        /// <param name="context">Current HTTP context in the pipeline.</param>
         /// <returns>AuthenticationProperties.</returns>
         private AuthenticationProperties BuildAuthenticationPropertiesForIncrementalConsent(
-            string[] scopes,
+            string[]? scopes,
             MsalUiRequiredException ex,
             HttpContext context)
         {
+            scopes ??= new string[0];
             var properties = new AuthenticationProperties();
 
             // Set the scopes, including the scopes that ADAL.NET / MSAL.NET need for the token cache
--- a/src/Microsoft.Identity.Web/CertificateManagement/DefaultCertificateLoader.cs
+++ b/src/Microsoft.Identity.Web/CertificateManagement/DefaultCertificateLoader.cs
@@ -27,19 +27,19 @@ namespace Microsoft.Identity.Web
                 switch (certificateDescription.SourceType)
                 {
                     case CertificateSource.KeyVault:
-                        certificateDescription.Certificate = LoadFromKeyVault(certificateDescription.Container, certificateDescription.ReferenceOrValue);
+                        certificateDescription.Certificate = LoadFromKeyVault(certificateDescription.Container!, certificateDescription.ReferenceOrValue!);
                         break;
                     case CertificateSource.Base64Encoded:
-                        certificateDescription.Certificate = LoadFromBase64Encoded(certificateDescription.ReferenceOrValue);
+                        certificateDescription.Certificate = LoadFromBase64Encoded(certificateDescription.ReferenceOrValue!);
                         break;
                     case CertificateSource.Path:
-                        certificateDescription.Certificate = LoadFromPath(certificateDescription.Container, certificateDescription.ReferenceOrValue);
+                        certificateDescription.Certificate = LoadFromPath(certificateDescription.Container!, certificateDescription.ReferenceOrValue!);
                         break;
                     case CertificateSource.StoreWithThumbprint:
-                        certificateDescription.Certificate = LoadFromStoreWithThumbprint(certificateDescription.ReferenceOrValue, certificateDescription.Container);
+                        certificateDescription.Certificate = LoadFromStoreWithThumbprint(certificateDescription.ReferenceOrValue!, certificateDescription.Container!);
                         break;
                     case CertificateSource.StoreWithDistinguishedName:
-                        certificateDescription.Certificate = LoadFromStoreWithDistinguishedName(certificateDescription.ReferenceOrValue, certificateDescription.Container);
+                        certificateDescription.Certificate = LoadFromStoreWithDistinguishedName(certificateDescription.ReferenceOrValue!, certificateDescription.Container!);
                         break;
                     default:
                         break;
@@ -52,7 +52,7 @@ namespace Microsoft.Identity.Web
             byte[] decoded = Convert.FromBase64String(certificateBase64);
             return new X509Certificate2(
                 decoded,
-                (string)null,
+                (string?)null,
                 X509KeyStorageFlags.MachineKeySet | X509KeyStorageFlags.EphemeralKeySet);
         }
 
@@ -79,7 +79,7 @@ namespace Microsoft.Identity.Web
             {
                 return new X509Certificate2(
                     certificate.Cer,
-                    (string)null,
+                    (string?)null,
                     X509KeyStorageFlags.MachineKeySet | X509KeyStorageFlags.EphemeralKeySet);
             }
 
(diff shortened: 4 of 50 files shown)


.editorconfig
```diff
@@ -527,20 +524,23 @@ dotnet_diagnostic.SA1614.severity = none
...
 dotnet_diagnostic.SA1649.severity = none
 #########################################################
 #########################################################
 
+# SA1011: Closing square brackets should be spaced correctly
+dotnet_diagnostic.SA1011.severity = none
+
+# SA1009: Closing parenthesis should be spaced correctly
+dotnet_diagnostic.SA1009.severity = none
+
```
pmaytak, line 543: Current version of StyleCop Analyzers has bugs related to `?` and `!` operators which are fixed in version 1.2.0 but it's still beta. So commented these rules out to reduce noise.

src/Microsoft.Identity.Web/AuthorityHelpers.cs
```diff
@@ -10,8 +10,8 @@ internal static class AuthorityHelpers
     {
         internal static string BuildAuthority(MicrosoftIdentityOptions options)
         {
-            var baseUri = new Uri(options.Instance);
-            var pathBase = baseUri.PathAndQuery.TrimEnd('/');
+            Uri baseUri = new Uri(options.Instance);
```
pmaytak, line 13: By default, `var` becomes the nullable reference type (`Uri?` in this case). So in some cases I used explicit type instead.

src/Microsoft.Identity.Web/CertificateManagement/DefaultCertificateLoader.cs
```diff
@@ -27,19 +27,19 @@ public void LoadIfNeeded(CertificateDescription certificateDescription)
                 switch (certificateDescription.SourceType)
                 {
                     case CertificateSource.KeyVault:
-                        certificateDescription.Certificate = LoadFromKeyVault(certificateDescription.Container, certificateDescription.ReferenceOrValue);
+                        certificateDescription.Certificate = LoadFromKeyVault(certificateDescription.Container!, certificateDescription.ReferenceOrValue!);
```
pmaytak, line 30: `!` overrides the conversion from nullable to non-nullable type, which otherwise would throw an exception. In this case, the properties shouldn't be null.

src/Microsoft.Identity.Web/CertificateManagement/DefaultCertificateLoader.cs
```diff
@@ -174,11 +174,7 @@ private static void ParseStoreLocationAndName(string storeDescription, ref Store
         /// <summary>
         /// Find a certificate by criteria.
         /// </summary>
-        /// <param name="x509Store"></param>
-        /// <param name="identifierCriterium"></param>
-        /// <param name="certificateIdentifier"></param>
-        /// <returns></returns>
-        private static X509Certificate2 FindCertificateByCriterium(
+        private static X509Certificate2? FindCertificateByCriterium(
```
pmaytak, line 177: Wondering if instead of returning `null` we throw an exception in this method, if the certificate is not found. Other load methods will throw exceptions if the certificate can't be created.

src/Microsoft.Identity.Web/CertificateManagement/DefaultCertificateLoader.cs
```diff
@@ -203,7 +199,7 @@ private static void ParseStoreLocationAndName(string storeDescription, ref Store
             DefaultCertificateLoader defaultCertificateLoader = new DefaultCertificateLoader();
             CertificateDescription certDescription = certificateDescription.First();
             defaultCertificateLoader.LoadIfNeeded(certDescription);
-            return certDescription?.Certificate;
+            return certDescription.Certificate!;
```
pmaytak, line 202: `Certificate` here, I assume, should never be null. But if it's not found in the cert store, it will be null. I think this will result in passing null to MSAL and to Wilson.

src/Microsoft.Identity.Web/InstanceDiscovery/IssuerConfigurationRetriever.cs
```diff
@@ -40,7 +42,7 @@ public async Task<IssuerMetadata> GetConfigurationAsync(string address, IDocumen
             };
 
             string doc = await retriever.GetDocumentAsync(address, cancel).ConfigureAwait(false);
-            return JsonSerializer.Deserialize<IssuerMetadata>(doc, options);
+            return JsonSerializer.Deserialize<IssuerMetadata>(doc, options)!;
```
pmaytak, line 45: This case is weird - the compiler says `Deserialize` can return null, but the docs don't mention it, say it just throws exceptions.

src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
```diff
@@ -16,17 +16,17 @@ public class MicrosoftIdentityOptions : OpenIdConnectOptions
         /// <summary>
         /// Gets or sets the Azure Active Directory instance, e.g. "https://login.microsoftonline.com".
         /// </summary>
-        public string Instance { get; set; }
+        public string Instance { get; set; } = null!;
```
pmaytak, line 19: This is a workaround for properties that are really non-nullable, except in once instance when the Data Transfer Objects are not yet initialized.

tests/.editorconfig
```diff
@@ -5,6 +5,9 @@
 # Code files
 [*.{cs,vb}]
 
+# SA0001: XML comment analysis disabled
+dotnet_diagnostic.SA0001.severity = none
```
pmaytak, line 9: This removes 4 warnings from the build pipeline related to disabled comment analysis.

src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
```diff
@@ -39,7 +39,7 @@ public class MicrosoftIdentityOptions : OpenIdConnectOptions
         /// If you don't specify the redirect URI, the redirect URI will be computed from the URL on which the app is
         /// deployed and the CallbackPath.
         /// </summary>
-        public string RedirectUri { get; set; }
+        public string? RedirectUri { get; set; }
```
jennyf19, line 42: > public string? RedirectUri { get; set; } [](start = 7, length = 41)

need to update w/master, these have been removed.

src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
```diff
@@ -70,25 +70,25 @@ public class MicrosoftIdentityOptions : OpenIdConnectOptions
...
 
         /// <summary>
         /// Gets the default user flow (which is signUpsignIn).
         /// </summary>
-        public string DefaultUserFlow => SignUpSignInPolicyId;
+        public string? DefaultUserFlow => SignUpSignInPolicyId;
 
         /// <summary>
-        /// Is considered B2C if the attribute SignUpSignInPolicyId is defined.
+        /// Gets a value indicating whether app is considered B2C if the attribute SignUpSignInPolicyId is defined.
```
jennyf19, line 91: nit: the previous comment is clearer imho

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -539,7 +535,7 @@ public void ReplyForbiddenWithWwwAuthenticateHeader(IEnumerable<string> scopes,
                 headers.Remove(HeaderNames.WWWAuthenticate);
             }
 
-            headers.Add(HeaderNames.WWWAuthenticate, v);
+            headers.Add(HeaderNames.WWWAuthenticate, new StringValues($"Bearer {parameterString}"));
         }
```
jennyf19, line 539: this has been changed as well in latest master.

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -502,8 +498,8 @@ private async Task<IConfidentialClientApplication> BuildConfidentialClientApplic
         /// the client can trigger an interaction with the user so that the user consents to more scopes.
         /// </summary>
         /// <param name="scopes">Scopes to consent to.</param>
-        /// <param name="msalServiceException"><see cref="MsalUiRequiredException"/> triggering the challenge.</param>
-        public void ReplyForbiddenWithWwwAuthenticateHeader(IEnumerable<string> scopes, MsalUiRequiredException msalServiceException)
+        /// <param name="msalServiceException">The <see cref="MsalUiRequiredException"/> that triggered the challenge.</param>
+        public async Task ReplyForbiddenWithWwwAuthenticateHeader(IEnumerable<string> scopes, MsalUiRequiredException msalServiceException)
         {
```
jennyf19, line 503: method name should end in `Async` if it's async now.

src/Microsoft.Identity.Web/WebAppCallsWebApiAuthenticationBuilderExtensions.cs
```diff
@@ -160,7 +174,7 @@ public static class WebAppCallsWebApiAuthenticationBuilderExtensions
                 options.Events.OnRedirectToIdentityProviderForSignOut = async context =>
                 {
                     // Remove the account from MSAL.NET token cache
-                    var tokenAcquisition = context.HttpContext.RequestServices.GetRequiredService<ITokenAcquisition>() as ITokenAcquisitionInternal;
+                    var tokenAcquisition = (ITokenAcquisitionInternal)context.HttpContext.RequestServices.GetRequiredService<ITokenAcquisition>();
```
jennyf19, line 177: is this part of a fix for [240](https://github.com/AzureAD/microsoft-identity-web/issues/240)? if so, can it go in a separate PR?

src/Microsoft.Identity.Web/CertificateManagement/DefaultCertificateLoader.cs
jennyf19: MSAL will handle the cert being null and throw an `ArgumentNullException`, we catch the exception

---
In reply to: [444031657](https://github.com/AzureAD/microsoft-identity-web/pull/231#discussion_r444031657) [](ancestors = 444031657)

src/Microsoft.Identity.Web/CertificateManagement/DefaultCertificateLoader.cs
jennyf19: the keyvault one will throw an exception, and if parsing the location fails, we throw an exception, otherwise we are returning the cert or null on all the other load methods.

---
In reply to: [444030354](https://github.com/AzureAD/microsoft-identity-web/pull/231#discussion_r444030354) [](ancestors = 444030354)

src/Microsoft.Identity.Web/InstanceDiscovery/IssuerConfigurationRetriever.cs
jennyf19: want to make a comment here then for future reference?

---
In reply to: [444034173](https://github.com/AzureAD/microsoft-identity-web/pull/231#discussion_r444034173) [](ancestors = 444034173)

src/Microsoft.Identity.Web/Resource/JwtBearerMiddlewareDiagnostics.cs
```diff
@@ -52,10 +52,7 @@ public JwtBearerMiddlewareDiagnostics(ILogger<JwtBearerMiddlewareDiagnostics> lo
         /// <param name="events">Events to subscribe to.</param>
         public JwtBearerEvents Subscribe(JwtBearerEvents events)
         {
-            if (events == null)
-            {
-                events = new JwtBearerEvents();
-            }
+            events ??= new JwtBearerEvents();
```
jennyf19, line 55: nit: space?

src/Microsoft.Identity.Web/Resource/OpenIdConnectMiddlewareDiagnostics.cs
```diff
@@ -80,6 +80,8 @@ public OpenIdConnectMiddlewareDiagnostics(ILogger<OpenIdConnectMiddlewareDiagnos
         /// <param name="events">Events to subscribe to.</param>
         public void Subscribe(OpenIdConnectEvents events)
         {
+            events ??= new OpenIdConnectEvents();
```
jennyf19, line 83: nit: space?

src/Microsoft.Identity.Web/Resource/RegisterValidAudience.cs
```diff
@@ -61,7 +61,12 @@ internal class RegisterValidAudience
             SecurityToken securityToken,
             TokenValidationParameters validationParameters)
         {
-            JwtSecurityToken token = securityToken as JwtSecurityToken;
+            JwtSecurityToken? token = securityToken as JwtSecurityToken;
+            if (token == null)
+            {
+                throw new SecurityTokenValidationException("Token is not JWT token.");
+            }
```
jennyf19, line 68: is that the right error message? when would the casting fail in this case and result in a null token value? the message should have some customer value, so how can they resolve or understand the issue to fix it.

jennyf19:
some comments. it's hard to review w/so many xml and spelling changes. in the future, can you do those in a separate PR, especially when you have so many? it tends to create noise and slows down the review. thank you.

src/Microsoft.Identity.Web/WebAppCallsWebApiAuthenticationBuilderExtensions.cs
pmaytak: No, this is just to make it non-nullable. #240 will change this line anyway.

pmaytak:
> some comments. it's hard to review w/so many xml and spelling changes. in the future, can you do those in a separate PR, especially when you have so many? it tends to create noise and slows down the review. thank you.

@jennyf19 Will do.

.editorconfig
jmprieur: why not leverage the beta? we are beta ourselves?

src/Microsoft.Identity.Web/AuthorizeForScopesAttribute.cs
```diff
@@ -32,38 +32,37 @@ public class AuthorizeForScopesAttribute : ExceptionFilterAttribute
...
+        public string? ScopeKeySection { get; set; }
 
         /// <summary>
-        /// Handles the MsalUiRequiredException.
+        /// Handles the <see cref="MsalUiRequiredException"/>.
         /// </summary>
         /// <param name="context">Context provided by ASP.NET Core.</param>
         public override void OnException(ExceptionContext context)
         {
-            // Do not re-use the attribute param Scopes. For more info: https://github.com/Azure-Samples/active-directory-aspnetcore-webapp-openidconnect-v2/issues/273
```
jmprieur, line 48: why remove this comment?

src/Microsoft.Identity.Web/ITokenAcquisition.cs
```diff
@@ -42,7 +42,7 @@ public interface ITokenAcquisition
         /// </summary>
         /// <param name="scopes">Scopes to consent to.</param>
         /// <param name="msalSeviceException"><see cref="MsalUiRequiredException"/> triggering the challenge.</param>
-        void ReplyForbiddenWithWwwAuthenticateHeader(
+        Task ReplyForbiddenWithWwwAuthenticateHeader(
```
jmprieur, line 45: postpone with Async?

.editorconfig
pmaytak: I can propose that change in a separate PR.

src/Microsoft.Identity.Web/AuthorizeForScopesAttribute.cs
jennyf19: it's further down.

src/Microsoft.Identity.Web/CertificateManagement/DefaultCertificateLoader.cs
pmaytak: If the certificate is not found, FromDisk, FromBase64, and FromKeyVault throw an exception; two FromCertStore methods return null. So my thought was to make the behavior consistent. Perhaps log that cert was not found and return null? This might benefit the certificate rotation logic (when implemented) when multiple certs are specified and we want to try until valid one is found.

src/Microsoft.Identity.Web/CertificateManagement/DefaultCertificateLoader.cs
pmaytak: - We do get an uncaught NPE when trying to set token decryption cert.
https://github.com/AzureAD/microsoft-identity-web/blob/6fe02dbb312bed2d7351ff9035a6e717cafc7574/src/Microsoft.Identity.Web/WebApiAuthenticationBuilderExtensions.cs#L117-L118

- Yes, we do catch the NPE from MSAL. But the exception is kind of general. Also, FromBase64 gives `Cannot find the requested object.`; FromDisk gives `The system cannot find the file specified.` Similar to my other comment, maybe we can update this to give better messaging and user experience when we implement cert rotation stuff.

src/Microsoft.Identity.Web/Resource/JwtBearerMiddlewareDiagnostics.cs
pmaytak: Where? Looks right.

src/Microsoft.Identity.Web/Resource/RegisterValidAudience.cs
pmaytak: I am not sure if it's always `JwtSecurityToken`. Is it? If it is, I can just explicitly cast it.

src/Microsoft.Identity.Web/AuthorizeForScopesAttribute.cs
pmaytak: It was moved few lines down.
https://github.com/AzureAD/microsoft-identity-web/blob/6fe02dbb312bed2d7351ff9035a6e717cafc7574/src/Microsoft.Identity.Web/AuthorizeForScopesAttribute.cs#L63-L64

src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
pmaytak: Yea, I agree. Reverted it. I'm not a huge fan of that analyzer suggestion myself.

src/Microsoft.Identity.Web/InstanceDiscovery/IssuerConfigurationRetriever.cs
pmaytak: Added.

src/Microsoft.Identity.Web/TokenAcquisition.cs
pmaytak: Good catch. Updated.

src/Microsoft.Identity.Web/ITokenAcquisition.cs
pmaytak: Good catch. Updated.

src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
pmaytak: Rebased.

src/Microsoft.Identity.Web/TokenAcquisition.cs
pmaytak: Rebased.

src/Microsoft.Identity.Web/CertificateManagement/DefaultCertificateLoader.cs
jennyf19: That's fine, but maybe something for a separate PR w/a new issue.

src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
jennyf19: The analyzer suggests that?

src/Microsoft.Identity.Web/Resource/RegisterValidAudience.cs
jennyf19: might be something to look into. the error message is misleading.

Issue #237, opened 2020-06-23
[Feature Request] Cookie options should be settable in `AddMicrosoftWebApp` 


jmprieur:
**Is your feature request related to a problem? Please describe.**
Cookie options should be settable. For this, `AddMicrosoftWebApp` should have a overload with an action to configure the `CookieAuthenticationOptions`

**Describe the solution you'd like**

The most complex override could be something like the following (note `configureCookieAuthenticationOptions` which is an `Action<CookieAuthenticationOptions>`)

```CSharp
 public static AuthenticationBuilder AddMicrosoftWebApp(
            this AuthenticationBuilder builder,
            Action<OpenIdConnectOptions> configureOpenIdConnectOptions,
            Action<MicrosoftIdentityOptions> configureMicrosoftIdentityOptions,
            Action<CookieAuthenticationOptions> configureCookieAuthenticationOptions,
            string openIdConnectScheme = OpenIdConnectDefaults.AuthenticationScheme,
            string cookieScheme = CookieAuthenticationDefaults.AuthenticationScheme,
            bool subscribeToOpenIdConnectMiddlewareDiagnosticsEvents = false)
        {
```

jennyf19:
Included in [0.2.0-preview release](https://github.com/AzureAD/microsoft-identity-web/releases/tag/0.2.0-preview)

Issue #234, opened 2020-06-23
[Bug] Microsoft.Identity.Web should not call BuildServiceProvider


jmprieur:
**Which Version of Microsoft Identity Web are you using ?**
Microsoft Identity Web 0.1.5-preview 

**Where is the issue?**
* Web App
    * [x ] Sign-in users
* Web API
    * [x ] Protected web APIs (Validating tokens)

**Repro**
See:

https://github.com/AzureAD/microsoft-identity-web/blob/5bc881bb6088f92db6e157f839bffb924627b215/src/Microsoft.Identity.Web/WebApiAuthenticationBuilderExtensions.cs#L145

https://github.com/AzureAD/microsoft-identity-web/blob/5bc881bb6088f92db6e157f839bffb924627b215/src/Microsoft.Identity.Web/WebAppAuthenticationBuilderExtensions.cs#L202

**Expected behavior**
NEVER call `BuildServiceProvider`. 

**Actual behavior**
**Microsoft.Identity.Web** calls `BuildServiceProvider`

**Discussion**
- It's a MUST FIX
- In NET 5.0, there is an overload of `AddJwtBearer` / `AddOpenIdConnect` with a service that you want to inject
- We need to come-up with alternatives for 3.0. @Tratcher will help

- in addition raise an issue with ASP.NET Core to have these diagnostics available by default without wrapping the events.

**Possible design**
(but feel free to do differently)
1. Add a constant in the Microsoft.Identity.Web.csproj file depending on the TargetFramework (maybe something like:

   ```XML
     <PropertyGroup Condition="'$(TargetFramework)' == 'netcoreapp3.1'">
       <DefineConstants>$(DefineConstants);DOTNET_CORE_31</DefineConstants>
     </PropertyGroup>
   ```

2. Based on the constant use one form of the other of AddOpenIdConnect and AddJwtBearer. For instance for AddOpenIdConnect we could have something like the following:

   ```CSharp
   #if DOTNET_CORE_31
            builder.AddOpenIdConnect(openIdConnectScheme, options =>
            {
             // Todo: replace by the work around that @Tratcher will provider
             IServiceProvider serviceProvider = builder.Services.BuildServiceProvider();
   #else
            builder.AddOpenIdConnect<IServiceProvider>(openIdConnectScheme, (options, serviceProvider) =>
            {
   #endif
   ```

Note that the aspnetcoreapp3.1 case would still use the BuildServiceProvider until Chris provides a workaround

[...]

PR #213, opened 2020-06-16
Jennyf/netcore5

jennyf19:
- [x] there are 6 tests failing w/net 5, getting a null ref. will be investigating that, but thought best to work from this branch

 @jmprieur

PR #257, opened 2020-06-25
Added ability to specify custom cookie options. Minor comment updates.

pmaytak:
Issue #237.
- Added the ability to specify custom cookie options.
- Added net5.0 and netcoreapp3.1 conditionals to test project.
- Updated the file path to XML docs in the project files.
- Minor comment fixes.

### DIFF
--- a/src/Microsoft.Identity.Web/WebApiCallsWebApiAuthenticationBuilderExtensions.cs
+++ b/src/Microsoft.Identity.Web/WebApiCallsWebApiAuthenticationBuilderExtensions.cs
@@ -12,18 +12,18 @@ using Microsoft.Identity.Client;
 namespace Microsoft.Identity.Web
 {
     /// <summary>
-    /// Extensions for AuthenticationBuilder for startup initialization of Web APIs.
+    /// Extensions for <see cref="AuthenticationBuilder"/> for startup initialization of web APIs.
     /// </summary>
     public static partial class WebApiCallsWebApiAuthenticationBuilderExtensions
     {
         /// <summary>
-        /// Protects the Web API with Microsoft identity platform (formerly Azure AD v2.0)
-        /// This supposes that the configuration files have a section named configSectionName (typically "AzureAD").
+        /// Protects the web API with Microsoft identity platform (formerly Azure AD v2.0).
+        /// This method expects the configuration file will have a section, named "AzureAd" as default, with the necessary settings to initialize authentication options.
         /// </summary>
-        /// <param name="builder">AuthenticationBuilder to which to add this configuration.</param>
-        /// <param name="configuration">Configuration.</param>
-        /// <param name="configSectionName">Section name in the config file (by default "AzureAD").</param>
-        /// <param name="jwtBearerScheme">Scheme for the JwtBearer token.</param>
+        /// <param name="builder">The <see cref="AuthenticationBuilder"/> to which to add this configuration.</param>
+        /// <param name="configuration">The configuration instance.</param>
+        /// <param name="configSectionName">The section name in the config file (by default "AzureAd").</param>
+        /// <param name="jwtBearerScheme">The scheme for the JWT bearer token.</param>
         /// <returns>The authentication builder to chain.</returns>
         public static AuthenticationBuilder AddMicrosoftWebApiCallsWebApi(
             this AuthenticationBuilder builder,
@@ -38,13 +38,12 @@ namespace Microsoft.Identity.Web
         }
 
         /// <summary>
-        /// Protects the Web API with Microsoft identity platform (formerly Azure AD v2.0)
-        /// This supposes that the configuration files have a section named configSectionName (typically "AzureAD").
+        /// Protects the web API with Microsoft identity platform (formerly Azure AD v2.0).
         /// </summary>
-        /// <param name="builder">AuthenticationBuilder to which to add this configuration.</param>
+        /// <param name="builder">The <see cref="AuthenticationBuilder"/> to which to add this configuration.</param>
         /// <param name="configureConfidentialClientApplicationOptions">The action to configure <see cref="ConfidentialClientApplicationOptions"/>.</param>
         /// <param name="configureMicrosoftIdentityOptions">The action to configure <see cref="MicrosoftIdentityOptions"/>.</param>
-        /// <param name="jwtBearerScheme">Scheme for the JwtBearer token.</param>
+        /// <param name="jwtBearerScheme">The scheme for the JWT bearer token.</param>
         /// <returns>The authentication builder to chain.</returns>
         public static AuthenticationBuilder AddMicrosoftWebApiCallsWebApi(
             this AuthenticationBuilder builder,
--- a/src/Microsoft.Identity.Web/WebAppAuthenticationBuilderExtensions.cs
+++ b/src/Microsoft.Identity.Web/WebAppAuthenticationBuilderExtensions.cs
@@ -15,7 +15,7 @@ using Microsoft.IdentityModel.Protocols.OpenIdConnect;
 namespace Microsoft.Identity.Web
 {
     /// <summary>
-    /// Extensions for AuthenticationBuilder for startup initialization.
+    /// Extensions for the <see cref="AuthenticationBuilder"/> for startup initialization.
     /// </summary>
     public static partial class WebAppAuthenticationBuilderExtensions
     {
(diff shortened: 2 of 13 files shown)


src/Microsoft.Identity.Web/WebAppAuthenticationBuilderExtensions.cs
```diff
@@ -65,10 +64,59 @@ public static partial class WebAppAuthenticationBuilderExtensions
             Action<MicrosoftIdentityOptions> configureMicrosoftIdentityOptions,
             string openIdConnectScheme = OpenIdConnectDefaults.AuthenticationScheme,
             string cookieScheme = CookieAuthenticationDefaults.AuthenticationScheme,
+            bool subscribeToOpenIdConnectMiddlewareDiagnosticsEvents = false) =>
```
pmaytak, line 67: @jmprieur I added an overload without custom cookie options. Should I remove it? We can just have 1 overload with config name and 1 with config actions?

src/Microsoft.Identity.Web/WebApiCallsWebApiAuthenticationBuilderExtensions.cs
```diff
@@ -38,13 +38,12 @@ public static partial class WebApiCallsWebApiAuthenticationBuilderExtensions
         }
 
         /// <summary>
-        /// Protects the Web API with Microsoft identity platform (formerly Azure AD v2.0)
-        /// This supposes that the configuration files have a section named configSectionName (typically "AzureAD").
+        /// Protects the web API with Microsoft identity platform (formerly Azure AD v2.0)
```
jennyf19, line 41:
```suggestion
        /// Protects the web API with Microsoft identity platform (formerly Azure AD v2.0).
```

pmaytak:
@jennyf19 

- I updated a test [here](https://github.com/AzureAD/microsoft-identity-web/blob/965ac569b82d24cdde231fac765c95177252756d/tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs#L102-L105). It only works for Core 3.1 though. In .NET 5, there seems to be no easy way to test. The internals of `AddCookies` method [are changed](https://github.com/dotnet/aspnetcore/blob/38f9b9abdbcb9c767f2bafd93a2fa1197fe8947f/src/Security/Authentication/Cookies/src/CookieExtensions.cs#L32-L45) to add `Action<CookieAuthenticationOptions, IServiceProvider>` rather than `Action<CookieAuthenticationOptions>`.
- Yea, if there are a lot or unrelated comment fixes, I'll move them to a separate PR.

tests/Microsoft.Identity.Web.Test/Microsoft.Identity.Web.Test.csproj
```diff
@@ -1,4 +1,11 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
+  
+  <PropertyGroup Condition="'$(TargetFramework)' == 'netcoreapp3.1'">
```
jmprieur, line 3: Aren't the constant obtained from Microsoft.Identity.Web ? (transitivity?)

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
```diff
@@ -90,14 +91,19 @@ public void AddMicrosoftWebApp_WithConfigActionParameters(bool subscribeToDiagno
...
+                .AddMicrosoftWebApp(_configureOidcOptions, _configureMsOptions, _configureCookieOptions, _oidcScheme, _cookieScheme, subscribeToDiagnostics);
 
             var provider = services.BuildServiceProvider();
 
             // Assert configure options actions added correctly
             var configuredOidcOptions = provider.GetServices<IConfigureOptions<OpenIdConnectOptions>>().Cast<ConfigureNamedOptions<OpenIdConnectOptions>>();
             var configuredMsOptions = provider.GetServices<IConfigureOptions<MicrosoftIdentityOptions>>().Cast<ConfigureNamedOptions<MicrosoftIdentityOptions>>();
 
+#if DOTNET_CORE_31
+            var configuredCookieOptions = provider.GetServices<IConfigureOptions<CookieAuthenticationOptions>>().Cast<ConfigureNamedOptions<CookieAuthenticationOptions>>();
```
jmprieur, line 103: Did you try to use IConfigureNamedOptions (interface) instead of an implementation?

henrik-me:
>         private Action<MicrosoftIdentityOptions> _configureMsOptions = (options) =>

unrelated to the change: this one is modified in a number of test cases, not sure how well this works with parallel test execution. Each test which needs something different should probably create a new options var.

this was the only one I spotted, but glancing over the top of this file there might be more... thinking there should be an effort to explore parallel test execution (easy to enable in VS) issues. Not very important if things generally works and test execution is acceptable.

---
Refers to: tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs:42 in 965ac56. [](commit_id = 965ac569b82d24cdde231fac765c95177252756d, deletion_comment = False)

pmaytak:
@henrik-me 
This works for xUnit because it makes a new instance of test class for to run each test method. (NUnit, on the other hand, creates one test class instance and runs all the methods in it.) By default, xUnit tests already run parallel.

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
pmaytak: Yea, but the service is of type `IConfigureOptions` and interfaces don't have the `Action` property.

jennyf19: did you see [this](https://github.com/AzureAD/microsoft-identity-web/pull/213/files#diff-36a4b8182e99fbf99b12e95c7a4e6dabR102), as an example? I would prefer to not have flags for .net 3.1/net 5 is possible.

tests/Microsoft.Identity.Web.Test/Microsoft.Identity.Web.Test.csproj
pmaytak: Tried adding constants only to main .Web project and it doesn't seem to propagate to the dependent projects.

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
pmaytak: Hmm, yea, I could just try to resolve the service for `IConfigureOptions<CookieAuthenticationOptions>` but that will only tell us that some cookie options were added and not whether our custom cookie action was added.

We might still need flags in general, like for #234 fix.

src/Microsoft.Identity.Web/WebAppAuthenticationBuilderExtensions.cs
jmprieur: I don't understand the question, @pmaytak

pmaytak: `WebAppAuthenticationBuilderExtensions` now has three methods. The difference between 2 and 3 is that 3 has cookie options parameter. I just thought maybe we should not add 2, and just have 1 and 3?
```csharp
public static AuthenticationBuilder AddMicrosoftWebApp(
            this AuthenticationBuilder builder,
            IConfiguration configuration,
            string configSectionName = "AzureAd",
            string openIdConnectScheme = OpenIdConnectDefaults.AuthenticationScheme,
            string cookieScheme = CookieAuthenticationDefaults.AuthenticationScheme,
            bool subscribeToOpenIdConnectMiddlewareDiagnosticsEvents = false)
```
```csharp
public static AuthenticationBuilder AddMicrosoftWebApp(
            this AuthenticationBuilder builder,
            Action<OpenIdConnectOptions> configureOpenIdConnectOptions,
            Action<MicrosoftIdentityOptions> configureMicrosoftIdentityOptions,
            string openIdConnectScheme = OpenIdConnectDefaults.AuthenticationScheme,
            string cookieScheme = CookieAuthenticationDefaults.AuthenticationScheme,
            bool subscribeToOpenIdConnectMiddlewareDiagnosticsEvents = false)
```
```csharp
public static AuthenticationBuilder AddMicrosoftWebApp(
            this AuthenticationBuilder builder,
            Action<OpenIdConnectOptions> configureOpenIdConnectOptions,
            Action<MicrosoftIdentityOptions> configureMicrosoftIdentityOptions,
            Action<CookieAuthenticationOptions> configureCookieAuthenticationOptions,
            string openIdConnectScheme = OpenIdConnectDefaults.AuthenticationScheme,
            string cookieScheme = CookieAuthenticationDefaults.AuthenticationScheme,
            bool subscribeToOpenIdConnectMiddlewareDiagnosticsEvents = false)
```

jmprieur: Oh, I understand. Yes, I agree. Only 1 and 3.
Thanks @pmaytak

PR #259, opened 2020-06-26
Fixes #253: Check/decide the behavior of VerifyUserHasAnyAcceptedScope()

### LINKED ISSUE #253: [Bug] Check the behavior of VerifyUserHasAnyAcceptedScope()
Check the behavior of VerifyUserHasAnyAcceptedScope()
[ScopesRequiredHttpContextExtensions.cs#L47-L49](https://github.com/AzureAD/microsoft-identity-web/blob/17f48c62410e08ea0d4c1fdb4f15a0713ed7f00c/src/Microsoft.Identity.Web/Resource/ScopesRequiredHttpContextExtensions.cs#L47-L49)

We thought that interest of throwing the `HttpRequestException` is that there is a lot of context associated with it, and this sends back the 401. 
Chris thinks that:
- HttpRequestException is not handled by ASP.NET Core. 
- Throwing an exception is not the right answer, because if not handled, results in a 505 not a 401

TODO:
- Jm to check the desired behavior. Rather 403? (there is an authenticated user with the wrong parameter)
- what do SAL/MISE do?

jmprieur:
Fixes #253:

- Now checks that the HttpContext is not null, otherwise throws an ArgumentNullException
- If there is no authenticated user, the HTTP response status code is set to 401
- If there is an authenticated user, the HTTP response code is set to 403 and a message tells which scopes to acquire

- Updated the tests with the new behavior.
- Updated the sample to surface an explicit message to the developer of the client.

  image

### DIFF
--- a/src/Microsoft.Identity.Web/Resource/ScopesRequiredHttpContextExtensions.cs
+++ b/src/Microsoft.Identity.Web/Resource/ScopesRequiredHttpContextExtensions.cs
@@ -19,14 +19,15 @@ namespace Microsoft.Identity.Web.Resource
         /// <summary>
         /// When applied to an <see cref="HttpContext"/>, verifies that the user authenticated in the
         /// web API has any of the accepted scopes.
+        /// If there is no authenticated user, the reponse is a 401 (Unauthenticated).
         /// If the authenticated user does not have any of these <paramref name="acceptedScopes"/>, the
-        /// method throws an HTTP Unauthorized with the message telling which scopes are expected in the token.
+        /// method updates the HTTP response providing a status code Forbidden (403)
+        /// and writes to the response body a message telling which scopes are expected in the token.
         /// </summary>
         /// <param name="context">HttpContext (from the controller).</param>
         /// <param name="acceptedScopes">Scopes accepted by this web API.</param>
-        /// <exception cref="HttpRequestException"> with a <see cref="HttpResponse.StatusCode"/> set to
-        /// <see cref="HttpStatusCode.Unauthorized"/>.
-        /// </exception>
+        /// <remarks>When the scopes don't match the response is a 403 (Forbidden), 
+        /// because the user is authenticated (hence not 401), but not authorized.</remarks>
         public static void VerifyUserHasAnyAcceptedScope(this HttpContext context, params string[] acceptedScopes)
         {
             if (acceptedScopes == null)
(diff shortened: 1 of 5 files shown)


src/Microsoft.Identity.Web/Resource/ScopesRequiredHttpContextExtensions.cs
```diff
@@ -19,34 +19,47 @@ public static class ScopesRequiredHttpContextExtensions
         /// <summary>
         /// When applied to an <see cref="HttpContext"/>, verifies that the user authenticated in the
         /// web API has any of the accepted scopes.
+        /// If there is no authenticated user, the reponse is a 401 (Unauthenticated).
         /// If the authenticated user does not have any of these <paramref name="acceptedScopes"/>, the
-        /// method throws an HTTP Unauthorized with the message telling which scopes are expected in the token.
+        /// method updates the HTTP response providing a status code Forbidden (403)
```
jmprieur, line 24: This is a behavioral change, BTW

mmacy, line 24:
```suggestion
        /// method updates the HTTP response providing a status code 403 (Forbidden)
```

Switching order to keep with convention of the others.

jmprieur:
Thanks @mmacy

src/Microsoft.Identity.Web/Resource/ScopesRequiredHttpContextExtensions.cs
```diff
@@ -19,34 +19,47 @@ public static class ScopesRequiredHttpContextExtensions
...
-        /// method throws an HTTP Unauthorized with the message telling which scopes are expected in the token.
+        /// method updates the HTTP response providing a status code 403 (Forbidden)
+        /// and writes to the response body a message telling which scopes are expected in the token.
         /// </summary>
         /// <param name="context">HttpContext (from the controller).</param>
         /// <param name="acceptedScopes">Scopes accepted by this web API.</param>
-        /// <exception cref="HttpRequestException"> with a <see cref="HttpResponse.StatusCode"/> set to
-        /// <see cref="HttpStatusCode.Unauthorized"/>.
-        /// </exception>
+        /// <remarks>When the scopes don't match the response is a 403 (Forbidden),
```
jennyf19, line 29:
```suggestion
        /// <remarks>When the scopes don't match, the response is a 403 (Forbidden),
```

src/Microsoft.Identity.Web/Resource/ScopesRequiredHttpContextExtensions.cs
```diff
@@ -19,34 +19,47 @@ public static class ScopesRequiredHttpContextExtensions
...
+            }
+            else
+            {
+                // Attempt with Scp claim
+                Claim? scopeClaim = context.User.FindFirst(ClaimConstants.Scp);
+
+                // Fallback to Scope claim name
+                if (scopeClaim == null)
+                {
+                    scopeClaim = context?.User?.FindFirst(ClaimConstants.Scope);
```
jennyf19, line 54: do you need to check for null here for the context, you've already done that the beginning

pmaytak:
@jmprieur This introduced regression bug. Anything after the call to Verify still executes even if scopes are missing. I remember Chris mentioning this sometime too.

```csharp
HttpContext.VerifyUserHasAnyAcceptedScope(scopeRequiredByApi);
TodoStore.Remove(id);
```

Issue #252, opened 2020-06-24
[Bug] Fixes in  MsalSessionTokenCacheProvider.cs


jmprieur:
**Which Version of Microsoft Identity Web are you using ?**
Microsoft Identity Web 0.1.5-preview

** Things to fix:
-	The lock object does not need to be static: [MsalSessionTokenCacheProvider.cs#L125](https://github.com/AzureAD/microsoft-identity-web/blob/17f48c62410e08ea0d4c1fdb4f15a0713ed7f00c/src/Microsoft.Identity.Web/TokenCacheProviders/Session/MsalSessionTokenCacheProvider.cs#L125)
-	The session should not be committed (for instance [MsalSessionTokenCacheProvider.cs#L96](https://github.com/AzureAD/microsoft-identity-web/blob/17f48c62410e08ea0d4c1fdb4f15a0713ed7f00c/src/Microsoft.Identity.Web/TokenCacheProviders/Session/MsalSessionTokenCacheProvider.cs#L96), but there are other lines). Commit is automatically called at the end of the request. If we do it prematurely, nobody can modify it any longer.

jennyf19:
Included in [0.2.0-preview release](https://github.com/AzureAD/microsoft-identity-web/releases/tag/0.2.0-preview)

PR #265, opened 2020-06-28
fix bugs in session cache provider.

jennyf19:
- remove static
- remove commit, as it's done automatically

for testing:
In Startup.cs:
	• using Microsoft.Identity.Web.TokenCacheProviders.Session; 
	•  `services.AddInMemoryTokenCaches();` replace with `services.AddSessionTokenCaches(); `
In Startup.cs, in the configure method: Configure
`app.UseSession();`
before 
`app.UseAuthentication();`

#252


jmprieur:
Tested (and updated the wiki: https://github.com/AzureAD/microsoft-identity-web/wiki/token-cache-serialization#session-token-cache)

Issue #251, opened 2020-06-24
[Bug] Token cache serialization should expose their own options


jmprieur:
**Which Version of Microsoft Identity Web are you using ?**
Microsoft Identity Web 0.1.5-preview

`MsalMemoryTokenCacheProvider` get its own options type but `MsalDistributedTokenCacheAdapter` injects `IOptions<DistributedCacheEntryOptions>` ?

https://github.com/AzureAD/microsoft-identity-web/blob/17f48c62410e08ea0d4c1fdb4f15a0713ed7f00c/src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs#L38

https://github.com/AzureAD/microsoft-identity-web/blob/17f48c62410e08ea0d4c1fdb4f15a0713ed7f00c/src/Microsoft.Identity.Web/TokenCacheProviders/InMemory/MsalMemoryTokenCacheProvider.cs#L39

It's historical and needs to be changed
We should not inject the `IOptions<DistributedCacheEntryOptions>` (because different components could be using different caches and would need to have different options)

## What needs to be done?

- `SessionTokenCache` does not need option, it uses the session options
- [`MsalMemoryTokenCacheProvider.cs`](https://github.com/AzureAD/microsoft-identity-web/blob/master/src/Microsoft.Identity.Web/TokenCacheProviders/InMemory/MsalMemoryTokenCacheProvider.cs) already has options, which we'll keep.
- Add a `MsalDistributedTokenCacheAdapterOptions` class inheriting from `DistributedCacheEntryOptions`, and use this one in lieu of `DistributedCacheEntryOptions` in MsalDistributedTokenCacheAdapter.cs (as other components would want to use the ASP.NET Core provided distributed cache)
- update the wiki documentation to mention the new option.

jennyf19:
Included in [0.2.0-preview release](https://github.com/AzureAD/microsoft-identity-web/releases/tag/0.2.0-preview)

PR #266, opened 2020-06-28
create new class to inherit from the DistributedCacheEntryOptions to …

jennyf19:
…provide flexiblity w/token cache lifetime

#251


jmprieur:
@pmaytak : the difference is this is an adapter for distributed caches (not token caches) implementation. so I would think Adapter is right?

Issue #249, opened 2020-06-24
[Bug] SingletonTokenAcquisition can be removed


jmprieur:
**Which Version of Microsoft Identity Web are you using ?**
Microsoft Identity Web 0.1.5-preview

AddTokenAcquisition isTokenAcquisitionSingleton parameter (never used)
https://github.com/AzureAD/microsoft-identity-web/blob/0b628e2a56e6525bbca682c50a7528da4d09cc22/src/Microsoft.Identity.Web/ServiceCollectionExtensions.cs#L32-L34 is a request from the Microsoft Graph SDK which want a singleton (whereas in general that’s not such a good idea)

See https://github.com/AzureAD/microsoft-identity-web/blob/master/tests/WebAppCallsMicrosoftGraph/appsettings.json#L9

There is an alternative which is that the GraphService registers the ITokenAcquisition as a singleton (the last which does the registration)

**Decision:**
- We keep AddTokenAcquisition as is
- We remove the [SingletonTokenAcquisition](https://github.com/AzureAD/microsoft-identity-web/blob/master/src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs#L35) property from the Microsoft.Identity.Options
- AddTokenAcquision is called with the default value (false) from the extension methods
- Remove the tests that were testing this.

jennyf19:
Included in [0.2.0-preview release](https://github.com/AzureAD/microsoft-identity-web/releases/tag/0.2.0-preview)

SviatoslavVilkovych:
So, there is no more posibility to have **TokenAcquisition** registered as a Singleton service, right?

pmaytak:
You still can. Call `services.AddTokenAcquisition(true)` at the end of `ConfigureServices` method. (The related PRs just removed `SingletonTokenAcquisition` property from `MicrosoftIdentityOptions`.)

https://github.com/AzureAD/microsoft-identity-web/blob/7df36791eb5314387c9aad3e084a94940ba24024/src/Microsoft.Identity.Web/ServiceCollectionExtensions.cs#L33-L35

PR #271, opened 2020-06-29
Jmprieur/related to249

jmprieur:
This PR is there to prepare the work on #249 which consists in removing the TokenAcquisitionSingleton property from the Microsoft.Identity.Options (and the associated management and tests). AddMicrosoftWebAppCallsWebApi / AddMicrosoftWebApiCallsWebApi would systematically do AddTokenAcquisition(), which means AddTokenAcquisition(false), as this the default value is false.

The solution below consists in having `AddMicrosoftGraph` call directly `AddTokenAcquisition(true)` (to set it up as a singleton), instead of having the user to specify a different option in the config file, which s/he could forget to do, and would require more documentation. Then `AddTokenAcquisition()` removes the previous service and readds it, if needed, with the right lifetime.

The last call to `AddTokenAcquisition()`  wins. Therefore the call to  `AddMicrosoftGraph` needs to happen after the calls to AddMicrosoftWebAppCallsWebApi / AddMicrosoftWebApiCallsWebApi

### DIFF
--- a/src/Microsoft.Identity.Web/ServiceCollectionExtensions.cs
+++ b/src/Microsoft.Identity.Web/ServiceCollectionExtensions.cs
@@ -46,7 +46,7 @@ namespace Microsoft.Identity.Web
             {
                 if (isTokenAcquisitionSingleton ^ (tokenAcquisitionService.Lifetime == ServiceLifetime.Singleton))
                 {
-                    // The service was added, but already, but not with the right lifetime
+                    // The service was already added, but not with the right lifetime
                     services.Remove(tokenAcquisitionService);
                     services.Remove(tokenAcquisitionInternalService);
                 }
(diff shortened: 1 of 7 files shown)


src/Microsoft.Identity.Web/ServiceCollectionExtensions.cs
```diff
@@ -39,6 +40,23 @@ public static class ServiceCollectionExtensions
                 throw new ArgumentNullException(nameof(services));
             }
 
+            ServiceDescriptor tokenAcquisitionService = services.FirstOrDefault(s => s.ServiceType == typeof(ITokenAcquisition));
+            ServiceDescriptor tokenAcquisitionInternalService = services.FirstOrDefault(s => s.ServiceType == typeof(ITokenAcquisitionInternal));
+            if (tokenAcquisitionService != null && tokenAcquisitionInternalService != null)
+            {
+                if (isTokenAcquisitionSingleton ^ (tokenAcquisitionService.Lifetime == ServiceLifetime.Singleton))
+                {
+                    // The service was added, but already, but not with the right lifetime
```
jennyf19, line 49:
```suggestion
                    // The service was already added, but not with the right lifetime
```

Issue #243, opened 2020-06-23
[Bug] RegisterValidAudience.ValidateAudience does not throw if user set custom audiences.


brentschmaltz:
**Which Version of Microsoft Identity Web are you using ?**
Microsoft Identity Web 0.1.5-preview

**Where is the issue?**
The normal call graph through asp.net results in a call to Validators.Validate where an exception is thrown if a validate audience is not found, even if the user has added custom audiences.

jmprieur:
Also fixed by #275

jennyf19:
Included in [0.2.0-preview release](https://github.com/AzureAD/microsoft-identity-web/releases/tag/0.2.0-preview)

Issue #241, opened 2020-06-23
[Bug] WebApiAuthenticationBuilderExtensions.cs should consider cloning TokenValidationParameters


brentschmaltz:
**Which Version of Microsoft Identity Web are you using ?**
Note that to get help, you need to run the latest version. 
Microsoft Identity Web 0.1.5-preview

**Where is the issue?**
`AddMicrosoftWebApi` uses the value of `TokenValidationParameters` from `JwtBearerOptions`. Any modifications will affect any other users of `JwtBearerOptions`. It would be better to use `TokenValidationParameters.Clone()`.

jmprieur:
Thanks @brentschmaltz !

jmprieur:
@brentschmaltz is it only about adding

```CSharp
options.TokenValidationParameters = options.TokenValidationParameters.Clone();
```

before starting using the options.TokenValidationParameters ? for instance here: https://github.com/AzureAD/microsoft-identity-web/blob/master/src/Microsoft.Identity.Web/WebApiAuthenticationBuilderExtensions.cs#L102 ?

cc: @jennyf19 @pmaytak

brentschmaltz:
@jmprieur i haven't reviewed the entire call graph, but the idea is if you modify TokenValidationParameters by adding an audience, issuer, etc, then you should call clone first.
When we developed JwtBearer, we created a local copy before any modifications. To get an idea see: https://github.com/dotnet/aspnetcore/blob/4a93863b8d865bb4d65f77b34ee335d7b23a79fc/src/Security/Authentication/JwtBearer/src/JwtBearerHandler.cs#L91 . This issue is to make sure we follow the same pattern. Let me know if you want me to review anything.

jmprieur:
Thanks @brentschmaltz 
@pmaytak I see you self-assigned the issue. please add Brent to the PR review.

jennyf19:
Included in [0.2.0-preview release](https://github.com/AzureAD/microsoft-identity-web/releases/tag/0.2.0-preview)

PR #275, opened 2020-06-30
New proposal to fix #211

### LINKED ISSUE #211: [Bug] Custom JwtBearerOptions not applied

jmprieur:
This is a new attempt to fix #211 
This time by delegating to Wilson once the default value is set according to the token version

### DIFF
--- a/src/Microsoft.Identity.Web/Resource/RegisterValidAudience.cs
+++ b/src/Microsoft.Identity.Web/Resource/RegisterValidAudience.cs
@@ -67,8 +67,7 @@ namespace Microsoft.Identity.Web.Resource
                 throw new SecurityTokenValidationException("Token is not JWT token.");
             }
 
-            TokenValidationParameters validationParametersClone = validationParameters.Clone();
-            validationParametersClone.AudienceValidator = null;
+            validationParameters.AudienceValidator = null;
 
             // Case of a default App ID URI (the developer did not provide explicit valid audience(s)
             if (string.IsNullOrEmpty(validationParameters.ValidAudience) &&
@@ -77,17 +76,17 @@ namespace Microsoft.Identity.Web.Resource
                 // handle v2.0 access token or Azure AD B2C tokens (even if v1.0)
                 if (IsB2C || token.Claims.Any(c => c.Type == Version && c.Value == V2))
                 {
-                    validationParametersClone.ValidAudience = $"{ClientId}";
+                    validationParameters.ValidAudience = $"{ClientId}";
                 }
 
                 // handle v1.0 access token
                 else if (token.Claims.Any(c => c.Type == Version && c.Value == V1))
                 {
-                    validationParametersClone.ValidAudience = $"api://{ClientId}";
+                    validationParameters.ValidAudience = $"api://{ClientId}";
                 }
             }
 
-            Validators.ValidateAudience(audiences, securityToken, validationParametersClone);
+            Validators.ValidateAudience(audiences, securityToken, validationParameters);
             return true;
         }
     }
--- a/src/Microsoft.Identity.Web/Resource/RegisterValidAudience.cs
+++ b/src/Microsoft.Identity.Web/Resource/RegisterValidAudience.cs
@@ -81,7 +81,7 @@ namespace Microsoft.Identity.Web.Resource
                 }
 
                 // handle v1.0 access token
-                if (token.Claims.Any(c => c.Type == Version && c.Value == V1))
+                else if (token.Claims.Any(c => c.Type == Version && c.Value == V1))
                 {
                     validationParametersClone.ValidAudience = $"api://{ClientId}";
                 }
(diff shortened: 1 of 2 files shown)


src/Microsoft.Identity.Web/Resource/RegisterValidAudience.cs
```diff
@@ -76,29 +77,18 @@ internal class RegisterValidAudience
...
-            // Cases where developers explicitly provided the valid audiences
-            else if (!string.IsNullOrEmpty(validationParameters.ValidAudience))
-            {
-                return audiences.Contains(validationParameters.ValidAudience);
-            }
-            else
-            {
-                return audiences.Intersect(validationParameters.ValidAudiences).Any();
-            }
+            Validators.ValidateAudience(audiences, securityToken, validationParametersClone);
```
jmprieur, line 90: Delegate to Wilson to be sure to do the right thing.

src/Microsoft.Identity.Web/WebApiAuthenticationBuilderExtensions.cs
```diff
@@ -99,7 +99,9 @@ public static partial class WebApiAuthenticationBuilderExtensions
                 // This is a Microsoft identity platform Web API
                 options.Authority = AuthorityHelpers.EnsureAuthorityIsV2(options.Authority);
 
-                if (options.TokenValidationParameters.AudienceValidator == null)
+                if (options.TokenValidationParameters.AudienceValidator == null
```
jmprieur, line 102: Only provide a default in the default case - when nothing was set.

src/Microsoft.Identity.Web/Resource/RegisterValidAudience.cs
```diff
@@ -67,7 +67,8 @@ internal class RegisterValidAudience
                 throw new SecurityTokenValidationException("Token is not JWT token.");
             }
 
-            string validAudience;
+            TokenValidationParameters validationParametersClone = validationParameters.Clone();
```
pmaytak, line 70: Should this line be removed from here and added in extension methods like you mentioned [here](https://github.com/AzureAD/microsoft-identity-web/issues/241#issuecomment-651828106)?

jmprieur: yes, probably, @pmaytak. That's assigned to you, right?

brentschmaltz: @pmaytak i would have to review the call graph, but i would think Clone() would be called higher in the stack.

src/Microsoft.Identity.Web/WebApiAuthenticationBuilderExtensions.cs
brentschmaltz: @jmprieur is your model that you want to exclude any additional audiences?
This might be a good place to pass options.TokenValidationParameters.Clone() as I assume this is called once at setup.

src/Microsoft.Identity.Web/Resource/RegisterValidAudience.cs
pmaytak: @brentschmaltz This method is the Validator for every request. To call this once, we would put it into `AddMicrosoftWebApi` method where we set up JwtBearer

jmprieur: changed. thanks for the feedback

pmaytak:
This also fixes #243

jmprieur:
yes, I forgot to link it, but that was the plan. thanks for calling it out, @pmaytak

Issue #239, opened 2020-06-23
[Bug] Remove the weird pattern for newing up redundant copies of MicrosoftIdentityOptions


jmprieur:
**Which Version of Microsoft Identity Web are you using ?
Microsoft Identity Web 0.1.5-preview

**Where is the issue?**
* Web App
    * [x ] Sign-in users
    * [ x] Sign-in users and call web APIs
* Web API
    * [x ] Protected web APIs
    * [x ] Protected web APIs call downstream web APIs

Other? - please describe;

**Repro**

Remove the weird pattern for newing up redundant copies of MicrosoftIdentityOptions
[WebApiAuthenticationBuilderExtensions.cs#L85-L88](https://github.com/AzureAD/microsoft-identity-web/blob/0e68799fc095c0611ccb3cec3a6758568b0f3f7c/src/Microsoft.Identity.Web/WebApiAuthenticationBuilderExtensions.cs#L85-L88)
[WebAppAuthenticationBuilderExtensions.cs#L74-L75](https://github.com/AzureAD/microsoft-identity-web/blob/0e68799fc095c0611ccb3cec3a6758568b0f3f7c/src/Microsoft.Identity.Web/WebAppAuthenticationBuilderExtensions.cs#L74-L75)
[WebApiCallsWebApiAuthenticationBuilderExtensions.cs#L63-L66](https://github.com/AzureAD/microsoft-identity-web/blob/0e68799fc095c0611ccb3cec3a6758568b0f3f7c/src/Microsoft.Identity.Web/WebApiCallsWebApiAuthenticationBuilderExtensions.cs#L63-L66)
[WebAppCallsWebApiAuthenticationBuilderExtensions.cs#L101-L102](https://github.com/AzureAD/microsoft-identity-web/blob/0e68799fc095c0611ccb3cec3a6758568b0f3f7c/src/Microsoft.Identity.Web/WebAppCallsWebApiAuthenticationBuilderExtensions.cs#L101-L102)

**Expected behavior**
-	Injecting something IOptions<MicrosoftIdentityOptions> from the container (in 5.0)
-	In 3.1. @Chris will provide an ugly work around.

See also https://github.com/AzureAD/microsoft-identity-web/issues/234 which leverages the same mechanism

**Proposed design**
- in https://github.com/AzureAD/microsoft-identity-web/blob/master/src/Microsoft.Identity.Web/WebApiAuthenticationBuilderExtensions.cs#L83, we already have a service provider. 
[...]

PR #277, opened 2020-07-01
remove odd pattern for newing up redundant copies of options

jennyf19:
#239

### DIFF
--- a/src/Microsoft.Identity.Web/Microsoft.Identity.Web.csproj
+++ b/src/Microsoft.Identity.Web/Microsoft.Identity.Web.csproj
@@ -55,13 +55,6 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <Nullable>enable</Nullable>
   </PropertyGroup>
-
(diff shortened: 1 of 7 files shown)


pmaytak:
@jennyf19 Can you verify that test projects work with this change? Both test client and test service projects throw below error for  me. I had to add `builder.AddOpenIdConnect(openIdConnectScheme, options => { });` before the new `AddOptions` method to fix this. (And similarly do `AddJwtBearer` in API extensions.)

`AddOpenIdConnect` adds more than just options. But we do call `AddAuthentication` in startup, so the error message is kind of misleading.

`System.InvalidOperationException: No authenticationScheme was specified, and there was no DefaultChallengeScheme found. The default schemes can be set using either AddAuthentication(string defaultScheme) or AddAuthentication(Action<AuthenticationOptions> configureOptions).`

jmprieur:
@pmaytak : thanks for the heads-up. This is now fixed by https://github.com/AzureAD/microsoft-identity-web/pull/284
(indeed there was a merge conflict)

Issue #131, opened 2020-04-27
[Feature Request] Add ValidateAppRole() in addition to VerifyUserHasAnyAcceptedScope


jmprieur:
**Is your feature request related to a problem? Please describe.**
Web API called by daemon app need to verify the app roles.
We could just use [Authorize(Roles="")]

But for symmetry proposing to add a ValidateAppRole() method similar to [VerifyUserHasAnyAcceptedScope()](https://github.com/AzureAD/microsoft-identity-web/blob/master/src/Microsoft.Identity.Web/Resource/ScopesRequiredHttpContextExtensions.cs#L31), but which would test for [Roles](https://github.com/AzureAD/microsoft-identity-web/blob/master/src/Microsoft.Identity.Web/ClaimConstants.cs#L71) and [Role](https://github.com/AzureAD/microsoft-identity-web/blob/master/src/Microsoft.Identity.Web/ClaimConstants.cs#L76), instead of Scp/Scopes.

**Note**
We'd also need to update https://docs.microsoft.com/en-us/azure/active-directory/develop/scenario-protected-web-api-verification-scope-app-roles#verify-app-roles-in-apis-called-by-daemon-apps.

jennyf19:
Included in [0.2.0-preview release](https://github.com/AzureAD/microsoft-identity-web/releases/tag/0.2.0-preview)

PR #278, opened 2020-07-01
add app roles

jennyf19:
#131


jennyf19:
I updated the daemon [sample](https://github.com/Azure-Samples/active-directory-dotnetcore-daemon-v2/pull/74)

Issue #261, opened 2020-06-26
Move all log and error messages into a separate file as constants


pmaytak:
Currently the log and error messages are just local string literals. Organize them as const strings into a file like in [Wilson](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/src/Microsoft.IdentityModel.Tokens/LogMessages.cs) or [MSAL](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/master/src/client/Microsoft.Identity.Client/MsalErrorMessage.cs).

Related #47

pmaytak:
This is included in [Microsoft Identity Web 0.2.1-preview release](https://github.com/AzureAD/microsoft-identity-web/releases/tag/0.2.1-preview).

PR #280, opened 2020-07-01
initial commit for creating a const files for errors, log messages an…

jennyf19:
…d string values
#261 
I had some spare time and thought to get started on this. not complete, but a start, so we can add to it as we go. have made files for constant string values, log messages and error messages. good idea @pmaytak


jennyf19:
@pmaytak sounds good. thanks.

src/Microsoft.Identity.Web/Constants/ErrorMessage.cs
```diff
@@ -0,0 +1,28 @@
...
+        public const string ScopeKeySectionIsProvidedButNotPresentInTheServicesCollection = "The {0} is provided but the IConfiguration instance is not present in the services collection";
+        public const string NoScopesProvided = "no scopes provided in scopes...";
+        public const string NeitherScopeOrRolesClaimFoundInToken = "Neither scope or roles claim was found in the bearer token.";
+        public const string IssuerMetadataURLIsRequired = "Azure AD Issuer metadata address URL is required";
+        public const string NoMetadataDocumentRetrieverProvided = "No metadata document retriever is provided";
+        public const string IssuerDoesNotMatchValidIssuers = "Issuer: '{0}', does not match any of the valid issuers provided for this application.";
+        public const string ClientInfoReturnedFromServerIsNull = "client info returned from the server is null";
+        public const string MissingClientCredentials = "missing_client_credentials";
+        public const string DuplicateClientCredentials = "duplicate_client_credentials";
```
henrik-me, line 19: These looks a bit like error codes, meaning something depends on these values being recognized? Perhaps those should be in a separate file.

src/Microsoft.Identity.Web/Constants/ErrorMessage.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Identity.Web
+{
+    internal static class ErrorMessage
```
henrik-me, line 6: >ErrorMessage [](start = 26, length = 12)

Would be interesting to try out adding an error identifier in the beginning of the error message, e.g. MIDW-0000001: the error text goes here. This way it will be easier to find the actual error when searching?

henrik-me:
@jennyf19 @pmaytak . a few late comments with ideas for a future iteration.

src/Microsoft.Identity.Web/Constants/ErrorMessage.cs
henrik-me, line 6: >ErrorMessage [](start = 26, length = 12)

A couple of late messages for consideration in a future PR

pmaytak: Yep, Wilson does this. Is there a standard on how to properly create these error codes?

henrik-me: not aware of a standard. but rules are, never re-use an id. it's ok to delete id's that are not used anymore, but it has to be ensured that id is not ever used by some other message (Wilson does this by commenting out the entry)

jennyf19: sounds good. thanks, henry

PR #287, opened 2020-07-02
Jmprieur/fix123

### LINKED ISSUE #123: [Feature Request] Simplify AuthenticationBuilder.AddSignIn extension method for web app
On the AddSignIn method signature https://github.com/AzureAD/microsoft-identity-web/blob/7d33e9f4351ac679382cd3c0f1bc61140376e34f/src/Microsoft.Identity.Web/WebAppAuthenticationBuilderExtensions.cs#L60-L66 is it necessary to have both the `configureOpenIdConnectOptions` and `configureMicrosoftIdentityOptions` parameters?

Given that `MicrosoftIdentityOptions` inherits from `OpenIdConnectOptions` I find this confusing as it is not clear which takes precedence or even if the inherited options must be set in the `configureOpenIdConnectOptions` delegate or if setting everything in the  `configureMicrosoftIdentityOptions` delegate is fine.

Remove the `configureOpenIdConnectOptions` parameter so there is a single point of configuration.

jmprieur:
Fixes #123 

Removes the need to specify the OpenIdConnectOptions in the Web apps calls. Instead the developer will only provide MicrosoftIdentityOptions (which derives from OpenIdConnectOptions)

I had to comment out 2 lines in the test, which I don't understand at all. @pmaytak if you could have a look at these, and propose the right fix if it's not deleting these lines

cc:  @henrik-me


src/Microsoft.Identity.Web/WebAppAuthenticationBuilderExtensions.cs
```diff
@@ -100,101 +95,175 @@ public static partial class WebAppAuthenticationBuilderExtensions
 
             builder.AddOpenIdConnect(openIdConnectScheme, options => { });
             builder.Services.AddOptions<OpenIdConnectOptions>(openIdConnectScheme)
-                .Configure<IServiceProvider>((options, serviceProvider) =>
+                .Configure<IServiceProvider, IOptions<MicrosoftIdentityOptions>>((options, serviceProvider, microsoftIdentityOptions) =>
```
pmaytak, line 98: Seems like this overload is good to use in other extension methods instead of explicitly calling GetRequiredService for options.

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
```diff
@@ -370,7 +368,7 @@ private async Task AddMicrosoftWebApp_TestRedirectToIdentityProviderEvent(IServi
             await oidcOptions.Events.RedirectToIdentityProvider(redirectContext).ConfigureAwait(false);
 
             // Assert properties set, events called
-            await redirectFunc.ReceivedWithAnyArgs().Invoke(Arg.Any<RedirectContext>()).ConfigureAwait(false);
+          //  await redirectFunc.ReceivedWithAnyArgs().Invoke(Arg.Any<RedirectContext>()).ConfigureAwait(false);
```
jmprieur, line 371: @pmaytak : do we need this line now?
If yes, how to fix it?

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
```diff
@@ -420,7 +418,7 @@ private async Task AddMicrosoftWebApp_TestB2cSpecificSetup(IServiceCollection se
             await oidcOptions.Events.RedirectToIdentityProvider(redirectContext).ConfigureAwait(false);
             await oidcOptions.Events.RemoteFailure(remoteFailureContext).ConfigureAwait(false);
 
-            await remoteFailureFuncMock.ReceivedWithAnyArgs().Invoke(Arg.Any<RemoteFailureContext>()).ConfigureAwait(false);
+ //           await remoteFailureFuncMock.ReceivedWithAnyArgs().Invoke(Arg.Any<RemoteFailureContext>()).ConfigureAwait(false);
```
jmprieur, line 421: @pmaytak : same thing

src/Microsoft.Identity.Web/WebAppAuthenticationBuilderExtensions.cs
```diff
@@ -100,101 +95,175 @@ public static partial class WebAppAuthenticationBuilderExtensions
...
            {
```
pmaytak, line 99: This brace needs to move one space to the right and then apply the VS formatting to fix the below spacing and indent changes.
```suggestion
            {
```

src/Microsoft.Identity.Web/WebAppAuthenticationBuilderExtensions.cs
```diff
@@ -100,101 +95,168 @@ public static partial class WebAppAuthenticationBuilderExtensions
...
+                   diags.Subscribe(options.Events);
+               }
+           });
 
             return builder;
         }
+
+        internal static void PopulateOpenIdOptionsFromMicrosoftIdentityOptions(OpenIdConnectOptions options, MicrosoftIdentityOptions microsoftIdentityOptions)
+        {
+            options.Authority = microsoftIdentityOptions.Authority;
```
pmaytak, line 197: Seems a bit too manual to copy each property but, I guess, based on Chris's [comment](https://github.com/AzureAD/microsoft-identity-web/issues/123#issuecomment-648925848) that's how we have to do...

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
pmaytak: Leave it commented for this PR and I will investigate and propose a different PR.

I would think it should work since we should be able to still resolve OIDC options. I will take a look afterwards.

src/Microsoft.Identity.Web/WebAppAuthenticationBuilderExtensions.cs
jmprieur: @pmaytak @jennyf19 
I've created a unit test (see below) to check if the properties get change in future versions of ASP.NET Core

src/Microsoft.Identity.Web/WebAppAuthenticationBuilderExtensions.cs
```diff
@@ -196,5 +190,76 @@ public static partial class WebAppAuthenticationBuilderExtensions
...
+            options.StateDataFormat = microsoftIdentityOptions.StateDataFormat;
+            options.StringDataFormat = microsoftIdentityOptions.StringDataFormat;
+            options.SecurityTokenValidator = microsoftIdentityOptions.SecurityTokenValidator;
+            options.TokenValidationParameters = microsoftIdentityOptions.TokenValidationParameters;
+            options.UseTokenLifetime = microsoftIdentityOptions.UseTokenLifetime;
+            options.SkipUnrecognizedRequests = microsoftIdentityOptions.SkipUnrecognizedRequests;
+            options.DisableTelemetry = microsoftIdentityOptions.DisableTelemetry;
+            options.NonceCookie = microsoftIdentityOptions.NonceCookie;
+            options.UsePkce = microsoftIdentityOptions.UsePkce;
+#if DOTNET_50
```
jmprieur, line 237: Fixing this as there were 2 additional properties in .NET 5 (discovered by the unit test)

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
```diff
@@ -526,5 +525,25 @@ private IConfigurationSection GetConfigSection(string configSectionName, bool in
             var configSection = configBuilder.Build().GetSection(configSectionName);
             return configSection;
         }
+
+        [Fact]
+        public void PreventChangesInOpenIdConnectOptionsToBeOverlooked()
```
jmprieur, line 530: This is the new unit test
with instructions of remediation when/if it fails

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
pmaytak: @jmprieur I sent a PR #291 to fix these tests. 
These tests and similar ones test that we call the custom event handlers set up by the dev.

jmprieur: Thanks @pmaytak

PR #291, opened 2020-07-06
Fix tests.

pmaytak:
Fixing the tests by calling the custom configure options properly.

These tests itself test that we call the custom event handlers set up by the dev.


jmprieur:
@pmaytak : I merged it so that I can merge my PR.
Thanks again!

PR #294, opened 2020-07-07
This is an improvement leveraging the fact that MSAL.NET now exposes …

jmprieur:
…TokenCacheNotificationArgs.HasTokens which enable applications (and Microsoft.Identity.Web) to remove cache entries when the cache is empty

Also OnBeforeAccessAsync tests if the SuggestedCacheKey is null (can happen in B2C cases, as we still call GetAccountsAsync()

### DIFF
--- a/tests/B2CWebAppCallsWebApi/Client/appsettings.json
+++ b/tests/B2CWebAppCallsWebApi/Client/appsettings.json
@@ -1,15 +1,15 @@
 {
-  "AzureAdB2C": {
-    "Instance": "https://fabrikamb2c.b2clogin.com",
-    "ClientId": "fdb91ff5-5ce6-41f3-bdbd-8267c817015d",
-    "Domain": "fabrikamb2c.onmicrosoft.com",
-    "SignedOutCallbackPath": "/signout/B2C_1_susi",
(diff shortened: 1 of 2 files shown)


tests/B2CWebAppCallsWebApi/Client/appsettings.json
```diff
@@ -1,15 +1,15 @@
...
-  },
+    "AzureAdB2C": {
+        "Instance": "https://fabrikamb2c.b2clogin.com",
+        "ClientId": "fdb91ff5-5ce6-41f3-bdbd-8267c817015d",
+        "Domain": "fabrikamb2c.onmicrosoft.com",
+        "SignedOutCallbackPath": "/signout/B2C_1_susi",
+        "SignUpSignInPolicyId": "b2c_1_susi",
+        "ResetPasswordPolicyId": "b2c_1_reset",
+        "EditProfilePolicyId": "b2c_1_edit_profile", // Optional profile editing policy
+        "ClientSecret": "X330F3#92!z614M4"
```
jennyf19, line 10:
```suggestion
        "ClientSecret": "[Copy the client secret added to the app from the Azure portal]"
```

PR #300, opened 2020-07-08
Jmprieur/templates with ms graph

jmprieur:


jmprieur:
@pmaytak @jennyf19 
merging (as this is the same as the previous template, and I broke the templates by my previous PR)
#elif instead of #elseif

PR #302, opened 2020-07-08
Use geographically neutral timeframe for expected GA

johndowns:
This PR clarifies the expected GA timeframe for those of us who are not in the United States or in the northern hemisphere :)


jennyf19:
Thank you @johndowns ! Appreciated

PR #308, opened 2020-07-09
add id prefix to error messages

jennyf19:
@henrik-me added the ID's you suggested to the error messages, but @jmprieur and I think it's too confusing to not have the error message included, and just the ID.


pmaytak:
@jennyf19 
What if we keep the variable names the same (just message), so it's easier to read in code and only add numbers to the text itself, so it's easier to search in the exception text? This way we get the benefits of both ways.

`ErrorMessage.ProvideEitherScopeKeySectionOrScopes`

`public const string ProvideEitherScopeKeySectionOrScopes = "IDW10101: Either provide the '{0}' or the '{1}' to the 'AuthorizeForScopes'. ";`

jennyf19:
sounds good.
@jmprieur can you take a look and sign-off, or not? thx.

---
In reply to: [445998438](https://github.com/AzureAD/microsoft-identity-web/pull/308#issuecomment-445998438) [](ancestors = 445998438)

PR #309, opened 2020-07-09
Made configure cookie action optional.

pmaytak:


pmaytak:
The [`AddCookie`](https://github.com/dotnet/aspnetcore/blob/5d170de7696694189bd531858c5f3410f1241975/src/Security/Authentication/Cookies/src/CookieExtensions.cs#L35) supports accepting null parameter and it checks for null action when it converts from `Action<CookieAuthenticationOptions>` to `Action<CookieAuthenticationOptions, IServiceProvider>`.

Issue #295, opened 2020-07-07
[Bug] Microsoft.Identity.Web should not call GetAccountsAsync()


jmprieur:
**Which Version of Microsoft Identity Web are you using ?**
Note that to get help, you need to run the latest version. 

**Where is the issue?**
B2C scenario

Other? - please describe;

**Repro**
TokenAcquisition calls GetAccountsAsync in the case of B2C to remove the accounts. However, this will call the token cache serialiation with no suggested cache key.

https://github.com/AzureAD/microsoft-identity-web/search?q=GetAccountsAsync&unscoped_q=GetAccountsAsync

**Expected behavior**
When signing out, the token for a user are removed from the cache

**Actual behavior**
The cache is not cleared as no account is found

**Possible Solution**
- We should probably think of adding all the tokens for all the policies of a given account in the same token cache, which we could clear at once.
 - then we should not call GetAccountsAsync(), but GetAccountAsync(account-without-policy)

**Additional context/ Logs / Screenshots**
Add any other context about the problem here, such as logs and screenshots.

jennyf19:
Part of this [pr](https://github.com/AzureAD/microsoft-identity-web/pull/331/)

pmaytak:
Included in [Microsoft Identity Web 0.2.1-preview release](https://github.com/AzureAD/microsoft-identity-web/releases/tag/0.2.1-preview).

Issue #27, opened 2020-02-19
ITokenAcquisition.GetTokenForUser() needs to take an optional userFlow parameter


jmprieur:
Why?
Today we don't specify in the controllers which policy/user flow to use to acquire the token, and therefore the B2C scenarios is not quite right.

What?
Expose a userFlow parameter and lookup the right account in the implementation

See also the 2 related issues

jennyf19:
moving some of the required cache look up to [MSAL .NET](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/1869)

jennyf19:
@jackheywood @grzegorzy @felickz

This work-in-progress [branch](https://github.com/AzureAD/microsoft-identity-web/tree/jennyf/b2cPolicyLookup) has some changes to how Microsoft Identity Web handles b2c, including a fix for this issue, which involves taking a userFlow parameter in `.GetTokenForUser()`. 
Each userFlow (policy) is a separate authorization server in AAD B2C, so they issue their own tokens. This means you'll have a variety of tokens per user in the cache. However, the web API is usually connected to the su_si policy, or another policy after the user has signed-in to the application.

In Microsoft Identity Web, we are using the HttpContext to determine the current userFlow, and using that userFlow . For example, I do su_si policy first, then I go to edit_profile. Then I hit "call the web API", Microsoft Identity Web will use the current userFlow, which is edit_profile,, which is incorrect because my web API is behind the su_si userFlow, we then catch an MsalUiRequiredException and use the su_si policy. However, If you want to specify the policy to use, you cannot do that, as mentioned [here](https://github.com/Azure-Samples/active-directory-aspnetcore-webapp-openidconnect-v2/issues/357). The branch linked above has the ability to specify the policy. 

You'll need to do two things to try it out.
- First, in the [TodoListController](https://github.com/AzureAD/microsoft-identity-web/blob/jennyf/b2cPolicyLookup/tests/B2CWebAppCallsWebApi/Client/Controllers/TodoListController.cs#L25), add in the userFlow value you want to target in the `AuthorizeForScopes` attribute.

```csharp
[AuthorizeForScopes(ScopeKeySection = "TodoList:TodoListScope", UserFlow = "b2c_1_susi")]
```
[...]

PR #324, opened 2020-07-15
Jennyf/b2c policy lookup

jennyf19:
Did manual testing, as we still haven't figured out how to mock the MSAL cache.
#27 #295 

**Manual Testing:**
test one | test two | 
--- | --- |
su_si | su_si
call api | edit_profile
sign-out | call api w/su_si
|| sign-out


jennyf19:
need to use this [msal .net build](https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=570140&view=artifacts&type=publishedArtifacts)

PR #330, opened 2020-07-17
fix b2c sign-in error

jennyf19:
I was getting build errors after creating the b2c web app, not sure if this is the right fix, as I also could not get the sign-in page to load. Also had build errors doing `dotnet build`, when trying to create just the web app.

And this:
image
In the --calls-graph one


jennyf19:
@jmprieur closing as you've already got the changes in the branch i reviewed.

Issue #274, opened 2020-06-30
[Bug] Token compatibility not handled correctly


nikkh:
**Which Version of Microsoft Identity Web are you using ?**
Microsoft.Identity.Web 0.1.2-preview

**Where is the issue?**
* Web App
    * [x] Sign-in users
    * [x] Sign-in users and call web APIs
* Web API
    * [ ] Protected web APIs (Validating tokens)
    * [ ] Protected web APIs (Validating scopes)
    * [ ] Protected web APIs call downstream web APIs
* Token cache serialization
     * [ ] In Memory caches
     * [ ] Session caches
     * [ ] Distributed caches

Other? - please describe;

**Is this a new or existing app?**
   This is a new app being built for production deployment.

**Repro**

I've just been asked to help an AAD B2C customer with a problem they have with logging on using AAD B2C in a dotnet core 3.1 web app they are developing.

Having run through their scenario in debug I have noticed that they have token compatibility set as described in https://docs.microsoft.com/en-us/azure/active-directory-b2c/tokens-overview#compatibility so that their issuer is https://xxxxxxxx.b2clogin.com/tfp/xxxxxxxx/xxxxxx_sign_up_in/v2.0/.

Looking at the source code in https://github.com/AzureAD/microsoft-identity-web/blob/master/src/Microsoft.Identity.Web/Resource/AadIssuerValidator.cs

```csharp
// The AAD "iss" claims contains the tenant ID in its value. The URI is {domain}/{tid}/v2.0
    private static string GetTenantIdFromIss(string iss)
    {
        if (string.IsNullOrEmpty(iss))
        {
            return string.Empty;
        }
        var uri = new Uri(iss);
        if (uri.Segments.Length > 1)
        {
            return uri.Segments[1].TrimEnd('/');
        }
        return string.Empty;
    }
```

**Expected behavior**
I expect that since the issuer is valid, the microsoft.identity.web package would enable sign in to complete normally.

**Actual behavior**
An exception is thrown indicating that the token issuer is invalid

**Possible Solution**
```
/if (uri.Segments.Length > 1)
        //{
        //    return uri.Segments[1].TrimEnd('/');
        //}

        if (uri.Segments.Length == 3)
        {
            return uri.Segments[1].TrimEnd('/');
        }
        if (uri.Segments.Length == 5)
        {
            return uri.Segments[2].TrimEnd('/');
        }
```

and 

```
[...]

PR #340, opened 2020-07-20
Fix validating B2C issuer with tfp in issuer URI.

pmaytak:
Issue #274 

Fix to validate the B2C issuer with `tfp` URI type (ex. `{domain}/{tfp}/{tid}/{userFlow}/v2.0/`) correctly.

More detail:
[B2C tokens compatibility](https://docs.microsoft.com/en-us/azure/active-directory-b2c/tokens-overview#compatibility)
[Configure B2C JWT token compatibility](https://docs.microsoft.com/en-us/azure/active-directory-b2c/configure-tokens#configure-jwt-token-compatibility)


jennyf19:
I agree, i'll ping b2c and see what they say about this. we don't want to necessarily support legacy stuff if we don't have to. i would like to get more context on that before moving forward.

pmaytak:
So, a couple of options:
1. Keep code as is - B2C tfp issuers with fail with generic invalid issuer exception.
2. Merge this PR - B2C tfp issuers will validate correctly.
3. At the end of the validate method, we can parse actual issuer URI, if it has tfp, then throw a more descriptive exception. 
https://github.com/AzureAD/microsoft-identity-web/blob/b694b720190fc2938ce58700b999ef35bf2ffd03/src/Microsoft.Identity.Web/Resource/AadIssuerValidator.cs#L128-L132

jmprieur:
@pmaytak. Given the reply from the B2C team and the customer, I think we could do 3.
- When you write parse, do you mean just verify that there is the /tpf/ string in it?
- We'd throw an exception explaining that this is not supported with an aka.ms link https://aka.ms/ms-id-web/b2c-issuer

B2C-Tfp-Issuer.md
```diff
@@ -0,0 +1,14 @@
+# Azure AD B2C issuer claim
```
pmaytak, line 1: Better title?

pmaytak:
@jmprieur @jennyf19 
Updated PR:
- throwing an exception with an aka.ms link.
- added B2C-Tfp-Issuer.md just to get feedback for the wiki (will not merge it)

B2C-Tfp-Issuer.md
jennyf19: i agree, i would put this in the wiki.

B2C-Tfp-Issuer.md
```diff
@@ -0,0 +1,14 @@
...
+
+Azure AD B2C user flows can be customized to emit issuer claim in two formats: `https://<domain>/<tenant>/v2.0/` and `https://<domain>/tfp/<tenant>/<userFlow>/v2.0/`. Originally `tfp` (which stands for Trust Framework Policy) in the URI was used to differentiate between Azure AD and Azure AD B2C . Currently the B2C team is working to streamline the experience and remove `tfp`. Therefore, the current guidance from Azure B2C team is to use `https://<domain>/<tenant>/v2.0/` format.
```
jennyf19, line 3: i don't think you need to explain this. Just mention we support the default value.

B2C-Tfp-Issuer.md
```diff
@@ -0,0 +1,14 @@
+# Azure AD B2C issuer claim
+
+Azure AD B2C user flows can be customized to emit issuer claim in two formats: `https://<domain>/<tenant>/v2.0/` and `https://<domain>/tfp/<tenant>/<userFlow>/v2.0/`. Originally `tfp` (which stands for Trust Framework Policy) in the URI was used to differentiate between Azure AD and Azure AD B2C . Currently the B2C team is working to streamline the experience and remove `tfp`. Therefore, the current guidance from Azure B2C team is to use `https://<domain>/<tenant>/v2.0/` format. 
+
+### Creating Azure B2C user flow
```
jennyf19, line 5: i'm not sure why we need this in our wiki, as it's in the b2c docs. plus user flow and policy are actually different, but we are using the term user flow to cover both, so i think this could lead to more confusion.

src/Microsoft.Identity.Web/Constants/IDWebErrorMessage.cs
```diff
@@ -32,6 +32,7 @@ internal static class IDWebErrorMessage
         public const string IssuerMetadataUrlIsRequired = "IDW10301: Azure AD Issuer metadata address URL is required. ";
         public const string NoMetadataDocumentRetrieverProvided = "IDW10302: No metadata document retriever is provided. ";
         public const string IssuerDoesNotMatchValidIssuers = "IDW10303: Issuer: '{0}', does not match any of the valid issuers provided for this application. ";
+        public const string B2CTfpIssuerNotSupported = "IDW10304: B2C issuer with 'tfp' in the URI is not supported. See https://aka.ms/ms-id-web/b2c-issuer for details.";
```
jennyf19, line 35: maybe something like "microsoft identity web does not support a B2C issuer with...."

B2C-Tfp-Issuer.md
pmaytak: Yes, it will be in the wiki. Added here just to get comments.

B2C-Tfp-Issuer.md
pmaytak: I'll remove this section.

I thought user flow is just a predefined policy, right?

jennyf19: you can have [custom policies](https://docs.microsoft.com/en-us/azure/active-directory-b2c/custom-policy-overview#comparing-user-flows-and-custom-policies), so we don't distinguish between the two, but b2c does.

B2C-Tfp-Issuer.md
pmaytak: @jennyf19 
I updated this wiki entry.

jennyf19:
let's not include the .md file for this. thx.

PR #349, opened 2020-07-22
Update to .NET Preview 7.

pmaytak:
Updates to [Preview 7](https://dotnet.microsoft.com/download/dotnet/5.0):
- JwtBearer and OpenIdConnect packages
- .NET version in the GitHub Build action

- This update adds 11 additional warnings related to nullable references.
- VS also seems to have trouble finding a specific version (3.7) of Microsoft.CodeAnalysis libraries in Preview 7 binaries.

### DIFF
--- a/src/Microsoft.Identity.Web/TokenCacheProviders/Session/SessionTokenCacheProviderExtension.cs
+++ b/src/Microsoft.Identity.Web/TokenCacheProviders/Session/SessionTokenCacheProviderExtension.cs
@@ -66,7 +66,16 @@ namespace Microsoft.Identity.Web.TokenCacheProviders.Session
 
             services.AddHttpContextAccessor();
             services.AddScoped<IMsalTokenCacheProvider, MsalSessionTokenCacheProvider>();
-            services.TryAddScoped(provider => provider.GetService<IHttpContextAccessor>().HttpContext.Session);
+            services.TryAddScoped(provider =>
+            {
+                var httpContext = provider.GetRequiredService<IHttpContextAccessor>().HttpContext;
+                if (httpContext == null)
+                {
+                    throw new InvalidOperationException(IDWebErrorMessage.HttpContextIsNull);
+                }
+
+                return httpContext.Session;
+            });
 
             return services;
         }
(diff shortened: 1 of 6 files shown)


jennyf19:
Will you fix the warnings?
would rather not merge something that might not work out of the box at this point.

pmaytak:
I'll fix in a bit. Didn't know if y'all wanted to have this merged in earlier for testing.

pmaytak:
Added fixes for warnings.

This Microsoft.CodeAnalysis version warnings seem to be coming from .NET Preview 7 compiler itself.
image

I created a new empty ASP.NET Core 3.1 web app without any additional Roslyn/FxCop analyzers. The warnings still appear. Seems like the compiler is looking for version 3.7 files in `C:\Program Files\dotnet\sdk\5.0.100-preview.7.20366.6\Sdks\Microsoft.NET.Sdk.Web\analyzers\cs\` directory. .NET 5 seems to point to [version 3.7.0-4.final](https://github.com/dotnet/aspnetcore/blob/8b7972064f068f76452eff7994263a5e4ee3ca46/eng/Versions.props#L219-L221). I tried mixing and matching lower and higher versions of packages, but it doesn't help. The only thing that fixed the warnings is overwriting the Preview 7 files in that directory with Preview 6 versions.

jmprieur:
My recommendation is to wait until Visual Studio supports Preview 7 completely, otherwise customers are going to have a hard time using Microsoft.Identity.Web, which we don't want;

src/Microsoft.Identity.Web/TokenCacheProviders/Session/SessionTokenCacheProviderExtension.cs
```diff
@@ -66,7 +66,16 @@ public static IServiceCollection AddSessionTokenCaches(this IServiceCollection s
 
             services.AddHttpContextAccessor();
             services.AddScoped<IMsalTokenCacheProvider, MsalSessionTokenCacheProvider>();
-            services.TryAddScoped(provider => provider.GetService<IHttpContextAccessor>().HttpContext.Session);
```
jmprieur, line 69: The fact of having a Scoped service does not work with the fact that for Microsoft Graph SDK, they require a singleton. I'm not sure we can do anything but warn developers that they cannot use a Session cache if they use the Microsoft Graph SDK ?

PR #362, opened 2020-07-23
Fix nullable warnings.

pmaytak:


src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -569,6 +569,11 @@ public async Task ReplyForbiddenWithWwwAuthenticateHeaderAsync(IEnumerable<strin
 
             string parameterString = string.Join(", ", parameters.Select(p => $"{p.Key}=\"{p.Value}\""));
 
+            if (CurrentHttpContext == null)
```
jennyf19, line 572: I don't think we want to do this, as for blazor, this will be null.

pmaytak: Good point. Suggestions? Without context, this method's kind of pointless. Will need a workaround for Blazor then, right?

jennyf19: that's what we've been working on.

src/Microsoft.Identity.Web/TokenCacheProviders/Session/SessionTokenCacheProviderExtension.cs
```diff
@@ -172,11 +194,20 @@ public static IServiceCollection AddSessionPerUserTokenCache(this IServiceCollec
...
             services.AddSession(option =>
-                {
-                    option.Cookie.IsEssential = true;
-                });
+            {
+                option.Cookie.IsEssential = true;
+            });
             services.AddScoped<IMsalTokenCacheProvider, MsalSessionTokenCacheProvider>();
-            services.TryAddScoped(provider => provider.GetService<IHttpContextAccessor>().HttpContext.Session);
+            services.TryAddScoped(provider =>
```
jmprieur, line 201: This won't work with Graph which is a singleton. Is there a possible mitigation?

src/Microsoft.Identity.Web/TokenAcquisition.cs
jmprieur: We now have a work around for Blazor, @pmaytak 
See [Managing incremental consent and conditional access](https://github.com/AzureAD/microsoft-identity-web/wiki/Managing-incremental-consent-and-conditional-access)

pmaytak: Created #380 to address this.

src/Microsoft.Identity.Web/TokenCacheProviders/Session/SessionTokenCacheProviderExtension.cs
pmaytak: > Because session token caches are added with scoped lifetime, they should not be used when `TokenAcquisition` is also used as a singleton (for example, when using Microsoft Graph SDK).

I added the above note to each method and to the [Token cache serialization](https://github.com/AzureAD/microsoft-identity-web/wiki/Token-cache-serialization) wiki.

In this method, I don't think we can reliably predict which services are added, since they can be modified in any order.

Maybe if `ServiceProvider` has some sort of validator or event that we can plug into, we could validate at that time.

pmaytak:
@jennyf19 @jmprieur FYI updated this PR.

Issue #20, opened 2020-02-18
RemoveAccount should remove all the policies for a given account from the cache


jmprieur:
from: @TiagoBrenck :
For B2C, the remove account might have a bug. Since one user can have many IAccounts (because of the policies), we need to find a solution that gets all of the accounts of that user (regardless the policy) and remove from the cache. The current implementation is not doing it.

jennyf19:
@jmprieur I checked the MSAL .NET code and they are removing all the accounts, regardless of policy. we do not enable removal of an account based on a policy. I will take a look at MS Identity Web's implementation next.

jennyf19:
I don't see multiple b2c accounts, even when targeting multiple policies, so on the surface, this seems to be a non-issue, as we do remove all the b2c accounts, but we only have one, which is not right. will require further investigation.

jmprieur:
@jennyf19 : this might be hard to do as we don't know how policies will have been used by a user (and the token caches are per policies).
If we do, we should have a cacheKey based only on the oid-without-policy and tid, and then all the policies for the same user would be in the same token cache.

I wonder if we really want to do that.

jennyf19:
pending work here in [msal .net](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/1940)

PR #363, opened 2020-07-23
0.2.1-preview changelog

pmaytak:
@jennyf19 Can you please suggest a summary for #27, #20, #295? I wasn't sure how to phrase them correctly.

Also since we added optional parameters to methods in `TokenAcquisition`, that makes it a runtime breaking change, right? Should I add it to the breaking changes section?

### DIFF
--- a/changelog.txt
+++ b/changelog.txt
@@ -1,18 +1,18 @@
 0.2.1-preview
 ============
-### New Features:
+### New features:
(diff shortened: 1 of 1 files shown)


pmaytak:
The [work](https://github.com/AzureAD/microsoft-identity-web/milestone/11) was related to B2C user flows/accounts, null HttpContext in TokenAcquisition, and templates.

Issue #389, opened 2020-08-02
[Bug] VerifyUserHasAnyAcceptedScope should throw UnauthorizedAccessException(message) (like roles)


jmprieur:
**Which version of Microsoft Identity Web are you using?**
0.2.1-preview

**Where is the issue?**
* Web app
    * [ ] Sign-in users
    * [ ] Sign-in users and call web APIs
* Web API
    * [x  ] Protected web APIs (validating tokens)
    * [ ] Protected web APIs (validating scopes)
    * [ ] Protected web APIs call downstream web APIs
* Token cache serialization
     * [ ] In-memory caches
     * [ ] Session caches
     * [ ] Distributed caches
* Other (please describe)

**Repro**

In **tests\WebAppCallsWebApiCallsGraph\TodoListService\Controllers\TodoListController.cs**, change line 22 to be

```csharp
static readonly string[] scopeRequiredByApi = new string[] { "access_as_users" };
```

Runt the Web app and Web API

**Expected behavior**
The scope being invalid, the Web API should not process anything after calling `HttpContext.VerifyUserHasAnyAcceptedScope(scopeRequiredByApi)`

**Actual behavior**
The error is written in the HttpResponse, but the processing in the controller action continues

**Possible solution**
throw UnauthorizedAccessException(message) like what was done for roles in [jennyf/roles](https://github.com/AzureAD/microsoft-identity-web/pull/386) for app roles

@jennyf19 : pre-assigning it to you as you've done the roles.

jennyf19:
Included in [0.2.2-preview release](https://github.com/AzureAD/microsoft-identity-web/releases/tag/0.2.2-preview)

PR #393, opened 2020-08-03
throw if scopes are not valid

jennyf19:
#389


src/Microsoft.Identity.Web/Resource/ScopesRequiredHttpContextExtensions.cs
```diff
@@ -60,6 +61,7 @@ public static void VerifyUserHasAnyAcceptedScope(this HttpContext context, param
                     context.Response.StatusCode = (int)HttpStatusCode.Forbidden;
                     string message = string.Format(CultureInfo.InvariantCulture, IDWebErrorMessage.MissingScopes, string.Join(",", acceptedScopes));
                     context.Response.WriteAsync(message);
```
pmaytak, line 63: Should this be removed, since the exception will have the message and end the response?

jmprieur: The message is sent to the client (on the client side).
The exception only affect the service side
I would leave it (that would also be a breaking change for our customers)

Issue #398, opened 2020-08-04
AuthorizeForScopesAttribute not detecting MsalUiRequiredException


cbrianball:
**Which version of Microsoft Identity Web are you using?**
Note that to get help, you need to run the latest version. 
v0.2.1-preview

**Where is the issue?**
* Web app
    * [ ] Sign-in users
    * [x] Sign-in users and call web APIs
* Web API
    * [ ] Protected web APIs (validating tokens)
    * [ ] Protected web APIs (validating scopes)
    * [ ] Protected web APIs call downstream web APIs
* Token cache serialization
     * [ ] In-memory caches
     * [ ] Session caches
     * [ ] Distributed caches
* Other (please describe)

**Is this a new or an existing app?**
Adding this library to an existing application. The version of this application that is in production is not using this library yet.

**Expected behavior**
The token has expired (or has been deleted) from the distributed cache. On the next request where the token is needed, the user should be redirected through the authentication flow (since a `MsalUiRequiredException` is thrown).

**Actual behavior**
The exception passes through the `AuthorizeForScopesAttribute` and is not acted upon

**Possible solution**
The current implementation of `AuthroizeForScopesAttribute` only looks at the current exception and its inner exception. It should look through all of the nested exceptions to determine if it is caused by `MsalUiRequiredException`.

I can write my own attribute that inherits from `AuthorizeForScopesAttribute`, perform the test myself, then call the base method (making sure the `MsalUiException` is set to the `context.Exception` property), and everything works as expected.

**Additional context / logs / screenshots**
![image](https://user-images.githubusercontent.com/640035/89328791-e6c10980-d65b-11ea-9be7-7af9b8633f21.png)

The order of the screenshot is the reverse nested order (the first exception listed is the innermost exception).

I can submit a PR if that would help, but wanted to double-check to make sure I wasn't doing anything wrong first.

jmprieur:
Thanks for reporting @cbrianball 
It is fixed and we'll release an update of Microsoft.Identity.Web with many bug fixes. No ETA yet as we need to release MSAL.NET first (which also has a fix for us)

jennyf19:
Included in [0.2.2-preview release](https://github.com/AzureAD/microsoft-identity-web/releases/tag/0.2.2-preview)

PR #399, opened 2020-08-04
Fix for https://github.com/AzureAD/microsoft-identity-web/issues/398

jmprieur:


src/Microsoft.Identity.Web/AuthorizeForScopesAttribute.cs
```diff
@@ -49,10 +49,7 @@ public override void OnException(ExceptionContext context)
         {
             if (context != null)
             {
-                MsalUiRequiredException? msalUiRequiredException =
-                    (context.Exception as MsalUiRequiredException)
-                    ?? (context.Exception?.InnerException as MsalUiRequiredException);
-
+                MsalUiRequiredException? msalUiRequiredException = FindMsalUiRequiredExceptionIfAny(context.Exception);
```
jmprieur, line 52: Customers rethrow exceptions and pass Microsoft.Identity.Web exceptions in as inner exceptions.
In 0.2.1-preview, we've added the `MicrosoftIdentityWebChallengeUserException`, but we overlooked the scenario where it would be nested (for instance by the Graph SDK)

Issue #392, opened 2020-08-03
[Bug] NullReferenceException when using Azure SignalR in server-side Blazor app


DanielMannInCycle:
**Which version of Microsoft Identity Web are you using?**
Note that to get help, you need to run the latest version. 

0.2.1-preview

**Where is the issue?**
* Web app
    * [ ] Sign-in users
    * [x] Sign-in users and call web APIs
* Web API
    * [ ] Protected web APIs (validating tokens)
    * [ ] Protected web APIs (validating scopes)
    * [ ] Protected web APIs call downstream web APIs
* Token cache serialization
     * [ ] In-memory caches
     * [ ] Session caches
     * [ ] Distributed caches
* Other (please describe)

**Is this a new or an existing app?**

This is an existing application. I upgraded to the latest version of the Microsoft Identity Web library. It works with self-hosted SignalR, but fails when integrated with Azure SignalR, presumably due to missing `HttpContext`.

**Repro**
I don't have a minimal repro. I am currently attempting to put one together; I will update this issue with a link to a repo if I'm successful.

**Expected behavior**
Azure SignalR works.

**Actual behavior**
NullReferenceException is thrown 

**Additional context / logs / screenshots**
We are registering the identity provider as follows:
Startup.cs:
```
public void ConfigureServices(IServiceCollection services)
        {
            services.AddOptions();

            _authenticationStrategy.AddAuthentication(services, Configuration);
            services.AddTransient<IAuthenticationStrategy>(provider => _authenticationStrategy);

            if (!Env.IsDevelopment() || !string.IsNullOrEmpty(Configuration.GetValue<string>("MsalDistributedTokenCache")))
            {
                services.AddStackExchangeRedisCache(options =>
                {
                    options.Configuration = Configuration.GetValue<string>("MsalDistributedTokenCache");
                });
                services.AddDistributedTokenCaches();
            }

            services.AddControllersWithViews(options =>
            {
                var policy = new AuthorizationPolicyBuilder()
                    .RequireAuthenticatedUser()
                    .Build();

                options.Filters.Add(new AuthorizeFilter(policy));
            })
                .AddMicrosoftIdentityUI();

            services.AddAuthorizationCore(authzOptions =>
            {
                foreach (var policyConfig in Configuration.GetSection("ElevatePolicies").GetChildren())
                {
                    var policyMapping = policyConfig.Get<AuthorizationPolicyGroupMapping>();
[...]

PR #407, opened 2020-08-05
fix blazor null httpcontext when using Azure SignalR

jennyf19:
fix for #392 
cc: @DanielMannInCycle @javiercn

### DIFF
--- a/src/Microsoft.Identity.Web/Constants/IDWebErrorMessage.cs
+++ b/src/Microsoft.Identity.Web/Constants/IDWebErrorMessage.cs
@@ -28,6 +28,8 @@ namespace Microsoft.Identity.Web
         public const string MissingRoles = "IDW10202:The 'roles' or 'role' claim does not contain roles '{0}' or was not found. ";
         public const string MissingScopes = "IDW10203:The 'scope' or 'scp' claim does not contain scopes '{0}' or was not found. ";
         public const string UnauthenticatedUser = "IDW10204:The user is unauthenticated. The HttpContext does not contain any claims. ";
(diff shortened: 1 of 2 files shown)


DanielMannInCycle:
When I integrated this code into my repro case, I started receiving this exception:
`MsalUiRequiredException: No account or login hint was passed to the AcquireTokenSilent call.`

When I wrapped the calls in a `try`/`catch` that calls `HandleException`, I get a `NullReferenceException`: >	```Microsoft.Identity.Web.dll!Microsoft.Identity.Web.MicrosoftIdentityConsentAndConditionalAccessHandler.HandleException(System.Exception exception) Line 140```

jennyf19:
@DanielMannInCycle ok. thank you. i think the NavigationManager is null, so it's throwing a null ref here:
`string redirectUri = NavigationManager.Uri;`

Will take a look.

DanielMannInCycle:
@jennyf19 This PR was merged with the issue I reported still occurring -- should I open a new issue to track that behavior?

jennyf19:
@DanielMannInCycle no...we are still working on that issue as well. I think we're getting further along. If needed, we will open a new issue w/details and let you know.

josheinstein:
In stepping through the debugger, it seems that MicrosoftIdentityConsentAndConditionalAccessHandler is trying to access NavigationManager before the MicrosoftIdentityServiceHandler has been instantiated, thus NavigationManager is still null. Can't NavigationManager be injected into the constructor instead of waiting for the circuit handler to set it? (I'm very new to Blazor, so my apologies if that's a stupid question.)

Issue #408, opened 2020-08-05
[Bug] Set name claim in Web API scenario, like is done in web app.


jennyf19:
like this in web app:
```csharp
Missing                 
// B2C doesn't have preferred_username claims 
 
              if (microsoftIdentityOptions.Value.IsB2C)
                {
                    options.TokenValidationParameters.NameClaimType = ClaimConstants.Name;
                }
                else
                {
                    options.TokenValidationParameters.NameClaimType = ClaimConstants.PreferredUserName;
                }
```
do it in web api as well, as in the web API controller:
```csharp
// GET: api/values
        [HttpGet]
        public async Task<IEnumerable<Todo>> GetAsync()
        {
            HttpContext.VerifyUserHasAnyAcceptedScope(scopeRequiredByApi);
            string owner = User.Identity.Name;

```

`User.Identity.Name` is null, there is no value, so we need to set it.__

Update. Based in @pmaytak's PR feedback, we decided to not take this issue now, as it would not work for v1.0 tokens.
but to change the test app to use .GetDisplayName() 

We'll take this issue later

jennyf19:
used .GetDisplayName() instead.

PR #409, opened 2020-08-05
set name claim in web api

jennyf19:
#408 found while testing the OBO stuff.

cc: @jmprieur


jennyf19:
@jmprieur I didn't check that yet. will do.

jmprieur:
Do you mind proposing a better fix, @pmaytak  ?

jennyf19:
@jmprieur I'll just use `.GetDispalyName()` in the controller and we can revisit this later. it's not high-priority.

jmprieur:
Good idea, @jennyf19 
thanks for the heads-up @pmaytak.

let's take the time to do the right thing here (not needed for this release as there is a good solution for the test apps/customer apps)

pmaytak:
@jennyf19 @jmprieur Yes, I'll create a lower priority issue to revisit this.

jennyf19:
I had already created an issue here: https://github.com/AzureAD/microsoft-identity-web/issues/408 and moved it to a general project board under a later milestone.

Issue #380, opened 2020-07-31
Handle possible null HttpContext in ReplyForbiddenWithWwwAuthenticateHeaderAsync


pmaytak:
Come up with a solution to handle possible null HttpContext in ReplyForbiddenWithWwwAuthenticateHeaderAsync in TokenAcquisition class.

https://github.com/AzureAD/microsoft-identity-web/blob/3c77e3c8f6fe1f8adad523b3278615a3585b8755/src/Microsoft.Identity.Web/TokenAcquisition.cs#L578-L582

jmprieur:
We should have an extra parameter with a default value of null.
If passed-in we'd use it in the method, otherwise we'd use the HttpContextAccessor as today.

```CSharp
public async Task ReplyForbiddenWithWwwAuthenticateHeaderAsync(IEnumerable<string> scopes, MsalUiRequiredException msalServiceException, HttpContext httpContext = null)
```

Alternatively we could just pass-in the HttpResponse:
1. Pro this is all it needs
2. in term of usability this could be more complicated for customers to find they need to use HttpContext.Reponse ?

@jennyf19 @pmaytak what do you think?

jmprieur:
@pmaytak
Let's go with providing the HttpResponse as a parameter

This will also fix Azure devops [Bug 1099665: Defect : CS8602, Component : src\Microsoft.Identity.Web\TokenAcquisition.cs (1 issue)](https://identitydivision.visualstudio.com/IDDP/_workitems/edit/1099665) and therefore assigning to you.

jmprieur:
Updated the wiki:
https://github.com/AzureAD/microsoft-identity-web/wiki/Managing-incremental-consent-and-conditional-access#handling-incremental-consent-or-conditional-access-in-web-apis

cc: @pmaytak @jennyf19

PR #412, opened 2020-08-06
Add optional HttpResponse parameter to ReplyForbiddenWithWwwAuthenticateHeaderAsync. Handle null context.

pmaytak:
Issue #380

### DIFF
--- a/src/Microsoft.Identity.Web/ITokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web/ITokenAcquisition.cs
@@ -50,12 +50,12 @@ namespace Microsoft.Identity.Web
         /// the client can trigger an interaction with the user so the user can consent to more scopes.
         /// </summary>
         /// <param name="scopes">Scopes to consent to.</param>
-        /// <param name="msalSeviceException"><see cref="MsalUiRequiredException"/> triggering the challenge.</param>
+        /// <param name="msalServiceException"><see cref="MsalUiRequiredException"/> triggering the challenge.</param>
         /// <param name="httpResponse">The <see cref="HttpResponse"/> to update.</param>
         /// <returns>A <see cref="Task"/> representing the asynchronous operation.</returns>
         Task ReplyForbiddenWithWwwAuthenticateHeaderAsync(
             IEnumerable<string> scopes,
-            MsalUiRequiredException msalSeviceException,
+            MsalUiRequiredException msalServiceException,
             HttpResponse? httpResponse = null);
     }
 }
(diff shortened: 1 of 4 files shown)


src/Microsoft.Identity.Web/ITokenAcquisition.cs
```diff
@@ -51,9 +51,11 @@ public interface ITokenAcquisition
         /// </summary>
         /// <param name="scopes">Scopes to consent to.</param>
         /// <param name="msalSeviceException"><see cref="MsalUiRequiredException"/> triggering the challenge.</param>
+        /// <param name="httpResponse">The <see cref="HttpResponse"/> to update.</param>
         /// <returns>A <see cref="Task"/> representing the asynchronous operation.</returns>
         Task ReplyForbiddenWithWwwAuthenticateHeaderAsync(
             IEnumerable<string> scopes,
-            MsalUiRequiredException msalSeviceException);
+            MsalUiRequiredException msalSeviceException,
```
jmprieur, line 58: there is a typo (pre-existing). msalSeviceException => msalServiceException.

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -533,8 +533,9 @@ private async Task<IConfidentialClientApplication> BuildConfidentialClientApplic
         /// </summary>
         /// <param name="scopes">Scopes to consent to.</param>
         /// <param name="msalServiceException">The <see cref="MsalUiRequiredException"/> that triggered the challenge.</param>
```
jmprieur, line 535: same thing here.

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -562,9 +563,10 @@ public async Task ReplyForbiddenWithWwwAuthenticateHeaderAsync(IEnumerable<strin
 
             string parameterString = string.Join(", ", parameters.Select(p => $"{p.Key}=\"{p.Value}\""));
 
-            if (CurrentHttpContext != null)
+            httpResponse ??= CurrentHttpContext?.Response;
+
+            if (httpResponse != null)
```
jmprieur, line 568: if the httpResponse is null, do we want to throw with an aka.ms link?

src/Microsoft.Identity.Web/ITokenAcquisition.cs
pmaytak: @jmprieur Do we want to include this breaking change into 0.2.2 or into 0.3.0?

jmprieur: you mean the name of the parameter?
I would think yes (we are in preview and I doubt many people would have used the named: way of calling the API.
I can take the risk @pmaytak

src/Microsoft.Identity.Web/TokenAcquisition.cs
pmaytak: @jmprieur That was my initial proposal - throw an Exception, something like InvalidOperationException. I think just a descriptive message will suffice (ex. HttpContext is null, pass HttpResponse as a parameter). What's your idea of what to put into a wiki article about this?

jmprieur: @pmaytak : Let's throw an InvalidOperationException with the message your proposed. Good idea!

Issue #413, opened 2020-08-06
[Feature Request] Allow to pass tenantId when requesting application access token


stvonolf:
**Is your feature request related to a problem? Please describe.**
We would use a multi-tenant application which is authorized in several tenants. In each tenant, the application would be added in the Dynamics CRM as an app user. For each of those, we need to request the token using the tenant authority.

**Describe the solution you'd like**
Add tenantId parameter to GetAccessTokenForAppAsync - similar as in GetAccessTokenForUserAsync

**Describe alternatives you've considered**
n/a

**Additional context**
https://github.com/AzureAD/microsoft-identity-web/blob/master/src/Microsoft.Identity.Web/TokenAcquisition.cs 

__

Addition. Given that for client credentials, we can only pass one scope (`resource/.default`), we'd want to change the signature to pass only one string (instead of a collection)

=> This is a breaking change. So proposing to take it (PM would say easy) in 0.3.0-preview
cc: @jennyf19 @henrik-me @pmaytak

jmprieur:
@stvonolf how urgent is it for you?

stvonolf:
If this is a use-case you want to support, it would be good to have it until the end of August if possible.

stvonolf:
We should clarify first though, if requesting application tokens for multiple tenants is considered an anti-pattern.

jmprieur:
Checked with @hpsin and here is what we agreed to:
- Change the signature of `GetAccessTokenForUserAsync` to take a `string` (instead of a `IEnumerable<string>`) as there is only one possible string for a given resource of App Id URI AppIdUri: "AppIdUri/.default". We can check that the resource ends in "./default"
- Add an additional optional parameter `tenant` to support this scenario, and verify that this tenant is not organizations (and of course common and consumers, which don't make sense)

jmprieur:
Taking it as there is an API breaking change so I'd like it to be in 0.3.0-preview

PR #424, opened 2020-08-09
Make GetTokenForAppAsync less confusing and allow to pass tenantId #413

jmprieur:
Checked with @hpsin and here is what we agreed to:
- Change the signature of `GetAccessTokenForUserAsync` to take a `string` (instead of a `IEnumerable<string>`) as there is only one possible string for a given resource of App Id URI AppIdUri: "AppIdUri/.default". Check that the resource ends in "./default"
- Add an additional optional parameter `tenant` to support this scenario, and verify that this tenant is not organizations (and of course common and consumers, which don't make sense)

```CSharp
public async Task<string> GetAccessTokenForAppAsync(string scope, string? tenant = null)
```

### DIFF
--- a/src/Microsoft.Identity.Web/Constants/IDWebErrorMessage.cs
+++ b/src/Microsoft.Identity.Web/Constants/IDWebErrorMessage.cs
@@ -45,7 +45,7 @@ namespace Microsoft.Identity.Web
         public const string ClientCredentialScopeParameterShouldEndInDotDefault =
        "IDW10404: 'scope' parameter should be of the form 'AppIdUri/.default'. See https://aka.ms/ms-id-web/daemon-scenarios.";
         public const string ClientCredentialTenantShouldBeTenanted =
-       "IDW10405: 'tenant' parameter should be a tenant ID or domain name, not 'common', 'organizations' or 'consumers'. See https://aka.ms/ms-id-web/daemon-scenarios.";
+       "IDW10405: 'tenant' parameter should be a tenant ID or domain name, not 'common', or 'organizations'. See https://aka.ms/ms-id-web/daemon-scenarios.";
 
         // MSAL IDW10500 = "IDW10500:"
         public const string ExceptionAcquiringTokenForConfidentialClient = "IDW10501: Exception acquiring token for a confidential client. ";
--- a/src/Microsoft.Identity.Web/Constants/IDWebErrorMessage.cs
+++ b/src/Microsoft.Identity.Web/Constants/IDWebErrorMessage.cs
@@ -42,6 +42,10 @@ namespace Microsoft.Identity.Web
         public const string TenantIdClaimNotPresentInToken = "IDW10401: Neither `tid` nor `tenantId` claim is present in the token obtained from Microsoft identity platform. ";
         public const string ClientInfoReturnedFromServerIsNull = "IDW10402: Client info returned from the server is null. ";
         public const string TokenIsNotJwtToken = "IDW10403: Token is not JWT token. ";
+        public const string ClientCredentialScopeParameterShouldEndInDotDefault =
+       "IDW10404: 'scope' parameter should be of the form 'AppIdUri/.default'.";
+        public const string ClientCredentialTenantShouldBeTenanted =
+       "IDW10405: 'tenant' parameter should be a tenant ID or domain name, not 'common', 'organizations' or 'consumers'.";
 
         // MSAL IDW10500 = "IDW10500:"
         public const string ExceptionAcquiringTokenForConfidentialClient = "IDW10501: Exception acquiring token for a confidential client. ";
--- a/src/Microsoft.Identity.Web/Constants/IDWebErrorMessage.cs
+++ b/src/Microsoft.Identity.Web/Constants/IDWebErrorMessage.cs
@@ -43,9 +43,9 @@ namespace Microsoft.Identity.Web
         public const string ClientInfoReturnedFromServerIsNull = "IDW10402: Client info returned from the server is null. ";
         public const string TokenIsNotJwtToken = "IDW10403: Token is not JWT token. ";
         public const string ClientCredentialScopeParameterShouldEndInDotDefault =
-       "IDW10404: 'scope' parameter should be of the form 'AppIdUri/.default'.";
+       "IDW10404: 'scope' parameter should be of the form 'AppIdUri/.default'. See https://aka.ms/ms-id-web/daemon-scenarios.";
         public const string ClientCredentialTenantShouldBeTenanted =
-       "IDW10405: 'tenant' parameter should be a tenant ID or domain name, not 'common', 'organizations' or 'consumers'.";
+       "IDW10405: 'tenant' parameter should be a tenant ID or domain name, not 'common', 'organizations' or 'consumers'. See https://aka.ms/ms-id-web/daemon-scenarios.";
 
         // MSAL IDW10500 = "IDW10500:"
         public const string ExceptionAcquiringTokenForConfidentialClient = "IDW10501: Exception acquiring token for a confidential client. ";
--- a/src/Microsoft.Identity.Web/ITokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web/ITokenAcquisition.cs
@@ -37,12 +37,15 @@ namespace Microsoft.Identity.Web
         /// Acquires a token from the authority configured in the app, for the confidential client itself (not on behalf of a user)
         /// using the client credentials flow. See https://aka.ms/msal-net-client-credentials.
         /// </summary>
-        /// <param name="scopes">scopes requested to access a protected API. For this flow (client credentials), the scopes
+        /// <param name="scope">scope requested to access a protected API. For this flow (client credentials), the scope
         /// should be of the form "{ResourceIdUri/.default}" for instance <c>https://management.azure.net/.default</c> or, for Microsoft
         /// Graph, <c>https://graph.microsoft.com/.default</c> as the requested scopes are defined statically with the application registration
-        /// in the portal, and cannot be overridden in the application.</param>
+        /// in the portal, cannot be overridden in the application, and you can request a token for only one resource at a time (use
+        /// several call to get tokens for other resources).</param>
+        /// <param name="tenant">Enables overriding of the tenant/account for the same identity. This is useful in the
+        /// cases where a given account is guest in other tenants, and you want to acquire tokens for a specific tenant, like where the user is a guest in.</param>
         /// <returns>An access token for the app itself, based on its scopes.</returns>
-        Task<string> GetAccessTokenForAppAsync(IEnumerable<string> scopes);
+        Task<string> GetAccessTokenForAppAsync(string scope, string? tenant);
 
         /// <summary>
         /// Used in Web APIs (which therefore cannot have an interaction with the user).
--- a/src/Microsoft.Identity.Web/TokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web/TokenAcquisition.cs
@@ -84,12 +84,14 @@ namespace Microsoft.Identity.Web
         /// <summary>
         /// meta-tenant identifiers which are not allowed in client credentials.
         /// </summary>
-        private readonly string[] unwantedTenantIdentifiersInClientCredentials = new string[]
-        {
-            "common",
-            "organizations",
-            "consumers",
-        };
+        private readonly ISet<string> _metaTenantIdentifiers = new HashSet<string>(
+            new[]
+            {
+                Constants.Common,
+                Constants.Organizations,
+                Constants.Consumers,
+            },
+            StringComparer.OrdinalIgnoreCase);
 
         /// <summary>
         /// This handler is executed after the authorization code is received (once the user signs-in and consents) during the
@@ -313,7 +315,7 @@ namespace Microsoft.Identity.Web
         /// <returns>An access token for the app itself, based on its scopes.</returns>
         public async Task<string> GetAccessTokenForAppAsync(string scope, string? tenant = null)
         {
-            if (scope == null)
+            if (string.IsNullOrEmpty(scope))
             {
                 throw new ArgumentNullException(nameof(scope));
             }
@@ -323,8 +325,7 @@ namespace Microsoft.Identity.Web
                 throw new ArgumentException(IDWebErrorMessage.ClientCredentialScopeParameterShouldEndInDotDefault, nameof(scope));
             }
 
-            if (!string.IsNullOrEmpty(tenant)
-                && unwantedTenantIdentifiersInClientCredentials.Any(u => u.Equals(tenant, StringComparison.InvariantCultureIgnoreCase)))
+            if (!string.IsNullOrEmpty(tenant) && _metaTenantIdentifiers.Contains(tenant))
             {
                 throw new ArgumentException(IDWebErrorMessage.ClientCredentialTenantShouldBeTenanted, nameof(tenant));
             }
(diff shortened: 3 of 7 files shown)


src/Microsoft.Identity.Web/ITokenAcquisition.cs
```diff
@@ -37,12 +37,15 @@ public interface ITokenAcquisition
         /// Acquires a token from the authority configured in the app, for the confidential client itself (not on behalf of a user)
         /// using the client credentials flow. See https://aka.ms/msal-net-client-credentials.
         /// </summary>
-        /// <param name="scopes">scopes requested to access a protected API. For this flow (client credentials), the scopes
+        /// <param name="scope">scope requested to access a protected API. For this flow (client credentials), the scope
```
jennyf19, line 40:
```suggestion
        /// <param name="scope">The scope requested to access a protected API. For this flow (client credentials), the scope
```

src/Microsoft.Identity.Web/ITokenAcquisition.cs
```diff
@@ -37,12 +37,15 @@ public interface ITokenAcquisition
...
         /// should be of the form "{ResourceIdUri/.default}" for instance <c>https://management.azure.net/.default</c> or, for Microsoft
         /// Graph, <c>https://graph.microsoft.com/.default</c> as the requested scopes are defined statically with the application registration
-        /// in the portal, and cannot be overridden in the application.</param>
+        /// in the portal, cannot be overridden in the application, and you can request a token for only one resource at a time (use
```
jennyf19, line 43:
```suggestion
        /// in the portal, cannot be overridden in the application, as you can request a token for only one resource at a time (use
```

src/Microsoft.Identity.Web/ITokenAcquisition.cs
```diff
@@ -37,12 +37,15 @@ public interface ITokenAcquisition
...
+        /// several call to get tokens for other resources).</param>
```
jennyf19, line 44:
```suggestion
        /// several calls to get tokens for other resources).</param>
```

src/Microsoft.Identity.Web/ITokenAcquisition.cs
```diff
@@ -37,12 +37,15 @@ public interface ITokenAcquisition
...
+        /// <param name="tenant">Enables overriding of the tenant/account for the same identity. This is useful in the
+        /// cases where a given account is guest in other tenants, and you want to acquire tokens for a specific tenant, like where the user is a guest in.</param>
```
jennyf19, line 46:
```suggestion
        /// cases where a given account is a guest in other tenants, and you want to acquire tokens for a specific tenant.</param>
```

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -81,6 +81,16 @@ internal class TokenAcquisition : ITokenAcquisition, ITokenAcquisitionInternal
             OidcConstants.ScopeOfflineAccess,
         };
 
+        /// <summary>
+        /// meta-tenant identifiers which are not allowed in client credentials.
+        /// </summary>
+        private readonly string[] unwantedTenantIdentifiersInClientCredentials = new string[]
+        {
```
jennyf19, line 88: nit: start w/`_` how about `_metaTenantIdentifiers` or `_invalidTenantIdentifiersInClientCredentials` or ...?
Also, we have a const file, so I would put these there as const values, we might use them other places and those can be replaced w/the constant value too.

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -293,25 +303,41 @@ internal class TokenAcquisition : ITokenAcquisition, ITokenAcquisitionInternal
...
+        /// several call to get tokens for other resources).</param>
+        /// <param name="tenant">Enables overriding of the tenant/account for the same identity. This is useful in the
+        /// cases where a given account is guest in other tenants, and you want to acquire tokens for a specific tenant, like where the user is a guest in.</param>
         /// <returns>An access token for the app itself, based on its scopes.</returns>
-        public async Task<string> GetAccessTokenForAppAsync(IEnumerable<string> scopes)
+        public async Task<string> GetAccessTokenForAppAsync(string scope, string? tenant = null)
         {
-            if (scopes == null)
+            if (scope == null)
             {
```
jennyf19, line 317: nit: you can do string.IsNullOrEmpty(scope) now

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -293,25 +303,41 @@ internal class TokenAcquisition : ITokenAcquisition, ITokenAcquisitionInternal
...
-                throw new ArgumentNullException(nameof(scopes));
+                throw new ArgumentNullException(nameof(scope));
+            }
+
+            if (!scope.EndsWith("/.default", true, CultureInfo.InvariantCulture))
+            {
```
jennyf19, line 322: nit: do you need the "/" or is ".default" enough?

src/Microsoft.Identity.Web/Constants/IDWebErrorMessage.cs
```diff
@@ -42,6 +42,10 @@ internal static class IDWebErrorMessage
         public const string TenantIdClaimNotPresentInToken = "IDW10401: Neither `tid` nor `tenantId` claim is present in the token obtained from Microsoft identity platform. ";
         public const string ClientInfoReturnedFromServerIsNull = "IDW10402: Client info returned from the server is null. ";
         public const string TokenIsNotJwtToken = "IDW10403: Token is not JWT token. ";
+        public const string ClientCredentialScopeParameterShouldEndInDotDefault =
+       "IDW10404: 'scope' parameter should be of the form 'AppIdUri/.default'.";
+        public const string ClientCredentialTenantShouldBeTenanted =
```
jennyf19, line 47: should an aka.ms link be included here w/more info on why this is the case? is it clear enough for developers to know how to fix it based on the error message? i'm not sure.

tests/Microsoft.Identity.Web.Test.Integration/AcquireTokenForAppIntegrationTests.cs
```diff
@@ -61,14 +62,30 @@ public async Task GetAccessTokenForApp_ReturnsAccessTokenAsync()
...
-                await _tokenAcquisition.GetAccessTokenForAppAsync(TestConstants.s_scopesForApp).ConfigureAwait(false);
+                await _tokenAcquisition.GetAccessTokenForAppAsync(TestConstants.s_scopeForApp).ConfigureAwait(false);
 
             // Assert
             Assert.NotNull(token);
 
             AssertAppTokenInMemoryCache(TestConstants.ConfidentialClientId, 1);
         }
 
+        [Fact]
```
jennyf19, line 73: you can do different arguments and replace the tenant value w/organziations, common, consumers.
```csharp
[Theory]
[InlineData(Constants.Common)]
[InlineData(Constants.Organizations)]
[InlineData(Constants.Consumers)]
public async Task GetAccessTokenForApp_WithMetaTenant(string metaTenant)
{
....
await _tokenAcquisition.GetAccessTokenForAppAsync(TestConstants.s_scopeForApp, metaTenant).ConfigureAwait(false);
....
}

File: src/Microsoft.Identity.Web/TokenAcquisition.cs [hunk 10]
```diff
@@ -293,25 +303,41 @@ internal class TokenAcquisition : ITokenAcquisition, ITokenAcquisitionInternal
         /// Acquires a token from the authority configured in the app, for the confidential client itself (not on behalf of a user)
         /// using the client credentials flow. See https://aka.ms/msal-net-client-credentials.
         /// </summary>
-        /// <param name="scopes">scopes requested to access a protected API. For this flow (client credentials), the scopes
+        /// <param name="scope">scope requested to access a protected API. For this flow (client credentials), the scope
```
jennyf19, line 306:
```suggestion
        /// <param name="scope">The scope requested to access a protected API. For this flow (client credentials), the scope
```

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -293,25 +303,41 @@ internal class TokenAcquisition : ITokenAcquisition, ITokenAcquisitionInternal
...
         /// should be of the form "{ResourceIdUri/.default}" for instance <c>https://management.azure.net/.default</c> or, for Microsoft
         /// Graph, <c>https://graph.microsoft.com/.default</c> as the requested scopes are defined statically with the application registration
-        /// in the portal, and cannot be overridden in the application.</param>
+        /// in the portal, cannot be overridden in the application, and you can request a token for only one resource at a time (use
```
jennyf19, line 309:
```suggestion
        /// in the portal, cannot be overridden in the application, as you can request a token for only one resource at a time (use
```

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -293,25 +303,41 @@ internal class TokenAcquisition : ITokenAcquisition, ITokenAcquisitionInternal
...
+        /// several call to get tokens for other resources).</param>
```
jennyf19, line 310:
```suggestion
        /// several calls to get tokens for other resources).</param>
```

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -293,25 +303,41 @@ internal class TokenAcquisition : ITokenAcquisition, ITokenAcquisitionInternal
...
+        /// <param name="tenant">Enables overriding of the tenant/account for the same identity. This is useful in the
+        /// cases where a given account is guest in other tenants, and you want to acquire tokens for a specific tenant, like where the user is a guest in.</param>
```
jennyf19, line 312:
```suggestion
        /// cases where a given account is a guest in other tenants, and you want to acquire tokens for a specific tenant.</param>
```

src/Microsoft.Identity.Web/TokenAcquisition.cs
jmprieur: Thanks for the tip.
Addressed!

src/Microsoft.Identity.Web/TokenAcquisition.cs
jmprieur: I think that /.default is probably better as the resource needs to be ended with /

src/Microsoft.Identity.Web/Constants/IDWebErrorMessage.cs
jmprieur: Thanks for the suggestion. I've added one:  See https://aka.ms/ms-id-web/daemon-scenarios.

tests/Microsoft.Identity.Web.Test.Integration/AcquireTokenForAppIntegrationTests.cs
jmprieur: We do already with the test that uses user scopes.

jmprieur: Addressed

jmprieur:
Thanks @jennyf19 for these suggestions.
I've addressed all your PR feedback: https://github.com/AzureAD/microsoft-identity-web/pull/424/commits/1df2f006ba9f367bf388062654bf29df51ee708b

hpsin:
Consumers is allowed, since it is a single tenant.
It's probably worth keeping the old signature around and updating its comments for future compatibility. If we suddenly start supporting multiple scopes (eg more than just /.default) then we'll want this to take multiple strings again.

src/Microsoft.Identity.Web/Constants/IDWebErrorMessage.cs
```diff
@@ -42,6 +42,10 @@ internal static class IDWebErrorMessage
         public const string TenantIdClaimNotPresentInToken = "IDW10401: Neither `tid` nor `tenantId` claim is present in the token obtained from Microsoft identity platform. ";
         public const string ClientInfoReturnedFromServerIsNull = "IDW10402: Client info returned from the server is null. ";
         public const string TokenIsNotJwtToken = "IDW10403: Token is not JWT token. ";
+        public const string ClientCredentialScopeParameterShouldEndInDotDefault =
+       "IDW10404: 'scope' parameter should be of the form 'AppIdUri/.default'. See https://aka.ms/ms-id-web/daemon-scenarios.";
+        public const string ClientCredentialTenantShouldBeTenanted =
+       "IDW10405: 'tenant' parameter should be a tenant ID or domain name, not 'common', 'organizations' or 'consumers'. See https://aka.ms/ms-id-web/daemon-scenarios.";
```
hpsin, line 48: Remove consumers.

src/Microsoft.Identity.Web/ITokenAcquisition.cs
```diff
@@ -37,12 +37,15 @@ public interface ITokenAcquisition
...
         /// </summary>
-        /// <param name="scopes">scopes requested to access a protected API. For this flow (client credentials), the scopes
+        /// <param name="scope">The scope requested to access a protected API. For this flow (client credentials), the scope
         /// should be of the form "{ResourceIdUri/.default}" for instance <c>https://management.azure.net/.default</c> or, for Microsoft
         /// Graph, <c>https://graph.microsoft.com/.default</c> as the requested scopes are defined statically with the application registration
-        /// in the portal, and cannot be overridden in the application.</param>
+        /// in the portal, cannot be overridden in the application, as you can request a token for only one resource at a time (use
+        /// several calls to get tokens for other resources).</param>
+        /// <param name="tenant">Enables overriding of the tenant/account for the same identity. This is useful in the
+        /// cases where a given account is a guest in other tenants, and you want to acquire tokens for a specific tenant.</param>
```
hpsin, line 46: There's no guest scenario for app only tokens, just SP presence in other tenants. Could say "useful for multi tenant apps" or daemons? Not sure that "same identity" will resonate, since we usually don't treat SPs as identities?

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -293,25 +305,40 @@ internal class TokenAcquisition : ITokenAcquisition, ITokenAcquisitionInternal
...
-            if (scopes == null)
+            if (string.IsNullOrEmpty(scope))
             {
-                throw new ArgumentNullException(nameof(scopes));
+                throw new ArgumentNullException(nameof(scope));
+            }
+
+            if (!scope.EndsWith("/.default", true, CultureInfo.InvariantCulture))
+            {
+                throw new ArgumentException(IDWebErrorMessage.ClientCredentialScopeParameterShouldEndInDotDefault, nameof(scope));
```
hpsin, line 325: This is good for the singleton scenario. I think that we should retain the original one and not do checks there, though, in an effort to remain OIDC/oauth compliant.

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -81,6 +81,18 @@ internal class TokenAcquisition : ITokenAcquisition, ITokenAcquisitionInternal
...
 
+        /// <summary>
+        /// meta-tenant identifiers which are not allowed in client credentials.
+        /// </summary>
+        private readonly ISet<string> _metaTenantIdentifiers = new HashSet<string>(
+            new[]
+            {
+                Constants.Common,
+                Constants.Organizations,
+                Constants.Consumers,
```
hpsin, line 92: Remove consumers.

jmprieur: Thansk @hspin. I addressed them all

PR #425, opened 2020-08-10
thought experiment...use just `MicrosoftIdentity` 

jennyf19:


src/Microsoft.Identity.Web/ITokenAcquisition.cs
```diff
@@ -37,12 +37,15 @@ public interface ITokenAcquisition
...
         /// should be of the form "{ResourceIdUri/.default}" for instance <c>https://management.azure.net/.default</c> or, for Microsoft
         /// Graph, <c>https://graph.microsoft.com/.default</c> as the requested scopes are defined statically with the application registration
-        /// in the portal, and cannot be overridden in the application.</param>
+        /// in the portal, cannot be overridden in the application, as you can request a token for only one resource at a time (use
+        /// several calls to get tokens for other resources).</param>
+        /// <param name="tenant">Enables overriding of the tenant/account for the same identity. This is useful in the
+        /// cases where a given account is a guest in other tenants, and you want to acquire tokens for a specific tenant.</param>
         /// <returns>An access token for the app itself, based on its scopes.</returns>
-        Task<string> GetAccessTokenForAppAsync(IEnumerable<string> scopes);
+        Task<string> GetAccessTokenForAppAsync(string scope, string? tenant);
```
pmaytak, line 48: The implementation signature added `tenant` as optional, which doesn't match interface signature.
```suggestion
        Task<string> GetAccessTokenForAppAsync(string scope, string? tenant = null));
```

src/Microsoft.Identity.Web/WebAppExtensions/MicrosoftIdentityWebAppAuthenticationBuilder.cs
```diff
@@ -0,0 +1,154 @@
...
+            IConfigurationSection? configurationSection)
+            : base(services, configurationSection)
+        {
+            OpenIdConnectScheme = openIdConnectScheme;
+            ConfigureMicrosoftIdentityOptions = configureMicrosoftIdentityOptions;
+
+            if (ConfigureMicrosoftIdentityOptions == null)
+            {
+                throw new ArgumentNullException(nameof(configureMicrosoftIdentityOptions));
+            }
```
pmaytak, line 42: Move the null check infront?
```suggestion
            if (configureMicrosoftIdentityOptions == null)
            {
                throw new ArgumentNullException(nameof(configureMicrosoftIdentityOptions));
            }
            OpenIdConnectScheme = openIdConnectScheme;
            ConfigureMicrosoftIdentityOptions = configureMicrosoftIdentityOptions;
```

src/Microsoft.Identity.Web/WebAppExtensions/MicrosoftIdentityWebAppAuthenticationBuilderExtensions.cs
```diff
@@ -0,0 +1,414 @@
...
+        /// <param name="builder">The <see cref="AuthenticationBuilder"/> to which to add this configuration.</param>
+        /// <param name="configureMicrosoftIdentityOptions">The action to configure <see cref="MicrosoftIdentityOptions"/>.</param>
+        /// <param name="configureCookieAuthenticationOptions">The action to configure <see cref="CookieAuthenticationOptions"/>.</param>
+        /// <param name="openIdConnectScheme">The OpenID Connect scheme name to be used. By default it uses "OpenIdConnect".</param>
+        /// <param name="cookieScheme">The cookie-based scheme name to be used. By default it uses "Cookies".</param>
+        /// <param name="subscribeToOpenIdConnectMiddlewareDiagnosticsEvents">
+        /// Set to true if you want to debug, or just understand the OpenID Connect events.
+        /// </param>
+        /// <returns>The authentication builder for chaining.</returns>
+        public static MicrosoftIdentityWebAppAuthenticationBuilder AddMicrosoftWebApp(
```
pmaytak, line 114:
```suggestion
        public static MicrosoftIdentityWebAppAuthenticationBuilder AddMicrosoftIdentityWebApp(
```

src/Microsoft.Identity.Web/ITokenAcquisition.cs
jmprieur: I think there was something weird with GitHub as the signature wasn't with a collection? Good catch for the default value

src/Microsoft.Identity.Web/WebAppExtensions/MicrosoftIdentityWebAppAuthenticationBuilder.cs
jennyf19: i didn't look at this closely enough and made the change, but it doesn't work as we have to set the options and then check if they are null after, so have left it as is.. let's use the "suggested changes" features for xml comments only, not code changes, unless you've tested and know it doesn't break things. sound good?

Issue #415, opened 2020-08-07
Domain hint only specified when login hint is not null or blank


glasody:
**Which version of Microsoft Identity Web are you using?**
0.2.1-preview

**Where is the issue?**

* Web app
 
  * [x]  Sign-in users
  * [ ]  Sign-in users and call web APIs
* Web API
  
  * [ ]  Protected web APIs (validating tokens)
  * [ ]  Protected web APIs (validating scopes)
  * [ ]  Protected web APIs call downstream web APIs
* Token cache serialization
  
  * [ ]  In-memory caches
  * [ ]  Session caches
  * [ ]  Distributed caches
* Other (please describe)

**Repro**
Create new net core app
Add identity web auth
Define AzureAd properties in config file with domain property defined but no login hint

Runt the Web app and Web API

**Expected behavior**
Domain hint to be added to the authorization request even if login hint isn't specified

**Actual behavior**
domain hint doesn't get added to the authorization call

**Possible solution**
Move domain hint definition outside the if clause

---
Hi,

I was working with this package to get a web app running with Azure AD, but found that defining the domain did not add it into my authentication request thus not directly opening the federation login page.

I found that this was because domain_hint is only being added when login_hint is not blank or null.
Is this a bug? I was under the assumption that domain_hint could be provided without a login_hint.

https://github.com/AzureAD/microsoft-identity-web/blob/bd0c266b0d6f42b84dab3cfb9fb1bc830097658b/src/Microsoft.Identity.Web/WebAppExtensions/WebAppAuthenticationBuilderExtensions.cs#L135-L146

pmaytak:
Hi @glasody. Thanks for reporting. We also appreciate outside contributions, if you'd like to submit a PR.

durandt:
I stumbled upon the same issue and it was a great time-saver that @glasody found it before me. 
I pushed a PR-suggestion. 

Regards,

jennyf19:
Included in [0.3.0-preview release](https://github.com/AzureAD/microsoft-identity-web/releases/tag/0.3.0-preview).

durandt:
Thank you for the heads-up. Works as expected 👍

jmprieur:
Thanks for the update, @durandt !

PR #428, opened 2020-08-10
Fix for #415

durandt:
domain_hint ingored when login_hint is set.
Break up code so that domain_hint and login_hint are treated
independently.

### DIFF
--- a/src/Microsoft.Identity.Web/WebAppExtensions/WebAppAuthenticationBuilderExtensions.cs
+++ b/src/Microsoft.Identity.Web/WebAppExtensions/WebAppAuthenticationBuilderExtensions.cs
@@ -136,12 +136,19 @@ namespace Microsoft.Identity.Web
                     if (!string.IsNullOrWhiteSpace(login))
                     {
                         context.ProtocolMessage.LoginHint = login;
-                        context.ProtocolMessage.DomainHint = context.Properties.GetParameter<string>(
-                            OpenIdConnectParameterNames.DomainHint);
 
-                        // delete the login_hint and domainHint from the Properties when we are done otherwise
+                        // delete the login_hint from the Properties when we are done otherwise
                         // it will take up extra space in the cookie.
                         context.Properties.Parameters.Remove(OpenIdConnectParameterNames.LoginHint);
+                    }
+
+                    var domainHint = context.Properties.GetParameter<string>(OpenIdConnectParameterNames.DomainHint);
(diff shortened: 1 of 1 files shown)


src/Microsoft.Identity.Web/WebAppExtensions/WebAppAuthenticationBuilderExtensions.cs
```diff
@@ -136,12 +136,19 @@ public static partial class WebAppAuthenticationBuilderExtensions
...
+                    }
+
+                    var domainHint = context.Properties.GetParameter<string>(OpenIdConnectParameterNames.DomainHint);
+                    if (!string.IsNullOrWhiteSpace(domainHint))
+                    {
+                        context.ProtocolMessage.DomainHint = domainHint;
+
+                        // delete the domainHint from the Properties when we are done otherwise
+                        // it will take up extra space in the cookie.
                         context.Properties.Parameters.Remove(OpenIdConnectParameterNames.DomainHint);
```
jennyf19, line 152: Thanks for the contribution, we already have a [test verifying the doman_hint is set](https://github.com/AzureAD/microsoft-identity-web/blob/master/tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs#L406), but not sure if we want to check one or the other as well, or this test is enough...personally, I'm okay with this current test and the work done here.

src/Microsoft.Identity.Web/WebAppExtensions/WebAppAuthenticationBuilderExtensions.cs
```diff
@@ -136,12 +136,19 @@ public static partial class WebAppAuthenticationBuilderExtensions
...
                         // it will take up extra space in the cookie.
                         context.Properties.Parameters.Remove(OpenIdConnectParameterNames.LoginHint);
+                    }
+
+                    var domainHint = context.Properties.GetParameter<string>(OpenIdConnectParameterNames.DomainHint);
+                    if (!string.IsNullOrWhiteSpace(domainHint))
+                    {
+                        context.ProtocolMessage.DomainHint = domainHint;
+
+                        // delete the domainHint from the Properties when we are done otherwise
```
jennyf19, line 150:
```suggestion
                        // delete the domain_hint from the Properties when we are done otherwise
```

jennyf19:
thanks again @durandt

durandt:
Well, that wasn't much job 😊 Thank you @jennyf19 , that was fast!

PR #431, opened 2020-08-11
Prototype of new API with the Graph and Downstream web API services

jmprieur:
Proposal for Microsoft.Graph support
--------------------------------------------
1. Rename `CallsWebApi` in `CallsWebApis`, as this becomes a hub for all what calls a Web API. The alternative design would have been to have CallsMicrosoftGraph and CallsDownstreamApi under AddMicrosoftIdentityWebApp / WebApi, however, this would have obliged the developer =to pass-in the `ConfidentialClientApplicationOptions` to each of the services they want to call (that is several times if they call several services, for instance both Graph, and a DownstreamApi for ARM, and a Downstream for https://myWebApi. 
This is now possible having several sections in the appsettings.json. See for instance the Blazor sample which call graph and another API (Graph really, but that would not matter)
2. Renamed `CallsMicrosoftGraph` into `AddMicrosoftGraphServiceClient` as GraphServiceClient this the type that the developer will use in the controller and is what they will look for in the documentation. This method is present under CallsWebApis, because otherwise we'd have to re-pass the initialization for CallsWebApis (the ConfidentialClientApplicationOptions)
3. Updates the tests to get a feel of the API (In particular **WebAppCallsGraph** and **BlazorServerCallsGraph**

Proposal for Downstream web API support
--------------------------------------------
- *Downstream API support** folder contains options, service, extensions
- Aligning the properties for Microsoft Graph options and Downstream API options
- Updated the Blazor sample to use both the Graph service and the Downstream API service (BlazorServer calls graph)

**Todo list sample**
- changed the namespace of the model (common to client and service). Ideally we'd put in a shared project. We might do that later.
- Updated the TodoListService class to leverage the DownstreamWebApi service. The code is way simpler!!!
- updated the AppSettings.json and the Startup.cs to leverage these services
- Fixed the Controller of the Web API which was not using the display name when adding an item (list always empty)


tests/WebAppCallsMicrosoftGraph/Startup.cs
```diff
@@ -29,10 +29,13 @@ public void ConfigureServices(IServiceCollection services)
         {
             services.AddAuthentication(OpenIdConnectDefaults.AuthenticationScheme)
                     .AddMicrosoftIdentityWebApp(Configuration.GetSection("AzureAd"))
-                        .CallsWebApi()
+                        .CallsWebApis()
+                           .AddMicrosoftGraphServiceClient()
+                           .AddMicrosoftGraphServiceClient(Configuration.GetSection("GraphBeta"))
```
jmprieur, line 34: This is really one or the other of these 3 lines;

tests/WebAppCallsWebApiCallsGraph/Client/Controllers/TodoListController.cs
```diff
@@ -6,21 +6,21 @@
...
 {
     public class TodoListController : Controller
     {
         private ITodoListService _todoListService;
 
-        public TodoListController(ITodoListService todoListService)
+        public TodoListController(IDownstreamWebApi downstreamWebApi)
         {
-            _todoListService = todoListService;
+            _todoListService = new TodoListService(downstreamWebApi);
```
jmprieur, line 19: I'm not too satisfied it, but I did not yet manage to inject the DownstreamApiService.
If we decided to go with this design, we'll have to fix that.

tests/WebAppCallsWebApiCallsGraph/Client/Services/ITodoListService.cs
```diff
@@ -3,7 +3,7 @@
 
 using System.Collections.Generic;
 using System.Threading.Tasks;
-using TodoListService.Models;
+using TodoList.Models;
```
jmprieur, line 6: I renamed the namespace because:
- it's common to the service and client (it's even the same file with a link)
- There is was a naming conflict between TodoListService (the namespace for this models), and the TodoListService class in the TodoListClient.

tests/WebAppCallsWebApiCallsGraph/Client/Services/TodoListService.cs
```diff
@@ -2,158 +2,96 @@
...
 namespace TodoListClient.Services
 {
     public static class TodoListServiceExtensions
     {
-        public static void AddTodoListService(this IServiceCollection services, IConfiguration configuration)
+        public static MicrosoftIdentityAppCallsWebApisAuthenticationBuilder AddTodoListService(
+            this MicrosoftIdentityAppCallsWebApisAuthenticationBuilder builder, 
+            IConfiguration configuration)
         {
+            builder.AddDownstreamApiService(TodoListService.ServiceName, configuration);
```
jmprieur, line 20: Just leverage the generic DownstreamApiService

tests/WebAppCallsWebApiCallsGraph/Client/Services/TodoListService.cs
```diff
@@ -2,158 +2,96 @@
...
-            if (response.StatusCode == HttpStatusCode.OK)
-            {
-                var content = await response.Content.ReadAsStringAsync();
-                todo = JsonSerializer.Deserialize<Todo>(content, _jsonOptions);
-
-                return todo;
-            }
-
-            throw new HttpRequestException($"Invalid status code in the HttpResponseMessage: {response.StatusCode}.");
+            return await _downstreamWebApi.CallWebApiForUserAsync<Todo, Todo>(
```
jmprieur, line 41: use strongly typed wrappers and express just what is different. The code is way simpler and no copy/paste any longer.

tests/WebAppCallsWebApiCallsGraph/Client/TodoListClient.csproj
```diff
@@ -23,7 +23,6 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.DataProtection.Abstractions" Version="3.1.1" />
-    <PackageReference Include="Microsoft.Graph" Version="1.16.0" />
```
jmprieur, line 26: Microsoft.Graph is now included in Microsoft.Identity.Web. We don't need to reference it any longer in the samples

tests/WebAppCallsWebApiCallsGraph/Client/Views/TodoList/Create.cshtml
```diff
@@ -1,4 +1,4 @@
-﻿@model TodoListService.Models.Todo
+﻿@model TodoList.Models.Todo
```
jmprieur, line 1: Consequence of the renaming of namespace ..

tests/WebAppCallsWebApiCallsGraph/TodoListService/Controllers/TodoListController.cs
```diff
@@ -76,7 +77,7 @@ public IActionResult Post([FromBody] Todo todo)
         {
             HttpContext.VerifyUserHasAnyAcceptedScope(scopeRequiredByApi);
             int id = TodoStore.Values.OrderByDescending(x => x.Id).FirstOrDefault().Id + 1;
-            Todo todonew = new Todo() { Id = id, Owner = HttpContext.User.Identity.Name, Title = todo.Title };
+            Todo todonew = new Todo() { Id = id, Owner = User.GetDisplayName(), Title = todo.Title };
```
jmprieur, line 80: Lol! we forgot this one when we updated the GetAsync()

tests/WebAppCallsWebApiCallsGraph/TodoListService/Startup.cs
```diff
@@ -33,7 +33,7 @@ public void ConfigureServices(IServiceCollection services)
             // Adds Microsoft Identity platform (AAD v2.0) support to protect this Api
             services.AddAuthentication(JwtBearerDefaults.AuthenticationScheme)
                     .AddMicrosoftIdentityWebApi(Configuration, "AzureAd")
-                        .CallsWebApi()
+                        .CallsWebApis()
```
jmprieur, line 36: Renaming to express that this is a hub for downstream APIs.
Note that CallsWebApis() really provides the ITokenAcquisitionService .... but I don't think we want to write that?

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
```diff
@@ -284,7 +284,7 @@ public async Task AddMicrosoftWebAppCallsWebApi_WithConfigActionParameters()
 
             var builder = services.AddAuthentication()
                 .AddMicrosoftWebApp(_configureMsOptions, null, OidcScheme)
-                .CallsWebApi(_configureAppOptions, initialScopes);
+                .CallsWebApis(_configureAppOptions, initialScopes);
```
jmprieur, line 287: See, in this case, if we had chosen the other design, we'd have to pass-in the _configureAppOptions in all the CallsMicrosoftGraph and CallsDownstreamApis, with a risk of errors (pass-in different things, whereas this would be the same service, really), and duplication of code.

tests/BlazorServerCallsGraph/appsettings.json
```diff
@@ -8,25 +8,29 @@
         "SignedOutCallbackPath ": "/signout-callback-oidc",
 
         // To call an API
-        "ClientSecret": "secret-goes-here"
+        "ClientSecret": "fvHwfoLA8Zls6Huj5MBUQUDrJ5HXrauFQKDd2CacxJQ="
+    },
+    "GraphBeta": {
```
jmprieur, line 13: This is the case of a sample that calls several sevices:
- Graph
- And a downstream API (which turns out to be Graph v1.0, but could be something else.

tests/BlazorServerCallsGraph/Startup.cs
```diff
@@ -26,15 +26,13 @@ public Startup(IConfiguration configuration)
...
         public void ConfigureServices(IServiceCollection services)
         {
-            string[] scopes = Configuration.GetValue<string>("CalledApi:CalledApiScopes")?.Split(' ');
             services.AddAuthentication(OpenIdConnectDefaults.AuthenticationScheme)
                     .AddMicrosoftIdentityWebApp(Configuration, "AzureAd")
-                        .CallsWebApi(scopes)
+                        .CallsWebApis()
+                            .AddMicrosoftGraphServiceClient(Configuration.GetSection("GraphBeta"))
+                            .AddDownstreamApiService("CalledApi", Configuration.GetSection("CalledApi"))
                         .AddInMemoryTokenCaches();
```
jmprieur, line 34: That's the kind of code that you would have if you want to call several services.

If you want to use ITokenAquisition directly, just use `.CallsWebApis()`
Otherwise:
- if you want to use the MicrosoftGraphServiceClient wrapper, use `.AddMicrosoftGraphServiceClient` under `.CallsWebApis()`
- if you want to use one or several instances of the downstream API service (they have a name, which here, I've chosen the same as the section in the config file to make it simpler, but that's not absolutely necessary), you woud have one or several calls to `.AddDownstreamApiService("NameOfTheService", Configuration.GetSection("SectionInConfig"))`

Then in the controller, you'd use IDownstreamWebApi.CallsWebApiForUser("NameOfTheService", and have the possiblity to override the configuration from the config file with a delegate.

tests/BlazorServerCallsGraph/Pages/CallWebApi.razor
```diff
@@ -20,13 +20,25 @@ else
...
+    private HttpResponseMessage response;
+    string apiResult;
 
     protected override async Task OnInitializedAsync()
     {
         try
         {
-            apiResult = await downstreamAPI.CallWebApiAsync("me");
+            response = await downstreamAPI.CallWebApiForUserAsync(
+                    "CalledApi",
```
jmprieur, line 31: Here we call the service "CalledApi", on behalf of the user (this maps to some configuration for this service: see below), and we override the relative path to call the "me" endpoint. All the other configurations are overridable as well.

tests/BlazorServerCallsGraph/Pages/CallWebApi.razor
```diff
@@ -20,13 +20,25 @@ else
...
+                    options => options.RelativePath = "me");
+
+            if (response.StatusCode == System.Net.HttpStatusCode.OK)
+            {
+                apiResult = await response.Content.ReadAsStringAsync();
+            }
```
jmprieur, line 37: Here we use the pattern recommended by Damien, but we could also use the strongly typed wrapper CallWebApiForUserAsync<InputType, ReturnType> which makes it even easier. It's used in the TodoListService wrapper in the TodoListClient project.

tests/BlazorServerCallsGraph/appsettings.json
```diff
@@ -8,25 +8,29 @@
         "SignedOutCallbackPath ": "/signout-callback-oidc",
 
         // To call an API
-        "ClientSecret": "secret-goes-here"
+        "ClientSecret": "fvHwfoLA8Zls6Huj5MBUQUDrJ5HXrauFQKDd2CacxJQ="
```
jmprieur, line 11:
```suggestion
        "ClientSecret": ""
```

src/Microsoft.Identity.Web/DownstreamApiSupport/CalledApiOptions.cs
```diff
@@ -0,0 +1,84 @@
...
+using System.Net.Http;
+
+namespace Microsoft.Identity.Web
+{
+    /// <summary>
+    /// Options passed-in to call web APIs. To call Microsoft Graph, see rather
+    /// <see cref="MicrosoftGraphOptions"/>.
+    /// </summary>
+    public class CalledApiOptions
+    {
```
jennyf19, line 13: not a fan of the name on this one. i will need to think about it to get some ideas.

src/Microsoft.Identity.Web/DownstreamApiSupport/DownstreamWebApi.cs
```diff
@@ -0,0 +1,172 @@
...
+            HttpClient httpClient)
+        {
+            _tokenAcquisition = tokenAcquisition;
+            _namedOptions = namedOptions;
+            _httpClient = httpClient;
+        }
+
+        /// <inheritdoc/>
+        public async Task<HttpResponseMessage> CallWebApiForUserAsync(
+            string optionsInstanceName,
```
jennyf19, line 50: question: should we stick w/consistency of the naming? `MicrosoftIdentityCallWebApiForUserAysnc()`....i don't like this much, but i think things should feel like they are part of MS identity web, or not...again, will need to think about it.

src/Microsoft.Identity.Web/DownstreamApiSupport/DownstreamWebApi.cs
```diff
@@ -0,0 +1,172 @@
...
+            where TOutput : class
+        {
+            StringContent? jsoncontent;
+            if (input != null)
+            {
+                var jsonRequest = JsonSerializer.Serialize(input);
+                jsoncontent = new StringContent(jsonRequest, Encoding.UTF8, "application/json");
+// case of patch?               jsoncontent = new StringContent(jsonRequest, Encoding.UTF8, "application/json-patch+json");
+            }
+            else
```
pmaytak, line 122: I think this `else` is unnecessary, `jsoncontent` would still be null here.

src/Microsoft.Identity.Web/DownstreamApiSupport/DownstreamWebApi.cs
```diff
@@ -0,0 +1,172 @@
...
+            {
+                string error = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
+                throw new HttpRequestException($"Invalid status code in the HttpResponseMessage: {response.StatusCode}: {error}");
+            }
+        }
+
+        /// <inheritdoc/>
+        public async Task<HttpResponseMessage> CallWebApiForAppAsync(
```
pmaytak, line 147: Do we also want to have a generic typed version for this method?

src/Microsoft.Identity.Web/DownstreamApiSupport/IDownstreamWebApi.cs
```diff
@@ -0,0 +1,78 @@
...
+        /// <param name="input">Input parameter to the API.</param>
+        /// <param name="calledApiOptionsOverride">Overides the options proposed in the configuration described
+        /// by <paramref name="optionsInstanceName"/>.</param>
+        /// <param name="user">[Optional] claims representing a user. This is useful in scenarios like Blazor
+        /// or Azure Signal R where the HttpContext is not available. In the other scenarios, the library
+        /// will find the user itself.</param>
+        /// <returns>The value returned by the API.</returns>
+        public Task<TOutput> CallWebApiForUserAsync<TInput, TOutput>(
+            string optionsInstanceName,
+            TInput input,
```
pmaytak, line 56: `input` > `requestContent`? 
Also maybe `TInput`, `TOutput` to `TRequestContent`, `TResponseContent`?

src/Microsoft.Identity.Web/DownstreamApiSupport/IDownstreamWebApi.cs
```diff
@@ -0,0 +1,78 @@
...
+        /// each for one API. You can pass-in null, but in that case <paramref name="calledApiOptionsOverride"/>
+        /// need to be set.</param>
+        /// <param name="input">Input parameter to the API.</param>
+        /// <param name="calledApiOptionsOverride">Overides the options proposed in the configuration described
+        /// by <paramref name="optionsInstanceName"/>.</param>
+        /// <param name="user">[Optional] claims representing a user. This is useful in scenarios like Blazor
+        /// or Azure Signal R where the HttpContext is not available. In the other scenarios, the library
+        /// will find the user itself.</param>
+        /// <returns>The value returned by the API.</returns>
+        public Task<TOutput> CallWebApiForUserAsync<TInput, TOutput>(
```
pmaytak, line 54: I would reorder parameters here to match `CallWebApiForAppAsync` and other methods:
`optionsInstanceName`, `calledApiOptionsOverride`, `input`, `user`.

src/Microsoft.Identity.Web/DownstreamApiSupport/IDownstreamWebApi.cs
```diff
@@ -0,0 +1,78 @@
...
+        /// need to be set.</param>
+        /// <param name="calledApiOptionsOverride">Overides the options proposed in the configuration described
+        /// by <paramref name="optionsInstanceName"/>.</param>
+        /// <param name="user">[Optional] claims representing a user. This is useful in scenarios like Blazor
+        /// or Azure Signal R where the HttpContext is not available. In the other scenarios, the library
+        /// will find the user itself.</param>
+        /// <param name="content">Http content in the case where <see cref="CalledApiOptions.HttpMethod"/> is
+        /// <see cref="HttpMethod.Patch"/>, <see cref="HttpMethod.Post"/>, <see cref="HttpMethod.Put"/>.</param>
+        /// <returns>An <see cref="HttpResponseMessage"/> that the application will process.</returns>
+        public Task<HttpResponseMessage> CallWebApiForUserAsync(
```
pmaytak, line 32: I would reorder parameters here to match `CallWebApiForAppAsync` and other methods:
`optionsInstanceName`, `calledApiOptionsOverride`, `content`, `user`.

src/Microsoft.Identity.Web/DownstreamApiSupport/IDownstreamWebApi.cs
```diff
@@ -0,0 +1,78 @@
...
+        /// need to be set.</param>
+        /// <param name="calledApiOptionsOverride">Overides the options proposed in the configuration described
+        /// by <paramref name="optionsInstanceName"/>.</param>
+        /// <param name="content">Http content in the case where <see cref="CalledApiOptions.HttpMethod"/> is
+        /// <see cref="HttpMethod.Patch"/>, <see cref="HttpMethod.Post"/>, <see cref="HttpMethod.Put"/>.</param>
+        /// <returns>An <see cref="HttpResponseMessage"/> that the application will process.</returns>
+        public Task<HttpResponseMessage> CallWebApiForAppAsync(
+            string optionsInstanceName,
+            Action<CalledApiOptions>? calledApiOptionsOverride = null,
+            StringContent? content = null);
```
pmaytak, line 76: Change `content` to  `requestContent` in all method signatures to be more descriptive?

src/Microsoft.Identity.Web/DownstreamApiSupport/CalledApiOptions.cs
jmprieur: Done

src/Microsoft.Identity.Web/DownstreamApiSupport/DownstreamWebApi.cs
jmprieur: Here this is a verb (method) on a class that we own, not an extension method on a class that we don't own (like Service Collection). I don't think it's abiguous. I would stick to `CallWebApiForUserAsync` in this particular case

src/Microsoft.Identity.Web/DownstreamApiSupport/DownstreamWebApi.cs
jmprieur: I tried but the compiler was not happy line 131.

src/Microsoft.Identity.Web/DownstreamApiSupport/IDownstreamWebApi.cs
jmprieur: No sure about this one. The content is the json. Here input and output are object types (like Todo)

src/Microsoft.Identity.Web/DownstreamApiSupport/IDownstreamWebApi.cs
jmprieur: Done

src/Microsoft.Identity.Web/DownstreamApiSupport/IDownstreamWebApi.cs
jmprieur: Done.

src/Microsoft.Identity.Web/DownstreamApiSupport/DownstreamWebApi.cs
jmprieur: It's more work.

src/Microsoft.Identity.Web/DownstreamApiSupport/IDownstreamWebApi.cs
jmprieur: Actually no as calledApiOptionsOverride is null by default

jennyf19:
This can be closed, correct? @jmprieur you did two smaller PRs instead right?

jmprieur:
No, there are the templates still

PR #442, opened 2020-08-12
fix PR feedback for xml comments and constants

jennyf19:


src/Microsoft.Identity.Web/MicrosoftGraph/MicrosoftGraphServiceExtensions.cs
```diff
@@ -10,16 +10,16 @@
 namespace Microsoft.Identity.Web
 {
     /// <summary>
-    /// Extensions methods on a MicrososoftAppCallingWebApiAuthenticationBuilder builder
+    /// Extensions methods on a MicrososoftIdentityAppCallingWebApiAuthenticationBuilder builder
```
pmaytak, line 13:
```suggestion
    /// Extensions methods on a MicrosoftIdentityAppCallingWebApiAuthenticationBuilder builder
```

jennyf19: lol. good catch!

jmprieur:
BTW, I much prefer a PR like that (a branch), where we see the modification, and can verify that this builds and that all the tests pass, rather than suggestions, which are hard to see depending on the tools used, and don't guaranty that all the changes were made.
cc: @jennyf19 @pmaytak @henrik-me

PR #443, opened 2020-08-12
fix warnings from earlier PR.

jennyf19:
Suppressing the warnings as the httpContext will not be null in this case

### DIFF
--- a/src/Microsoft.Identity.Web/MicrosoftIdentityCircuitHandler.cs
+++ b/src/Microsoft.Identity.Web/MicrosoftIdentityCircuitHandler.cs
@@ -51,7 +51,9 @@ namespace Microsoft.Identity.Web
     {
         private ClaimsPrincipal? _user = null;
         private string? _baseUri = null;
+#pragma warning disable CS8602 // Dereference of a possibly null reference. HttpContext will not be null in this case.
         private IHttpContextAccessor _httpContextAccessor;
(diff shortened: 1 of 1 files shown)


src/Microsoft.Identity.Web/MicrosoftIdentityCircuitHandler.cs
```diff
@@ -75,7 +77,9 @@ public ClaimsPrincipal User
             get
             {
                 return _user ??
+#pragma warning disable CS8602 // Dereference of a possibly null reference. HttpContext will not be null in this case.
                     (!IsBlazorServer ? _httpContextAccessor.HttpContext.User :
```
pmaytak, line 81: Instead of pragmas, I would use a [! null-forgiving operator](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-forgiving) here.
Something like `_httpContextAccessor!.HttpContext!.User` probably.

jennyf19: I know, i just have to add them everywhere, even in creating the request...and then there more and more and more. so it's not really necessary in this case, so that's why i went w/the pragmas.

jmprieur:
Merging this one, @jennyf19 as there is a O365 issue with them

Issue #426, opened 2020-08-10
AddInMemoryTokenCaches should take options and delegates


jmprieur:
From DevDiv review
AddInMemoryTokenCaches should take options and delegates

jennyf19:
Included in [0.3.0-preview release](https://github.com/AzureAD/microsoft-identity-web/releases/tag/0.3.0-preview).

PR #450, opened 2020-08-14
Add an options delegate for methods that add caches.

pmaytak:
Issue #426


pmaytak:
> is this updated w/master, you should be getting the warnings anymore.
> :shipit:

It is. Thanks.

PR #469, opened 2020-08-19
migration aid

jennyf19:


jmprieur (approved):
Thanks @jennyf19 
LGTM

we might want to add an aka.ms in the obsolete attributes, pointing to the 0.3.0-preview wiki article

PR #472, opened 2020-08-20
Fix for Authorization header format in Graph calls

pmaytak:
When adding Authorization header to Graph requests, space was missing, so requests weren't going through.


jennyf19:
Oops. My bad on that one. Glad you caught it. You're further along in testing 😃

Issue #429, opened 2020-08-10
[Bug] 'Scheme already exists: Bearer' when trying to setup both AAD and AAD B2C auth


pheuter:
**Which version of Microsoft Identity Web are you using?**
0.2.2-preview

**Where is the issue?**

I'm trying to make my ASP.NET Core Web API compatible with both AAD tokens issued on behalf of applications as well as AAD B2C tokens issued on behalf of users, but run into errors when trying to configure both entries in my appsettings.json file. If I only initialize `AddMicrosoftWebApi` once, then I get issues verifying JWT signature when the token is generated using the identity provider that was left out.

**Is this a new or an existing app?**
c. This is a new app

**Repro**

```csharp
services.AddAuthentication(JwtBearerDefaults.AuthenticationScheme)
             .AddMicrosoftWebApi(this.Configuration, "AzureAd")
             .AddMicrosoftWebApi(this.Configuration, "AzureAdB2C")
```

**Expected behavior**
ASP.NET Core app is setup to validate tokens issued from both identity providers.

**Actual behavior**
Error during startup: `InvalidOperationException: 'Scheme already exists: Bearer'`

jmprieur:
You added twice the same authentication scheme. You'd need to use a different name for the `jwtBearerScheme` parameter in one of your calls. for instance:

```csharp
services.AddAuthentication(JwtBearerDefaults.AuthenticationScheme)
             .AddMicrosoftWebApi(this.Configuration, "AzureAd")
             .AddMicrosoftWebApi(this.Configuration, "AzureAdB2C", "jwtBearerScheme2")
```

BTW, the default one will be "Bearer" (JwtBearerDefaults.AuthenticationScheme).

pheuter:
@jmprieur Got it, that makes sense! I am seeing a new error now:

```
info: Microsoft.AspNetCore.Authentication.JwtBearer.JwtBearerHandler[1]
Failed to validate the token. 
Microsoft.IdentityModel.Tokens.SecurityTokenSignatureKeyNotFoundException: IDX10501: Signature validation failed. Unable to match key: kid: 'huN95IvPfehq34GzBDZ1GXGirnM'.
```

pheuter:
One thought that just occurred to me is if it's even possible to support passing different `Authorization: Bearer` tokens to the same Web Api that are issued by different providers, i.e. one from Active Directory and another from AAD B2C?

pmaytak:
[...]

PR #475, opened 2020-08-20
Fix: API Multiple Identity Providers

timClyburn:
Changed AddMicrosoftIdentityWebApiImplementation to
add named configuration options for MicrosoftIdentityOptions.

Changed AddOptions Configure to use IOptionsMonitor to get
and use the named options.

Attempt at fix for #429

### DIFF
--- a/src/Microsoft.Identity.Web/WebApiExtensions/MicrosoftIdentityWebApiAuthenticationBuilderExtensions.cs
+++ b/src/Microsoft.Identity.Web/WebApiExtensions/MicrosoftIdentityWebApiAuthenticationBuilderExtensions.cs
@@ -153,7 +153,7 @@ namespace Microsoft.Identity.Web
             bool subscribeToJwtBearerMiddlewareDiagnosticsEvents)
         {
             builder.AddJwtBearer(jwtBearerScheme, configureJwtBearerOptions);
-            builder.Services.Configure(configureMicrosoftIdentityOptions);
+            builder.Services.Configure(jwtBearerScheme, configureMicrosoftIdentityOptions);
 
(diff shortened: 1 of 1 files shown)


src/Microsoft.Identity.Web/WebApiExtensions/MicrosoftIdentityWebApiAuthenticationBuilderExtensions.cs
```diff
@@ -166,11 +166,13 @@ public static partial class MicrosoftIdentityWebApiAuthenticationBuilderExtensio
 
             // Change the authentication configuration to accommodate the Microsoft identity platform endpoint (v2.0).
             builder.Services.AddOptions<JwtBearerOptions>(jwtBearerScheme)
-                .Configure<IServiceProvider, IOptions<MicrosoftIdentityOptions>>((options, serviceProvider, microsoftIdentityOptions) =>
+                .Configure<IServiceProvider, IOptionsMonitor<MicrosoftIdentityOptions>>((options, serviceProvider, microsoftIdentityOptionsMonitor) =>
```
pmaytak, line 169: Does this Configure method run twice for you?

It only runs once for me - for `Bearer` scheme. So whichever scheme (AAD or B2C) I set up as default `Bearer` is the only one that works.

timClyburn: Interestingly @pmaytak, yes it only runs once. I was expecting it to be ran once for each options that we are adding. but it seems to only get fired once for the last schema added.

PR #495, opened 2020-08-25
Update CodeQL task to run on schedule

pmaytak:
Removed push and PR triggers.
Added schedule trigger to run every day at 4am UTC.

(Reference](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#scheduled-events)

> Scheduled workflows run on the latest commit on the default or base branch.


jmprieur (approved):
LGTM. Thanks @pmaytak

Issue #496, opened 2020-08-25
[Bug] Cannot resolve IHttpContextAccessor scoped service in blazorserver web apps that sign-in users


jmprieur:
**Which version of Microsoft Identity Web are you using?**
0.3.0-preview
With the fixed Blazorserver UI changes

**Where is the issue?**
Blazor server B2C or single org

**Repro**
dotnet new blasorserver2 --auth SingleOrg
configure the app coordinates.

**Expected behavior**
This works

**Actual behavior**
System.AggregateException: 'Some services are not able to be constructed (Error while validating the service descriptor 'ServiceType: Microsoft.AspNetCore.Components.Server.Circuits.CircuitHandler Lifetime: Scoped ImplementationType: Microsoft.Identity.Web.MicrosoftIdentityServiceHandler': Unable to resolve service for type 'Microsoft.AspNetCore.Http.IHttpContextAccessor' while attempting to activate 'Microsoft.Identity.Web.MicrosoftIdentityConsentAndConditionalAccessHandler'.) (Error while validating the service descriptor 'ServiceType: Microsoft.Identity.Web.MicrosoftIdentityConsentAndConditionalAccessHandler Lifetime: Scoped ImplementationType: Microsoft.Identity.Web.MicrosoftIdentityConsentAndConditionalAccessHandler': Unable to resolve service for type 'Microsoft.AspNetCore.Http.IHttpContextAccessor' while attempting to activate 'Microsoft.Identity.Web.MicrosoftIdentityConsentAndConditionalAccessHandler'.)'

**Root cause**
This is because when the initialization of the `MicrosoftIdentityConsentAndConditionalAccessHandler` happens, we are in pre-rendering, and the IHttpContextAccessor is not available.

Note that our test apps call web apis ... which is why we did not catch this

**Possible solution**
Just inject an `IServiceProvider`, and attempt to get the `IHttpContextAccess` as a service. If this returns null, that's ok as the CircuitHandler will provide all the necessary information (there are tests for `IsBlazorServer` everywhere in `MicrosoftIdentityConsentAndConditionalAccessHandler`)

jennyf19:
Included in [0.3.0-preview release](https://github.com/AzureAD/microsoft-identity-web/releases/tag/0.3.0-preview).

Issue #490, opened 2020-08-24
[Bug] Fix .NET Core 3.1 blazorwasm2 and blazorserver project template


jmprieur:
**Which version of Microsoft Identity Web are you using?**
0.3.0-preview

**Where is the issue?**
blazorwasm project template

**Repro**
Create a blasorwasm project from the template

**Expected behavior**
The expected UI appears with the menu bar on the left side

**Actual behavior**
The UI is not great. There is an error in the browser that `_framework/scoped.styles.css` cannot be found

**Investigation**
The templates only work for .NET Core 5.0 Preview 8+

**Proposed fix**
See example in https://github.com/AzureAD/microsoft-identity-web/pull/487/files, detailed as follows

#### ComponentsWebAssembly-CSharp/Client/Program.cs
- line 25 replace `"#app"` by `"app"`

#### ComponentsWebAssembly-CSharp/Client/MainLayout-Auth.razor
- line 3 and line 18 remove the <div class="page"> and the corresponding </div>

#### ComponentsWebAssembly-CSharp/Client/MainLayout-NoAuth.razor
- line 3 and line 18 remove the <div class="page"> and the corresponding </div>
(same changes as previous file)

#### delete the following files
- ComponentsWebAssembly-CSharp/Client/Shared/MainLayout.razor.css
- ComponentsWebAssembly-CSharp/Client/Shared/NavMenu.razor.css 

####   ComponentsWebAssemblyCSharp/Client/wwwroot/css/app.css
Replace the content by the one from **tests/blazorwasm2-b2c-hosted/Client/wwwroot/css/app.css**

#### ComponentsWebAssembly-CSharp/Client/wwwroot/index.html 
- remove line 11: https://github.com/AzureAD/microsoft-identity-web/blob/1bb75bd2a0ec532ba35f07360b53ed294e915cea/ProjectTemplates/templates/ComponentsWebAssembly-CSharp/Client/wwwroot/index.html#L11
- line 15: replace `<div id="app">Loading...</div>` by `<app>Loading...</app>`

jennyf19:
Included in [0.3.0-preview release](https://github.com/AzureAD/microsoft-identity-web/releases/tag/0.3.0-preview).

PR #497, opened 2020-08-25
Fixes issues with BlazorServer (template and running)

jmprieur:
Fixes issues with BlazorServer (template and running)
- #496 (See the minimal change in **MicrosoftIdentityCircuitHandler.cs**)
- #490 extended to blazorserver2 templates, as they also had UI.

Improve the configuration of the generated templates (by test-templates.bat) for the Blazor templates (specific B2C_Blazorwasm_ClientId and B2C_HelloApi_ClientId) different from the regular B2C ones.


ProjectTemplates/configuration.json
```diff
@@ -11,6 +11,9 @@
         "B2C_WebApi_ClientId": "90c0fe63-bcf2-44d5-8fb7-b8bbc0b29dc6",
         "B2C_WebApi_Port": "44332",
 
+        "B2C_Blazorwasm_ClientId": "25f7fc07-bb5b-49ad-8179-6d486d094ca6",
```
jmprieur, line 14: This is to make it simpler to test the projects generated from the B2C blazor templates (because our test apps would require to change the route in the controller, etc ...). Using a new client makes it easier

Issue #506, opened 2020-08-27
Is it possible to use Microsoft.Graph.Beta instead of Microsoft.Graph?


schmid37:
Both Libraries are using the same namespaces… We have to use the Beta because of the new Intune fuctions and it produces reference errors.

Please evaluate if this is possible.

schmid37:
@jmprieur how do you think Microsoft should handle this? The namespace in Microsoft.Graph and Microsoft.Graph.Beta are the same... We definatelly need the new functions of Microsoft.Graph.Beta to integration Win32Lob apps in intune. This is not possible with Microsoft.Graph. This is very inconsistent.

jmprieur:
@darrelmiller: what would you suggest?
when would Intune be available in Graph v1.0 ?

schmid37:
@jmprieur Intune Graph v1.0 is available since I think the beginning. The whole world is waiting for the release of the actual Beta 😆

darrelmiller:
@schmid37   Microsoft.Graph.Beta is a superset of the Microsoft.Graph library so you could use the Microsoft.Graph.Beta library instead of Microsoft.Graph.  It would mean that your client is calling beta for all of the resources.  However, due to the fact that we don't allow breaking changes, calling the beta APIs would not likely cause issues.

The alternate approach to use `extern alias` to reference both assemblies even though they use the same namespace.  https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/extern-alias

schmid37:
@darrelmiller Thanks for the reply. We use the Microsoft.Graph.Beta for sure. Because we need to integrate Win32Lob apps in Intune. This is only possible with the beta. 

The question is, what impacts would @jmprieur have when they would use Microsoft.Graph.Beta instead of Microsoft.Graph?

An other question is, when will you find the consistency between Microsoft.Graph and Microsoft.Graph.Beta? Maybe through a release? Maybe through the possibility to provide previews within only **one** NuGet package? 

Nobody understands why there are two libraries with the same namespace. I don't know any company, we work with, which is amused about this situation. Companies which needs to integrate Win32Lob apps in Intune are reliant on using an beta... This is not really acceptable.

The alternate approach is a workaround with limits... So I already tried this and it won't really work with referenced projects in a solution.

gavinmnz:
[...]

PR #515, opened 2020-08-28
Changed NuGet package from Microsoft.Graph to Microsoft.Graph.Beta

schmid37:
Changed NuGet package from Microsoft.Graph to Microsoft.Graph.Beta #506


jmprieur:
@schmid37 : we cannot take this PR because Microsoft.Identity.Web is used in the ASP.NET Core 5.0 templates (from Preview 8), and it should leverage only released (GAed) libraries from RC2.
We'll GA Microsoft.Identity.Web ASAP, but Microsoft.Graph.Beta will always remain in preview.

jmprieur (commented):
We cannot take this PR, unfortunately as Microsoft.Graph.Beta will always remain in preview whereas Microsoft.Identity.Web needs to only leverage GA libraries as referenced from .NET 5 project templates.

Issue #441, opened 2020-08-12
[Question] How to get OBO token from an event handler


hrazmsft:
How can I get an OBO access token in an event handler like:

```
[HttpPost]
public IActionResult ProcessData(string data)
{
    var processor = new Processor();

    processor.Finished += async (s, e) =>
    {
        var accessToken = await this
            ._tokenAcquisition
            .GetAccessTokenForUserAsync(new[] { "Files.ReadWrite" }, user: this.User);

        await WriteToOneDrive(accessToken, e.Data);
    };

    processor.processAsync(data); // Run async without awaiting

    return NoContent();
}
```

In this case the `this.User` is already disposed, and even without it, `GetAccessTokenForUserAsync` crashes with NullObjectException.

My process takes a very long time (hours).

How can I get the token inside a callback when I need it?

jmprieur:
@hraz-msft when your start your long running process, you should keep some claims about the user (at least the home tenant id and home object id), and then you can use `ClaimsPrincipalFactory.FromTenantIdAndObjectId` to get back a claims principal, and pass it instead of this.User.

https://github.com/AzureAD/microsoft-identity-web/blob/41195539af65836b464b063c186842896ab15e0e/src/Microsoft.Identity.Web/ClaimsPrincipalFactory.cs#L36


So I would try:

```CSharp
[HttpPost]
public IActionResult ProcessData(string data)
{
    var processor = new Processor();
    string oid = User.GetObjectId();
    string tid = User.GetTenantId();

    processor.Finished += async (s, e) =>
    {
        var accessToken = await this
            ._tokenAcquisition
            .GetAccessTokenForUserAsync(
              new[] { "Files.ReadWrite" }, 
              user: ClaimsPrincipalFactory.FromTenantIdAndObjectId(tid, oid));

        await WriteToOneDrive(accessToken, e.Data);
    };

    processor.processAsync(data); // Run async without awaiting

    return NoContent();
}
```

hraz-msft:
Hmm, it still crushing internally... The `tid` and `oid` are valid. 

![image](https://user-images.githubusercontent.com/66222191/90116437-28d60380-dd5e-11ea-80c2-ac692bb1fcdb.png)

jmprieur:
I think that the problem is that the HttpContext is null, or HttpContext.Items is null:

[...]

PR #535, opened 2020-09-02
E2e tests

jennyf19:
- Adding support for e2e tests. 
- Right now, testing token acquisition - obo, validation of tokens, web api scope validation

more tests need to be added, but doing a PR w/the ground work.

also fix for #441 (could do in a separate PR, but then this one won't work w/out it, so leaving it here)


jmprieur (approved):
LGTM @jennyf19 
It's a really nice way of testing a Web API end to end.

jmprieur:
BTW, WheatherForecast is no longer needed @jennyf19 ?

jennyf19:
true. thank you. have removed

---
In reply to: [685853967](https://github.com/AzureAD/microsoft-identity-web/pull/535#issuecomment-685853967) [](ancestors = 685853967)

PR #542, opened 2020-09-03
Add support for the AzureSDK

jmprieur:
See https://github.com/tamram/storage-dotnet-azure-ad-msal/tree/tamram-0818

```CSharp
        [AuthorizeForScopes(Scopes = new string[] { "https://storage.azure.com/user_impersonation" })]
        public async Task<IActionResult> Blob()
        {
            var scopes = new string[] { "https://storage.azure.com/user_impersonation" }; // I guess the Blob SDK knows already?
            ViewData["Message"] = await CreateBlob(new TokenAcquisitionTokenCredential(_tokenAcquisition),);
            return View();
        }

        private static async Task<string> CreateBlob(TokenAcquisitionTokenCredential tokenCredential)
        {
            // Replace the URL below with the URL to your blob.
            Uri blobUri = new Uri("https://storagesamples.blob.core.windows.net/sample-container/blob1.txt");
            BlobClient blobClient = new BlobClient(blobUri, tokenCredential);

            // Create a blob on behalf of the user.
            string blobContents = "Blob created by Azure AD authenticated user.";
            byte[] byteArray = Encoding.ASCII.GetBytes(blobContents);

            using (MemoryStream stream = new MemoryStream(byteArray))
            {
                await blobClient.UploadAsync(stream);
            }
            return "Blob successfully created";
        }
```

### DIFF
--- a/src/Microsoft.Identity.Web/AzureSdkSupport/TokenAcquisitionTokenCredential.cs
+++ b/src/Microsoft.Identity.Web/AzureSdkSupport/TokenAcquisitionTokenCredential.cs
@@ -1,0 +1,43 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Threading;
+using System.Threading.Tasks;
+using Azure.Core;
+using Microsoft.Identity.Client;
+
+namespace Microsoft.Identity.Web
+{
+    /// <summary>
+    /// Azure SDK token credential based on the ITokenAcquisition service.
+    /// </summary>
+    public class TokenAcquisitionTokenCredential : TokenCredential
+    {
+        private ITokenAcquisition _tokenAcquisition;
+
+        /// <summary>
+        /// Constructor from an ITokenAcquisition service.
+        /// </summary>
+        /// <param name="tokenAcquisition">Token acquisition.</param>
+        public TokenAcquisitionTokenCredential(ITokenAcquisition tokenAcquisition)
+        {
+            _tokenAcquisition = tokenAcquisition;
+        }
+
+        /// <inheritdoc/>
+        public override AccessToken GetToken(TokenRequestContext requestContext, CancellationToken cancellationToken)
+        {
+            AuthenticationResult result = _tokenAcquisition.GetAuthenticationResultForUserAsync(requestContext.Scopes)
+                .GetAwaiter()
+                .GetResult();
+            return new AccessToken(result.AccessToken, result.ExpiresOn);
+        }
+
+        /// <inheritdoc/>
+        public override async ValueTask<AccessToken> GetTokenAsync(TokenRequestContext requestContext, CancellationToken cancellationToken)
+        {
+            AuthenticationResult result = await _tokenAcquisition.GetAuthenticationResultForUserAsync(requestContext.Scopes).ConfigureAwait(false);
+            return new AccessToken(result.AccessToken, result.ExpiresOn);
+        }
+    }
+}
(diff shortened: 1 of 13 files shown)


jmprieur:
@schaabs : do you want to have a look?

jmprieur:
@jennyf19 
I met with @schaabs yesterday night and he likes this.

src/Microsoft.Identity.Web/AzureSdkSupport/TokenAcquisitionTokenCredential.cs
```diff
@@ -0,0 +1,43 @@
...
+                .GetAwaiter()
+                .GetResult();
+            return new AccessToken(result.AccessToken, result.ExpiresOn);
+        }
+
+        /// <inheritdoc/>
+        public override async ValueTask<AccessToken> GetTokenAsync(TokenRequestContext requestContext, CancellationToken cancellationToken)
+        {
+            AuthenticationResult result = await _tokenAcquisition.GetAuthenticationResultForUserAsync(requestContext.Scopes).ConfigureAwait(false);
+            return new AccessToken(result.AccessToken, result.ExpiresOn);
```
jennyf19, line 40: same as above, and how do we test this?

jennyf19:
Included in [1.9.0 release](https://github.com/AzureAD/microsoft-identity-web/releases/tag/1.9.0)

harris-boyce:
@jmprieur What would it look like when developing a multi-tenant application with requirements to authenticate as an application/service principal?  Would it be best to use `Client[Secret|Certificate]Credential` and let Azure.Identity handle token cache?

maliksahil:
@harris-boyce please see this for authenticating using managed identity. Note that SPs currently do not work cross-tenant in the way you describe. https://docs.microsoft.com/en-us/azure/storage/blobs/authorize-managed-identity

jmprieur:
Microsoft.Identity.Web leverages Azure.Identity to get the client certs and decrypt certs from keyvault.
What else would you need. @harris-boyce ?

harris-boyce:
> Microsoft.Identity.Web leverages Azure.Identity to get the client certs and decrypt certs from keyvault. What else would you need. @harris-boyce ?

@jmprieur My scenarios cover the need to leverage the Azure SDK management libraries to operate against Azure resources as both the app principal and on-behalf-of the signed in user.  For example:

 - User -> OBO token (`tokenAcquisition.GetAccessTokenForUserAsync(tenantId)`) -> `GET https://mgmt.az.com/subscriptions` -> `PUT https://mgmt.az.com/subs/rg/Ms.Authz/roleAssignment`
 - App -> App token (`tokenAcquisition.GetAccessTokenForAppAsync(tenantId)`) -> `GET https://mgmt.az.com/subscriptions`

In both cases above, `tenantId` is intended to be the home tenant of the signed in user and the Web App is registered as a multi-tenant application (SaaS scenario).

I have worked around this thus far by creating my own implementations of `TokenAcquisitionTokenCredential` that call `GetAuthenticationResultFor[User|App]Async()`, where appropriate, and supply the tenant Id explicitly in `GetToken`.

I would be willing to contribute this code to the project as it feels like this could be a valid use case beyond my scenarios. Thoughts?

> @harris-boyce please see this for authenticating using managed identity. Note that SPs currently do not work cross-tenant in the way you describe. https://docs.microsoft.com/en-us/azure/storage/blobs/authorize-managed-identity

@maliksahil I'm not sure why this would be the case; my understanding is that managed identities are simply special implementations of Service Principals, so why wouldn't I be able to grant RBAC Data Reader rights to a service principal and use it to perform operations against a storage account?

jmprieur:
@harris-boyce : sure, feel free to contribute. Would it be a change on TokenAcquisitionTokenCredential?

PR #558, opened 2020-09-08
fix strong name issue

jennyf19:
thanks @jmprieur for noticing this.


jmprieur (approved):
LGTM.
Thanks @jennyf19

Issue #589, opened 2020-09-17
[Bug] Error when attempting to use PCKE


calbert82uhah:
**Which version of Microsoft Identity Web are you using?**
0.4.0-preview

**Where is the issue?**
* Web app
    * [X] Sign-in users
    * [ ] Sign-in users and call web APIs
* Web API
    * [ ] Protected web APIs (validating tokens)
    * [ ] Protected web APIs (validating scopes)
    * [ ] Protected web APIs call downstream web APIs
* Token cache serialization
     * [ ] In-memory caches
     * [ ] Session caches
     * [ ] Distributed caches
* Other (please describe)

**Is this a new or an existing app?**
a. The app is in production and I have upgraded to a new version of Microsoft Identity Web.

**Repro**

            services.AddMicrosoftIdentityWebAppAuthentication(Configuration)
                .EnableTokenAcquisitionToCallDownstreamApi()
                .AddMicrosoftGraph(Configuration.GetSection("DownstreamApi"))
                .AddInMemoryTokenCaches();

            services.Configure<MicrosoftIdentityOptions>(options =>
            {
                options.AuthenticationMethod = OpenIdConnectRedirectBehavior.RedirectGet;
                options.UsePkce = true;
                options.ResponseType = OpenIdConnectResponseType.Code;
            });

            services.AddMvc();

            services.AddControllersWithViews(options =>
            {
                var policy = new AuthorizationPolicyBuilder()
                    .RequireAuthenticatedUser()
                    .Build();
                options.Filters.Add(new AuthorizeFilter(policy));
            });

**Expected behavior**
I try to upgrade from Microsoft.AspNetCore.Authentication.AzureAd.UI to Microsoft.Identity.Web.
I reuse the app registration, which works fine with an Authorization Code Flow.
I expect that the app successfully logs in the user using Auth Code Flow.

**Actual behavior**
When trying to log in, I get the following error:

`OpenIdConnectProtocolException: Message contains error: 'unsupported_response_type', error_description: 'AADSTS700054: response_type 'id_token' is not enabled for the application.`

I don't really understand the error message. Is it trying to perform an implicit flow? 

With the older package, the initial redirect to `https://login.microsoftonline.com` contains the query `&response_type=code`; with Microsoft.Identity.Web it contains` response_type=code%20id_token`; why is it doing that?

jmprieur:
[...]

PR #591, opened 2020-09-17
Attempt to avoid using the IDToken when the web app calls a web API.

jmprieur:
Investigation for https://github.com/AzureAD/microsoft-identity-web/issues/589

@jennyf19 I tested all the dev samples except a blazor one.

### DIFF
--- a/src/Microsoft.Identity.Web/AzureADB2COpenIDConnectEventHandlers.cs
+++ b/src/Microsoft.Identity.Web/AzureADB2COpenIDConnectEventHandlers.cs
@@ -41,7 +41,14 @@ namespace Microsoft.Identity.Web
                 }
                 else
                 {
-                    context.ProtocolMessage.ResponseType = OpenIdConnectResponseType.CodeIdToken;
+                    if (Options.IsB2C)
+                    {
+                        context.ProtocolMessage.ResponseType = OpenIdConnectResponseType.CodeIdToken;
+                    }
+                    else
+                    {
+                        context.ProtocolMessage.ResponseType = OpenIdConnectResponseType.Code;
+                    }
                 }
             }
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.Identity.Web/AzureADB2COpenIDConnectEventHandlers.cs
```diff
@@ -41,7 +41,14 @@ public Task OnRedirectToIdentityProvider(RedirectContext context)
...
                 else
                 {
-                    context.ProtocolMessage.ResponseType = OpenIdConnectResponseType.CodeIdToken;
+                    if (Options.IsB2C)
+                    {
+                        context.ProtocolMessage.ResponseType = OpenIdConnectResponseType.CodeIdToken;
+                    }
+                    else
+                    {
+                        context.ProtocolMessage.ResponseType = OpenIdConnectResponseType.Code;
```
jmprieur, line 50: With AAD this can work without the IDToken.
For B2C when we ask for another policy we need to have the IDToken

src/Microsoft.Identity.Web/WebAppExtensions/MicrosoftIdentityWebAppAuthenticationBuilder.cs
```diff
@@ -93,7 +93,8 @@ public class MicrosoftIdentityWebAppAuthenticationBuilder : MicrosoftIdentityBas
             services.AddOptions<OpenIdConnectOptions>(openIdConnectScheme)
                  .Configure<IServiceProvider>((options, serviceProvider) =>
                  {
-                     options.ResponseType = OpenIdConnectResponseType.CodeIdToken;
+                     options.ResponseType = OpenIdConnectResponseType.Code;
```
jmprieur, line 96: By default when the ResponseType is Code, it seems PKCE is set to true by default.
Given MSAL.NET does not support it yet, switching it to false for the moment.

jennyf19 (approved):

src/Microsoft.Identity.Web/AzureADB2COpenIDConnectEventHandlers.cs
jennyf19: i'll try to test this out today.

Issue #573, opened 2020-09-12
[Bug] Looping redirect with Azure Signalr Service


jassent:
**Which version of Microsoft Identity Web are you using?**
Microsoft.Identity.Web (0.4.0-preview)

**Where is the issue?**
* Web app
      [X] Sign-in users and call web APIs

**Is this a new or an existing app?**
This is a new app or an experiment.

**Repro**
This sample app demonstrates the bug and uses Microsoft.Identity.Web, Blazor Server, Azure Signalr Service and Microsoft Graph:
https://github.com/jassent/Example-Blazor-Graph-Microsoft-Identity-Web/tree/master/tests/BlazorServerCallsGraph

The goal is to add Azure Signalr Service. Adding the following code to startup.cs causes an endless looping redirect when calling Microsoft Graph once the Blazor Server app is using Azure Signalr Service.

```csharp
services.AddSignalR().AddAzureSignalR();
```

**Expected behavior**
The expected behavior is to be able to use Microsoft.Identity.Web in a Blazor Server web app that uses Microsoft Azure Signalr Service. 

**Actual behavior**

Build the app and visit: https://localhost:44314/showprofile

Web App works fine when Blazor Server is run self-hosted. Migrating the Web App to Azure Signalr Service causes the problem. Adding "services.AddSignalR().AddAzureSignalR();" to Startup.cs causes the web page calling the Graph Api to continuously reload repeatedly calling https://login.microsoftonline.com/{tenantid}/oauth2/v2.0/token

There is an exception thrown in the visual studio debug window but no meaningful information:
Exception thrown: 'Microsoft.Graph.ServiceException' in System.Private.CoreLib.dll System.Net.Http.HttpClient.Default.LogicalHandler: Information: Start processing HTTP request POST https://login.microsoftonline.com/{tenantid}/oauth2/v2.0/token 

System.Net.Http.HttpClient.Default.ClientHandler: Information: Sending HTTP request POST https://login.microsoftonline.com/{tenantid}/oauth2/v2.0/token

System.Net.Http.HttpClient.Default.ClientHandler: Information: Received HTTP response after 201.8975ms - OK 

System.Net.Http.HttpClient.Default.LogicalHandler: Information: End processing HTTP request after 209.1404ms - OK

The browser shows this error in the console window:
blazor.server.js:19 [2020-09-12T12:35:54.880Z] Error: Circuit has been shut down due to error.

Graph Api data is returned to the application between each login loop.

[...]

Issue #404, opened 2020-08-05
[Feature Request] AddTokenAcquisition should take a delegate rather than a boolean


jmprieur:
**Is your feature request related to a problem? Please describe.**
AddTokenAcquisition is taking a boolean (to use a singleton service instead of a scoped one)

**Describe the solution you'd like**
This is not the pattern that the ASP.NET Core team uses.
Please have an option + a delegate.

**Suggested design**
Add an option class:

```C#
    public class TokenAcquisitionOptions
    {
        public bool IsSingleton { get; set} = false;
    }
```

Change `AddTokenAcquisition` to take an optional delegate to configure the option, rather than the boolean

```C#
       public static IServiceCollection AddTokenAcquisition(
            this IServiceCollection services,
            Action<TokenAcquisitionOptions> initializeTokenAcquisitionOptions = null)
```

Change the implementation to use the options ...

jmprieur:
@pmaytak : proposing to assign this one to you.

jmprieur:
We are actually not going to do this issue because a singleton has some bad effects:
- [Performance: when IConfidentialClientApplication is used as a singleton service, the cache size grows #454](https://github.com/AzureAD/microsoft-identity-web/issues/454) for the moment
- [Looping redirect with Azure Signalr Service #573](https://github.com/AzureAD/microsoft-identity-web/issues/573)

PR #593, opened 2020-09-18
fix blazor loop w/azure signalR service

jennyf19:
fix for #573 and also addressing #404 

FYI: @jmprieur for the singleton vs scoped issue

### DIFF
--- a/src/Microsoft.Identity.Web.MicrosoftGraph/MicrosoftGraphExtensions.cs
+++ b/src/Microsoft.Identity.Web.MicrosoftGraph/MicrosoftGraphExtensions.cs
@@ -66,7 +66,6 @@ namespace Microsoft.Identity.Web
 
             // https://docs.microsoft.com/en-us/dotnet/standard/microservices-architecture/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests
             builder.Services.AddOptions<MicrosoftGraphOptions>().Configure(configureMicrosoftGraphOptions);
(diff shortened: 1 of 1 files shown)


jmprieur (approved):
LGTM.
Thanks @jennyf19. This was very subtle!
Tested

src/Microsoft.Identity.Web.MicrosoftGraph/MicrosoftGraphExtensions.cs
```diff
@@ -66,9 +66,9 @@ public static class MicrosoftGraphExtensions
 
             // https://docs.microsoft.com/en-us/dotnet/standard/microservices-architecture/implement-resilient-applications/use-httpclientfactory-to-implement-resilient-http-requests
             builder.Services.AddOptions<MicrosoftGraphOptions>().Configure(configureMicrosoftGraphOptions);
-            builder.Services.AddTokenAcquisition(true);
+            builder.Services.AddTokenAcquisition(false);
```
pmaytak, line 69: It's `false` by default so it can just be omitted.

pmaytak (approved):

src/Microsoft.Identity.Web.MicrosoftGraph/MicrosoftGraphExtensions.cs
jennyf19, line 69:
```suggestion
```

PR #596, opened 2020-09-18
add token acquisition options

jennyf19:
need to test, so marking as draft.

### DIFF
--- a/src/Microsoft.Identity.Web/DownstreamWebApiSupport/DownstreamWebApi.cs
+++ b/src/Microsoft.Identity.Web/DownstreamWebApiSupport/DownstreamWebApi.cs
@@ -3,13 +3,11 @@
 
 using System;
 using System.Globalization;
-using System.Net;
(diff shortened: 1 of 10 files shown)


src/Microsoft.Identity.Web/DownstreamWebApiSupport/DownstreamWebApi.cs
```diff
@@ -74,7 +74,9 @@ public class DownstreamWebApi : IDownstreamWebApi
             string accessToken = await _tokenAcquisition.GetAccessTokenForUserAsync(
                 effectiveOptions.GetScopes(),
                 effectiveOptions.Tenant,
-                userflow)
+                userflow,
+                user,
```
jmprieur, line 78: good catch for missing `user`

jmprieur (approved):
Thanks @jennyf19 
Those are good additions that our 1P partners were expecting :)

LGTM

PR #599, opened 2020-09-21
fix changelog format

jennyf19:


jmprieur (approved):
Thanks @jennyf19 
It's much better

PR #601, opened 2020-09-22
remove net 5.0 targetframework in templates, as we are in the .net 5 …

jennyf19:
…templates already


jmprieur (approved):
LGTM 
Thanks @jennyf19

PR #604, opened 2020-09-22
fixing up a few some doc related things

jennyf19:


jmprieur (commented):
good catch, @jennyf19. Thanks!

PR #609, opened 2020-09-23
fix webapi2-b2c default su_si value

jennyf19:
@jmprieur i think we could release the template package & rerun that, but not run a new overall release build.


jmprieur (approved):
LGTM.
Thanks @jennyf19 
Good catch for the default policy

PR #614, opened 2020-09-25
Jmprieur/perf with scalable Public Client Application cache

jmprieur:
Jmprieur/perf with scalable Public Client Application cache in a different file for each account

### DIFF
--- a/tests/Perf/Microsoft.Identity.Web.Perf.Client/TestRunner.cs
+++ b/tests/Perf/Microsoft.Identity.Web.Perf.Client/TestRunner.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
 using System.Linq;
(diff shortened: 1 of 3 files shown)


tests/Perf/Microsoft.Identity.Web.Perf.Client/TestRunner.cs
```diff
@@ -129,6 +123,13 @@ private async Task<AuthenticationResult> AcquireTokenAsync(int userIndex)
             var scopes = new string[] { _configuration["ApiScopes"] };
             var upn = $"{NamePrefix}{userIndex}@{_configuration["TenantDomain"]}";
 
+            var _msalPublicClient = PublicClientApplicationBuilder
```
jmprieur, line 126: Recreating a PublicClientApplication each time to avoid the MSAL.NET performance bug which makes the cache grow with the number of users even if we store it in a different file (singleton)

tests/Perf/Microsoft.Identity.Web.Perf.Client/TokenCacheHelper.cs
```diff
@@ -14,29 +14,21 @@ static class TokenCacheHelper
         /// </summary>
         public static readonly string s_cacheFilePath = System.Reflection.Assembly.GetExecutingAssembly().Location + ".msalcache.bin3";
 
-        private static readonly object s_fileLock = new object();
-
         public static void BeforeAccessNotification(TokenCacheNotificationArgs args)
         {
```
jmprieur, line 18: The lock is not necessary and takes time

jennyf19: yes, i already removed that

Issue #509, opened 2020-08-27
[Feature Request] In the Blazorwasm templates hosted, show how to call graph and a downstream api


jmprieur:
**Is your feature request related to a problem? Please describe.**
The Blasorwasm template has all the initialization to call a downstream API, or Graph, but we did not expose pages that actually call them

jmprieur:
Work to be done (in the Client):
- in the Pages sub folder, add the CallWebApi.razor and ShowProfile.razor pages from https://github.com/AzureAD/microsoft-identity-web/tree/master/ProjectTemplates/templates/BlazorServerWeb-CSharp/Pages
- in the Shared sub folder, add the NavMenu.* pages from https://github.com/AzureAD/microsoft-identity-web/tree/master/ProjectTemplates/templates/BlazorServerWeb-CSharp/Shared
- in the template.json file, add the conditional inclusion of the pages: https://github.com/AzureAD/microsoft-identity-web/blob/5d3c277ed53e060b91f0c1c36d041e09e6055388/ProjectTemplates/templates/BlazorServerWeb-CSharp/.template.config/template.json#L131-L173

jennyf19:
Included in [1.1.0 release](https://github.com/AzureAD/microsoft-identity-web/releases/tag/1.1.0).

PR #622, opened 2020-09-29
separate out graph and call web api for blazorwasm hosted

jennyf19:
Fixes https://github.com/AzureAD/microsoft-identity-web/issues/509


ProjectTemplates/templates/ComponentsWebAssembly-CSharp/.template.config/template.json
```diff
@@ -205,11 +205,56 @@
...
+          "exclude": [
+            "Client/Shared/NavMenu.CallsMicrosoftGraph.razor",
+            "Client/Pages/ShowProfile.razor",
+            "Server/Controllers/ShowProfileController.cs"
+            ]
+        },
+        {
+          "condition": "(!GenerateApiOrGraph)",
+          "rename": {
+            "Client/Shared/NavMenu.NoGraphOrApi.razor": "Client/Shared/NavMenu.razor"
```
jmprieur, line 233: In **blazorwasm2-singleorg-hosted.Client**, under Client\Shared we don't have the **NavMenu.razor** but the **NavMenu.NoGraphorAPI.razor.razor**

My hypothesis is that the template engine did not do the work correctly because the file is named NavMenu.NoGraphorAPI.razor :( on disk [note the or instead of Or)

See below **ProjectTemplates/templates/ComponentsWebAssembly-CSharp/Client/Shared/NavMenu.NoGraphorApi.razor**

Because on Windows the file name are case insensitive, to fix this, you would need to send 2 commits:
- one commits that deletes the NavMenu.NoGraphorApi.razor file
- another commit to re-add it.
The best would be to move it to another folder do the first commit, rename it in the other folder, and move it back where it should be.

ProjectTemplates/AspNetCoreMicrosoftIdentityWebProjectTemplates.csproj
```diff
@@ -36,4 +36,11 @@
     <Content Include="templates\**\*" Exclude="templates\**\bin\**;templates\**\obj\**" />
     <Compile Remove="**\*" />
   </ItemGroup>
+
```
jmprieur, line 39: I don't understand these 4 lines? 
I think it's a tooling (Visual Studio) issue.

ProjectTemplates/templates/ComponentsWebAssembly-CSharp/.template.config/template.json
```diff
@@ -205,11 +205,56 @@
           ]
         },
         {
-          "condition": "(!IndividualLocalAuth || !Hosted)",
+            "condition": "(!IndividualLocalAuth || !Hosted)",
```
jmprieur: There is an indentation issue (one less space)

ProjectTemplates/templates/ComponentsWebAssembly-CSharp/.template.config/template.json
```diff
@@ -205,11 +205,56 @@
...
+                "Server/Controllers/OidcConfigurationController.cs",
+                "Server/Models/ApplicationUser.cs"
+            ]
+        },
+         {
+          "condition": "(!GenerateApi)",
+            "exclude": [
+              "Client/Pages/CallWebApi.razor",
+              "Server/Controllers/CallWebApiController.cs"
+            ]
```
jmprieur: I think we also want to exclude "Client/Shared/NavMenu.CallsWebApi.razor"

jmprieur (commented):
This is awesome work @jennyf19.
That was much needed.

There are a couple of issues to fix.

ProjectTemplates/templates/ComponentsWebAssembly-CSharp/.template.config/template.json
jmprieur: note that I pushed a small commit to fix a small indentation issue because GitHub was very confused, and lines were appearing at the bottom, which are at the top

ProjectTemplates/templates/ComponentsWebAssembly-CSharp/.template.config/template.json
```diff
@@ -211,6 +211,51 @@
...
+        {
+          "condition": "(!GenerateGraph)",
+          "exclude": [
+            "Client/Shared/NavMenu.CallsMicrosoftGraph.razor",
+            "Client/Pages/ShowProfile.razor",
+            "Server/Controllers/ShowProfileController.cs"
+            ]
+        },
+        {
+          "condition": "(!GenerateApiOrGraph)",
```
jmprieur, line 231: For the renaming, we need to change this condition into two (because the layout on disk is different when not-hosted (all flat), and hosted (sub folder client, server, etc ...)

image


```json
        {
          "condition": "(!GenerateApiOrGraph && Hosted)",
          "rename": {
            "Client/Shared/NavMenu.NoGraphOrApi.razor": "Client/Shared/NavMenu.razor"
          },
          "exclude": [
            "Client/Shared/NavMenu.CallsMicrosoftGraph.razor",
            "Client/Shared/NavMenu.CallsWebApi.razor"
          ]
        },
        {
          "condition": "(!GenerateApiOrGraph && !Hosted)",
          "rename": {
            "Client/Shared/NavMenu.NoGraphOrApi.razor": "Shared/NavMenu.razor"
          },
          "exclude": [
            "Client/Shared/NavMenu.CallsMicrosoftGraph.razor",
            "Client/Shared/NavMenu.CallsWebApi.razor"
          ]
        },
```
```

File: ProjectTemplates/AspNetCoreMicrosoftIdentityWebProjectTemplates.csproj [hunk 6]
```diff
@@ -36,4 +36,11 @@
...
+  <ItemGroup>
```
jmprieur, line 40: I thought I had provided a comment here, but it disappeared
These files should not be excluded.

This is due to a bug in Visual Studio (when you did a renaming)
Just remove these 6 lines (the ItemGroup)

ProjectTemplates/templates/ComponentsWebAssembly-CSharp/.template.config/template.json
jennyf19: code flow is not opening for me atm. but i addressed this.

ProjectTemplates/templates/ComponentsWebAssembly-CSharp/.template.config/template.json
jennyf19: got it. thanks.

jmprieur (approved):
Awesome, @jennyf19 
I retested all the blasorwasm projects and this all works
:shipit:

Issue #617, opened 2020-09-29
[Bug] add tid long-claim check in AadIssuerValidator.GetTenantIdFromToken


jmprieur:
**Which version of Microsoft Identity Web are you using?**
1.0.0

**Where is the issue?**
* Web app
    * [x ] Sign-in users
    * [ ] Sign-in users and call web APIs
* Web API
    * [ x] Protected web APIs (validating tokens)
    * [ ] Protected web APIs (validating scopes)
    * [ ] Protected web APIs call downstream web APIs
* Token cache serialization
     * [ ] In-memory caches
     * [ ] Session caches
     * [ ] Distributed caches
* Other (please describe)

**Is this a new or an existing app?**
Not a regression. Found by @GeoK when reviewing the code. No customer impact known

**Repro**
See https://github.com/AzureAD/microsoft-identity-web/blob/5d3c277ed53e060b91f0c1c36d041e09e6055388/src/Microsoft.Identity.Web/Resource/AadIssuerValidator.cs#L177

There is no long-claim check

**Expected behavior**
Have a long-claim tid check in addition to the short tid claim check

**Actual behavior**
long tid claims not checked from the token. No know customer impact (hence P2)

jennyf19:
Included in [1.1.0 release](https://github.com/AzureAD/microsoft-identity-web/releases/tag/1.1.0).

PR #629, opened 2020-09-30
add long form check of tenant id

jennyf19:
#617

### DIFF
--- a/src/Microsoft.Identity.Web/Resource/AadIssuerValidator.cs
+++ b/src/Microsoft.Identity.Web/Resource/AadIssuerValidator.cs
@@ -174,7 +174,13 @@ namespace Microsoft.Identity.Web.Resource
         {
             if (securityToken is JwtSecurityToken jwtSecurityToken)
             {
-                if (jwtSecurityToken.Payload.TryGetValue(ClaimConstants.Tid, out object? tenantId))
+                if (jwtSecurityToken.Payload.TryGetValue(ClaimConstants.Tid, out object? tid))
+                {
+                    return (string)tid;
+                }
+
+                jwtSecurityToken.Payload.TryGetValue(ClaimConstants.TenantId, out object? tenantId);
+                if (tenantId != null)
                 {
                     return (string)tenantId;
                 }
--- a/src/Microsoft.Identity.Web/Resource/AadIssuerValidator.cs
+++ b/src/Microsoft.Identity.Web/Resource/AadIssuerValidator.cs
@@ -191,6 +191,12 @@ namespace Microsoft.Identity.Web.Resource
                     return tid;
                 }
 
+                jsonWebToken.TryGetPayloadValue(ClaimConstants.TenantId, out string? tenantId);
+                if (tenantId != null)
+                {
+                    return tenantId;
+                }
+
                 // Since B2C doesn't have "tid" as default, get it from issuer
                 return GetTenantIdFromIss(jsonWebToken.Issuer);
             }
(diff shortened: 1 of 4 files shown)


jmprieur (approved):
LGTM.
I'll let @GeoK verify

tests/Microsoft.Identity.Web.Test/Resource/AadIssuerValidatorTests.cs
```diff
@@ -100,11 +98,22 @@ public void Validate_NullOrEmptyTenantId_ThrowsException()
             Assert.Equal(IDWebErrorMessage.TenantIdClaimNotPresentInToken, exception.Message);
         }
 
-        [Fact]
-        public void Validate_IssuerMatchedInValidIssuer_ReturnsIssuer()
+        [Theory]
+        [InlineData(true)]
+        [InlineData(false)]
+        public void Validate_IssuerMatchedInValidIssuer_ReturnsIssuer(bool shortTidClaim)
```
pmaytak, line 104: Instead of a bool, can pass the claim name itself as a parameter, and then into a constructor.

src/Microsoft.Identity.Web/Resource/AadIssuerValidator.cs
```diff
@@ -191,6 +191,12 @@ private static string GetTenantIdFromToken(SecurityToken securityToken)
                     return tid;
                 }
 
+                jsonWebToken.TryGetPayloadValue(ClaimConstants.TenantId, out string? tenantId);
```
pmaytak, line 194: Does this also have to be added inside the above `if (securityToken is JwtSecurityToken jwtSecurityToken)`?

tests/Microsoft.Identity.Web.Test/Resource/AadIssuerValidatorTests.cs
jennyf19: yes. nice suggestion.

src/Microsoft.Identity.Web/Resource/AadIssuerValidator.cs
jennyf19: yes. good catch.

GeoK: Are you looking to provide support for JsonWebToken as well? I see that on some places you have support for both JwtSecurityToken and JsonWebToken.

src/Microsoft.Identity.Web/Resource/AadIssuerValidator.cs
```diff
@@ -174,7 +174,13 @@ private static string GetTenantIdFromToken(SecurityToken securityToken)
         {
             if (securityToken is JwtSecurityToken jwtSecurityToken)
             {
-                if (jwtSecurityToken.Payload.TryGetValue(ClaimConstants.Tid, out object? tenantId))
+                if (jwtSecurityToken.Payload.TryGetValue(ClaimConstants.Tid, out object? tid))
```
GeoK, line 177: nit: not consistent with other if-conditions that are checking for tid/tenantId.

GeoK (approved):

src/Microsoft.Identity.Web/Resource/AadIssuerValidator.cs
jennyf19: @GeoK the `TryGetPayloadValue` isn't exposed on the `JwtSecurityToken`, like it is on the `JsonWebToken` object. hence the difference.

Issue #639, opened 2020-10-02
Include MemoryCacheOptions w/the memory cache for performance improvement


jennyf19:
Why?
The InMemoryCache grows for ever unless it's limited.

What?
Provide information to the in-memory cache so that it can limit its memory consumption

See https://docs.microsoft.com/en-us/aspnet/core/performance/caching/memory

jennyf19:
Included in [1.1.0 release](https://github.com/AzureAD/microsoft-identity-web/releases/tag/1.1.0).

PR #640, opened 2020-10-02
add memory cache options

jennyf19:
#639


jmprieur (approved):
LGTM. @jennyf19 
do we need a unit test? (we cannot know the size of the in memory cache, so that might be a bit pointless?)

pmaytak (approved):
👍

jennyf19:
i added a small one. thx.

---
In reply to: [501235649](https://github.com/AzureAD/microsoft-identity-web/pull/640#issuecomment-501235649) [](ancestors = 501235649)

Issue #626, opened 2020-09-30
[Bug] ConsentHandler in MVC Partial View Produces Malformed Redirect Uri


phil4business:
**Which version of Microsoft Identity Web are you using?**
Microsoft Identity Web 1.0.0

**Where is the issue?**
* Web app
    * [x] Sign-in users and call web APIs
* Token cache serialization
    * [x] Distributed caches

**Is this a new or an existing app?**
c. This is a new app or an experiment.

**Repro**

StartUp.cs
```csharp
services.Configure<AuthenticationOptions>(Configuration.GetSection("AzureAd"));
string[] initialScopes = Configuration.GetSection("ServerRequestApi:Scopes").Get<string[]>();

services.AddMicrosoftIdentityWebAppAuthentication(Configuration)
    .EnableTokenAcquisitionToCallDownstreamApi(initialScopes)
    .AddDistributedTokenCaches();

services.TryAddScoped<MicrosoftIdentityConsentAndConditionalAccessHandler>();

services.AddDistributedSqlServerCache(options =>
{
    options.ConnectionString = Configuration.GetConnectionString("DefaultConnection");
    options.SchemaName = "dbo";
    options.TableName = "TokenCache";
});

services.AddHttpClient<IApiHealthService, ApiHealthService>();

services.AddHttpContextAccessor();

services.AddControllersWithViews(options =>
{
    var policy = new AuthorizationPolicyBuilder()
        .RequireAuthenticatedUser()
        .Build();
    options.Filters.Add(new AuthorizeFilter(policy));
}).AddMicrosoftIdentityUI();
```
ApiHealthService.cs
```csharp
[AuthorizeForScopes(ScopeKeySection = "ServerRequestApi:Scopes)")]
public class ApiHealthService : IApiHealthService
{
    private readonly HttpClient _apiClient;
    private readonly ILogger<ApiHealthService> _logger;
    private readonly IConfiguration _configuration;
    private readonly ITokenAcquisition _tokenAcquisition;
    private readonly MicrosoftIdentityConsentAndConditionalAccessHandler _consentHandler;

    private readonly string _remoteServiceBaseUrl;
    private readonly string[] _scopes;

    public ApiHealthService(
        HttpClient httpClient,
        ILogger<ApiHealthService> logger,
        IConfiguration configuration,
        ITokenAcquisition tokenAcquisition,
        MicrosoftIdentityConsentAndConditionalAccessHandler consentHandler)
    {
[...]

PR #655, opened 2020-10-05
ensure correct redirect links

robi26:
Ensuring the redirect URI is correct, fixing #626


jennyf19 (approved):
:shipit: thanks @robi26 great fix.

Issue #583, opened 2020-09-15
[Bug] Fix newly failing test


jennyf19:
W/same version of .net core (5.0.100-preview.7.20366.6) and VS 2019, the OpenIdConnectOptions properties
returns `MapInboundClaims` as a 57th claim, when we should be getting 56. This happens when running in Github as well.

this [PR](https://github.com/AzureAD/microsoft-identity-web/pull/582) skips the test for now.

jennyf19:
Included in [1.1.0 release](https://github.com/AzureAD/microsoft-identity-web/releases/tag/1.1.0).

PR #656, opened 2020-10-05
reenable test

jennyf19:
#583


pmaytak (commented):
@jennyf19 What do you think about updating our GitHub build .NET version to RC1, adding the new property to `PopulateOpenIdOptionsFromMicrosoftIdentityOptions` method and updating the test?

We'll still have to do this once stable .NET 5 comes out.

jmprieur:
The issue @pmaytak is that would require us (and our customers) to install the preview 3 version of Visual Studio 16.8, otherwise we won't be able to build locally

Issue #643, opened 2020-10-03
[Fundamentals] Add additional metrics/counters to the PerformanceTestService


pmaytak:
Add metrics:
- Cache item size
- Id.Web Cache size in MB, and in number of items, 
- Id.Web Cache hits/misses, 
- Id.Web Cache load/save timing numbers, 
- Id.Web Errors (by error code)
- MSAL cache hits/misses,  
- MSAL cache load/save timing numbers
- MSAL Errors (by error code)

Note: Might need to update MSAL to send the telemetry.

Also add metrics for System.Runtime ([here](https://github.com/AzureAD/microsoft-identity-web/blob/6bbc143b33a07ec6fb34562ace6b8d0afbe244af/tests/PerformanceTests/PerformanceTestService/Startup.cs#L42-L53)) and verify if they are needed/more accurate than the default ones in the App Insights:
- cpu-usage
- working-set

Additionally should think if it's better to use EventCounters or TelemetryClient to send metrics.

bgavrilMS:
No work is needed on MSAL to get these metrics. For cache hit / miss, there exists `AuthenticationResult.ResultMetadata.TokenSource` which will be set to Cache in case of a hit. 

What is missing here is the time spent in network calls, which can be calculated by injecting an HTTPClient that monitors the network, similar to  [this one](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/5000f4a5624c1acd4cf1c07b25c6abb5eedf746c/tests/Microsoft.Identity.Test.Common/Core/Helpers/HttpSnifferClientFactory.cs#L11))

Overall, we care more about time spent in MSAL / M.I.W than on the network and in the cache, although total time using a well-known cache implementation (e.g. Redis, as you propose) is what customers want to see.

pmaytak:
I added some related thoughts on this in [Feature Request High-Level perf assement scenario #2146](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/2146).
`AuthenticationResult.ResultMetadata.TokenSource` could be checked in the test runner that directly calls MSAL (as opposed to in the service). We do that in MIW perf client.

pmaytak:
Closing as the cache related metrics were added. Postponing adding error metrics (created #735 for tracking that).

PR #669, opened 2020-10-08
Add perf counters

pmaytak:
Added cache related counters: cache size, read misses, read and write durations.
Refactored BenchmarkMsalDistributedTokenCacheAdapter, BenchmarkMsalMemoryTokenCacheProvider to directly call the caches because the old way (making these classes a wrapper) was breaking up the method signatures.
Other minor refactorings.

Issue #643 partially.


tests/PerformanceTests/PerformanceTestService/EventSource/BenchmarkMsalDistributedTokenCacheAdapter.cs
```diff
@@ -25,10 +26,17 @@ public class BenchmarkMsalDistributedTokenCacheAdapter : MsalDistributedTokenCac
...
         /// <returns>A <see cref="Task"/> that completes when key removal has completed.</returns>
-        protected override async Task RemoveKeyAsync(string cacheKey)
+        protected override Task RemoveKeyAsync(string cacheKey)
         {
+            var bytes = base.ReadCacheBytesAsync(cacheKey).GetAwaiter().GetResult();
+
+            if (bytes != null)
+            {
+                MemoryCacheEventSource.Log.DecrementSize(bytes.Length);
+            }
```
henrik-me, line 36: what is the cpu cost of doing this?

pmaytak: As per docs, EventSource is low overhead. AppInsights SDK gets the data every 60 seconds. If we want to collect periodically, there's no way to directly get item count and size from the memory caches, so seems like we would still have to have some sort of a counter running.

tests/PerformanceTests/PerformanceTestService/EventSource/MemoryCacheEventSource.cs
```diff
@@ -54,32 +63,80 @@ public void IncrementReadCount()
...
+        /// <remarks>
+        /// This is the convention for initializing counters in the RuntimeEventSource (lazily on the first enable command).
+        /// They aren't disabled afterwards
+        /// </remarks>
         protected override void OnEventCommand(EventCommandEventArgs command)
         {
             if (command.Command == EventCommand.Enable)
             {
-                _cacheItemCounter ??= new PollingCounter(CacheItemCounterName, this, () => _totalItemCount)
+                _cacheItemCounter ??= new PollingCounter(CacheItemCounterName, this, () => Volatile.Read(ref _totalItemCount))
```
gfoidl, line 105: FYI there was a discussion whether `Volatile.Read` or `Interlocked.Read` should be used.
The [consensus](https://github.com/dotnet/docs/pull/20984#discussion_r502028017) was to use `Interlocked.Read` in this case.

pmaytak: Awesome, thanks @gfoidl. This is very helpful.

tests/PerformanceTests/PerformanceTestService/EventSource/BenchmarkMsalDistributedTokenCacheAdapter.cs
```diff
@@ -40,8 +72,18 @@ protected override async Task RemoveKeyAsync(string cacheKey)
         /// (account or app).</returns>
         protected override async Task<byte[]> ReadCacheBytesAsync(string cacheKey)
         {
+            var stopwatch = Stopwatch.StartNew();
+            var bytes = await _distributedCache.GetAsync(cacheKey).ConfigureAwait(false);
+            stopwatch.Stop();
+
             MemoryCacheEventSource.Log.IncrementReadCount();
-            return await base.ReadCacheBytesAsync(cacheKey);
+            MemoryCacheEventSource.Log.AddReadDuration(stopwatch.Elapsed.TotalMilliseconds);
```
jmprieur, line 80: Is it to measure the time of the the distributed store read?

jmprieur (approved):
LGTM.
Thanks @pmaytak

tests/PerformanceTests/PerformanceTestService/EventSource/BenchmarkMsalDistributedTokenCacheAdapter.cs
pmaytak: Yes.

Issue #673, opened 2020-10-09
[Bug] FormatException using the TokenAcquisitionCredentialProvider


davesmits:
**Which version of Microsoft Identity Web are you using?**
Note that to get help, you need to run the latest version. 

**Where is the issue?**

* Web API
    * [x] Protected web APIs call downstream web APIs

**Is this a new or an existing app?**

This is a new app or an experiment.

**Repro**

```csharp

[HttpGet]
        public async Task<ActionResult> Register()
        {
            var result = await _graphServiceClient.Subscriptions.Request()
                .GetAsync();

            var notificationChannel = result.FirstOrDefault();

            if (notificationChannel is null)
            {
                var r = await _graphServiceClient.Subscriptions.Request().AddAsync(new Microsoft.Graph.Subscription
                {
                    ChangeType = "created",
                    NotificationUrl = "https://teamer.servicebus.windows.net/dave-apidesktop-ojevm42/api/webhook",
                    Resource = "/chats/getAllMessages",
                    ExpirationDateTime = new DateTimeOffset(DateTime.UtcNow.AddMinutes(30)),
                    ClientState = "TeamerNotifications",
                });
            }
            else
            {

            }

            return Ok();

```

**Expected behavior**
Subscription in the graph is created or an exception from the Graph API it self

**Actual behavior**
Exception in .net code

```
fail: Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddleware[1]
      An unhandled exception has occurred while executing the request.
Status Code: 0
Microsoft.Graph.ServiceException: Code: generalException
Message: An error occurred sending the request.

 ---> System.FormatException: Cannot add value because header 'Authorization' does not support multiple values.
   at System.Net.Http.Headers.HttpHeaders.ParseAndAddValue(HeaderDescriptor descriptor, HeaderStoreItemInfo info, String value)
   at System.Net.Http.Headers.HttpHeaders.Add(String name, String value)
   at Microsoft.Identity.Web.TokenAcquisitionCredentialProvider.AuthenticateRequestAsync(HttpRequestMessage request)
   at Microsoft.Graph.AuthenticationHandler.SendRetryAsync(HttpResponseMessage httpResponseMessage, IAuthenticationProvider authProvider, CancellationToken cancellationToken)
   at Microsoft.Graph.AuthenticationHandler.SendAsync(HttpRequestMessage httpRequestMessage, CancellationToken cancellationToken)
[...]

PR #674, opened 2020-10-10
Removing existing authorization token 

davesmits:
fixing bug #673 

I tried to leave it as much intact as is with the string format but personally i like this syntax more but not sure if I am missing something with the string format

```
request.Headers.Authorization = new AuthenticationHeaderValue("Bearer", "token");
```

### DIFF
--- a/src/Microsoft.Identity.Web.MicrosoftGraph/TokenAcquisitionCredentialProvider.cs
+++ b/src/Microsoft.Identity.Web.MicrosoftGraph/TokenAcquisitionCredentialProvider.cs
@@ -30,7 +30,7 @@ namespace Microsoft.Identity.Web
         /// <returns>A Task (as this is an async method).</returns>
         public async Task AuthenticateRequestAsync(HttpRequestMessage request)
         {
-            // In case of a retry there already can be a authorization header; replace it with new one to avoid using expired token
+            // In case of a retry there is already an authorization header; replace it with new one to avoid using expired token
             if (request.Headers.Contains(Constants.Authorization))
             {
                 request.Headers.Remove(Constants.Authorization);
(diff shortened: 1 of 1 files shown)


src/Microsoft.Identity.Web.MicrosoftGraph/TokenAcquisitionCredentialProvider.cs
```diff
@@ -30,6 +30,12 @@ public TokenAcquisitionCredentialProvider(ITokenAcquisition tokenAcquisition, IE
         /// <returns>A Task (as this is an async method).</returns>
         public async Task AuthenticateRequestAsync(HttpRequestMessage request)
         {
+            // In case of a retry there is already an authorization header; replace it with new one to avoid using expired token
+            if (request.Headers.Contains(Constants.Authorization))
```
jmprieur, line 34: What about the following?

```CSharp
request.Headers[Constants.Authorization] = string.Format(
                    CultureInfo.InvariantCulture,
                    "{0} {1}",
                    Constants.Bearer,
                    await _tokenAcquisition.GetAccessTokenForUserAsync(_initialScopes).ConfigureAwait(false)));
```

jmprieur (approved):
Thanks for this PR, @davesmits 
This looks good.
I've provided a suggestion for improvement.

src/Microsoft.Identity.Web.MicrosoftGraph/TokenAcquisitionCredentialProvider.cs
davesmits: the HttpHeaders doesn't have a indexer 

image

jmprieur: Ok. Thanks @davesmits

Issue #648, opened 2020-10-03
[Fundamentals] Enable perf test runner to send requests in parallel


pmaytak:
Enable parallelization in test runner when it acquires tokens using MSAL and when it makes calls to the service.

PR #679, opened 2020-10-14
Add options class for config.

pmaytak:
Smaller incremental PR related to adding parallelization (#648)
- Adds `TestRunnerOptions` class for config


jmprieur (approved):
LGTM. Thanks @pmaytak 
the code is much nicer :)

jennyf19 (approved):
Will it be possible to pull the secret from key vault?  Then we can have the tests running in vsts.

pmaytak:
> Will it be possible to pull the secret from key vault? Then we can have the tests running in vsts.

Yes, we should.

PR #684, opened 2020-10-15
add claims to token acquisition options

jennyf19:


jennyf19:
closing, might do something different.

PR #690, opened 2020-10-16
add challengeUser for conditional access processing

jennyf19:
also update graph sdk. 

@jmprieur also a public api addition, so you might want to try it.


jmprieur (approved):
LGTM. Thanks @jennyf19 
I tested all the test apps.

Issue #537, opened 2020-09-02
[Enhancement] Improved public API for IDownstreamWebApi 


jmprieur:
> Thanks for your suggestion, @NikolaosWakem, and for being an early adopter.
> Would you see value in having other generic methods (for instance CallWebApiForUserAsync?)

Thanks for getting back to me... I just use the following... 

```
public static class MicrosoftIdentityWebExtensions
    {
        public static async Task<TOutput> GetWebApiForUserAsync<TOutput>(this IDownstreamWebApi downstreamWebApi, string serviceName, string relativePath)
        {
            return await downstreamWebApi.CallWebApiForUserAsync<TOutput>(serviceName, options =>
            {
                options.HttpMethod = HttpMethod.Get;
                options.RelativePath = relativePath;
            });
        }

        public static async Task PostWebApiForUserAsync(this IDownstreamWebApi downstreamWebApi, string serviceName, string relativePath)
        {
            await downstreamWebApi.CallWebApiForUserAsync(serviceName, options =>
            {
                options.HttpMethod = HttpMethod.Post;
                options.RelativePath = relativePath;
            });
        }

        public static async Task<TOutput> PostWebApiForUserAsync<TOutput, TInput>(this IDownstreamWebApi downstreamWebApi, string serviceName, string relativePath, TInput data)
        {
            return await downstreamWebApi.CallWebApiForUserAsync<TInput, TOutput>(serviceName, data, options =>
            {
                options.HttpMethod = HttpMethod.Post;
                options.RelativePath = relativePath;
            });
        }

        public static async Task PutWebApiForUserAsync<TInput>(this IDownstreamWebApi downstreamWebApi, string serviceName, string relativePath, TInput data)
        {
            await downstreamWebApi.CallWebApiForUserAsync(serviceName, data, options =>
            {
                options.HttpMethod = HttpMethod.Put;
                options.RelativePath = relativePath;
            });
        }

        public static async Task<TOutput> PutWebApiForUserAsync<TOutput, TInput>(this IDownstreamWebApi downstreamWebApi, string serviceName, string relativePath, TInput data)
        {
            return await downstreamWebApi.CallWebApiForUserAsync<TInput, TOutput>(serviceName, data, options =>
            {
                options.HttpMethod = HttpMethod.Put;
                options.RelativePath = relativePath;
            });
        }

[...]

PR #695, opened 2020-10-19
Improve public api w/extensions for downstream web api

jennyf19:
#537


jmprieur (approved):
LGTM.
Thanks @jennyf19 

I love how the TodoListController got simplified

When we have time we should apply the same to the B2C one :)

Issue #551, opened 2020-09-04
[Bug] Signing in users breaks when using backchannel proxy


mjnorman:
**Which version of Microsoft Identity Web are you using?**
Note that to get help, you need to run the latest version. 
0.3.1-preview

**Where is the issue?**
* Web app
    * [x] Sign-in users
    * [ ] Sign-in users and call web APIs
* Web API
    * [ ] Protected web APIs (validating tokens)
    * [ ] Protected web APIs (validating scopes)
    * [ ] Protected web APIs call downstream web APIs
* Token cache serialization
     * [ ] In-memory caches
     * [ ] Session caches
     * [ ] Distributed caches
* Other (please describe)

**Is this a new or an existing app?**

a. The app is in production and I have upgraded to a new version of Microsoft Identity Web.

**Repro**
I have a public repro which is mainly a new project created with `dotnet new mvc2 -singleauth` template, then modified to handle Cloud Foundry variable parsing to obtain proxy information, and create the back channel proxy.  Relevant code for Proxy is below. --  [repro here](https://github.com/mjnorman/identity-web-repro)

I am not sure why, but simply changing back to `.AddAzureAD`, changing options to `OpenIdConnectOptions`, and no other changes, the back channel communication works properly.

```csharp
        public void PostConfigure(string name, MicrosoftIdentityOptions options)
        {
            Log.Debug("Starting PostConfigure");
            var proxyURI = VCAPHelper.GetSquidUri();
            Log.Debug("Setting Proxy URI to {Proxy}", proxyURI);

            if (!string.IsNullOrEmpty(proxyURI))
            {
                var webProxy = new WebProxy()
                {
                    Address = new Uri($"http://{VCAPHelper.GetSquidHost()}:{VCAPHelper.GetSquidPort()}"),
                    BypassProxyOnLocal = true,
                    UseDefaultCredentials = false,
                    Credentials = new NetworkCredential(VCAPHelper.GetSquidUsername(), VCAPHelper.GetSquidPassword())
                };
                var httpClientHandler = new HttpClientHandler()
                {
                    Proxy = webProxy,
                    UseDefaultCredentials = true,
                    UseProxy = true
                };
                options.Backchannel = new HttpClient(httpClientHandler, true);
            }
```

**Expected behavior**
I would expect back channel behavior to work normally when access through a proxy is required to access metadata at login.microsoft.com.

**Actual behavior**
Receive an error below presumably due to some communication on the back channel not working through the proxy. 
[...]

PR #696, opened 2020-10-19
initial commit for adding aadIssuerValidatorOptions

jennyf19:
#551

### DIFF
--- a/src/Microsoft.Identity.Web/Resource/AadIssuerValidatorOptions.cs
+++ b/src/Microsoft.Identity.Web/Resource/AadIssuerValidatorOptions.cs
@@ -1,0 +1,17 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Identity.Web
+{
+    /// <summary>
+    /// Options passed-in to create the AadIssuerValidator object.
+    /// </summary>
+    public class AadIssuerValidatorOptions
+    {
+        /// <summary>
+        /// Sets the name of the HttpClientFactory to use with the configuration manager.
+        /// Needed when setting up a proxy.
+        /// </summary>
+        public string? HttpClientFactoryName { get; set; } = null!;
+    }
+}
--- a/src/Microsoft.Identity.Web/Resource/AadIssuerValidatorOptions.cs
+++ b/src/Microsoft.Identity.Web/Resource/AadIssuerValidatorOptions.cs
@@ -9,9 +9,9 @@ namespace Microsoft.Identity.Web
     public class AadIssuerValidatorOptions
     {
         /// <summary>
-        /// Sets the name of the HttpClientFactory to use with the configuration manager.
-        /// Needed when setting up a proxy.
+        /// Sets the name of the HttpClient to get from the IHttpClientFactory for use with the configuration manager.
+        /// Needed when customizing the client such as configuring a proxy.
         /// </summary>
-        public string? HttpClientFactoryName { get; set; } = null!;
+        public string? HttpClientName { get; set; } = null!;
     }
 }
--- a/src/Microsoft.Identity.Web/Resource/AadIssuerValidatorOptions.cs
+++ b/src/Microsoft.Identity.Web/Resource/AadIssuerValidatorOptions.cs
@@ -12,6 +12,6 @@ namespace Microsoft.Identity.Web
         /// Sets the name of the HttpClient to get from the IHttpClientFactory for use with the configuration manager.
         /// Needed when customizing the client such as configuring a proxy.
         /// </summary>
-        public string? HttpClientName { get; set; } = null!;
+        public string? HttpClientName { get; set; }
     }
 }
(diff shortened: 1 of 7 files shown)


src/Microsoft.Identity.Web/Resource/AadIssuerValidatorOptions.cs
```diff
@@ -0,0 +1,17 @@
...
+    /// <summary>
+    /// Options passed-in to create the AadIssuerValidator object.
+    /// </summary>
+    public class AadIssuerValidatorOptions
+    {
+        /// <summary>
+        /// Sets the name of the HttpClientFactory to use with the configuration manager.
+        /// Needed when setting up a proxy.
+        /// </summary>
+        public string? HttpClientFactoryName { get; set; } = null!;
```
Tratcher, line 15:
```suggestion
        /// Sets the name of the HttpClient to get from the IHttpClientFactory for use with the configuration manager.
        /// Needed when customizing client such as configuring a proxy.
        /// </summary>
        public string? HttpClientName { get; set; } = null!;
```

src/Microsoft.Identity.Web/Resource/MicrosoftIdentityIssuerValidatorFactory.cs
```diff
@@ -15,9 +17,20 @@ namespace Microsoft.Identity.Web.Resource
...
     {
+        public MicrosoftIdentityIssuerValidatorFactory(
+            IOptions<AadIssuerValidatorOptions> aadIssuerValidatorOptions,
+            IHttpClientFactory httpClientFactory)
+        {
+            _configManager =
+            new ConfigurationManager<IssuerMetadata>(
+                Constants.AzureADIssuerMetadataUrl,
+                new IssuerConfigurationRetriever(),
+                httpClientFactory?.CreateClient(aadIssuerValidatorOptions?.Value?.HttpClientFactoryName));
```
Tratcher, line 28: What's with all the `?`? None of these things should be null when activated from DI.

src/Microsoft.Identity.Web/Resource/MicrosoftIdentityIssuerValidatorFactory.cs
```diff
@@ -15,9 +17,20 @@ namespace Microsoft.Identity.Web.Resource
     /// </summary>
     internal class MicrosoftIdentityIssuerValidatorFactory
     {
+        public MicrosoftIdentityIssuerValidatorFactory(
+            IOptions<AadIssuerValidatorOptions> aadIssuerValidatorOptions,
+            IHttpClientFactory httpClientFactory)
```
Tratcher, line 22: Did you need to add a package dependency for this? You at least need to register it in DI.

src/Microsoft.Identity.Web/Resource/AadIssuerValidatorOptions.cs
jmprieur: @Tratcher: the thing I don't fully get is the proxy configuration is a static property of HttpClient? 
But I understand that for the default headers etc ... this is important.

src/Microsoft.Identity.Web/Resource/MicrosoftIdentityIssuerValidatorFactory.cs
jmprieur: no need to add a package dependency. IHttpClientFactory is already used by Microsoft.Identity.Web

src/Microsoft.Identity.Web/Resource/AadIssuerValidatorOptions.cs
Tratcher: Yeah, you don't want the static, you want to set the instance property on SocketsHttpHandler when creating the HttpClient.
https://docs.microsoft.com/en-us/dotnet/api/system.net.http.socketshttphandler.proxy?view=netcore-3.1#System_Net_Http_SocketsHttpHandler_Proxy

src/Microsoft.Identity.Web/Resource/MicrosoftIdentityIssuerValidatorFactory.cs
jennyf19: that's for the unit test to pass, otherwise i have to configure all the service, etc..., which i might do or leave as an integration test instead.

src/Microsoft.Identity.Web/Resource/MicrosoftIdentityIssuerValidatorFactory.cs
jennyf19: right, what @jmprieur said.

src/Microsoft.Identity.Web/Resource/AadIssuerValidatorOptions.cs
jennyf19: @Tratcher i see. makes sense. will add to our documentation when we release.

jmprieur (approved):
LGTM
Thanks @jennyf19 !

jennyf19:
@jmprieur i need to do some more testing, so will update accordingly.

tests/Microsoft.Identity.Web.Test/Resource/HttpClientFactoryTest.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Microsoft.Identity.Web.Test.Resource
+{
+    public class HttpClientFactoryTest : IHttpClientFactory
+    {
+        public Dictionary<string, HttpClient> dictionary = new Dictionary<string, HttpClient>();
+
+        public HttpClient CreateClient(string name)
+        {
+            SocketsHttpHandler socketsHttpHandler = new SocketsHttpHandler();
```
jmprieur, line 15: @jennyf19 you might want to suppress this warning,

jmprieur (approved):
Thanks @jennyf19 
More happy customers :)

🚢

PR #697, opened 2020-10-19
update to 4.21.1 & xml updates

jennyf19:


jmprieur (approved):
LGTM. Thanks @jennyf19 
I'm not sure why the XML file changed, but ... it will be regenerated anyway ...

PR #699, opened 2020-10-20
updates to downstream api

jennyf19:


jmprieur (approved):
LGTM.

I used this [comparison](https://github.com/AzureAD/microsoft-identity-web/compare/978232b410077765f8169ef631dd605ba6fe658e...jennyf/downstream) to understand what the change was with before the previous PR.

The interface is unchanged and we have new additional extension methods.

PR #701, opened 2020-10-20
renaming

jennyf19:


src/Microsoft.Identity.Web/AppServicesAuth/AppServicesAuthenticationTokenAcquisition.cs
```diff
@@ -49,16 +52,17 @@ private async Task<IConfidentialClientApplication> GetOrCreateApplication()
             {
                 ConfidentialClientApplicationOptions options = new ConfidentialClientApplicationOptions()
                 {
-                    ClientId = AppServiceAuthenticationInformation.ClientId,
-                    ClientSecret = AppServiceAuthenticationInformation.ClientSecret,
-                    Instance = AppServiceAuthenticationInformation.Issuer,
+                    ClientId = AppServicesAuthenticationInformation.ClientId,
```
jennyf19, line 55: we should consider some refactoring in tokenAcqusition so we don't have to repeat this code. not something for this PR, but something for the future we shoudl consider.

src/Microsoft.Identity.Web/AppServicesAuth/AppServicesAuthenticationHandler.cs
```diff
@@ -53,10 +53,10 @@ protected override Task<AuthenticateResult> HandleAuthenticateAsync()
                     ClaimsPrincipal claimsPrincipal = new ClaimsPrincipal(new ClaimsIdentity(
                         jsonWebToken.Claims,
                         idp,
-                        "name", // v1.0
+                        Constants.NameClaim, // v1.0
```
jmprieur, line 56: That makes me think that we need to check the "ver" claim, and use the right constant (NameClaim or prefered_username, depending on the version.

jmprieur (approved):
Thanks so much @jennyf19

Issue #654, opened 2020-10-05
[Feature Request] It should be possible to use GraphServiceClient to call graph APIs requiring app only permissions 


jmprieur:
**Is your feature request related to a problem? Please describe.**
When developers want to call graph to call an app only method, the current GraphServiceClient cannot be used as it uses GetTokenForUserAsync. See https://github.com/AzureAD/microsoft-identity-web/blob/f609eeef9e69cee9c997bb73fb98d7c180d845af/src/Microsoft.Identity.Web.MicrosoftGraph/TokenAcquisitionCredentialProvider.cs#L39

number of customers hitting this issue so far: 2

**Describe the solution you'd like**
Have a way to call app only methods (corresponding to App-only scopes)

**Describe alternatives you've considered**
- enable a second GraphServiceClient to be injected (with a different type ?). Maybe a class named `AppOnlyGraphServiceClient` which would have only one property which would be the GraphServiceClient, or which would inherit from GraphServiceClient.
- have an extension method on GraphServiceClient which would specify that the developer wants to have an app-only and then use it GetTokenForUserAsync. See https://github.com/AzureAD/microsoft-identity-web/blob/f609eeef9e69cee9c997bb73fb98d7c180d845af/src/Microsoft.Identity.Web.MicrosoftGraph/TokenAcquisitionCredentialProvider.cs#L39

**Additional context**
See the [ASP.NET Core Graph web hooks](https://github.com/microsoftgraph/aspnetcore-webhooks-sample/blob/52cb9fcdb5c962928538766e792916745fc3b999/src/GraphWebhooks-Core/Controllers/NotificationController.cs#L172-L176) sample where the GraphServiceClient can only be used for delegated scopes:

davesmits:
Maybe little bit of topic; but I want to avoid using an app account as they very often (if not always) need Admin consent from the tenant. For client applications it can be helpful is the gateway way api to the graph will do some background work

So imagine:
native client app -> user signs in and goes with token to gateway api

gateway api -> with the token from the native client is able to exchange it to a token for the graph api. If the offline scope is present this new access token will have a refrsh token too. Now want to 'store' it somewhere so a background process is  able to get new access tokens to do work in the background for this specific user (in my case it was updating webhook subscription)

Currently the api takes so many concerns out of your hands that you don't see/ have access to this refresh token at all. Making it hard to get a background process alive for a user

jmprieur:
[...]

PR #705, opened 2020-10-21
Proposal for enabling AppOnly access to Microsoft Graph

darrelmiller:
This PR proposes a number of updates to the Graph integration with Identity.Web.
It addresses https://github.com/AzureAD/microsoft-identity-web/issues/654, plus more.

It creates two new overloads of AddMicrosoftGraph that provide full control to the user for configuring the GraphServiceClient while still abstracting away the auth configuration.

This configures a GraphServiceClient to use delegated by default
```
.AddMicrosoftGraph(authProvider => new GraphServiceClient(authProvider), initialScopes);
```
In order to make the occasional appOnly request the `WithAppOnly()` method can be used,

```
graphServiceClient.Users[userId].Request().WithAppOnly().GetAsync();
```

This configures a HTTP client to use AppOnly by default,
```
,AddMicrosoftGraphAppOnly(authProvider => new GraphServiceClient(authProvider));
```

I'm still not completely satisfied with this approach because you cannot call both of these and inject them both into a controller.  We would need some way of registering the two instances separately.  But this at least gives us a place to start.

This PR also introduced the use of TokenAcquisitionProviderOption which is the class that can contain the configuration options passed on a per-request basis.,

I also renamed TokenAcquistionCredentialProvider because that mixes the names of AuthenticationProviders and TokenCredentials. As we will be moving to TokenCredentials in the future, it gets very confusing.

#654


baywet (approved):
great work! so if I understood things well, in this design people would only have one service (either app only or delegated) and locally override the context as needed. Not two services?

src/Microsoft.Identity.Web.MicrosoftGraph/BaseRequestExtensions.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// and can be used to perform incremental scope consent.
+        /// This only works with the default authentication handler and default set of Microsoft graph authentication providers.
+        /// If you use a custom authentication handler or authentication provider, you have to handle it's retrieval in your implementation.
+        /// </summary>
+        /// <param name="baseRequest">The <see cref="IBaseRequest"/>.</param>
+        /// <param name="scopes">Microsoft graph scopes used to authenticate this request.</param>
+        public static T WithScopes<T>(this T baseRequest, string[] scopes) where T : IBaseRequest
+        {
+            string authHandlerOptionKey = typeof(TokenAcquisitionAuthenticationProviderOption).ToString();
+            AuthenticationHandlerOption authHandlerOptions = baseRequest.MiddlewareOptions[authHandlerOptionKey] as AuthenticationHandlerOption ?? new AuthenticationHandlerOption();
```
jmprieur, line 21: Using `[ ]` when there is nothing in the dictionary will lead to an exception: `KeyNotFoundException: The given key 'Microsoft.Identity.Web.TokenAcquisitionAuthenticationProviderOption' was not present in the dictionary`.

We'd need to test baseRequest.MiddlewareOptions.ContainsKey...

src/Microsoft.Identity.Web.MicrosoftGraph/BaseRequestExtensions.cs
```diff
@@ -0,0 +1,46 @@
...
+            authHandlerOptions.AuthenticationProviderOption = msalAuthProviderOption;
+            baseRequest.MiddlewareOptions[authHandlerOptionKey] = authHandlerOptions;
+
+            return baseRequest;
+        }
+
+        public static T WithAppOnly<T>(this T baseRequest, bool appOnly = true) where T : IBaseRequest
+        {
+            string authHandlerOptionKey = typeof(TokenAcquisitionAuthenticationProviderOption).ToString();
+            AuthenticationHandlerOption authHandlerOptions = baseRequest.MiddlewareOptions[authHandlerOptionKey] as AuthenticationHandlerOption ?? new AuthenticationHandlerOption();
```
jmprieur, line 35: Same comment as above

jmprieur (commented):
Thanks so much, @darrelmiller 
here are some initial comments.

jmprieur:
@jennyf19 
I propose that we merge this PR in a branch (different from master), and fix there the couple of comments I had.

I tested the WebAppCallsGraph and added the possibility of getting the mails (`Me.Messages`, with the "Mail.Read" scope

```CSharp
try
{
 var messages = await _graphServiceClient.Me.Messages.Request().WithScopes(new[] { "Mail.Read" }).GetAsync();
}
catch(Exception ex)
{
 // No access to mailbox
}
```

src/Microsoft.Identity.Web.MicrosoftGraph/BaseRequestExtensions.cs
```diff
@@ -0,0 +1,46 @@
...
+        /// <param name="baseRequest">The <see cref="IBaseRequest"/>.</param>
+        /// <param name="scopes">Microsoft graph scopes used to authenticate this request.</param>
+        public static T WithScopes<T>(this T baseRequest, string[] scopes) where T : IBaseRequest
+        {
+            string authHandlerOptionKey = typeof(TokenAcquisitionAuthenticationProviderOption).ToString();
+            AuthenticationHandlerOption authHandlerOptions = baseRequest.MiddlewareOptions[authHandlerOptionKey] as AuthenticationHandlerOption ?? new AuthenticationHandlerOption();
+            TokenAcquisitionAuthenticationProviderOption msalAuthProviderOption = authHandlerOptions?.AuthenticationProviderOption as TokenAcquisitionAuthenticationProviderOption ?? new TokenAcquisitionAuthenticationProviderOption();
+
+            msalAuthProviderOption.Scopes = scopes;
+
```
jennyf19, line 25: null check here on msalAuthProviderOptions and on line 26? (authHandlerOptoins)

jennyf19 (changes requested):
:clock1:

jennyf19 (approved):
:shipit:

jennyf19:
this is cool. i think if you know the changes to make, let's just merge to master and then make a new PR w/those changes.

---
In reply to: [713628028](https://github.com/AzureAD/microsoft-identity-web/pull/705#issuecomment-713628028) [](ancestors = 713628028)

PR #716, opened 2020-10-22
make issuer validator factory public and add obsolete messaging

jennyf19:


jmprieur (approved):
LGTM.
Thanks @jennyf19

PR #718, opened 2020-10-23
fix link

jennyf19:
@jmprieur needs to be updated in aka.ms, i don't have access to update this particular link. thx.


jmprieur (approved):
LGTM.
Thanks @jennyf19

PR #734, opened 2020-10-30
Do not require scopes for app only tokens

radistmorse:
scopes are ignored anyway for an app-only requests, there is no reason to require them.


jmprieur (approved):
LGTM. Thanks @radistmorse 
cc: @jennyf19

jennyf19 (approved):
thank you @radistmorse

PR #737, opened 2020-11-02
add net461 target for cache extensions

jennyf19:
@jmprieur as discussed

### DIFF
--- a/src/Microsoft.Identity.Web/ClaimsPrincipalExtensions.cs
+++ b/src/Microsoft.Identity.Web/ClaimsPrincipalExtensions.cs
@@ -209,5 +209,12 @@ namespace Microsoft.Identity.Web
 
             return claimsPrincipal.FindFirstValue(ClaimConstants.NameIdentifierId);
         }
+
+#if NET472
+        private static string? FindFirstValue(this ClaimsPrincipal claimsPrincipal, string type)
+        {
+             return claimsPrincipal.FindFirst(type)?.Value;
+        }
+#endif
     }
 }
--- a/src/Microsoft.Identity.Web/Microsoft.Identity.Web.csproj
+++ b/src/Microsoft.Identity.Web/Microsoft.Identity.Web.csproj
@@ -47,9 +47,8 @@
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
   </ItemGroup>
 
-
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp3.1; net5.0</TargetFrameworks>
+    <TargetFrameworks>netcoreapp3.1; net461; net5.0</TargetFrameworks>
     <SignAssembly>true</SignAssembly>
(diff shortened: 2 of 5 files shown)


jmprieur (approved):
This is awesome @jennyf19, and this opens the way to delivering even more value for the ASP.NET customers.

@henrik-me : is it an option to support .NET 4.7.2 instead of .NET 4.6.1 ? this would also give us the certificate management with only one line of conditional compilation (with 4.6.1 the crypto is a bit different)

henrik-me:
> This is awesome @jennyf19, and this opens the way to delivering even more value for the ASP.NET customers.
> 
> @henrik-me : is it an option to support .NET 4.7.2 instead of .NET 4.6.1 ? this would also give us the certificate management with only one line of conditional compilation (with 4.6.1 the crypto is a bit different)

4.7.2 is recommended in many ways. Not sure every one is there yet though and what the timeline will be (probably most by end of CY). Will require a bit of investigation on current usages.

Perhaps we can start with 4.7.2?

jennyf19:
@henrik-me @jmprieur ok. thanks.

henrik-me (approved):
:shipit:

henrik-me:
perhaps a utest validating that this the expected is available for the 4.7.2 target.

jmprieur:
We could even expose the certificate management (including getting them from keyVault), by changing one line in DefaultCertificateLoader.cs:

line 97 
from `string[] segments = certificate.SecretId.AbsolutePath.Split('/', StringSplitOptions.RemoveEmptyEntries);``
to `string[] segments = certificate.SecretId.AbsolutePath.Split(new[] { '/' }, StringSplitOptions.RemoveEmptyEntries);`

jmprieur:
@jennyf19 : tested with https://github.com/Azure-Samples/ms-identity-aspnet-webapp-openidconnect/pull/43
:shipit:

jennyf19:
i'm going to add a test app instead. so we can verify all the classes taht are added e2e.

---
In reply to: [720773691](https://github.com/AzureAD/microsoft-identity-web/pull/737#issuecomment-720773691) [](ancestors = 720773691)

jennyf19:
maybe in a next PR?

---
In reply to: [721111217](https://github.com/AzureAD/microsoft-identity-web/pull/737#issuecomment-721111217) [](ancestors = 721111217)

henrik-me:
I don't really like the test apps for something like this, but can be clarified as required validation step when checking the API. 
I would prefer a utest project that validates this. We should not require any manual validation of this. 
Also the sample should demonstrate the usage of this (I believe JM already made some of those changes to the sample)

---
In reply to: [721338397](https://github.com/AzureAD/microsoft-identity-web/pull/737#issuecomment-721338397) [](ancestors = 721338397,720773691)

jmprieur:
yes, this is demonstrated in the sample: https://github.com/Azure-Samples/ms-identity-aspnet-webapp-openidconnect/pull/43
I've also updated the release spreadsheet

jennyf19:
@henrik-me https://github.com/AzureAD/microsoft-identity-web/pull/754 for adding test.

PR #740, opened 2020-11-03
fix some sonar cloud warnings. more to do.

jennyf19:


jmprieur (approved):
LGTM
Thanks @jennyf19 
Some good catches.

PR #745, opened 2020-11-04
possible improvements to ms.id.web for net472

jmprieur:
This enables the certificates to be added to the net472 target if we want to later
(to avoid losing this change)

Also committing the DGML graph showing which classes could be exposed to net472 (with time, not a priority)

### DIFF
--- a/Classes depending on ASP.NET Core.dgml
+++ b/Classes depending on ASP.NET Core.dgml
@@ -1,0 +1,2099 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<DirectedGraph DataVirtualized="True" Layout="Sugiyama" ZoomLevel="-1" xmlns="http://schemas.microsoft.com/vs/2009/dgml">
+  <Nodes>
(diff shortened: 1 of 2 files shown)


Classes depending on ASP.NET Core.dgml
```diff
@@ -0,0 +1,2099 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<DirectedGraph DataVirtualized="True" Layout="Sugiyama" ZoomLevel="-1" xmlns="http://schemas.microsoft.com/vs/2009/dgml">
```
bgavrilMS, line 2: I recommend against using this deprecated tool, which doesn't even work on recent VS. Consider using https://app.diagrams.net/ for diagrams - it generates a model file and pictures, and you can include both in your source.

src/Microsoft.Identity.Web/CertificateManagement/DefaultCertificateLoader.cs
```diff
@@ -94,7 +94,7 @@ private static X509Certificate2 LoadFromKeyVault(string keyVaultUrl, string cert
             }
 
             // Parse the secret ID and version to retrieve the private key.
-            string[] segments = certificate.SecretId.AbsolutePath.Split('/', StringSplitOptions.RemoveEmptyEntries);
+            string[] segments = certificate.SecretId.AbsolutePath.Split(new[] { '/' }, StringSplitOptions.RemoveEmptyEntries);
```
bgavrilMS, line 97: Why don't you leave the string.Split as is ?

bgavrilMS (approved):

src/Microsoft.Identity.Web/CertificateManagement/DefaultCertificateLoader.cs
jmprieur: Because the particular override that was used is available in netcoreapp3.1 and net5.0, but not in net461/net472

jennyf19 (approved):
:shipit:

Classes depending on ASP.NET Core.dgml
jmprieur: @bgavrilMS : this deprecated tool generates a graph from code. For a graph if this size this would have taken me 1 week to get this precision (think of listing all the classes and their dependencies, all that without making mistakes).

PR #757, opened 2020-11-10
Update runner to persist exceptions into a file and update metrics across processes.

pmaytak:
- Added functionality to keep a record of global metrics across threads.
- Added `PersistExceptions` method to save exceptions to a file thread-safely.
- Passing cancellation tokens into `CreateStopProcessingByTimeoutTask` and `CreateStopProcessingByUserRequestTask` so they can be cancelled immediately.
- Renamed `ScalableTokenCacheHelper` to `TokenCacheHelper`.
- Moved `Log` method into `Logger` class.

[Board card](https://github.com/AzureAD/microsoft-identity-web/projects/14#card-48719139)

### DIFF
--- a/tests/PerformanceTests/Microsoft.Identity.Web.Perf.Client/Logger.cs
+++ b/tests/PerformanceTests/Microsoft.Identity.Web.Perf.Client/Logger.cs
@@ -1,0 +1,75 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.IO;
+using System.Text;
+using Microsoft.Identity.Client;
+
+namespace Microsoft.Identity.Web.Perf.Client
+{
+    public class Logger
+    {
+        private static readonly string s_logsFolder = Path.Combine(Path.GetDirectoryName(System.Reflection.Assembly.GetExecutingAssembly().Location), "Logs");
+        private static string s_exceptionsFile = s_logsFolder + $"\\{DateTime.Now:yyyyMMdd}.exceptions.log";
+        private static string s_msalLogsFile = System.Reflection.Assembly.GetExecutingAssembly().Location + ".msalLogs.txt";
+
+        private static StringBuilder s_msalLog = new StringBuilder();
+        private static volatile bool s_isMsalLogging = false;
+
+        private static object s_exceptionsLock = new object();
+        private static object s_msalLogLock = new object();
+
+        public static void PersistExceptions(StringBuilder stringBuilderContent)
+        {
+            if (stringBuilderContent.Length == 0)
+            {
+                return;
+            }
+
+            lock (s_exceptionsLock)
+            {
+                Directory.CreateDirectory(s_logsFolder);
+                File.AppendAllText(s_exceptionsFile, stringBuilderContent.ToString());
+            }
+        }
+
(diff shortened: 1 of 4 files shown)


tests/PerformanceTests/Microsoft.Identity.Web.Perf.Client/TokenCacheHelper.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.Identity.Web.Perf.Client
     /// Token cache writing on disk one cache per account
     /// WARNING: this version is not encrypted
     /// </summary>
-    static class ScalableTokenCacheHelper
+    static class TokenCacheHelper
```
jmprieur, line 17: why change its name?

jmprieur (approved):
LGTM.

tests/PerformanceTests/Microsoft.Identity.Web.Perf.Client/Logger.cs
```diff
@@ -0,0 +1,75 @@
...
+// Licensed under the MIT License.
+
+using System;
+using System.IO;
+using System.Text;
+using Microsoft.Identity.Client;
+
+namespace Microsoft.Identity.Web.Perf.Client
+{
+    public class Logger
```
jennyf19, line 11: do we need this? are we going to be analyzing the logs on failures, or do we have enough info with the pre-existing dashboards in app services? could be one more thing for someone to have to go through. just a thought.

tests/PerformanceTests/Microsoft.Identity.Web.Perf.Client/TestRunner.cs
```diff
@@ -221,38 +228,59 @@ private async Task SendRequestsAsync(int userStartIndex, int userEndIndex, Cance
...
+            }
+        }
+
+        private void DisplayProgress()
+        {
+            var stringBuilder = new StringBuilder();
+            stringBuilder.AppendLine("------------------------------------------------------------------------");
+            stringBuilder.AppendLine($"Run time: {_processingStartTime} - {DateTime.Now} = {DateTime.Now - _processingStartTime}.");
+            stringBuilder.AppendLine($"Total number of users: {_options.NumberOfUsersToTest} users from {_options.StartUserIndex} to {_options.StartUserIndex + _options.NumberOfUsersToTest - 1}.");
+
```
jennyf19, line 268: nit: remove extra line

jennyf19 (approved):

tests/PerformanceTests/Microsoft.Identity.Web.Perf.Client/TokenCacheHelper.cs
pmaytak: That was feedback from one of the previous PRs.

tests/PerformanceTests/Microsoft.Identity.Web.Perf.Client/Logger.cs
pmaytak: Persisting test runner exceptions is needed I think, although it should be updated to push that data to App Insights rather than a file. The MSAL logging is set to Info level right now, which does generate a lot of logs. I would set it to Error level and also try to update it to push to App Insights. I don't know if we need Info level logging for MSAL in a client, as we do have other related counters.

Issue #760, opened 2020-11-11
AccountController implementing MicrosoftIdentity/Account/xxx endpoints does not honour redirectUrl


rollandjb:
**Which version of Microsoft Identity Web are you using?**
Note that to get help, you need to run the latest version. 
Microsoft Identity Web 1.2.0

**Where is the issue?**
* Web app
    * [x] Sign-in users
    * [ ] Sign-in users and call web APIs
* Web API
    * [ ] Protected web APIs (validating tokens)
    * [ ] Protected web APIs (validating scopes)
    * [ ] Protected web APIs call downstream web APIs
* Token cache serialization
     * [ ] In-memory caches
     * [ ] Session caches
     * [ ] Distributed caches
* Other (please describe)

**Is this a new or an existing app?**
This is an app in development

**Repro**

```html
<a href="MicrosoftIdentity/Account/SignIn?redirectUrl=yyy">Sign In</a>
```

**Expected behavior**
Clicking on the link should take me to the page at `yyy`

**Actual behavior**
Takes me to the route of the application, `/`

**Possible solution**
Set the `redirectUrl` property of AuthenticationProperties to the incoming query value

```csharp
[HttpGet("{scheme?}/{redirectUrl?}")]
        public IActionResult SignIn([FromRoute] string scheme, [FromQuery] string redirectUrl)
        {
            scheme ??= OpenIdConnectDefaults.AuthenticationScheme;
            // var redirectUrl = Url.Content("~/");
            return Challenge(
                new AuthenticationProperties { RedirectUrl = redirectUrl ?? Url.Content("~/") },
                scheme);
        }
```

All endpoints should honour `redirectUrl`

**Additional context / logs / screenshots**

This will allow clients to use `MicrosoftIdentity/Account/xxx` endpoints and land the user on desired page.

jmprieur:
Thanks for the heads-up @rollandjb

rollandjb:
My pleasure @jmprieur

jmprieur:
@rollandjb I tried to repro this bug, but cannot.
Using https://github.com/AzureAD/microsoft-identity-web/tree/master/tests/B2CWebAppCallsWebApi, and signing out each time, I'm always redirected to the page where the sign-in was triggered.

Do you have repro steps? Do you want to force another redirect URL?

rollandjb:
@jmprieur 

I don't know ASP.NET Core web apps well enough to know how the page url is being remembered.

[...]

PR #767, opened 2020-11-17
Honor redirect a/AccountController

jennyf19:
needs a little more testing, but have some some w/the dev apps.
#760


pmaytak:
Should this change be also applied to `EditProfile` and `ResetPassword`, since we're redirecting to root there also?

jmprieur:
what does this mean @pmaytak: what do you suggest.? (how does this concretely impact this draft PR?)

pmaytak:
@jmprieur 

For example, add the `redirectUri` argument to `EditProfile` action.
```csharp
[HttpGet("{scheme?}/{redirectUri?}")]
public IActionResult EditProfile([FromRoute] string scheme, [FromQuery] string redirectUri)
```
 Because the current behavior is
- Go to a non root page, like `/TodoList`.
- Go through the `edit profile` flow.
- We redirect to root `/`, instead of where the user was before.

jmprieur:
Thanks for providing the full explanation, @pmaytak. This is useful.

This think a customer was trying to do that at some point.
@jennyf19 : would you mind updating the EditProfile as well. I'm not too sure for reset password as it's an interrupt from the sign-in flow. We'd need to test thorughly.

jennyf19:
thanks. yes was wondering about b2c when testing and was going to bring it up in stand-up today.

sven5:
@jennyf19 
When is this included in a release?

PR #789, opened 2020-11-26
Removes a double /tfp in B2C web apps calling web APIs when instance ends in /tfp/

jmprieur:
For a b2C web app **that calls a web API**, if the developer provides the tfp in the authority in the appsettings.json, the sign-in part of the Oauth code flow works, but Microsoft.Identity.Web adds a second /tfp in the authority, and MSAL fails.

```Json
{
  "AzureAdB2C": {
    "Instance": "https://login.microsoftonline.com/tfp/",
```

This PR provides a work around for this case (and only this case).

This is an issue because the ASP.NET Core templates do that.

Not sure if we want to add a unit test?


jennyf19:
@jmprieur i understand the concern here. i think we can move this logic elsewhere though. i'll propose something.

jmprieur:
Thanks @jennyf19 
then closing this PR.

PR #790, opened 2020-11-27
Add support for tenant selection when using AppOnly Microsoft Graph

hajekj:
When using AppOnly call with multi-tenant application, you won't be able to obtain the token for target tenant due to not being able to choose the tenant. This pull request adds support to explicitly choose tenant when obtaining an app-only token like this:

```csharp
var organization = await _microsoftGraph.Organization
    .Request()
    .WithAppOnly(tenant: "thenetw.org") // or tenantId
    .GetAsync();
```

### DIFF
--- a/src/Microsoft.Identity.Web.MicrosoftGraph/BaseRequestExtensions.cs
+++ b/src/Microsoft.Identity.Web.MicrosoftGraph/BaseRequestExtensions.cs
@@ -30,10 +30,15 @@ namespace Microsoft.Identity.Web
         /// <typeparam name="T">Type of the request.</typeparam>
         /// <param name="baseRequest">Request.</param>
         /// <param name="appOnly">Should the permissions be app only or not.</param>
+        /// <param name="tenant">Tenant ID or domain for which we want to make the call..</param>
         /// <returns></returns>
-        public static T WithAppOnly<T>(this T baseRequest, bool appOnly = true) where T : IBaseRequest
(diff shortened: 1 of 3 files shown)


jmprieur (approved):
LGTM. 
Thanks @hajekj. This is a good addition/scenario.

src/Microsoft.Identity.Web.MicrosoftGraph/BaseRequestExtensions.cs
```diff
@@ -30,10 +30,15 @@ public static class BaseRequestExtensions
...
+        /// <param name="tenant">Tenant ID or domain for which we want to make the call..</param>
         /// <returns></returns>
-        public static T WithAppOnly<T>(this T baseRequest, bool appOnly = true) where T : IBaseRequest
+        public static T WithAppOnly<T>(this T baseRequest, bool appOnly = true, string? tenant = null) where T : IBaseRequest
         {
-            return SetParameter(baseRequest, options => options.AppOnly = appOnly);
+            return SetParameter(baseRequest, options =>
+            {
+                options.AppOnly = appOnly;
+                options.Tenant = tenant;
```
jmprieur, line 40: We might want to throw an ArgumentException if tenant is 'common' or 'organizations' as these should not be used.
cc: @jennyf19, @bgavrilMS

jennyf19: @jmprieur i'll merge but we should implement your suggestion before a release.

jennyf19 (approved):
:shipit:

PR #795, opened 2020-12-01
Throw an ArgumentException if tenant is 'common' or 'organizations' for acquire token for app scenarios

### LINKED ISSUE #793: [Bug] throw an ArgumentException if tenant is 'common' or 'organizations' for acquire token for app scenarios
see #790 for details.

hajekj:
* Added `common` tenant to meta tenant identifiers since `client_credentials` flow cannot be used without explicit tenant.
* Check `TenantId` value against the above list as well, like with explicitly passed tenant.

Fixes #793

### DIFF
--- a/src/Microsoft.Identity.Web/TokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web/TokenAcquisition.cs
@@ -86,6 +86,7 @@ namespace Microsoft.Identity.Web
         private readonly ISet<string> _metaTenantIdentifiers = new HashSet<string>(
             new[]
             {
+                Constants.Common,
                 Constants.Organizations,
                 Constants.Consumers,
             },
@@ -274,6 +275,11 @@ namespace Microsoft.Identity.Web
             if (!string.IsNullOrEmpty(tenant) && _metaTenantIdentifiers.Contains(tenant))
             {
                 throw new ArgumentException(IDWebErrorMessage.ClientCredentialTenantShouldBeTenanted, nameof(tenant));
+            }
+
+            if (!string.IsNullOrEmpty(_microsoftIdentityOptions.TenantId) && _metaTenantIdentifiers.Contains(_microsoftIdentityOptions.TenantId))
+            {
+                throw new ArgumentException(IDWebErrorMessage.ClientCredentialTenantShouldBeTenanted, nameof(_microsoftIdentityOptions.TenantId));
             }
 
             // Use MSAL to get the right token to call the API
--- a/src/Microsoft.Identity.Web/TokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web/TokenAcquisition.cs
@@ -271,14 +271,14 @@ namespace Microsoft.Identity.Web
                 throw new ArgumentException(IDWebErrorMessage.ClientCredentialScopeParameterShouldEndInDotDefault, nameof(scope));
             }
 
+            if (string.IsNullOrEmpty(tenant))
+            {
+                tenant = _applicationOptions.TenantId ?? _microsoftIdentityOptions.TenantId;
+            }
+
             if (!string.IsNullOrEmpty(tenant) && _metaTenantIdentifiers.Contains(tenant))
             {
                 throw new ArgumentException(IDWebErrorMessage.ClientCredentialTenantShouldBeTenanted, nameof(tenant));
-            }
-
-            if (!string.IsNullOrEmpty(_microsoftIdentityOptions.TenantId) && _metaTenantIdentifiers.Contains(_microsoftIdentityOptions.TenantId))
-            {
-                throw new ArgumentException(IDWebErrorMessage.ClientCredentialTenantShouldBeTenanted, nameof(_microsoftIdentityOptions.TenantId));
             }
 
             // Use MSAL to get the right token to call the API
--- a/src/Microsoft.Identity.Web/TokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web/TokenAcquisition.cs
@@ -88,7 +88,6 @@ namespace Microsoft.Identity.Web
             {
                 Constants.Common,
                 Constants.Organizations,
-                Constants.Consumers,
             },
             StringComparer.OrdinalIgnoreCase);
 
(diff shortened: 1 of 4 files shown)


jmprieur (commented):
Thanks for your PR, @hajekj 
actually, I believe that Consumers should be allowed (for some first party scenarios)

hajekj:
So what would be the best way to approach this? It's the first time I hear, that it's possible with consumers endpoint. So you think it would be the best to skip the validation for Consumers endpoint? Because right now, the check is being done against Consumers and Organizations and Common was skipped out (https://github.com/AzureAD/microsoft-identity-web/blob/master/src/Microsoft.Identity.Web/TokenAcquisition.cs#L86).

bgavrilMS:
`consumers` is essentially a single (huge) tenant, while `organizations` and `common` are not tenants, they are descriptors to tell AAD "I don't know what tenant I want, you figure it out for me". These 2 descriptors should not be used in conjunction with client_creds grant.

hajekj:
Gotcha, makes sense! I have updated the code to omit the `consumers` endpoint from validation. Would that be okay now?

jmprieur:
Thanks @hajekj. I think this is ok.
Let's have @jennyf19 review and merge.

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -276,6 +276,11 @@ internal class TokenAcquisition : ITokenAcquisitionInternal
                 throw new ArgumentException(IDWebErrorMessage.ClientCredentialTenantShouldBeTenanted, nameof(tenant));
             }
 
+            if (!string.IsNullOrEmpty(_microsoftIdentityOptions.TenantId) && _metaTenantIdentifiers.Contains(_microsoftIdentityOptions.TenantId))
```
jennyf19, line 279: This isn't needed. We already are checking the tenant in the lines above (starting line 274), and the developer needs to specifically pass in the tenant in this method, so it doesn't matter what the tenant is with the options. They could even be different tenants, we just care about the one passed in.

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -86,8 +86,8 @@ internal class TokenAcquisition : ITokenAcquisitionInternal
         private readonly ISet<string> _metaTenantIdentifiers = new HashSet<string>(
             new[]
             {
+                Constants.Common,
```
jennyf19, line 89: this is the only fix that is needed. Can you add a test as well? or modify an existing test?

hajekj: Sure! Will add it tomorrow.

src/Microsoft.Identity.Web/TokenAcquisition.cs
hajekj: I am not 100% sure about that, since it either takes the provided tenant or uses the Authority (https://github.com/AzureAD/microsoft-identity-web/blob/master/src/Microsoft.Identity.Web/TokenAcquisition.cs#L703) passed from within the options passed to MSAL, that's why I added this part.

The reason is to cover a scenario, where someone sets up the authority in settings as `common` for example, and would then call `GetAccessTokenForAppAsync(scope)` without explicitly provided `tenant` which won't work and AAD will return an error, so this check prevents the request reaching AAD at all.

I am perfectly fine with removing it of course. Please let me know, what makes more sense.

jennyf19: @hajekj I see. thanks for the explanation. I would propose something like this starting on line 274:
```csharp
if (string.IsNullOrEmpty(tenant))
{
        tenant = _applicationOptions.TenantId ?? _microsoftIdentityOptions.TenantId;
}

if (!string.IsNullOrEmpty(tenant) && _metaTenantIdentifiers.Contains(tenant))
{
}
```

then we'll always have a value for the tenant, or not, and it will be handled before we create the authority. and the authority will have the tenanted value.

jennyf19: PS...i haven't tested it, just a thought.

src/Microsoft.Identity.Web/TokenAcquisition.cs
jennyf19: @hajekj Actually, nevermind on the tests, you'll need access to our KeyVault to run the tests that would cover this. If you can take a look at the comment above, and then once we merge this, i will add the test(s). If you can test manually though and share the results (assuming it works), that would be helpful.

jennyf19: I added a [branch](https://github.com/AzureAD/microsoft-identity-web/blob/jennyf/addTenantTests/tests/Microsoft.Identity.Web.Test.Integration/AcquireTokenForAppIntegrationTests.cs#L74) with some additional tests for the work you have here. will merge after we merge this.

src/Microsoft.Identity.Web/TokenAcquisition.cs
hajekj: Looks good to me. I have added the change and tested locally with very limited scenario. Seems to work fine, but would be great to hook it up with some more tests, especially when the application is configured as multi-tenant with `TenantId` set to `common` or `organizations` in configuration.

src/Microsoft.Identity.Web/TokenAcquisition.cs
hajekj: Looks good to me!

jennyf19 (approved):
thank you so much for all your work in this area @hajekj much appreciated.

jennyf19:
@hajekj nice [blog post](https://hajekj.net/2020/12/12/microsoft-identity-web-and-azure-functions/)! mind if we link to it from our wiki?

jennyf19:
@hajekj  also, in regards to this...
image
if you got it to work, we'll help support it (what a great surprise for us). just open issues here if you need help w/anything. we'll probably try it out in the coming weeks. thanks again!

hajekj:
@jennyf19 Feel free to link or re-use in any way you want! I will update the blog post later today to update the information about the support!

PR #796, opened 2020-12-01
fix tfp double up in legacy b2c domains

jennyf19:


jmprieur (approved):
LGTM
Thanks @jennyf19 

tested on the offending app (B2C web app calling web API, where the Instance was set to http://login.microsoftonline.com/tfp/ and this worked.

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -505,6 +505,16 @@ private async Task<IConfidentialClientApplication> BuildConfidentialClientApplic
             }
         }
 
+        private void PrepareAuthorityInstanceForMsal()
+        {
+            if (_microsoftIdentityOptions.IsB2C && _applicationOptions.Instance.EndsWith("/tfp/"))
```
pmaytak, line 510: Should it be checked in the case of the instance ending with `/tfp` also?

jennyf19: I know. @jmprieur and i discussed this and we are just focusing on a fix for the template, which has `/`. developers should not organically use this, it's not even in any documentation. this is a quick fix for an issue we have, and a longer term fix is more involved w/msal .net fixes and asp .net templates. if customer have issues, we can look at this, or move forward faster w/the fixes needed in the other libs.

Issue #784, opened 2020-11-24
The provided ConfidentialClientApplicationOptions.RedirectUri is not honored when Web App requests access token for the downstream API


vh-vahan:
**Which version of Microsoft Identity Web are you using?**
Note that to get help, you need to run the latest version. 
Microsoft Identity Web 1.3.0

**Where is the issue?**
* Web app
    * [ ] Sign-in users
    * [x] Sign-in users and call web APIs
* Web API
    * [ ] Protected web APIs (validating tokens)
    * [ ] Protected web APIs (validating scopes)
    * [ ] Protected web APIs call downstream web APIs
* Token cache serialization
     * [ ] In-memory caches
     * [ ] Session caches
     * [ ] Distributed caches
* Other (please describe)

**Is this a new or an existing app?**
This is a new app in development

**Repro**

```csharp
services.AddAuthentication("OpenIdConnect")
	.AddMicrosoftIdentityWebApp(
		msIdentityOptions =>
		{
			msIdentityOptions.ClientId = myWebAppClientId;
			msIdentityOptions.Scope.Add(myWebApiScope);

			msIdentityOptions.Events.OnRedirectToIdentityProvider = context =>
			{
				context.ProtocolMessage.RedirectUri = clusterLoadBalancedUri;
				return Task.CompletedTask;
			};

			msIdentityOptions.Instance = Instance;
			msIdentityOptions.Domain = Domain;
			msIdentityOptions.TenantId = TenantId;
                        .........
		})
	.EnableTokenAcquisitionToCallDownstreamApi(
		confidentialClientApplicationOptions =>
		{
			confidentialClientApplicationOptions.RedirectUri = clusterLoadBalancedUri;

			confidentialClientApplicationOptions.ClientId = myWebAppClientId;
			confidentialClientApplicationOptions.Instance = Instance;
			confidentialClientApplicationOptions.TenantId = TenantId;
		})
	.AddInMemoryTokenCaches();
```

**Expected behavior**
confidentialClientApplicationOptions.RedirectUri is used when provided.

**Actual behavior**
I have an ASP.NET Core web app that calls a downstream web API, and being deployed into Service Fabric cluster. The Web App's redirect URI in app registration portal  is configured to be the load balanced URI of the cluster and also localhost. 

It works on local environment when the app is deployed locally.
But I get exception when it runs on the cluster -

Exception occurred while processing the request.
System.Exception: An error was encountered while handling the remote login.
[...]

PR #802, opened 2020-12-02
take cca options redirecturi if provided

jennyf19:
#784


jmprieur (approved):
LGTM
Thanks @jennyf19

PR #800, opened 2020-12-02
fix login/domain hint constant

jennyf19:
#798

Issue #798, opened 2020-12-02
[Bug] Constants.LoginHint should value login_hint, not loginHint (and domain_hint for DomainHint)


jmprieur:
**Which version of Microsoft Identity Web are you using?**
1.3.0

**Where is the issue?**
* Web app
    * [ x ] Sign-in users
    * [ x ] Sign-in users and call web APIs

**Repro**

Call `AccountController.Challenge` with a loginHint.

```csharp
var your = (code) => here;
```

**Expected behavior**
The call to AAD should show login_hint

**Actual behavior**
It's not

**Possible solution**
Change the value of the LoginHint to `login_hint`
https://github.com/AzureAD/microsoft-identity-web/blob/a4bde02483a792bc0f3af68f2da702be88175e88/src/Microsoft.Identity.Web/Constants/Constants.cs#L15

Same for DomainHint which should value `domain_hint`

**Additional context / logs / screenshots**
Add any other context about the problem here, such as logs and screenshots.

jmprieur:
@jennyf19 do you agree?

jennyf19:
Included in [1.4 Release](https://github.com/AzureAD/microsoft-identity-web/releases/tag/1.4.0).

PR #812, opened 2020-12-05
Fixes the login_hint and domain_hint challenge

jmprieur:
Fixes the login_hint and domain_hint challenge.
This is a complement to the fix done in https://github.com/AzureAD/microsoft-identity-web/issues/798, which was not sufficient

The issue was:
- after fixing #800, `MicrosoftIdentityConsentAndConditionalAccessHandler.ChallengeUser` was now calling `AccountController.Challenge` with the `login_hint` and `domain_hint` parameters names (that is the **OIDC** parameter names), whereas the name of the **method** parameters are `loginHint` and `domainHint`. Added 2 new constants in `Constants` for these two method parameters names.
- The `AccountController.Challenge` method was sending the `login_hint` and `domaint_hint` to the STS as **properties** whereas they should be **parameters** (this is a very subtle behavior of the OpenIdConnectMessage)

Tested on Caleb's application. The user is no longer requested to sign-in.
We probably need to check on the tests samples, but this should really improve the SSO experience

### DIFF
--- a/build/template-install-dotnet-core.yaml
+++ b/build/template-install-dotnet-core.yaml
@@ -5,7 +5,7 @@ steps:
 - task: UseDotNet@2
   displayName: 'Use .Net Core SDK 5'
   inputs:
-    version: 5.0.0
+    version: 5.0.100
 - task: UseDotNet@2
   displayName: 'Use .Net Core SDK 3.1.404'
   inputs:
--- a/build/template-install-dotnet-core.yaml
+++ b/build/template-install-dotnet-core.yaml
@@ -3,7 +3,11 @@
 
 steps:
 - task: UseDotNet@2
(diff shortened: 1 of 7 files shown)


jennyf19 (approved):

build/template-install-dotnet-core.yaml
```diff
@@ -5,10 +5,9 @@ steps:
 - task: UseDotNet@2
   displayName: 'Use .Net Core SDK 5'
   inputs:
-    version: 5.0.100-preview.7.20366.6
-    includePreviewVersions: true # Required for preview versions
+    version: 5.0.0
```
jennyf19, line 8:
```suggestion
    version: 5.0.100
```

Issue #809, opened 2020-12-04
This should not try to configure 'everything' especially setting Cookie Authentication


Shazwazza:
**Which version of Microsoft Identity Web are you using?**
1.3.0

**Where is the issue?**
* Other (please describe)

It's not possible to disable this library from adding cookie authentication which is problematic if you already have it. This library seems to be trying to do 'everything' which I understand can mean simplicity for some users but it reduces the flexibility. The work around for now is to pass in a "Fake" cookie scheme that doesn't exist, else you will get exceptions if you pass in an existing one since it already exists. 

It would be nice to split this method up with additional public methods for more flexibility. See here https://github.com/AzureAD/microsoft-identity-web/blob/c75ea22dd4cf99ff6fc30c805d3d0cffcbc76879/src/Microsoft.Identity.Web/WebAppExtensions/MicrosoftIdentityWebAppAuthenticationBuilderExtensions.cs#L238 it's doing all sorts of stuff. Ideally there would just be a method to configure the OpenIdConnect options here https://github.com/AzureAD/microsoft-identity-web/blob/c75ea22dd4cf99ff6fc30c805d3d0cffcbc76879/src/Microsoft.Identity.Web/WebAppExtensions/MicrosoftIdentityWebAppAuthenticationBuilderExtensions.cs#L254

**Is this a new or an existing app?**
c. This is a new app or an experiment.

**Expected behavior**
Be able to add/configure the openidconnect options without configuring a bunch of unwanted or already existing services.

**Actual behavior**
There is not an option to do this

Shazwazza:
I'm happy to create a PR for this

jmprieur:
@Shazwazza 
Thanks for the feedback. This is also to support samesite workarounds ... which are a bit of a pain to develop.

Yes, you could provide an experimental PR so that we understand your proposed design. We'd like to avoid breaking changes if possible, and also we want simples scenarios to remain simple, and complex ones to be possible. 
Would that work for you?

On the other hand if you have very advanced needs, and know what you are doing, you might not even need to use AddMicrosoftIdentityWebApp. you could iinitialize OpenIdConnect yourself? (with `builder.AddOpenIdConnect`). 

What is your scenario? 
What part of Microsoft.Identity.Web do you want to leverage, and which don't you want?

jennyf19:
[...]

Issue #133, opened 2020-04-28
[Bug] signInManager.GetExternalLoginInfoAsync() always returns null


ManuelHaas:
**Which Version of Microsoft Identity Web are you using ?**
Microsoft Identity Web 0.1.1-preview

**Application**
1. Create new project -> ASP.NET Core Web Application -> Web Application (Model-View-Controller). Authentication: **Individual User Accounts** (Store user accounts in-app).
2. Replace AddAzureAd with AddSignIn

for details see https://github.com/AzureAD/microsoft-identity-web/issues/133#issuecomment-621042225 below

In this I only added two things:
**Where is the issue?**
-> Sign-in user

In our application we use .NET Core Identity Framework (local DB) and external login provider. So we have multiple authentication provider and some users sign in directly (via .NET Core Identity Framework) and some via AAD.
Calling signInManager.GetExternalLoginInfoAsync() leads to null return value with the updated code.
Maybe it has to do with the appsettings property `"CookieSchemeName": "Identity.External"`. Because this was necessary to get external AAD login working. As stated [here](https://stackoverflow.com/questions/60198623/how-to-use-both-azure-ad-authentication-and-identity-on-asp-net-core-3). Maybe it will be ignored by now.

So we replaced the following code:
```csharp
services.AddAuthentication(AzureADDefaults.AuthenticationScheme)
    .AddAzureAD(options => Configuration.Bind("AzureAd", options));
```
with that one:
```csharp
services.AddAuthentication(OpenIdConnectDefaults.AuthenticationScheme)
	.AddSignIn(Configuration, "AzureAd", subscribeToOpenIdConnectMiddlewareDiagnosticsEvents: true);
```

appsettings.json section was not modified:
```
"AzureAd": {
	"Instance": "https://login.microsoftonline.com/",
	"Domain": "xyz.com",
	"TenantId": "AEEE9884-9E13-41E7-A25F-304C44407A78",
	"ClientId": "787EBDED-B77E-4CFC-BAA0-2D2B00C41B4F",
	"CallbackPath": "/signin-oidc",
	"SignedOutCallbackPath ": "/signout-callback-oidc",
	"CookieSchemeName": "Identity.External"
}
```

Login with the "old" code leads to the following log entries:

```
Anforderung gestartet: "POST" https://localhost:44338/Account/ExternalLogin?returnurl=%2F
[...]

PR #817, opened 2020-12-06
fix cookie issue

jennyf19:
#809 & #133 need to do more testing, so WIP


jmprieur (approved):
LGTM
Tested on the `dotnet new --auth Individual` scenario.

PR #825, opened 2020-12-08
expose cache extensions for net 472

jennyf19:


jmprieur (approved):
LGTM. Thanks @jennyf19 
Tested with the v1-to-v2 sample

PR #831, opened 2020-12-11
Activate MSAL logs via application options

jennyf19:
Pros:
- no mapping between `Microsoft.Extensions.Logging.LogLevel` and `Client.LogLevel`
- if enabling pii logs, devs will have to set this value anyway. `"EnablePiiLogging": true` in appsettings.json

Cons:
- have to use the enums values in MSAL - Info, Verbose, Warning, Error exactly, any different throws exception
- have to have two sets of logs now, so `"LogLevel": "Info"` in AzureAd and something like this 
```csharp
"Logging": {
        "LogLevel": {
            "Default": "Information",
...
```

### DIFF
--- a/src/Microsoft.Identity.Web/Microsoft.Identity.Web.xml
+++ b/src/Microsoft.Identity.Web/Microsoft.Identity.Web.xml
@@ -1424,15 +1424,6 @@
             </summary>
             The default is <c>false.</c>
         </member>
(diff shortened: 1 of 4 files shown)


jennyf19:
closing as we are going w/this [option](https://github.com/AzureAD/microsoft-identity-web/pull/832)

PR #832, opened 2020-12-11
doing logs via default logging

jennyf19:
Pros:
- define the loglevel in the regular logs, 
```csharp
"Logging": {
        "LogLevel": {
            "Microsoft.Identity.Web": "Information",
...
```
and we pick it up: 
```csharp
loglevel = configuration.GetValue<Microsoft.Extensions.Logging.LogLevel>("Logging:LogLevel:Microsoft.Identity.Web");
```
- using natural logging system for asp .net

Cons:
- seems to be mapping to a default of `Trace` even with nothing selected. We would not want MSAL logs on all the time, as this is expensive. We'd have to figure out a different option if they want logs, but not MSAL logs.
- mapping between `Microsoft.Extensions.Logging.LogLevel` and `Client.LogLevel`, and the enum values are different
- if enabling pii logs, devs will have to set this value `"EnablePiiLogging": true` in appsettings.json

### DIFF
--- a/src/Microsoft.Identity.Web/TokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web/TokenAcquisition.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.ComponentModel;
 using System.Globalization;
 using System.IdentityModel.Tokens.Jwt;
 using System.Linq;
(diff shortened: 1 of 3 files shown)


jennyf19:
@jmprieur I will do a wiki page for setting this up asap.

jmprieur (approved):
LGTM
Thanks @jennyf19 
Awesome feature!

PR #833, opened 2020-12-11
Updating version of Graph package

GeertVL:
Changed to the latest version of the Microsoft.Graph.Beta package


jmprieur (approved):
Thanks for this update, @GeertVL 
cc: @jennyf19

PR #854, opened 2020-12-29
allow for separate sign up and sign in user flows

dan-giddins:
I am currently building a Blazor Server Web App using Azure B2C for identify management.

For my web app, it would be nice if I could have a separate 'Sign Up' and 'Sign In' user flow, as opposed having to use a single Sign Up/Sign in user flow, taking me to 'Sign In' page, in which new users then have to click through to the small 'Sign up now' link in the bottom. It would be much better if I could have separate buttons on my home page for both.

In order to add this feature, this will require:
* The addition of separate 'Sign Up' and 'Sign In' user flow options
* Moving the old 'MicrosoftIdentity/Account/SignIn' endpoint, (which is actually currently using the combined SignUpSignIn user flow), being moved to something like MicrosoftIdentity/Account/SignUpSignIn
* Creating new 'MicrosoftIdentity/Account/SignUp' and 'MicrosoftIdentity/Account/SignIn' endpoints, just for signing up and signing in.

I have tried to tackle all these in this pull request (which I am currently working on at the point of creating this draft), but please have a look and see if there is anything else I need to add.

### DIFF
--- a/ProjectTemplates/templates/BlazorServerWeb-CSharp/.template.config/dotnetcli.host.json
+++ b/ProjectTemplates/templates/BlazorServerWeb-CSharp/.template.config/dotnetcli.host.json
@@ -13,8 +13,16 @@
       "shortName": ""
     },
     "SignUpSignInPolicyId": {
-      "longName": "susi-policy-id",
-      "shortName": "ssp"
+        "longName": "susi-policy-id",
+        "shortName": "ssp"
+    },
+    "SignUpPolicyId": {
+        "longName": "signup-policy-id",
+        "shortName": "sup"
+    },
+    "SignInPolicyId": {
+      "longName": "signin-policy-id",
+      "shortName": "sip"
     },
     "ResetPasswordPolicyId": {
       "longName": "reset-password-policy-id",
--- a/ProjectTemplates/templates/BlazorServerWeb-CSharp/.template.config/template.json
+++ b/ProjectTemplates/templates/BlazorServerWeb-CSharp/.template.config/template.json
@@ -215,11 +215,25 @@
       "description": "The Azure Active Directory B2C instance to connect to (use with IndividualB2C auth)."
     },
     "SignUpSignInPolicyId": {
+        "type": "parameter",
+        "datatype": "string",
+        "defaultValue": "b2c_1_susi",
+        "replaces": "MySignUpSignInPolicyId",
+        "description": "The sign-in and sign-up policy ID for this project (use with IndividualB2C auth)."
+    },
+    "SignUpPolicyId": {
+        "type": "parameter",
+        "datatype": "string",
+        "defaultValue": "b2c_1_signup",
+        "replaces": "MySignUpPolicyId",
+        "description": "The sign-up policy ID for this project (use with IndividualB2C auth)."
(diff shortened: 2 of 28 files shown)


dan-giddins:
Before I do any more work on this, I have some questions,, as you can see from the comments I have added.

What makes 'IsB2C' true? Currently this is true if DefaultUserFlow (which is currently just SignUpSignInPolicyId) is not NullOrWhitespace, but my intuition says that if you had both a sperate 'SignUp' and 'SignIn' user flow, (and didn't use the combined SignUpSignIn user flow) this would also be 'B2C'. This then affects a quite a few bits of code, as well as a fair few test cases.

As an aside, I think the DefaultUserFlow should be SignUpSignInPolicyId, but I've noticed a few places in the code (which I have flagged up with comments) where it is assumed:
* DefaultUserFlow is SignUpSignInPolicyId
* DefaultUserFlow is the only 'B2C' user flow that is being used
I think that first point is mainly for the test cases, in which case this might be fine? I'm not really sure the standard is for meta-assumptions in test cases...
The second point means that core logic will need to be changed, but I think we should resolve the 'What makes something 'B2C' question first, as I think that will have knock on effects as to how the code is structured.

I am also still unsure as to how 'MicrosoftIdentity/Account/SignIn' URL is currently resolved in the code to be passed on the to SignUpSignIn policy endpoint. As I originally posted, I want to change this so that you can access any of /SignUp, /SignIn, and /SignUpSignIn.

I also cannot see what part of the build is failing as I don't have permissions to the pipeline. Could someone past the details of exactly what is failing please so I can fix it :)

jmprieur:
@dan-giddins 
Thanks for your proposal.

I wonder if just changing this method https://github.com/AzureAD/microsoft-identity-web/blob/c5681fa91e444efca09621e1a5a2fadf9cc5dd0a/src/Microsoft.Identity.Web.UI/Areas/MicrosoftIdentity/Controllers/AccountController.cs#L43 and adding the policy as an optional parameter would not do the trick (then from your home page you could call sign-in with either the sign-up or sign-in policy?

Let's take the modifications of the protect templates aside, as we need to discuss with the ASP.NET Core team about that

For the built, it's meant to fail as tests require access to KeyVault that you don't have.

dan-giddins:
Ah yes, we now use the inbuilt Account controller, I see!

@jmprieur I would prefer to be able to simply call different endpoints, as in my app I just use `href="MicrosoftIdentity/Account/SignIn"`. I would be nice to also be able to just call `href="MicrosoftIdentity/Account/SignOut"` or `href="MicrosoftIdentity/Account/SignInSignUp"`.

I do kind of see where you are coming from, as passing an argument would allow for more flexibility and fewer account methods, but in order to add a new user flow, you have to do quite a bit of code anyway to allow for the extra user flows. Also, there is only a discreet number of user flow types you can have, so it might be nice to have one account method corresponding to each of the following in Azure:
image

dan-giddins:
I have tested locally and it seems to work. Personally, I have opted to use the SignUpSignIn and SignUp endpoints in my webapp, but I checked SignIn and this also works. I have still left a few comments however about various things that I think might be worth talking about :)

PR #863, opened 2021-01-12
fix TSA warning

jennyf19:


jmprieur (approved):
I agree.
I spent 15 mins trying to figure out what could be null given we test against null in 25 for httpContext and 33 for result ..

It's a false positive!

PR #868, opened 2021-01-14
Allowing AuthorizeForScopesAttribute to specify an alternate AuthenticationScheme name

DrEsteban:
Our web project doesn't have OpenIdConnect set as the default AuthenticationScheme for Challenges, so this attribute is currently unusable. (We support several types of authentication schemes)

This change allows consumers to specify the AuthenticationScheme to challenge against if the login session needs to be refreshed.


jmprieur:
@jennyf19. I think that we should take it. This is going to make it easier to have web apps that sign-in users with several authentication providers

The only question is whether the authentication scheme could be already in the context when the exception is processed. Something to verify
Proposing to take it for 1.6.0

jennyf19 (approved):

PR #874, opened 2021-01-15
rename folder

jennyf19:


jmprieur (approved):
LGTM.
Thanks @jennyf19. that's a good idea. The solution was a bit overwhelming!

PR #877, opened 2021-01-15
fix changelog format

jennyf19:


jennyf19:
@jmprieur i think i'm missing something for azure functions...does that seem right? thoughts on what to say? thx.

jmprieur (approved):
LGTM @jennyf19

I created an issue for the work done to support Azure Functions is https://github.com/AzureAD/microsoft-identity-web/issues/878.
We could say.

Microsoft.Identity.Web now helps writing Azure Functions protected with Azure AD or Azure AD B2C. See https://github.com/AzureAD/microsoft-identity-web/issues/878.

PR #914, opened 2021-01-30
fix warnings in integration test

jennyf19:


pmaytak (approved):

tests/Microsoft.Identity.Web.Test.Integration/AcquireTokenForAppIntegrationTests.cs
```diff
@@ -46,7 +46,7 @@ public class AcquireTokenForAppIntegrationTests
             else
             {
                 _output.WriteLine("Connection to Key Vault failed. No secret returned. ");
-                throw new ArgumentNullException(nameof(_ccaSecret), "No secret returned from Key Vault. ");
+                throw new ArgumentNullException(message: "No secret returned from Key Vault. ", null);
             }
```
jmprieur, line 50: I don't understand this fix?

jmprieur (commented):
Maybe the null argument exception fix woukd need explanation?

tests/Microsoft.Identity.Web.Test.Integration/AcquireTokenForAppIntegrationTests.cs
jennyf19: _ccaSecret is not a parameter, so i'm just returning the message instead. ParamName is specifically for a parameter that is null.

jmprieur: I see. Strickly speaking this probably shouldn't be an `ArgumentNullException`, then. But I don't have a better idea.

Issue #846, opened 2020-12-17
[Feature Request] AuthorizeForGroups and AuthorizeForRoles implementation


hajekj:
I think it would be great, if Microsoft.Identity.Web provided two extension methods - `AuthorizeForGroupsAsync(string[] groupIds)` and `AuthorizeForTenantRolesAsync(string[] roleIds)`.

# AuthorizeForGroupsAsync
I am well aware that the token from AAD can contain groups, but the limit is [200](https://docs.microsoft.com/en-us/azure/active-directory/hybrid/how-to-connect-fed-group-claims#:~:text=Azure%20Active%20Directory%20limits%20the,group%20information%20is%20included%20instead.) and transitive membership is not considered.
The method would be an extension for `HttpContext` like [`ValidateAppRole`](https://github.com/AzureAD/microsoft-identity-web/blob/4ab0da4c93ef554fab397cb4f1a873ff62d392c3/src/Microsoft.Identity.Web/Resource/RolesRequiredHttpContextExtensions.cs#L28) and would either output the group ids the user is member of or an `UnauthorizedAccessException`. The limit for `checkMemberGroups` is [20](https://docs.microsoft.com/en-us/graph/api/user-checkmembergroups?view=graph-rest-1.0&tabs=http) so this would be the limit for the input length.

# AuthorizeForTenantRolesAsync
This would allow to easily enable only for example tenant admin access into the application. The input is a list of `roleTemplateId`s as per this [list](https://docs.microsoft.com/en-us/azure/active-directory/roles/permissions-reference#role-template-ids) provided as a built-in enum with support to specify custom directory roles as well (maybe add this in future).

# Implementation
* Since it would call Microsoft Graph behind the scenes, it would probably be best to bundle this with MicrosoftGraph package instead of making it directly into Microsoft.Identity.Web.
* I would consider making this both a regular extension method and an attribute for use within Web Apps. The extension method is important when you want to do your custom RBAC based on groups listed in a database. The attribute can be helpful if you want to only let tenant admins access a specific Area of the application.

I already have implemented something similar in the past, so if we agree on where it should be located (eg. MicrosoftGraph package), I am perfectly fine with implementing this. I have done a very small POC in a private project already and it works just fine.

jennyf19:
Thanks @hajekj seems related to this [issue](https://github.com/AzureAD/microsoft-identity-web/issues/369)

jmprieur:
[...]

PR #921, opened 2021-02-03
[WIP] HttpContext authorization extension methods for Microsoft Graph

hajekj:
#846 

Exposes following method:
```
bool HttpContext.AuthorizeUserAsync(string[]? groupIds = null, string[]? tenantRoleTemplateIds = null, bool useAppOnly = false, string? tenant = null)
```

And following attribute to be used with controllers and pages:
```
[MicrosoftGraphAuthorization(string[]? groups = null, string[]? roles = null, bool useAppOnly = false, string? tenant = null)]
```

It is part of `Microsoft.Identity.Web.MicrosoftGraph` package since it has direct dependency on Microsoft Graph itself for group and role checks.


jennyf19:
@hajekj changing the base and merging to a temp branch that i can work off of. thanks again, will keep you posted.

PR #925, opened 2021-02-04
fix downstream api spelling in some templates

jennyf19:
- did not test -


jmprieur (commented):
LGTM. Thanks @jennyf19 
Good catch!

PR #930, opened 2021-02-05
fix func template issue

jennyf19:
- the template csproj was not loading
 - appsettings.json was not loading from the tool


jmprieur (approved):
LGTM. Thanks @jennyf19 
I tested. both issues are solved.

PR #933, opened 2021-02-05
make func2 based on webapi2

jennyf19:
can't base the template on worker2, as it uses different endpoints:
image
so going to use webapi2 instead.


jmprieur:
Does the function have a launchsettings.json?

jmprieur (approved):
LGTM.
Thanks @jennyf19

jennyf19:
> aunchsettings.json

no it doesn't, but i'll see what happens.

PR #946, opened 2021-02-10
update templates for new scope attribute (initial commit)

jennyf19:
@jmprieur  let's discuss, i have a few questions.

### DIFF
--- a/ProjectTemplates/templates/ComponentsWebAssembly-CSharp/Server/Controllers/CallWebApiController.cs
+++ b/ProjectTemplates/templates/ComponentsWebAssembly-CSharp/Server/Controllers/CallWebApiController.cs
@@ -11,7 +11,7 @@ namespace AspNetCoreMicrosoftIdentityWebProjectTemplates.templates.ComponentsWeb
     [Authorize]
     [ApiController]
     [Route("[controller]")]
-    [RequiredScope("access_as_user", IsReusable = true)] // The web API will only accept tokens 1) for users, and 2) having the "access_as_user" scope for this API
+    [RequiredScope("access_as_user")] // The web API will only accept tokens 1) for users, and 2) having the "access_as_user" scope for this API
     public class CallWebApiController : Controller
     {
         private readonly IDownstreamWebApi _downstreamWebApi;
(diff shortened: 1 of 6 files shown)


ProjectTemplates/templates/ComponentsWebAssembly-CSharp/Server/Controllers/CallWebApiController.cs
```diff
@@ -11,11 +11,9 @@ namespace AspNetCoreMicrosoftIdentityWebProjectTemplates.templates.ComponentsWeb
     [Authorize]
     [ApiController]
     [Route("[controller]")]
+    [RequiredScope("access_as_user", IsReusable = true)] // The web API will only accept tokens 1) for users, and 2) having the "access_as_user" scope for this API
```
jmprieur, line 14: As proposed in the other PR, let's embed IsReusable=true in the RequiredScope attribute itself (as this is a perf optimization)

We need to decide if we prefer:
```CSharp
[RequiredScope("access_as_user"] // The web API will only accept tokens 1) for users, and 2) having the "access_as_user" scope for this API
```

or 
```CSharp
[RequiredScope(CallWebApiController.scopeRequiredByApi]
public class CallWebApiController : Controller
    {
        // The web API will only accept tokens 1) for users, and 2) having the "access_as_user" scope for this API
        public const string[] scopeRequiredByApi = new string[] { "access_as_user" };
```

ProjectTemplates/templates/Worker-CSharp/GreeterService.cs
```diff
@@ -40,10 +34,10 @@ public class GreeterService : Greeter.GreeterBase
         }
 
         [Authorize]
+        [RequiredScope("access_as_user", IsReusable = true)] // The web API will only accept tokens 1) for users, and 2) having the "access_as_user" scope for this API
```
jmprieur, line 37: Would it make sense to have it on the GreeterService?

jmprieur (commented):
LGTM.
We have a couple of decisions to make.

ProjectTemplates/templates/ComponentsWebAssembly-CSharp/Server/Controllers/CallWebApiController.cs
jennyf19: fixed in other pr, so will update

---
In reply to: [573604272](https://github.com/AzureAD/microsoft-identity-web/pull/946#discussion_r573604272) [](ancestors = 573604272)

PR #959, opened 2021-02-11
fix grpc template

jennyf19:


jmprieur (approved):
LGTM
Tested on --auth SingleOrg and this works fine!

Issue #938, opened 2021-02-08
Access Sign Up Directly From Web App


SwagStork:
Currently, the web app documentation shows a "Sign In / Up" button, however there are many instances where it would be beneficial to send the user directly to the Sign Up page. However, I cannot seem to figure out how to do that.

Is this possible with Microsoft Identity Web?

 If so, I would be happy to add to the docs to help others learn as well.

jmprieur:
@SwagStork : You are right. That's not what Microsoft identity web does today.
There is a PR, however, proposed by @dan-giddins to achieve the same, and which we still need to look at more deeply, 
https://github.com/AzureAD/microsoft-identity-web/pull/854

MarcAnnous:
Would be helpful indeed. Thank you @dan-giddins and @jmprieur .

TMUNYU:
How far with the PR @jmprieur. Do we have a potential ETA on this?

dan-giddins:
I have updated my PR #968 off the current master branch, and managed to get it working again.

jmprieur:
This is done service side (AAD/AAD B2C), by providing a sign-up UI.
Closing for now

PR #968, opened 2021-02-13
Allow for separate sign up and sign in user flows - Minimal changes

dan-giddins:
I've created a version of my PR (#854) with only the required changes, as I noticed someone else had the this issue (#938)

### DIFF
--- a/src/Microsoft.Identity.Web.UI/Areas/MicrosoftIdentity/Controllers/AccountController.cs
+++ b/src/Microsoft.Identity.Web.UI/Areas/MicrosoftIdentity/Controllers/AccountController.cs
@@ -141,11 +141,9 @@ namespace Microsoft.Identity.Web.UI.Areas.MicrosoftIdentity.Controllers
                 { Constants.DomainHint, domainHint },
             };
 
-            OAuthChallengeProperties oAuthChallengeProperties = new OAuthChallengeProperties(items, parameters)
-            {
-                Scope = scope?.Split(" "),
-                RedirectUri = redirectUri
-            };
+            OAuthChallengeProperties oAuthChallengeProperties = new OAuthChallengeProperties(items, parameters);
+            oAuthChallengeProperties.Scope = scope?.Split(" ");
+            oAuthChallengeProperties.RedirectUri = redirectUri;
 
             return Challenge(
                 oAuthChallengeProperties,
@@ -204,7 +202,7 @@ namespace Microsoft.Identity.Web.UI.Areas.MicrosoftIdentity.Controllers
         public async Task<IActionResult> EditProfile([FromRoute] string scheme)
         {
             scheme ??= OpenIdConnectDefaults.AuthenticationScheme;
-            AuthenticateResult? authenticated = await HttpContext.AuthenticateAsync(scheme).ConfigureAwait(false);
+            var authenticated = await HttpContext.AuthenticateAsync(scheme).ConfigureAwait(false);
             if (!authenticated.Succeeded)
             {
                 return Challenge(scheme);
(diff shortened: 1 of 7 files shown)


dan-giddins:
If this isn't useful, feel free to close

jmprieur:
@dan-giddins : sorry for the delay. We have some urgent feature to deliver, and for your PR, we want to take the time to look at all the implications on the scenarios.
We'll get there. Thanks for your patience!
cc: @jennyf19

dan-giddins:
@jmprieur no worries! I just thought I would make this option available to possibly make things easier, as I realise there is quite a lot of stuff in the other PR that isn't required for the core functionality of the feature to work. Its up to you when and how you merge!

dan-giddins:
@jmprieur I've returned to a Blazor project with a need for separate sign up/sign in user flow, and would love to see this feature added to the core libs (so I don't have to set up my own building of the libs in the CI/CD). Do you think you'd be able to add this any time soon? I have updated the branch from master, so its worth going over the code again, as I notice some similar flows have changed.

dan-giddins:
After merging from master, I now get the following error:
`IOException: IDX20807: Unable to retrieve document from: 'https://login.microsoftonline.com/<tennant>.onmicrosoft.com/B2C_1_SignUpSignIn/v2.0/.well-known/openid-configuration'. HttpResponseMessage: 'StatusCode: 404`. This happens even when going down the `MicrosoftIdentity/Account/SignUp` link.

I found a solution online where I need to add `"Authority": "https://PropSaleLtd.b2clogin.com/tfp/PropSaleLtd.onmicrosoft.com/B2C_1_SignUp/v2.0/"`, in my appsettings.

This makes it work, but the problem is now that all the links, even for other flows seem to use this. I will need to look at further modifying this library to make this work for all user flows, maybe by dynamically generating/manipulating this 'Authority' property based on which link/flow the user follows. 

@jmprieur happy to take steer from you as to what you think the appropriate architecture is for this.

dan-giddins:
Okay, it looks like an issue is that `OpenIdConnectOptions.Authority` is set in the action that is passed to `builder.Services.AddOptions<OpenIdConnectOptions>` in `AddMicrosoftIdentityWebAppInternal()`. This however is done before we know exactly which authority/userflow will be needed.

dan-giddins:
A lot of my previous comments are moot, as the `IOptionsMonitor<MicrosoftIdentityOptions>` injected into `AccountController` doesn't seem to have either an `Authority` value, or any of the user flow policy ids present anyway. I will need to fix this.

dan-giddins:
@jmprieur I have managed to get it working, but I have no idea if the architecture is appropriate (it probably isn't).

I have added the mapping logic for the policy ids to the `MicrosoftIdentityOptions` dependency in the `.AddOptions<OpenIdConnectOptions>().Configure` in `AddMicrosoftIdentityWebAppInternal`. I then tried to find the values in `IOptionsMonitor<MicrosoftIdentityOptions>.CurrentValue` in the `AccountController` but they still weren't populated, so I just injected `MicrosoftIdentityOptions` directly. I could then see and use the policy ids, passing them on to the challenge.

I'm sure I've missed something and that there must be a better way to do this. Why are these values not present in the `IOptionsMonitor<MicrosoftIdentityOptions>`? I am still trying to fully get my head around DI so I've probably just overlooked something!

jmprieur:
@dan-giddins : do you want to provide a test app? or maybe update this project? https://github.com/AzureAD/microsoft-identity-web/tree/master/tests/B2CWebAppCallsWebApi/Client

dan-giddins:
> @dan-giddins : do you want to provide a test app? or maybe update this project? https://github.com/AzureAD/microsoft-identity-web/tree/master/tests/B2CWebAppCallsWebApi/Client

@jmprieur I have just added new sign in buttons to the `B2CWebAppCallsWebApi` app. I get some strange results when testing. I think someone who has access needs to check that https://fabrikamb2c.b2clogin.com has user flows for b2c_1_su and b2c_1_si, and that they are set up correctly. I am also missing a client secret.

jennyf19:
@dan-giddins appreciate the contribution, but we are going to close this PR due to lack of activity and ability on our end to thoroughly test the proposal. We are working to do better at working with external contributions, and have some [guidelines](https://github.com/AzureAD/microsoft-identity-web/blob/master/CONTRIBUTING.md) to hopefully make this less painful for you in the future. Apologies for any cycles lost on your end.

dan-giddins:
@jennyf19 Fair enough. I found this work useful when I was working on a project, so I might come back to my branch one day.

PR #980, opened 2021-02-17
Jmprieur/fixing linux issues

jmprieur:
Jmprieur/fixing Linux issues:
- \\ is a Windows separator
- the name of the project should be case sensitively the same.


jennyf19 (approved):
cool.

PR #988, opened 2021-02-18
Updating the tool to account for non-default values.

jmprieur:
this PR updates the tool to account for the fact that, now, the reader sets all the read values even when they are the same as the default values in the template, 

Tested with the E2E tests.


jennyf19 (approved):
:shipit:

PR #991, opened 2021-02-19
fix for issue 23

jennyf19:
Fixes https://github.com/jmprieur/app-provisonning-tool/issues/23

## How to test?

Under tools\app-provisioning-tool:

### Install or update the tool

```Shell
dotnet tool uninstall --global ms-identity-app
dotnet pack
dotnet tool install --global --add-source app-provisioning-tool\nupkg ms-identity-app
```

### Run the tool (from anywhere)

```Shell
ms-identity-app --help
```

### DIFF
--- a/tools/app-provisioning-tool/app-provisioning-tool/Program.cs
+++ b/tools/app-provisioning-tool/app-provisioning-tool/Program.cs
@@ -29,12 +29,12 @@ namespace DotnetTool
         /// <param name="clientSecret">Client secret to use as a client credential.</param>
         /// <returns></returns>
         static public async Task Main(
-            string? tenantId = null,
-            string? username = null,
-            string? clientId = null,
-            bool? unregister = false,
-            string? folder = null,
-            string? clientSecret = null)
+            string? tenantId,
+            string? username,
+            string? clientId,
+            string? folder,
+            string? clientSecret,
+            bool? unregister)
         {
             // Read options
             ProvisioningToolOptions provisioningToolOptions = new ProvisioningToolOptions


jmprieur (approved):
LGTM.
Actually the user experience is good, the command line library does not oblige the parameters to be entered.


Alternatively, we could have had:

```CSharp
static public async Task Main(
            string? username,
            string? clientId,
            string? clientSecret,
            bool? unregister,
            string? tenantId = "home tenant id",
            string? folder = "current folder"
            )
{ 
 if (tenantId == "home tenant id")
 {
  tenantId = null;
 }

 if (folder == "current folder")
 {
  folder = null;
 }
 // ...
```

which produces:
 ```Text
ms-identity-app:
  Creates or updates an AzureAD/Azure AD B2C application, and updates the code, using
   the developer credentials (Visual Studio, Azure CLI, Azure RM PowerShell, VS Code).

Usage:
  ms-identity-app [options]

Options:
  --username <username>              Username to use to connect to the Azure AD or Azure AD B2C tenant.
                                      It's only needed when you are signed-in in Visual Studio, or Azure CLI with several identities.
                                      In that case, the username param is used to disambiguate which identity to use.
  --client-id <client-id>            Client ID of an existing application from which to update the code. This is
                                      used when you don't want to register a new app in AzureAD/AzureAD B2C, but want to configure the
                                      code from an existing application (which can also be updated by the tool).
  --client-secret <client-secret>    Client secret to use as a client credential.
  --unregister                       Unregister the application, instead of registering it.
  --tenant-id <tenant-id>            Azure AD or Azure AD B2C tenant in which to create/update the app.
                                      If specified, the tool will create the application in the specified tenant.
                                      Otherwise it will create the app in your home tenant ID. [default: home tenant id]
  --folder <folder>                  When specified, will analyze the application code in the specified folder.
                                      Otherwise analyzes the code in the current directory. [default: current folder]
  --version                          Show version information
  -?, -h, --help                     Show help and usage information
```

However, I think that the order of the parameters is more important (the tenantId is the one that will be changed the most often from the scenarios expressed in the Readme.md).

So I prefer your solution.
Thank you!

jennyf19:
@jmprieur i see...i wasn't sure how to test it actually, but glad it works. i wasn't sure if this is what we would want. glad we don't have to do a bunch of string comparisons.

PR #999, opened 2021-02-20
fix error message

jennyf19:
**Update**
Working now for everything but MSA, w/the `.../.default` scope

----
@jmprieur i have to have an account w/MFA enabled it seems. but anyway, here is the experience based on the updated error message. I thought it would be helpful for us to be able to get the actual error returned from MSAL, instead of relying on a stack trace, which might be harder to get later.
image

After setting up MFA, i get this error...but, it's because the template was already configured for a Microsoft tenant, when I removed the pre configuration, it worked.

image

When creating an MSA account via VS UI, getting an Msal Service Exception that consumers is not available.
image

### DIFF
--- a/tools/app-provisioning-tool/app-provisioning-lib/DeveloperCredentials/MsalTokenCredential.cs
+++ b/tools/app-provisioning-tool/app-provisioning-lib/DeveloperCredentials/MsalTokenCredential.cs
@@ -102,7 +102,11 @@ namespace Microsoft.Identity.App.DeveloperCredentials
                     "Error returned: {1}",
                     account?.Username ?? "Account not specified, sign-in to Visual Studio",
                     ex.Message);
-                Environment.Exit(1);
+                result = await app.AcquireTokenInteractive(requestContext.Scopes)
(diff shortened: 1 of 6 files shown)


tools/app-provisioning-tool/app-provisioning-lib/DeveloperCredentials/MsalTokenCredential.cs
```diff
@@ -98,13 +99,22 @@ public override async ValueTask<AccessToken> GetTokenAsync(TokenRequestContext r
             }
             catch (MsalUiRequiredException ex)
             {
-                Console.WriteLine("Please re-sign-in in Visual Studio. ");
+                Console.WriteLine("No valid tokens found in the cache.\nPlease sign-in to Visual Studio with this account:\n\n{0}.\n\nAfter signing-in, re-run the tool.\n" +
```
jmprieur, line 102: Awesome improvement, @jennyf19

jmprieur (approved):
LGTM.
Thanks @jennyf19

PR #1005, opened 2021-02-22
catch graph service exception

jennyf19:


jmprieur (approved):
LGTM; Thanks @jennyf19

PR #1010, opened 2021-02-23
Fixing the end to end test (which were no longer compiled)

jmprieur:
Fixing a CA issue.


jennyf19 (approved):
thanks. i had this locally but not pushed a branch.

PR #1012, opened 2021-02-23
set LegacyCacheCompat w/ADAL to false & as option

jennyf19:


jmprieur (approved):
LGTM. Thanks @jennyf19

PR #1013, opened 2021-02-24
re-adding disabling of null check, as we keep getting a warning, even…

jennyf19:
… though false positive.


jmprieur (approved):
LGTM.
I've no idea why the analyzer presents a false positive here.
Looking with the tooltips, Visual Studio clearly says httpContext is not null and result is not null.

PR #1024, opened 2021-02-25
update cert

jennyf19:
It's now possible to specify the `X509KeyStorageFlags` in the certificate description (both in the config file, or programmatically). This way if customers want to use other storage flags than the default, this is possible;.


jmprieur (approved):
LGTM
Thanks @jennyf19

jmprieur (approved):
LGTM

Issue #957, opened 2021-02-11
[Feature Request] Provide a more sophisticated token cache L1/L2 (L1 in InMemoryCache, L2 IDistributedCache, with encryption strategy)


jmprieur:
**Is your feature request related to a problem? Please describe.**
@henrik-me to update or @jennyf19 / @jmprieur  after discussing with @henrik-me and/or @GeoK

**Describe the solution you'd like**
TBD

**Describe alternatives you've considered**
TBD

**Additional context**
See MISE

damienbod:
It would be great that when the required access token has expired, or does not exist in the cache, the lib just tries to get a new access token. We do this with other OIDC, OAUTH clients. Would this be possible? 

Greetings Damien

jmprieur:
@damienbod : in a web app? or in a web API?
in a web app, Microsoft.Identity.Web manages it already with the [AuthorizeForScopes] attribute. See https://github.com/AzureAD/microsoft-identity-web/wiki/Managing-incremental-consent-and-conditional-access

damienbod:
Hi @jmprieur In both, I don't see this from the docs. But you say then that if I don't use a persistent cache, then I can use the  [AuthorizeForScopes] attribute on my API and it will get me a new token if the the cache has been reset?

Correct?

Thanks Damien

jmprieur:
Thanks @damienbod 
this is on the backlog: https://github.com/AzureAD/microsoft-identity-web/issues/685
(today if you have a web API, you'd need handle the exception and pass it to ITokenAcquisition.ReplyForbiddenWithWwwAuthenticateHeaderAsync. See https://github.com/AzureAD/microsoft-identity-web/wiki/web-apis#handle-conditional-access

jennyf19:
Included in [1.8.0 release](https://github.com/AzureAD/microsoft-identity-web/releases/tag/1.8.0)

PR #1037, opened 2021-03-03
Add L1/L2 cache implementation

jennyf19:
#957 

Manual testing done:
- [x] L1 w/L2 redis cache in docker container w/the perf runner w/10k users

TODO:
- handle redis errors/failures - probably in a second PR

### DIFF
--- a/src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
+++ b/src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
@@ -2,14 +2,19 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Diagnostics;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Caching.Distributed;
+using Microsoft.Extensions.Caching.Memory;
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
+using Microsoft.Identity.Web.TokenCacheProviders.InMemory;
 
 namespace Microsoft.Identity.Web.TokenCacheProviders.Distributed
 {
     /// <summary>
-    /// An implementation of the token cache for both Confidential and Public clients backed by MemoryCache.
+    /// An implementation of the token cache for both Confidential and Public clients backed by a DistributedCache.
+    /// The DistributedCache, by default create a Memory Cache, for faster look up, two level cache.
     /// </summary>
     /// <seealso>https://aka.ms/msal-net-token-cache-serialization</seealso>
     public class MsalDistributedTokenCacheAdapter : MsalAbstractTokenCacheProvider
@@ -18,28 +23,49 @@ namespace Microsoft.Identity.Web.TokenCacheProviders.Distributed
         /// .NET Core Memory cache.
         /// </summary>
         private readonly IDistributedCache _distributedCache;
+        private readonly IMemoryCache _memoryCache;
+        private readonly ILogger<MsalDistributedTokenCacheAdapter> _logger;
+
+        /// <summary>
+        /// MSAL distributed token cache options.
+        /// </summary>
+        private readonly MsalDistributedTokenCacheAdapterOptions _distributedCacheOptions;
 
         /// <summary>
         /// MSAL memory token cache options.
         /// </summary>
-        private readonly MsalDistributedTokenCacheAdapterOptions _cacheOptions;
+        private readonly MsalMemoryTokenCacheOptions _memoryCacheOptions;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="MsalDistributedTokenCacheAdapter"/> class.
         /// </summary>
-        /// <param name="memoryCache">Distributed cache instance to use.</param>
-        /// <param name="cacheOptions">Options for the token cache.</param>
+        /// <param name="distributedCache">Distributed cache instance to use.</param>
+        /// <param name="distributedCacheOptions">Options for the token cache.</param>
+        /// <param name="logger">MsalDistributedTokenCacheAdapter logger.</param>
+        /// <param name="memoryCache">Memory cache instance to use.</param>
+        /// <param name="memoryCacheOptions">Memory cache options.</param>
         public MsalDistributedTokenCacheAdapter(
-                                            IDistributedCache memoryCache,
-                                            IOptions<MsalDistributedTokenCacheAdapterOptions> cacheOptions)
+                                            IDistributedCache distributedCache,
+                                            IOptions<MsalDistributedTokenCacheAdapterOptions> distributedCacheOptions,
+                                            ILogger<MsalDistributedTokenCacheAdapter> logger,
+                                            IMemoryCache? memoryCache = null,
+                                            IOptions<MsalMemoryTokenCacheOptions>? memoryCacheOptions = null)
         {
-            if (cacheOptions == null)
+            if (distributedCacheOptions == null)
             {
-                throw new ArgumentNullException(nameof(cacheOptions));
+                throw new ArgumentNullException(nameof(distributedCacheOptions));
             }
 
-            _distributedCache = memoryCache;
-            _cacheOptions = cacheOptions.Value;
+            if (memoryCacheOptions == null)
+            {
+                memoryCacheOptions = Options.Create(new MsalMemoryTokenCacheOptions());
+            }
+
+            _distributedCache = distributedCache;
+            _distributedCacheOptions = distributedCacheOptions.Value;
+            _memoryCache = memoryCache ?? new MemoryCache(new MemoryCacheOptions());
+            _memoryCacheOptions = memoryCacheOptions.Value;
+            _logger = logger;
         }
 
         /// <summary>
@@ -50,7 +76,16 @@ namespace Microsoft.Identity.Web.TokenCacheProviders.Distributed
         /// <returns>A <see cref="Task"/> that completes when key removal has completed.</returns>
         protected override async Task RemoveKeyAsync(string cacheKey)
         {
+            // remove in both
+            var stopwatch = new Stopwatch();
+            stopwatch.Start();
+            _memoryCache.Remove(cacheKey);
+            stopwatch.Stop();
+            _logger.LogInformation($"[IdWebCache]: Remove cacheKey {cacheKey} MemoryCache time: {stopwatch.Elapsed.TotalMilliseconds}. ");
+            stopwatch.Start();
             await _distributedCache.RemoveAsync(cacheKey).ConfigureAwait(false);
+            stopwatch.Stop();
+            _logger.LogInformation($"[IdWebCache]: Remove cacheKey {cacheKey} DistributedCache time: {stopwatch.Elapsed.TotalMilliseconds}. ");
         }
 
         /// <summary>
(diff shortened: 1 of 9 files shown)


src/Microsoft.Identity.Web/TokenCacheProviders/MeasureDurationResult.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace Microsoft.Identity.Web.TokenCacheProviders
+{
+    internal struct MeasureDurationResult
```
jennyf19, line 6: we aren't using these right now, but going to leave them for if we want to later (copied from MISE).

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
```diff
@@ -4,42 +4,57 @@
...
             {
-                throw new ArgumentNullException(nameof(cacheOptions));
+                throw new ArgumentNullException(nameof(distributedCacheOptions));
             }
 
-            _distributedCache = memoryCache;
-            _cacheOptions = cacheOptions.Value;
+            _distributedCache = distributedCache;
+            _distributedCacheOptions = distributedCacheOptions.Value;
+            _memoryCache = new MemoryCache(new MemoryCacheOptions { SizeLimit = _distributedCacheOptions.L1CacheSizeLimit * (1024 * 1024) });
```
bgavrilMS, line 51: Is there a scenario where the user would want to overwrite this? Magic numbers and all...

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
```diff
@@ -4,42 +4,57 @@
...
+            _logger = logger;
+
+            if (_distributedCacheOptions.AbsoluteExpirationRelativeToNow != null)
+            {
+                _expirationTime = TimeSpan.FromMilliseconds(_distributedCacheOptions.AbsoluteExpirationRelativeToNow.Value.TotalMilliseconds * _distributedCacheOptions.L1ExpirationTimeRatio);
```
bgavrilMS, line 56: Is it possible that the L1ExpirationTimeRatio is set to a number > 1? Should there be a validation?

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
```diff
@@ -50,7 +65,12 @@ public class MsalDistributedTokenCacheAdapter : MsalAbstractTokenCacheProvider
         /// <returns>A <see cref="Task"/> that completes when key removal has completed.</returns>
         protected override async Task RemoveKeyAsync(string cacheKey)
         {
+            var startTicks = Utility.Watch.Elapsed.Ticks;
```
bgavrilMS, line 68: nit: here's a little trick you can do achieve the same but with nicer syntax - https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/ee169d69820a92fbf9538020592e30a9972ae33d/src/client/Microsoft.Identity.Client/Internal/Logger/DurationLogHelper.cs#L10

syntax would be 

```csharp
using (_logger.LogDuration("[IdWebCache] Do stuff"))
{
     // do stuff
}
```

This would log: 

[IdWebCache] Do stuff started
[IdWebCache] Do stuff finished in 30 ms

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
```diff
@@ -62,7 +82,37 @@ protected override async Task RemoveKeyAsync(string cacheKey)
...
+                    };
+
+                    _logger.LogDebug($"[IdWebCache] Back propagate from Distributed to Memory, Byte size: {result?.Length}");
+                    _memoryCache.Set(cacheKey, result, memoryCacheEntryOptions);
+                    _logger.LogDebug($"[IdWebCache] MemoryCache: Count: {_memoryCache.Count}");
+                }
+            }
+
+            _logger.LogDebug($"[IdWebCache] Read caches for {cacheKey} returned Bytes {result?.Length} Time in Ticks: {Utility.Watch.Elapsed.Ticks - startTicks}. ");
+#pragma warning disable CS8603 // Possible null reference return.
```
bgavrilMS, line 113: You could return an empty byte[] (i.e. new byte[0]) instead of null and get rid of this warning.

bgavrilMS (commented):
LGTM, minor comments

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
jmprieur: Yes, @bgavrilMS : 
we'd want the customer/services to decide which amount of memory they want to use for the L1 cache. By default this is 500Mb, but it's up to them to decide. It's a design decision.
See the MsalDistributedTokenCacheAdapterOptions properties that provides options which seem to make sense to avoid the L1 cache to grow forever (otherwise it would not be useful to have the L2 and all the memory will be exhausted)

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
jmprieur: What's the contract with MSAL.NET, @bgavrilMS : I thought that null needed to be returned when the cache does not exist?

here this is because Ms.Id.Web overrides an MSAL.NET method that does not use nullable (and therefore the signature does no have it)

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapterOptions.cs
```diff
@@ -12,5 +12,17 @@ namespace Microsoft.Identity.Web.TokenCacheProviders.Distributed
...
+        /// Default is 500 Mb.
+        /// </summary>
+        public long L1CacheSizeLimit { get; set; } = 500;
+
+        /// <summary>
+        /// Value more than 0, less than 1, to set the In Memory (L1) cache
+        /// expiration time values relative to the Distributed (L2) cache.
+        /// Default is 1.
+        /// </summary>
+        public double L1ExpirationTimeRatio { get; set; } = 1;
```
henrik-me, line 26: I know this is used in MISE, I'm not convinced about this time ratio. would be good with an example in the summary which explains what it means for the L2 cache size in actual numbers.

henrik-me (approved):
:shipit:

henrik-me:
Would be good with some numbers from the perf run posted in the description.

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapterOptions.cs
jennyf19: It's actually not using MISE, we invented this. will add more details.

---
In reply to: [586554302](https://github.com/AzureAD/microsoft-identity-web/pull/1037#discussion_r586554302) [](ancestors = 586554302)

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
jennyf19: my understanding was to return null as well, at least that's the current behavior.

---
In reply to: [586403545](https://github.com/AzureAD/microsoft-identity-web/pull/1037#discussion_r586403545) [](ancestors = 586403545)

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
jennyf19: thanks for the info and link. I took a look, but really we are doing a lot more in the log messages, like logging cache size, cache key, count, and more, not just time, so I think i'll stick to what the MISE team is doing for now and we might revisit this later.

---
In reply to: [586320555](https://github.com/AzureAD/microsoft-identity-web/pull/1037#discussion_r586320555) [](ancestors = 586320555)

jennyf19:
agree, but the perf stuff needs some work

---
In reply to: [789836536](https://github.com/AzureAD/microsoft-identity-web/pull/1037#issuecomment-789836536) [](ancestors = 789836536)

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
jennyf19: What @jmprieur said. we are trying to keep a lid on the memory growth, but allow it to be configurable.

---
In reply to: [586402386](https://github.com/AzureAD/microsoft-identity-web/pull/1037#discussion_r586402386) [](ancestors = 586402386)

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
jennyf19: fixed. nice catch.

---
In reply to: [586317374](https://github.com/AzureAD/microsoft-identity-web/pull/1037#discussion_r586317374) [](ancestors = 586317374)

jmprieur (approved):
LGTM
Thanks @jennyf19

jmprieur (approved):
LGTM
Thanks @jennyf19 

SonarCloud is going to be super happy :)

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapterOptions.cs
henrik-me: I don't think we should expose this

jmprieur: Why @henrik-me : don't we want the L1 cache to expire before the L2 one ?

jennyf19: I made it internal for now. [https://github.com/AzureAD/microsoft-identity-web/pull/1039](https://github.com/AzureAD/microsoft-identity-web/pull/1039)

---
In reply to: [586726432](https://github.com/AzureAD/microsoft-identity-web/pull/1037#discussion_r586726432) [](ancestors = 586726432)

jmprieur: BTW, @henrik-me : this was about time, not size. There are 2 dimensions in these caches: the freshness, and the size limit.

PR #1040, opened 2021-03-03
Add try/catch to handle L2 connection issues

jennyf19:
@henrik-me 
If you want to test this, use the dev app, which has the redis cache set up already, just uncomment `//#define UseRedisCache` in `Startup.cs` of the client, and install [docker desktop](https://docs.docker.com/docker-for-windows/install/). You can start the L2, sign-in, then stop the L2 and sign-out, or just have start stopped.
From command prompt:
```shell
docker run --name my-redis -p 5002:6379 -d redis
docker exec -it my-redis sh
redis-cli
```
to stop the connection:
```shell
exit
exit
docker ps -a
docker stop [enter 1st three digits of the redis cache name]
```

Before adding the try/catch, unhandled exception and app failure:
image

After, we can't catch a specific exception, as there are different types of L2 caches, so we can't catch an exception specific to a distribute cache implementation (ex. `RedisConnectionException`) (will discuss w/asp net core team on a possible feature to allow us to detect the connection availability of a Distributed cache implementation), so just catching `Exception`. There is no common base class for such exceptions for the L2.

The user experience now is that we log an Error (see below) and the L1 (Memory cache) takes over, so no interruption of service from a user perspective.

image

### DIFF
--- a/src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
+++ b/src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
@@ -74,8 +74,15 @@ namespace Microsoft.Identity.Web.TokenCacheProviders.Distributed
             _memoryCache.Remove(cacheKey);
             _logger.LogDebug($"[IdWebCache] MemoryCache: Remove cacheKey {cacheKey} Time in Ticks: {Utility.Watch.Elapsed.Ticks - startTicks}. ");
 
-            await _distributedCache.RemoveAsync(cacheKey).ConfigureAwait(false);
-            _logger.LogDebug($"[IdWebCache] DistributedCache: Remove cacheKey {cacheKey} Time in Ticks: {Utility.Watch.Elapsed.Ticks - startTicks}. ");
+            try
+            {
+                await _distributedCache.RemoveAsync(cacheKey).ConfigureAwait(false);
+                _logger.LogDebug($"[IdWebCache] DistributedCache: Remove cacheKey {cacheKey} Time in Ticks: {Utility.Watch.Elapsed.Ticks - startTicks}. ");
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError($"[IdWebCache] Connection issue encountered with Distributed cache. Currently using In Memory cache only. Error message: {ex.Message} ");
+            }
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


jmprieur (approved):
LGTM
Thanks @jennyf19. The Distributed token cache adapter just gained some reliability!

henrik-me:
> <?xml version="1.0"?>

Gave up on reviewing changes here. Anything in particular to look out for?

---
Refers to: src/Microsoft.Identity.Web/Microsoft.Identity.Web.xml:1 in 0ab296b. [](commit_id = 0ab296b17cfe9cc2de0435048a3c70b6b27fa805, deletion_comment = False)

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
```diff
@@ -141,8 +155,15 @@ protected override async Task WriteCacheBytesAsync(string cacheKey, byte[] bytes
             _logger.LogDebug($"[IdWebCache] MemoryCache: Write cacheKey {cacheKey} Byte size: {bytes?.Length} Time in Ticks: {Utility.Watch.Elapsed.Ticks - startTicks}. ");
             _logger.LogDebug($"[IdWebCache] MemoryCache: Count: {_memoryCache.Count}");
 
-            await _distributedCache.SetAsync(cacheKey, bytes, _distributedCacheOptions).ConfigureAwait(false);
-            _logger.LogDebug($"[IdWebCache] DistributedCache: Write cacheKey {cacheKey} Byte size {bytes?.Length} Time in Ticks: {Utility.Watch.Elapsed.Ticks - startTicks}. ");
+            try
+            {
+                await _distributedCache.SetAsync(cacheKey, bytes, _distributedCacheOptions).ConfigureAwait(false);
```
henrik-me, line 160: I don't think we should await this one, we should let it execute async, and we can also handle the exception async as well. We should treat L2 as optimistic - delays to the incoming call is too big.

henrik-me:
Good to see the additional protection around errors. We should explore how and where we can trigger re-connect, it's a real issue, and customers will end up bypassing our code if we can't give them a way for this. Perhaps something we have to trigger to make applications aware of the connection issue. I would go as far... if we detect it's redis we can do special handling.

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
jmprieur: By not awaiting, do we have a risk of not respecting causality ?

jennyf19: I'd like to discuss this one more. I'll merge this PR and then open another w/this change.

PR #1043, opened 2021-03-04
fix async call in cache write

jennyf19:
for discussion.


jennyf19:
@henrik-me i think we're still discussing this, but this branch has the latest that we (@jmprieur ) discussed.

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
```diff
@@ -74,6 +74,24 @@
...
+                Task.Run(async () =>
+                {
+                    await DistributedCacheRemoveKeyAsync(cacheKey, startTicks).ConfigureAwait(false);
+                });
+#pragma warning restore CS4014 // Because this call is not awaited, execution of the current method continues before the call is completed
+                await Task.CompletedTask.ConfigureAwait(false);
+            }
+            else
+            {
+                await DistributedCacheRemoveKeyAsync(cacheKey, startTicks).ConfigureAwait(false);
```
henrik-me, line 89: >await DistributedCacheRemoveKeyAsync(cacheKey, startTicks).ConfigureAwait(false); [](start = 16, length = 81)

should be an action, and the work should be added to an queue, and there should be an event processor to guarantee ordering of events. should do a bit more work here

jennyf19:
not doing this.

PR #1046, opened 2021-03-04
L2 Cache retry initial commit

### LINKED ISSUE #1042: [Feature Request] Enable apps to decide what to do when the L2 cache fails (Retry)
Today, when the L2 cache fails, an error is logged, but given the L1 cache was able to proceed, we don't re-throw (think of Redis cache connection error)

Have a new property in the `MsalDistributedTokenCacheAdapterOptions`, so that the application can handle the failure ?
 
```CSharp
 /// <summary>
 /// Callback offered to the app to be notified when the L2 cache fails.
 /// This way the app is given the possibility to act on the L2 cache,
 /// for instance, in the case of Redis, to reconnect. This is left to the application as it's
 /// the only one that knows about the real implementation of the L2 cache.
 /// The handler should return <c>true</c> if the cache should try again the operation, and
 /// <c>false</c> otherwise. When <c>true</c> is passed and the retry fails, an exception
 /// will be thrown.
 /// </summary>
 public Func<Exception, bool>? OnL2CacheFailure {get;set;}
```

Here is an example of how this would be hooked from the Startup.cs:
 
 ```CSharp
services.Configure<MsalDistributedTokenCacheAdapterOptions>(options =>
{
 options.L1CacheSizeLimit = 10 * 1024 * 1024; // 10 Mb
 options.OnL2CacheFailure = (ex) =>
 {
   if (ex is StackExchange.Redis.RedisConnectionException)
   {
     // Attempt to act on the redis cache if at all possible?
     // Put here your reconnected code
     return true; // Retry
   } 
   return false;  // Don't retry.
  };
});
```

We could also decide of the behavior when it fails again on retry (we could be silent instead of throwing)

### PR DESCRIPTION
**Pull request by jennyf19**
Fixes #1042 1042
-TODO - needs testing and further work, just to start discussion

### DIFF
--- a/src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
+++ b/src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
@@ -70,15 +70,13 @@ namespace Microsoft.Identity.Web.TokenCacheProviders.Distributed
         /// <returns>A <see cref="Task"/> that completes when key removal has completed.</returns>
         protected override async Task RemoveKeyAsync(string cacheKey)
         {
-            var startTicks = Utility.Watch.Elapsed.Ticks;
             _memoryCache.Remove(cacheKey);
-            _logger.LogDebug($"[MsIdWeb] MemoryCache: Remove cacheKey {cacheKey} Time in Ticks: {Utility.Watch.Elapsed.Ticks - startTicks}. ");
+            _logger.LogDebug($"[MsIdWeb] MemoryCache: Remove cacheKey {cacheKey}. ");
 
             await L2OperationWithRetryOnFailureAsync(
                 "Remove",
                 (cacheKey) => _distributedCache.RemoveAsync(cacheKey),
-                cacheKey,
-                startTicks).ConfigureAwait(false);
+                cacheKey).ConfigureAwait(false);
         }
 
         /// <summary>
@@ -90,20 +88,22 @@ namespace Microsoft.Identity.Web.TokenCacheProviders.Distributed
         /// (account or app).</returns>
         protected override async Task<byte[]> ReadCacheBytesAsync(string cacheKey)
         {
-            var startTicks = Utility.Watch.Elapsed.Ticks;
-
             // check memory cache first
             byte[]? result = (byte[])_memoryCache.Get(cacheKey);
             _logger.LogDebug($"[MsIdWeb] MemoryCache: Read {cacheKey} cache size: {result?.Length}. ");
 
             if (result == null)
             {
-                // not found in memory, check distributed cache
-                result = await L2OperationWithRetryOnFailureAsync(
-                    "Read",
-                    (cacheKey) => _distributedCache.GetAsync(cacheKey),
-                    cacheKey,
-                    startTicks).ConfigureAwait(false);
+                var measure = await Task.Run( async () =>
+                {
+                    // not found in memory, check distributed cache
+                    result = await L2OperationWithRetryOnFailureAsync(
+                        "Read",
+                        (cacheKey) => _distributedCache.GetAsync(cacheKey),
+                        cacheKey).ConfigureAwait(false);
+                }).Measure().ConfigureAwait(false);
+
+                _logger.LogDebug($"[MsIdWeb] DistributedCache: Read time in Ticks: {measure.Ticks}");
 
                 // back propagate to memory cache
                 if (result != null)
@@ -120,7 +120,6 @@ namespace Microsoft.Identity.Web.TokenCacheProviders.Distributed
                 }
             }
 
-            _logger.LogDebug($"[MsIdWeb] Read caches for {cacheKey} returned cache size: {result?.Length} Time in Ticks: {Utility.Watch.Elapsed.Ticks - startTicks}. ");
 #pragma warning disable CS8603 // Possible null reference return.
             return result;
 #pragma warning restore CS8603 // Possible null reference return.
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by henrik-me** [comment]
I think this direction is good

File: src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs [hunk 1]
```diff
@@ -82,6 +82,12 @@ protected override async Task RemoveKeyAsync(string cacheKey)
             catch (Exception ex)
             {
                 _logger.LogError($"[IdWebCache] Connection issue encountered with Distributed cache. Currently using In Memory cache only. Error message: {ex.Message} ");
```
henrik-me, line 84: >Currently using In Memory cache only [](start = 101, length = 36)

Forgot to mention this last time...
Replace: Remove cache entry for (key) failed. Don't have to mention anything about the memory cache here.
(same comment a few other places)

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
```diff
@@ -82,6 +82,12 @@ protected override async Task RemoveKeyAsync(string cacheKey)
...
+
+                if (_distributedCacheOptions.OnL2CacheFailure != null && _distributedCacheOptions.OnL2CacheFailure(ex))
+                {
+                    _logger.LogDebug($"[IdWebCache] DistributedCache: Retry to remove cacheKey {cacheKey}. ");
+                    await _distributedCache.RemoveAsync(cacheKey).ConfigureAwait(false);
```
henrik-me, line 89: >await _distributedCache.RemoveAsync(cacheKey).ConfigureAwait(false); [](start = 20, length = 68)

have to catch this one as well

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
```diff
@@ -82,6 +82,12 @@ protected override async Task RemoveKeyAsync(string cacheKey)
             catch (Exception ex)
             {
                 _logger.LogError($"[IdWebCache] Connection issue encountered with Distributed cache. Currently using In Memory cache only. Error message: {ex.Message} ");
+
+                if (_distributedCacheOptions.OnL2CacheFailure != null && _distributedCacheOptions.OnL2CacheFailure(ex))
```
henrik-me, line 86: Good pattern to apply everywhere. Try/catch can be refactored into a method/generic which takes an action

henrik-me:
>             var startTicks = Utility.Watch.Elapsed.Ticks;

nit: Usage of timer/ticks/time spend can be wrapped up in a helper. Allowing you to have time since method start as well as time for each operation of needed. Also you have now one more place where you want to put timing data into the log. Can also consider adding the total call time for the method.

---
Refers to: src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs:73 in 92fdc0e. [](commit_id = 92fdc0e6e55a0a38eed2b40d71706f2b46ed2207, deletion_comment = False)

jennyf19:
>             var startTicks = Utility.Watch.Elapsed.Ticks;

i've used the Measure() static method for this instead.

---
In reply to: [790928037](https://github.com/AzureAD/microsoft-identity-web/pull/1046#issuecomment-790928037) [](ancestors = 790928037)

---
Refers to: src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs:73 in 92fdc0e. [](commit_id = 92fdc0e6e55a0a38eed2b40d71706f2b46ed2207, deletion_comment = False)

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
jennyf19: done & i agree...was more for testing too to see where we were. 

---
In reply to: [587807465](https://github.com/AzureAD/microsoft-identity-web/pull/1046#discussion_r587807465) [](ancestors = 587807465)

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
jennyf19: fixed.

---
In reply to: [587808630](https://github.com/AzureAD/microsoft-identity-web/pull/1046#discussion_r587808630) [](ancestors = 587808630)

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
jennyf19: yes. fixed.

---
In reply to: [587807756](https://github.com/AzureAD/microsoft-identity-web/pull/1046#discussion_r587807756) [](ancestors = 587807756)

jmprieur (approved):
LGTM.
I tried it out with the Redis cache. It works great.

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
```diff
@@ -70,19 +70,13 @@ public class MsalDistributedTokenCacheAdapter : MsalAbstractTokenCacheProvider
         /// <returns>A <see cref="Task"/> that completes when key removal has completed.</returns>
         protected override async Task RemoveKeyAsync(string cacheKey)
         {
-            var startTicks = Utility.Watch.Elapsed.Ticks;
             _memoryCache.Remove(cacheKey);
-            _logger.LogDebug($"[IdWebCache] MemoryCache: Remove cacheKey {cacheKey} Time in Ticks: {Utility.Watch.Elapsed.Ticks - startTicks}. ");
+            _logger.LogDebug($"[MsIdWeb] MemoryCache: Remove cacheKey {cacheKey}. ");
```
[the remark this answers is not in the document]
jennyf19: fixed.

PR #1065, opened 2021-03-15
Jennyf/l2 refresh

jennyf19:


jennyf19:
was testing and this needs to be handled same as the others if l2 is offline.

jmprieur (approved):
LGTM
good catch @jennyf19 !
Thanks

PR #1073, opened 2021-03-16
update MSAL & Graph SDKs

jennyf19:
@jmprieur I've only tested web app calls web API & web app calls Graph


jmprieur (approved):
LGTM.
Thanks @jennyf19

PR #1082, opened 2021-03-23
Fix thread safety issues in TokenAcquisition classes

DavidParks8:
There were some cases where highly multithreaded environments could cause issues due to non thread-safe code. This PR attempts to resolve the thread safety issues with an eye on maintaining performant execution.

### DIFF
--- a/src/Microsoft.Identity.Web/AppServicesAuth/AppServicesAuthenticationTokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web/AppServicesAuth/AppServicesAuthenticationTokenAcquisition.cs
@@ -18,6 +18,11 @@ namespace Microsoft.Identity.Web
     /// </summary>
     public class AppServicesAuthenticationTokenAcquisition : ITokenAcquisition
     {
+        private readonly object _applicationSyncObj = new object();
+        /// <summary>
+        ///  Please call GetOrBuildConfidentialClientApplication instead of accessing this field directly.
+        /// </summary>
+        private IConfidentialClientApplication? _application;
         private readonly IHttpContextAccessor _httpContextAccessor;
         private readonly IMsalHttpClientFactory _httpClientFactory;
         private readonly IMsalTokenCacheProvider _tokenCacheProvider;
@@ -69,19 +74,29 @@ namespace Microsoft.Identity.Web
 
         private IConfidentialClientApplication GetOrCreateApplication()
         {
-            var options = new ConfidentialClientApplicationOptions()
+            if (_application == null)
+            {
+                lock (_applicationSyncObj)
                 {
-                    ClientId = AppServicesAuthenticationInformation.ClientId,
-                    ClientSecret = AppServicesAuthenticationInformation.ClientSecret,
-                    Instance = AppServicesAuthenticationInformation.Issuer,
-                };
-            var confidentialClientApplication = ConfidentialClientApplicationBuilder.CreateWithApplicationOptions(options)
-                .WithHttpClientFactory(_httpClientFactory)
-                .Build();
-            _tokenCacheProvider.Initialize(_confidentialClientApplication.AppTokenCache);
-            _tokenCacheProvider.Initialize(_confidentialClientApplication.UserTokenCache);
+                    if (_application == null)
+                    {
+                        var options = new ConfidentialClientApplicationOptions()
+                        {
+                            ClientId = AppServicesAuthenticationInformation.ClientId,
+                            ClientSecret = AppServicesAuthenticationInformation.ClientSecret,
+                            Instance = AppServicesAuthenticationInformation.Issuer,
+                        };
+                        _application = ConfidentialClientApplicationBuilder.CreateWithApplicationOptions(options)
+                            .WithHttpClientFactory(_httpClientFactory)
+                            .Build();
+                        _tokenCacheProvider.Initialize(_application.AppTokenCache);
+                        _tokenCacheProvider.Initialize(_application.UserTokenCache);
+                    }
+                }
+            }
 
-            return confidentialClientApplication;
+
+            return _application;
         }
 
         /// <inheritdoc/>
--- a/src/Microsoft.Identity.Web/AppServicesAuth/AppServicesAuthenticationTokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web/AppServicesAuth/AppServicesAuthenticationTokenAcquisition.cs
@@ -19,8 +19,9 @@ namespace Microsoft.Identity.Web
     public class AppServicesAuthenticationTokenAcquisition : ITokenAcquisition
     {
         private readonly object _applicationSyncObj = new object();
+
         /// <summary>
-        ///  Please call GetOrBuildConfidentialClientApplication instead of accessing this field directly.
+        ///  Please call GetOrCreateApplication instead of accessing this field directly.
         /// </summary>
         private IConfidentialClientApplication? _application;
         private readonly IHttpContextAccessor _httpContextAccessor;
@@ -94,7 +95,6 @@ namespace Microsoft.Identity.Web
                     }
                 }
             }
-
 
             return _application;
         }
@@ -196,6 +196,13 @@ namespace Microsoft.Identity.Web
                 scopes,
                 tokenAcquisitionOptions != null ? tokenAcquisitionOptions.CorrelationId : Guid.Empty);
             return authenticationResult;
+        }
+
+        /// <inheritdoc/>
+        public Task ReplyForbiddenWithWwwAuthenticateHeaderAsync(IEnumerable<string> scopes, MsalUiRequiredException msalServiceException, HttpResponse? httpResponse = null)
+        {
+            // Not implemented for the moment
+            throw new NotImplementedException();
         }
 
         /// <inheritdoc/>
--- a/src/Microsoft.Identity.Web/AppServicesAuth/AppServicesAuthenticationTokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web/AppServicesAuth/AppServicesAuthenticationTokenAcquisition.cs
@@ -5,6 +5,7 @@ using System;
 using System.Collections.Generic;
 using System.Net.Http;
 using System.Security.Claims;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Http;
 using Microsoft.Identity.Client;
@@ -15,9 +16,15 @@ namespace Microsoft.Identity.Web
     /// <summary>
     /// Implementation of ITokenAcquisition for App Services authentication (EasyAuth).
     /// </summary>
-    public class AppServicesAuthenticationTokenAcquisition : ITokenAcquisition
+    public class AppServicesAuthenticationTokenAcquisition : ITokenAcquisition, IDisposable
     {
+        private readonly SemaphoreSlim _confidentialClientApplicationSyncObj = new SemaphoreSlim(/* max concurrent locks */ 1);
+        /// <summary>
+        /// Do no read or set this directly.
+        /// Instead, use GetOrCreateApplication which is thread-safe
+        /// </summary>
         private IConfidentialClientApplication? _confidentialClientApplication;
+        private bool _disposedValue;
         private readonly IHttpContextAccessor _httpContextAccessor;
         private readonly IMsalHttpClientFactory _httpClientFactory;
         private readonly IMsalTokenCacheProvider _tokenCacheProvider;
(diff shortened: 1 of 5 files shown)


src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -435,20 +440,49 @@ public async Task RemoveAccountAsync(RedirectContext context)
         /// </summary>
         internal /* for testing */ async Task<IConfidentialClientApplication> GetOrBuildConfidentialClientApplicationAsync()
         {
+            // Use a double checking lock to avoid locking in most situations
             if (_application == null)
```
henrik-me, line 444: > if (_application == null) [](start = 11, length = 26)

I would suggest to use the lazy load pattern instead

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -435,20 +440,49 @@ public async Task RemoveAccountAsync(RedirectContext context)
...
 
             return _application;
         }
 
+        
+        private string BuildCurrentUriFromRequest(HttpContext httpContext, HttpRequest request)
+        {
+            // need to lock to avoid threading issues with code outside of this library
+            // https://docs.microsoft.com/en-us/aspnet/core/performance/performance-best-practices?#do-not-access-httpcontext-from-multiple-threads
+            lock(httpContext)
```
henrik-me, line 469: >httpContext [](start = 17, length = 11)

I'm not sure where values from the httpcontext is read here?

src/Microsoft.Identity.Web/HttpContextExtensions.cs
```diff
@@ -26,7 +30,11 @@ internal static void StoreTokenUsedToCallWebAPI(this HttpContext httpContext, Jw
         /// <returns><see cref="JwtSecurityToken"/> used to call the web API.</returns>
         internal static JwtSecurityToken? GetTokenUsedToCallWebAPI(this HttpContext httpContext)
         {
-            return httpContext.Items[Constants.JwtSecurityTokenUsedToCallWebApi] as JwtSecurityToken;
+            // lock due to https://docs.microsoft.com/en-us/aspnet/core/performance/performance-best-practices?#do-not-access-httpcontext-from-multiple-threads
+            lock(httpContext)
```
henrik-me, line 34: I don't see any big harm in doing this, however I also don't think this is needed. as the write only happens once during validation of the incoming token. and reads only happens in the controllers which happens after the token validation. I agree there can be multiple reads happening concurrently.

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -144,12 +151,12 @@ internal class TokenAcquisition : ITokenAcquisitionInternal
 
             try
             {
-                _application = await GetOrBuildConfidentialClientApplicationAsync().ConfigureAwait(false);
+                var application = await GetOrBuildConfidentialClientApplicationAsync().ConfigureAwait(false);
```
henrik-me, line 154: >var application = await GetOrBuildConfidentialClientApplicationAsync().ConfigureAwait(false); [](start = 16, length = 93)

I think it makes sense to create a new CC object every time

jmprieur: TokenAcquisition is a scoped service. We don't want to recreate it several times per request?

src/Microsoft.Identity.Web/AppServicesAuth/AppServicesAuthenticationTokenAcquisition.cs
```diff
@@ -87,16 +106,29 @@ private async Task<IConfidentialClientApplication> GetOrCreateApplication()
...
+            string accessToken;
+            if (httpContext != null)
+            {
+                // Need to lock due to https://docs.microsoft.com/en-us/aspnet/core/performance/performance-best-practices?#do-not-access-httpcontext-from-multiple-threads
+                lock (httpContext)
+                {
+                    accessToken = GetAccessToken(httpContext.Request.Headers);
+                }
+            }
+            else
```
bgavrilMS, line 126: nit: just initialize the accessToken var with string.emtpy to avoid having this "else" clause.

src/Microsoft.Identity.Web/AppServicesAuth/AppServicesAuthenticationTokenAcquisition.cs
```diff
@@ -87,16 +106,29 @@ private async Task<IConfidentialClientApplication> GetOrCreateApplication()
...
-        public async Task<string> GetAccessTokenForUserAsync(
+        public Task<string> GetAccessTokenForUserAsync(
             IEnumerable<string> scopes,
             string? tenantId = null,
             string? userFlow = null,
             ClaimsPrincipal? user = null,
             TokenAcquisitionOptions? tokenAcquisitionOptions = null)
         {
-            string accessToken = GetAccessToken(CurrentHttpContext?.Request.Headers);
+            var httpContext = CurrentHttpContext;
```
bgavrilMS, line 116: nit: do you need this extra variable? Can't you just lock on the CurrentHttpContext?

src/Microsoft.Identity.Web/AppServicesAuth/AppServicesAuthenticationTokenAcquisition.cs
```diff
@@ -87,16 +106,29 @@ private async Task<IConfidentialClientApplication> GetOrCreateApplication()
...
+            string accessToken;
+            if (httpContext != null)
+            {
+                // Need to lock due to https://docs.microsoft.com/en-us/aspnet/core/performance/performance-best-practices?#do-not-access-httpcontext-from-multiple-threads
+                lock (httpContext)
```
bgavrilMS, line 121: Is this the pattern that should be used throughout the entire library when dealing with HttpContext?

From what I see, if 2 different requests occur at the same time, the HttpContext is different and no locking occurs.

If the same top-level request ends up accessing the Headers twice in this method, then the locking occurs. Is this scenario even possible?

src/Microsoft.Identity.Web/AppServicesAuth/AppServicesAuthenticationTokenAcquisition.cs
```diff
@@ -15,9 +16,15 @@ namespace Microsoft.Identity.Web
     /// <summary>
     /// Implementation of ITokenAcquisition for App Services authentication (EasyAuth).
     /// </summary>
-    public class AppServicesAuthenticationTokenAcquisition : ITokenAcquisition
+    public class AppServicesAuthenticationTokenAcquisition : ITokenAcquisition, IDisposable
```
bgavrilMS, line 19: Although not a breaking change, apps consuming this type will incur a compiler warning that they should dispose of this object properly.

src/Microsoft.Identity.Web/HttpContextExtensions.cs
bgavrilMS: Yes, this is my observation as well. Not sure even if multiple reads would happen concurrently...

src/Microsoft.Identity.Web/TokenAcquisition.cs
bgavrilMS: Async lazy is a bit more complex, see https://blog.stephencleary.com/2012/08/asynchronous-lazy-initialization.html - but ultimately the code will be more readable.

src/Microsoft.Identity.Web/TokenAcquisition.cs
bgavrilMS: Yes. Might be more readable to just lock on the `CurrentHttpContext` variable instead of passing around an HttpContext param.

src/Microsoft.Identity.Web/TokenAcquisition.cs
DavidParks8: Async  lazy will not perform as well as this code, and this code is technically lazy loading. Only the first call takes the hit to load the _application. Subsequent calls will reuse that loaded value. The double checking lock is referring to the atomicity model rather than whether or not the data is lazy.

src/Microsoft.Identity.Web/TokenAcquisition.cs
DavidParks8: The example of what not to do in the docs link in the comment explicitly shows accessing a property on the HttpContext.Request object in parallel. The reason httpContext was passed in was mainly to guarantee that CurrentHttpContext was accessed only once, and to not have to do the null check again. I thought about locking on the httprequest object, but then realized that nearly all locks elsewhere will be done on the httpContext, so the highest chance of reducing threading issues would be to also lock on that.

src/Microsoft.Identity.Web/HttpContextExtensions.cs
DavidParks8: The reason this is needed is because code outside the library could be accessing httpContext.Items at the same time. Application insights may do this, but they are already locking on the httpContext, so if we lock as well, issues can be avoided. I'd like to leave the lock in place as a defensive measure against external code which may corrupt the http context.

src/Microsoft.Identity.Web/AppServicesAuth/AppServicesAuthenticationTokenAcquisition.cs
DavidParks8: It is very unlikely, but still possible that subsequent calls to CurrentHttpContext would yield a different instance. This can happen if someone registers a custom IHttpContextAccessor. For that reason, the httpContext is stored in a var and reused.

src/Microsoft.Identity.Web/AppServicesAuth/AppServicesAuthenticationTokenAcquisition.cs
DavidParks8: According to the linked docs page, it is something we should protect against. The http context is completely not thread safe and neither are the request/response properties of it. Even if this library does not access the httpContext in parallel, external libraries can. I took a look at the HttpContext source code and found that even reading certain fields in parallel can corrupt it due to how it lazy adds HttpFeatures to an internal collection.

src/Microsoft.Identity.Web/AppServicesAuth/AppServicesAuthenticationTokenAcquisition.cs
DavidParks8: The reason it isn't initialized to empty is because doing so creates a chance that the cpu cycles to assign the variable to empty were wasted when it is overwritten. Using the else clause, we guarantee the variable is only assigned once.

src/Microsoft.Identity.Web/HttpContextExtensions.cs
Tratcher: > The reason this is needed is because code outside the library could be accessing httpContext.Items at the same time. Application insights may do this, but they are already locking on the httpContext, so if we lock as well, issues can be avoided. I'd like to leave the lock in place as a defensive measure against external code which may corrupt the http context.

This assumes the other code is also locking on the HttpContext instance rather than using some other synchronization method like SimaphoreSlim like you do elsewhere. All multi-threading code needs to have an explicit contract for how that process is made safe. In practice you can prepare for concurrent executions of your own APIs, but not somebody else's. I'd remove the lock here.

We've made a similar point to ApplicationInsights and they've gotten better about pre-fetching fields they need when they have definitive control of the context.

src/Microsoft.Identity.Web/TokenAcquisition.cs
henrik-me: Yup both will be lazy load, get the good old double locking. Trying to avoid the need to dispose. I don't find async lazy more complex. Both are fine, I just prefer the other one.

src/Microsoft.Identity.Web/TokenAcquisition.cs
henrik-me: depends on what we mean by scoped, my understanding is the current scope is "while handling everything triggered by the incomming call" and now you have multiple threads accessing the same object. MSAL can handle that for these scenarios, however I'm not sure what is best for guaranteeing the proper locks on accessing the cache. It may be either or, but if we include the resource in the obo cache I think the situation will be better if we have a CC object for each concurrent call. Probably worth testing out.

---
In reply to: [600316109](https://github.com/AzureAD/microsoft-identity-web/pull/1082#discussion_r600316109) [](ancestors = 600316109)

src/Microsoft.Identity.Web/HttpContextExtensions.cs
henrik-me: Makse sense. I believe what David suggest for locking on httpcontext has been consistently using the httpcontext for locking, the semaphore is used around creation of MSAL objects. Seems reasonable to me that we are using the HttpContext for locking in these cases.

---
In reply to: [600634668](https://github.com/AzureAD/microsoft-identity-web/pull/1082#discussion_r600634668) [](ancestors = 600634668)

henrik-me:
Would like to see us test this out. I would like though to split out the change done for creating new CC objects from the change on locking.

src/Microsoft.Identity.Web/TokenAcquisition.cs
henrik-me: I don't think Azync Lazy is needed, just Lazy as this is only for object initialization, not for the actual async operations. 

---
In reply to: [600648216](https://github.com/AzureAD/microsoft-identity-web/pull/1082#discussion_r600648216) [](ancestors = 600648216)

DavidParks8: I did some profiling. Anything under 4 parallel initializations will not benefit from lazy loading. Over 4 and a locked implementation starts to show benefit.

src/Microsoft.Identity.Web/AppServicesAuth/AppServicesAuthenticationTokenAcquisition.cs
DavidParks8: Updated to remove IDisposable now that initialization is synchronous.

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -144,12 +143,12 @@ internal class TokenAcquisition : ITokenAcquisitionInternal
 
             try
             {
-                _application = GetOrBuildConfidentialClientApplication();
+                var application = BuildConfidentialClientApplication();
```
jennyf19, line 146: @DavidParks8 The main issue @jmprieur  and I see here is that we could have 100s of tokenAcquisitions in the controller, and we won't want to recreate the cca object each time. Is there a reason why you removed `_application` and create it each time?

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -500,11 +500,7 @@ private IConfidentialClientApplication BuildConfidentialClientApplication()
 
             if (request != null && string.IsNullOrEmpty(currentUri))
             {
-                currentUri = UriHelper.BuildAbsolute(
-                    request.Scheme,
-                    request.Host,
-                    request.PathBase,
-                    _microsoftIdentityOptions.CallbackPath.Value ?? string.Empty);
+                currentUri = BuildCurrentUriFromRequest(httpContext!, request);
```
jennyf19, line 503: so nice. thank you.

henrik-me:
Can we make this change only about httpcontext and take the application objection creation in the other PR or is that other one replacing this? [https://github.com/AzureAD/microsoft-identity-web/pull/1084](https://github.com/AzureAD/microsoft-identity-web/pull/1084)

---
In reply to: [805976220](https://github.com/AzureAD/microsoft-identity-web/pull/1082#issuecomment-805976220) [](ancestors = 805976220)

src/Microsoft.Identity.Web/TokenAcquisition.cs
DavidParks8: Locking has been added back for lazy loading.

jennyf19:
going to merge to another branch and we will fix up anything as we test if needed. thank you so much @DavidParks8

PR #1084, opened 2021-03-24
[do not merge] feedback PR for perf improvements

jennyf19:


src/Microsoft.Identity.Web/AppServicesAuth/AppServicesAuthenticationInformation.cs
```diff
@@ -10,173 +10,173 @@
...
+				throw new ArgumentNullException(nameof(headers));
+			}
 
-            string? idToken = null;
-            if (headers.ContainsKey(AppServicesAuthIdTokenHeader))
-            {
-                idToken = headers[AppServicesAuthIdTokenHeader];
-            }
+			string? idToken = null;
+			if (headers.ContainsKey(AppServicesAuthIdTokenHeader))
```
DavidParks8, line 114: TryGetValue results in a single dictionary lookup rather than two lookups when using ContainsKey with subsequent access.
```suggestion
			if (headers.TryGetValue(AppServicesAuthIdTokenHeader, out var idToken))
```

src/Microsoft.Identity.Web/AppServicesAuth/AppServicesAuthenticationTokenAcquisition.cs
```diff
@@ -13,181 +13,181 @@
...
-            throw new NotImplementedException();
-        }
+		public async Task<AuthenticationResult> GetAuthenticationResultForUserAsync(
+			IEnumerable<string> scopes,
+			string? tenantId = null,
+			string? userFlow = null,
+			ClaimsPrincipal? user = null,
+			TokenAcquisitionOptions? tokenAcquisitionOptions = null)
+		{
+			string? idToken = AppServicesAuthenticationInformation.GetIdToken(CurrentHttpContext?.Request?.Headers!);
```
DavidParks8, line 156: CurrentHttpContext needs to be locked on to avoid threading issues with external code.

src/Microsoft.Identity.Web/AppServicesAuth/AppServicesAuthenticationTokenAcquisition.cs
```diff
@@ -13,181 +13,181 @@
...
+			ClaimsPrincipal? userClaims = AppServicesAuthenticationInformation.GetUser(CurrentHttpContext?.Request?.Headers!);
+			string accessToken = await GetAccessTokenForUserAsync(scopes, tenantId, userFlow, user, tokenAcquisitionOptions).ConfigureAwait(false);
+			string expiration = userClaims.FindFirstValue("exp");
+			DateTimeOffset dateTimeOffset = (expiration != null)
+				? DateTimeOffset.FromUnixTimeSeconds(long.Parse(expiration, CultureInfo.InvariantCulture))
```
DavidParks8, line 161: long.parse could throw if the claim has a broken expiration string. long.TryParse would be safer.

src/Microsoft.Identity.Web/AppServicesAuth/AppServicesAuthenticationTokenAcquisition.cs
```diff
@@ -13,181 +13,181 @@
...
+				: DateTimeOffset.Now;
+
+			AuthenticationResult authenticationResult = new AuthenticationResult(
+				accessToken,
+				isExtendedLifeTimeToken: false,
+				userClaims?.GetDisplayName(),
+				dateTimeOffset,
+				dateTimeOffset,
+				userClaims?.GetTenantId(),
+				userClaims != null ? new Account(userClaims) : null,
```
DavidParks8, line 171: there are three null checks on userClaims here. 
userClaims?.GetDisplayName()
userClaims?.GetTenantId()
userClaims != null

It would be faster to do the check once and set three variables.
string? displayName;
string? tenantId;
Account? account;
if (userClaims != null)
{
   displayName = userClaims.GetDisplayName();
   tenantId = userClaims.GetTenantId();
   account = new Account(userClaims);
}
else
{
   displayName = null;
   tenantId = null;
   account = null;
}

src/Microsoft.Identity.Web/AppServicesAuth/AppServicesAuthenticationTokenAcquisition.cs
```diff
@@ -13,181 +13,181 @@
...
+				idToken,
+				scopes,
+				tokenAcquisitionOptions != null ? tokenAcquisitionOptions.CorrelationId : Guid.Empty);
+			return authenticationResult;
+		}
+
+		/// <inheritdoc/>
+		public async Task ReplyForbiddenWithWwwAuthenticateHeaderAsync(IEnumerable<string> scopes, MsalUiRequiredException msalServiceException, HttpResponse? httpResponse = null)
```
DavidParks8, line 179: the async keyword can be removed when all the function does is throw. Doing so would avoid the need to suppress a warning about async functions needing something awaited.

DavidParks8: Oh wait, I already did that in this file in my other PR.

src/Microsoft.Identity.Web/AuthorityHelpers.cs
```diff
@@ -6,35 +6,35 @@
...
-            {
-                return new Uri(baseUri, new PathString($"{pathBase}/{tenantId}/v2.0")).ToString();
-            }
-        }
+			if (options.IsB2C)
+			{
+				var userFlow = options.DefaultUserFlow;
+				return new Uri(baseUri, new PathString($"{pathBase}/{domain}/{userFlow}/v2.0")).ToString();
+			}
+			else
```
DavidParks8, line 23: This doesn't really matter that much, but the else keyword and block nesting isn't needed because the `if (options.IsB2C)` block had a return statement.

src/Microsoft.Identity.Web/AuthorityHelpers.cs
```diff
@@ -6,35 +6,35 @@
...
-        {
-            authority = authority.Trim().TrimEnd('/');
-            if (!authority.EndsWith("v2.0", StringComparison.InvariantCulture))
-            {
-                authority += "/v2.0";
-            }
+		internal static string EnsureAuthorityIsV2(string authority)
+		{
+			authority = authority.Trim().TrimEnd('/');
+			if (!authority.EndsWith("v2.0", StringComparison.InvariantCulture))
```
DavidParks8, line 32: For strings that are language and culture agnostic, StringComparison.Ordinal should be used. It is faster and more targeted to the usecase here: https://docs.microsoft.com/en-us/dotnet/standard/base-types/best-practices-strings

src/Microsoft.Identity.Web/AuthorityHelpers.cs
```diff
@@ -6,35 +6,35 @@
...
-        internal static string EnsureAuthorityIsV2(string authority)
-        {
-            authority = authority.Trim().TrimEnd('/');
-            if (!authority.EndsWith("v2.0", StringComparison.InvariantCulture))
-            {
-                authority += "/v2.0";
-            }
+		internal static string EnsureAuthorityIsV2(string authority)
+		{
+			authority = authority.Trim().TrimEnd('/');
```
DavidParks8, line 31: String allocations tend to be a death by 1000 cuts issue at scale. Instead of TrimEnd('/'), I suggest adding "v2.0" or "/v2.0" based upon the presence of the trailing slash. Doing so should skip one of the string reallocations.

src/Microsoft.Identity.Web/AuthorizeForScopesAttribute.cs
```diff
@@ -14,152 +14,152 @@
...
+				if (msalUiRequiredException != null &&
+					IncrementalConsentAndConditionalAccessHelper.CanBeSolvedByReSignInOfUser(msalUiRequiredException))
+				{
+					// the users cannot provide both scopes and ScopeKeySection at the same time
+					if (!string.IsNullOrWhiteSpace(ScopeKeySection) && Scopes != null && Scopes.Length > 0)
+					{
+						throw new InvalidOperationException(
+							string.Format(
+								CultureInfo.InvariantCulture,
+								IDWebErrorMessage.ProvideEitherScopeKeySectionOrScopes,
```
DavidParks8, line 69: The error messages should really be localized in order to promote the most inclusive api experiences.

DavidParks8: Once localized, InvariantCulture is probably not the correct culture to use.

src/Microsoft.Identity.Web/AuthorizeForScopesAttribute.cs
```diff
@@ -14,152 +14,152 @@
...
+					}
+
+					// Do not re-use the property Scopes. For more info: https://github.com/Azure-Samples/active-directory-aspnetcore-webapp-openidconnect-v2/issues/273
+					string[]? incrementalConsentScopes;
+
+					// If the user wishes us to pick the Scopes from a particular config setting.
+					if (!string.IsNullOrWhiteSpace(ScopeKeySection))
+					{
+						// Load the injected IConfiguration
+						IConfiguration configuration = context.HttpContext.RequestServices.GetRequiredService<IConfiguration>();
```
DavidParks8, line 81: GetRequiredService will throw if the service isn't registered. I recommend changing it to GetService to avoid an unhelpful error.

src/Microsoft.Identity.Web/AzureADB2COpenIDConnectEventHandlers.cs
```diff
@@ -10,103 +10,103 @@
...
+				context.ProtocolMessage.IssuerAddress = BuildIssuerAddress(context, defaultUserFlow, userFlow);
+				context.Properties.Items.Remove(OidcConstants.PolicyKey);
+
+				if (!Options.HasClientCredentials)
+				{
+					context.ProtocolMessage.ResponseType = OpenIdConnectResponseType.IdToken;
+				}
+				else
+				{
+					if (Options.IsB2C)
```
DavidParks8, line 50: This if block can be less nested by combining towards "else if" which will slightly improve readability.

src/Microsoft.Identity.Web/AzureADB2COpenIDConnectEventHandlers.cs
```diff
@@ -10,103 +10,103 @@
...
+			context.HandleResponse();
+
+			// Handle the error code that Azure Active Directory B2C throws when trying to reset a password from the login page
+			// because password reset is not supported by a "sign-up or sign-in user flow".
+			// Below is a sample error message:
+			// 'access_denied', error_description: 'AADB2C90118: The user has forgotten their password.
+			// Correlation ID: f99deff4-f43b-43cc-b4e7-36141dbaf0a0
+			// Timestamp: 2018-03-05 02:49:35Z
+			// ', error_uri: 'error_uri is null'.
+			if (context.Failure is OpenIdConnectProtocolException && context.Failure.Message.Contains(ErrorCodes.B2CForgottenPassword))
```
DavidParks8, line 75: context.Failure is OpenIdConnectProtocolException can be evaluated once rather than re-evaluated in multiple if statements.

src/Microsoft.Identity.Web/AzureADB2COpenIDConnectEventHandlers.cs
```diff
@@ -10,103 +10,103 @@
...
+		private string BuildIssuerAddress(RedirectContext context, string? defaultUserFlow, string userFlow)
+		{
+			if (!_userFlowToIssuerAddress.TryGetValue(userFlow, out var issuerAddress))
+			{
+				_userFlowToIssuerAddress[userFlow] = context.ProtocolMessage.IssuerAddress.ToLowerInvariant()
+					.Replace($"/{defaultUserFlow?.ToLowerInvariant()}/", $"/{userFlow.ToLowerInvariant()}/");
+			}
+
+			return _userFlowToIssuerAddress[userFlow];
+		}
```
DavidParks8, line 110: Doing this will be twice as fast:
```suggestion
                        if (!_userFlowToIssuerAddress.TryGetValue(userFlow, out var issuerAddress))
			{
			        issuerAddress = context.ProtocolMessage.IssuerAddress.ToLowerInvariant()
					.Replace($"/{defaultUserFlow?.ToLowerInvariant()}/", $"/{userFlow.ToLowerInvariant()}/");
				_userFlowToIssuerAddress[userFlow] = issuerAddress;
			}

			return issuerAddress;
```

src/Microsoft.Identity.Web/AzureFunctionsAuthenticationHttpContextExtension.cs
```diff
@@ -9,42 +9,42 @@
...
-            }
-        }
-    }
+				return (true, null);
+			}
+			else
+			{
+				return (false, new UnauthorizedObjectResult(new ProblemDetails
+				{
+					Title = "Authorization failed.",
```
DavidParks8, line 44: The title string needs to be localized.

src/Microsoft.Identity.Web/Base64UrlHelpers.cs
```diff
@@ -7,80 +7,80 @@
...
+	{
+		private const char Base64PadCharacter = '=';
+		private const char Base64Character62 = '+';
+		private const char Base64Character63 = '/';
+		private const char Base64UrlCharacter62 = '-';
+		private const char Base64UrlCharacter63 = '_';
+		private static readonly Encoding TextEncoding = Encoding.UTF8;
 
-        private static readonly string DoubleBase64PadCharacter = string.Format(CultureInfo.InvariantCulture, "{0}{0}", Base64PadCharacter);
+		private static readonly string DoubleBase64PadCharacter = string.Format(CultureInfo.InvariantCulture, "{0}{0}", Base64PadCharacter);
```
DavidParks8, line 19: This is a good example of where string.Format isn't needed
```suggestion
		private const string DoubleBase64PadCharacter = Base64PadCharacter + Base64PadCharacter;
```

src/Microsoft.Identity.Web/CertificateManagement/DefaultCertificateLoader.cs
```diff
@@ -13,265 +13,265 @@
...
+	///		   .Build();
+	/// </code></example>
+	/// </summary>
+	public class DefaultCertificateLoader : ICertificateLoader
+	{
+		/// <summary>
+		/// User assigned managed identity client ID (as opposed to system assigned managed identity)
+		/// See https://docs.microsoft.com/azure/active-directory/managed-identities-azure-resources/how-to-manage-ua-identity-portal.
+		/// </summary>
+		public static string? UserAssignedManagedIdentityClientId { get; set; }
```
DavidParks8, line 37: Having this be static and externally settable opens up race conditions where someone thinks it is populated when it may not be, and then makes incorrect decisions based upon that. Saying it is not used that way now is not sufficient. This is a high risk of misuse for future contributors. Recommendation is to make it an instance field on a new interface that is not settable externally, then access it through dependency injection.

src/Microsoft.Identity.Web/ClaimsPrincipalFactory.cs
```diff
@@ -5,42 +5,42 @@
...
+		/// <code>
+		/// private async Task GetChangedMessagesAsync(IEnumerable&lt;Notification&gt; notifications)
+		/// {
+		///  foreach (var notification in notifications)
+		///  {
+		///   SubscriptionStore subscription =
+		///		   subscriptionStore.GetSubscriptionInfo(notification.SubscriptionId);
+		///  HttpContext.User = ClaimsPrincipalExtension.FromTenantIdAndObjectId(subscription.TenantId,
+		///																	  subscription.UserId);
+		///  string accessToken = await tokenAcquisition.GetAccessTokenForUserAsync(scopes);
```
DavidParks8, line 33: This sample looks a bit broken with missing end brackets and overwriting of HttpContext.User with each iteration.

src/Microsoft.Identity.Web/ClientInfo.cs
```diff
@@ -7,37 +7,37 @@
...
-                PropertyNameCaseInsensitive = true,
-            };
-
-            return JsonSerializer.Deserialize<ClientInfo>(jsonByteArray, options);
-        }
-    }
+	internal class ClientInfo
+	{
+		[JsonPropertyName(ClaimConstants.UniqueObjectIdentifier)]
+		public string UniqueObjectIdentifier { get; set; } = null!;
```
DavidParks8, line 13: setting to null! seems like we are incorrectly trying to override the type. The reality is that json can omit the field which would result in a null string.

src/Microsoft.Identity.Web/ClaimsPrincipalExtensions.cs
```diff
@@ -6,215 +6,215 @@
...
+		}
+
+		/// <summary>
+		/// Gets the user flow ID associated with the <see cref="ClaimsPrincipal"/>.
+		/// </summary>
+		/// <param name="claimsPrincipal">The <see cref="ClaimsPrincipal"/> from which to retrieve the user flow ID.</param>
+		/// <returns>User flow ID of the identity, or <c>null</c> if it cannot be found.</returns>
+		public static string? GetUserFlowId(this ClaimsPrincipal claimsPrincipal)
+		{
+			if (claimsPrincipal == null)
```
DavidParks8, line 154: All of these extension methods could be deduplicated into calling a single private shared method:

```cs
private static string? GetClaimValue(ClaimsPrincipal? claimsPrincipal, params string[] claimNames)
{
    if (claimsPrincipal == null)
    {
        throw new ArgumentNullException(nameof(claimsPrincipal));
    }

    for (var i = 0; i < claimNames.Length; i++)
    {
        var currentValue = claimsPrincipal.FindFirstValue(claimNames[i]);
        if (!string.IsNullOrEmpty(currentValue))
        {
            return currentValue;
        }
    }

    return null;
}
```

src/Microsoft.Identity.Web/Constants/Constants.cs
```diff
@@ -3,127 +3,127 @@
...
-        // Metadata
-        internal const string PreferredNetwork = "preferred_network";
-        internal const string PreferredCache = "preferred_cache";
-        internal const string Aliases = "aliases";
-
-        // AadIssuerValidator
+	/// <summary>
+	/// General constants for Microsoft Identity Web.
+	/// </summary>
+	public static class Constants
```
DavidParks8, line 9: There are many constants throughout the codebase. It may be worthwhile to rename this class to be more descriptive about the specific constants it holds.

src/Microsoft.Identity.Web/Constants/LogMessages.cs
```diff
@@ -3,25 +3,25 @@
...
+		public const string MethodBegin = "Begin {0}. ";
+		public const string MethodEnd = "End {0}. ";
 
-        // Caching
-        public const string DeserializingSessionCache = "Deserializing session {0}, cache key {1}. ";
-        public const string SessionCacheKeyNotFound = "Cache key {0} not found in session {1}. ";
-        public const string SerializingSessionCache = "Serializing session {0}, cache key {1}. ";
-        public const string ClearingSessionCache = "Clearing session {0}, cache key {1}. ";
+		// Caching
+		public const string DeserializingSessionCache = "Deserializing session {0}, cache key {1}. ";
```
DavidParks8, line 20: Please consider using LoggerMessage as described at https://docs.microsoft.com/en-us/dotnet/core/extensions/high-performance-logging instead of doing costly message formats into strings.

src/Microsoft.Identity.Web/CookiePolicyOptionsExtensions.cs
```diff
@@ -9,175 +9,175 @@
...
+				}
+
+				return IsChromiumBased() &&
+					IsChromiumVersionAtLeast(51) &&
+					!IsChromiumVersionAtLeast(67);
+			}
+
+			bool IsIosVersion(int major)
+			{
+				string regex = @"\(iP.+; CPU .*OS (\d+)[_\d]*.*\) AppleWebKit\/";
```
DavidParks8, line 104: This code is in the hotpath of requests. As such, it would likely speed up by 5-7x if regular expressions were avoided where possible. We did a similar change in a planet scale service and sped the code up by 7x.

src/Microsoft.Identity.Web/DownstreamWebApiSupport/DownstreamWebApi.cs
```diff
@@ -13,208 +13,208 @@
...
+		public DownstreamWebApi(
+			ITokenAcquisition tokenAcquisition,
+			IOptionsMonitor<DownstreamWebApiOptions> namedDownstreamWebApiOptions,
+			HttpClient httpClient,
+			IOptions<MicrosoftIdentityOptions> microsoftIdentityOptions)
+		{
+			_tokenAcquisition = tokenAcquisition;
+			_namedDownstreamWebApiOptions = namedDownstreamWebApiOptions;
+			_httpClient = httpClient;
 #pragma warning disable CA1062 // Validate arguments of public methods
```
DavidParks8, line 42: This warning suppression might be able to be removed now that strict null types is enabled.

src/Microsoft.Identity.Web/DownstreamWebApiSupport/DownstreamWebApi.cs
```diff
@@ -13,208 +13,208 @@
...
+			{
+				if (content != null)
+				{
+					httpRequestMessage.Content = content;
+				}
+
+				httpRequestMessage.Headers.Add(
+					Constants.Authorization,
+					authResult.CreateAuthorizationHeader());
+				response = await _httpClient.SendAsync(httpRequestMessage).ConfigureAwait(false);
```
DavidParks8, line 96: the response var isn't used outside the scope of the using block. It can be removed and replaced by a direct return statement in front of the awaited call.

src/Microsoft.Identity.Web/DownstreamWebApiSupport/DownstreamWebApi.cs
```diff
@@ -13,208 +13,208 @@
...
+			ClaimsPrincipal? user = null)
+			where TOutput : class
+		{
+			HttpResponseMessage response = await CallWebApiForUserAsync(
+				serviceName,
+				downstreamWebApiOptionsOverride,
+				user,
+				new StringContent(JsonSerializer.Serialize(input), Encoding.UTF8, "application/json")).ConfigureAwait(false);
+
+			try
```
DavidParks8, line 116: Using exceptions for control flow is a very inefficient way of doing things. I recommend instead checking if (!response.IsSuccessStatusCode) and only then throwing the  HttpRequestException.

src/Microsoft.Identity.Web/DownstreamWebApiSupport/DownstreamWebApiGenericExtensions.cs
```diff
@@ -10,305 +10,305 @@
...
-            using StringContent? input = ConvertFromInput(inputData);
+			using StringContent? input = ConvertFromInput(inputData);
 
-            await downstreamWebApi.CallWebApiForUserAsync(
-              serviceName,
-              PrepareOptions(relativePath, downstreamWebApiOptionsOverride, HttpMethod.Put),
-              user,
-              input).ConfigureAwait(false);
-        }
+			await downstreamWebApi.CallWebApiForUserAsync(
```
DavidParks8, line 132: This can be returned directly without the await keyword.

src/Microsoft.Identity.Web/DownstreamWebApiSupport/DownstreamWebApiGenericExtensions.cs
```diff
@@ -10,305 +10,305 @@
...
+			 input).ConfigureAwait(false);
+		}
 
-        private static StringContent ConvertFromInput<TInput>(TInput input)
-        {
-            return new StringContent(JsonSerializer.Serialize(input), Encoding.UTF8, "application/json");
-        }
+		private static StringContent ConvertFromInput<TInput>(TInput input)
+		{
+			return new StringContent(JsonSerializer.Serialize(input), Encoding.UTF8, "application/json");
```
DavidParks8, line 259: Assuming the input is intended to be default json is something which disallows calling downstream apis using other formats such as msgpack. Msgpack is a popular format used within internal containerized servers that need more performance when communicating amongst each other.

src/Microsoft.Identity.Web/DownstreamWebApiSupport/DownstreamWebApiOptions.cs
```diff
@@ -5,97 +5,97 @@
...
-    {
-        /// <summary>
-        /// Base URL for the called downstream web API. For instance <c>"https://graph.microsoft.com/beta/".</c>.
-        /// </summary>
-        public string BaseUrl { get; set; } = Constants.GraphBaseUrlV1;
+	/// <summary>
+	/// Options passed-in to call downstream web APIs. To call Microsoft Graph, see rather
+	/// <c>MicrosoftGraphOptions</c> in the <c>Microsoft.Identity.Web.MicrosoftGraph</c> assembly.
+	/// </summary>
+	public class DownstreamWebApiOptions
```
DavidParks8, line 12: This is clonable, so why not have it implement ICloneable?

src/Microsoft.Identity.Web/DownstreamWebApiSupport/IDownstreamWebApi.cs
```diff
@@ -8,101 +8,101 @@
...
+		/// <param name="user">[Optional] Claims representing a user. This is useful in platforms like Blazor
+		/// or Azure Signal R, where the HttpContext is not available. In other platforms, the library
+		/// will find the user from the HttpContext.</param>
+		/// <returns>The value returned by the downstream web API.</returns>
+		/// <example>
+		/// A list method that returns an IEnumerable&lt;MyItem&gt;&gt;.
+		/// <code>
+		/// public async Task&lt;IEnumerable&lt;MyItem&gt;&gt; GetAsync()
+		/// {
+		///  return await _downstreamWebApi.CallWebApiForUserAsync&lt;object, IEnumerable&lt;MyItem&gt;&gt;(
```
DavidParks8, line 59: to illustrate best practices, the task should be returned instead of awaited in the example.

src/Microsoft.Identity.Web/Extensions.cs
```diff
@@ -3,27 +3,27 @@
...
+	internal static class Extensions
+	{
+		/// <summary>Determines whether the specified string collection contains any.</summary>
+		/// <param name="searchFor">The search for.</param>
+		/// <param name="stringCollection">The string collection.</param>
+		/// <returns>
+		///   <c>true</c> if the specified string collection contains any; otherwise, <c>false</c>.</returns>
+		public static bool ContainsAny(this string searchFor, params string[] stringCollection)
+		{
+			foreach (string str in stringCollection)
```
DavidParks8, line 18: This can be done faster, but probably shouldn't be. The faster implementation I'm thinking of involves char operations and a trie built from the string collection inputs..

src/Microsoft.Identity.Web/IdHelper.cs
```diff
@@ -8,35 +8,35 @@
...
-           () =>
-           {
+	internal static class IdHelper
+	{
+		private static readonly Lazy<string> s_idWebVersion = new Lazy<string>(
+		   () =>
+		   {
 #pragma warning disable CS8600 // Converting null literal or possible null value to non-nullable type.
-               string fullVersion = typeof(IdHelper).GetTypeInfo().Assembly.FullName;
+			   string fullVersion = typeof(IdHelper).GetTypeInfo().Assembly.FullName;
```
DavidParks8, line 17: Why is the FullName field parsed rather than using the .Version property of the Assembly object?

src/Microsoft.Identity.Web/IncrementalConsentAndConditionalAccessHelper.cs
```diff
@@ -11,90 +11,90 @@
...
+				 OidcConstants.ScopeOfflineAccess,
+				 OidcConstants.ScopeProfile,
+			};
 
-            properties.SetParameter<ICollection<string>>(
-                OpenIdConnectParameterNames.Scope,
-                scopes.Union(additionalBuiltInScopes).ToList());
+			properties.SetParameter<ICollection<string>>(
+				OpenIdConnectParameterNames.Scope,
+				scopes.Union(additionalBuiltInScopes).ToList());
```
DavidParks8, line 73: LINQ is super slow. Like, so slow, I would recommend people almost never use it when there is a non-verbose alternative. In this case, there is a data structure that perfectly does what we want in an optimized way. That data structure is the HashSet<string> which happens to have a UnionWith method and already implements ICollection<string>.

src/Microsoft.Identity.Web/MicrosoftIdentityCircuitHandler.cs
```diff
@@ -17,257 +17,257 @@
...
 
-    /// <summary>
-    /// Handler for Blazor specific APIs to handle incremental consent
-    /// and conditional access.
-    /// </summary>
+	/// <summary>
+	/// Handler for Blazor specific APIs to handle incremental consent
+	/// and conditional access.
+	/// </summary>
 #pragma warning disable SA1402 // File may only contain a single type
```
DavidParks8, line 68: Can the type be moved to its own file in order to make it easier to find on the github folder view?

src/Microsoft.Identity.Web/MicrosoftIdentityCircuitHandler.cs
```diff
@@ -17,257 +17,257 @@
...
+		/// Base URI to use in forming the redirect.
+		/// </summary>
+		public string? BaseUri
+		{
+			get
+			{
+				return _baseUri ??
 #pragma warning disable CS8602 // Dereference of a possibly null reference. HttpContext will not be null in this case
-                    (!IsBlazorServer ? CreateBaseUri(_httpContextAccessor.HttpContext.Request) :
+					(!IsBlazorServer ? CreateBaseUri(_httpContextAccessor.HttpContext.Request) :
```
DavidParks8, line 120: Everytime HttpContext is touched, it will need somesort of lock in this code.

src/Microsoft.Identity.Web/MicrosoftIdentityCircuitHandler.cs
```diff
@@ -17,257 +17,257 @@
...
-            }
-        }
+				ChallengeUser(
+					scopes.ToArray(),
+					claims,
+					userflow);
+			}
+			else
+			{
+				throw exception;
```
DavidParks8, line 178: This line is a problem because it causes the original stacktrace to be lost. The exception needs to be thrown with `throw;` instead of `throw exception;` which means it can't be done inside this method and instead needs to be done from a catch block directly.

src/Microsoft.Identity.Web/MicrosoftIdentityCircuitHandler.cs
```diff
@@ -17,257 +17,257 @@
...
-            };
+			string[] additionalBuiltInScopes =
+			{
+				 OidcConstants.ScopeOpenId,
+				 OidcConstants.ScopeOfflineAccess,
+				 OidcConstants.ScopeProfile,
+			};
 
-            effectiveScopes = effectiveScopes.Union(additionalBuiltInScopes).ToArray();
+			effectiveScopes = effectiveScopes.Union(additionalBuiltInScopes).ToArray();
```
DavidParks8, line 207: ToArray results in an unnecessary allocation.

src/Microsoft.Identity.Web/MicrosoftIdentityCircuitHandler.cs
```diff
@@ -17,257 +17,257 @@
...
-            else
-            {
+			string redirectUri;
+			if (IsBlazorServer)
+			{
+				redirectUri = NavigationManager.Uri;
+			}
+			else
+			{
 #pragma warning disable CS8602 // Dereference of a possibly null reference. HttpContext will not be null in this case.
```
DavidParks8, line 216: HttpContext can definitely be null here if the library is attempted to be used within a worker service. With that in mind, the warning should be addressed.

src/Microsoft.Identity.Web/MicrosoftIdentityCircuitHandler.cs
```diff
@@ -17,257 +17,257 @@
...
-            Service.BaseUri = Manager.BaseUri.TrimEnd('/');
-            Service.NavigationManager = Manager;
-            await base.OnCircuitOpenedAsync(circuit, cancellationToken).ConfigureAwait(false);
-        }
-    }
+		public override async Task OnCircuitOpenedAsync(Circuit circuit, CancellationToken cancellationToken)
+		{
+			var state = await Provider.GetAuthenticationStateAsync().ConfigureAwait(false);
+			Service.User = state.User;
+			Service.IsBlazorServer = true;
```
DavidParks8, line 267: IsBlazorServer might not be correct or am I reading this wrong?

src/Microsoft.Identity.Web/Resource/AadIssuerValidator.cs
```diff
@@ -10,198 +10,198 @@
...
+				return _issuerAliases.Contains(issuerFromTemplateUri.Authority) &&
+					   // "iss" authority is in the aliases
+					   _issuerAliases.Contains(actualIssuerUri.Authority) &&
+					  // Template authority ends in the tenant ID
+					  IsValidTidInLocalPath(tenantId, issuerFromTemplateUri) &&
+					  // "iss" ends in the tenant ID
+					  IsValidTidInLocalPath(tenantId, actualIssuerUri);
+			}
+			catch
+			{
```
DavidParks8, line 112: faults should minimally be logged.

src/Microsoft.Identity.Web/Resource/AadIssuerValidator.cs
```diff
@@ -10,198 +10,198 @@
...
+				if (jwtSecurityToken.Payload.TryGetValue(ClaimConstants.Tid, out object? tid))
+				{
+					return (string)tid;
+				}
+
+				jwtSecurityToken.Payload.TryGetValue(ClaimConstants.TenantId, out object? tenantId);
+				if (tenantId != null)
+				{
+					return (string)tenantId;
```
DavidParks8, line 141: in case someone is trying to break things, the cast would be safer as `tenantId as string;`

src/Microsoft.Identity.Web/Resource/JwtBearerMiddlewareDiagnostics.cs
```diff
@@ -9,99 +9,99 @@
...
+		/// <param name="logger">Logger.</param>
+		public JwtBearerMiddlewareDiagnostics(ILogger<JwtBearerMiddlewareDiagnostics> logger)
+		{
+			_logger = logger;
+		}
+
+		/// <summary>
+		/// Invoked if exceptions are thrown during request processing. The exceptions will be re-thrown after this event unless suppressed.
+		/// </summary>
+		private Func<AuthenticationFailedContext, Task> s_onAuthenticationFailed = null!;
```
DavidParks8, line 32: This naming convention doesn't match the rest of the codebase. Historically, s_ meant static, which makes this naming a bit misleading.

src/Microsoft.Identity.Web/Resource/MicrosoftIdentityIssuerValidatorFactory.cs
```diff
@@ -12,74 +12,74 @@
...
 
-            // Add issuer aliases of the chosen authority to the cache
-            IEnumerable<string> aliases = issuerMetadata.Metadata
-                .Where(m => m.Aliases.Any(a => string.Equals(a, authorityHost, StringComparison.OrdinalIgnoreCase)))
-                .SelectMany(m => m.Aliases)
-                .Append(authorityHost) // For B2C scenarios, the alias will be the authority itself
-                .Distinct();
-            _issuerValidators[authorityHost] = new AadIssuerValidator(aliases);
+			// Add issuer aliases of the chosen authority to the cache
+			IEnumerable<string> aliases = issuerMetadata.Metadata
```
DavidParks8, line 75: This could use the unLINQ treatment.

src/Microsoft.Identity.Web/Resource/OpenIdConnectMiddlewareDiagnostics.cs
```diff
@@ -10,206 +10,206 @@
...
+			_logger.LogDebug(string.Format(CultureInfo.InvariantCulture, LogMessages.MethodBegin, nameof(OnRedirectToIdentityProviderAsync)));
+
+			await s_onRedirectToIdentityProvider(context).ConfigureAwait(false);
+
+			_logger.LogDebug("   Sending OpenIdConnect message:");
+			DisplayProtocolMessage(context.ProtocolMessage);
+			_logger.LogDebug(string.Format(CultureInfo.InvariantCulture, LogMessages.MethodEnd, nameof(OnRedirectToIdentityProviderAsync)));
+		}
+
+		private void DisplayProtocolMessage(OpenIdConnectMessage message)
```
DavidParks8, line 138: Can this please be ifdef'd out of existence? The reflection used in this method makes it very slow.

src/Microsoft.Identity.Web/Resource/RequiredScopeFilter.cs
```diff
@@ -13,108 +13,108 @@
...
 
-        private void ValidateEffectiveScopes(AuthorizationFilterContext context)
-        {
-            if (_effectiveAcceptedScopes == null || !_effectiveAcceptedScopes.Any())
-            {
-                throw new InvalidOperationException(IDWebErrorMessage.MissingRequiredScopesForAuthorizationFilter);
-            }
+		private void ValidateEffectiveScopes(AuthorizationFilterContext context)
+		{
+			if (_effectiveAcceptedScopes == null || !_effectiveAcceptedScopes.Any())
```
DavidParks8, line 53: .Any() is much slower than string[].Length > 0;

src/Microsoft.Identity.Web/Resource/RequiredScopeFilter.cs
```diff
@@ -13,108 +13,108 @@
...
+				if (scopeClaim == null || !scopeClaim.Value.Split(' ').Intersect(_effectiveAcceptedScopes).Any())
+				{
+					context.HttpContext.Response.StatusCode = (int)HttpStatusCode.Forbidden;
+					string message = string.Format(
+						CultureInfo.InvariantCulture,
+						IDWebErrorMessage.MissingScopes,
+						string.Join(",", _effectiveAcceptedScopes));
+					context.HttpContext.Response.WriteAsync(message);
+					context.HttpContext.Response.CompleteAsync();
+					throw new UnauthorizedAccessException(message);
```
DavidParks8, line 83: Why does an exception need to be thrown when the response was already completed in the line above?

src/Microsoft.Identity.Web/Resource/RolesRequiredHttpContextExtensions.cs
```diff
@@ -11,52 +11,52 @@
...
+	{
+		/// <summary>
+		/// When applied to an <see cref="HttpContext"/>, verifies that the application
+		/// has the expected roles.
+		/// </summary>
+		/// <param name="context">HttpContext (from the controller).</param>
+		/// <param name="acceptedRoles">Roles accepted by this web API.</param>
+		/// <remarks>When the roles don't match, the response is a 403 (Forbidden),
+		/// because the app does not have the expected roles.</remarks>
+		public static void ValidateAppRole(this HttpContext context, params string[] acceptedRoles)
```
DavidParks8, line 28: httpContext needs locking to avoid concurrent reads with external code.

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
```diff
@@ -10,210 +10,210 @@
...
+		/// (account or app).</returns>
+		protected override async Task<byte[]> ReadCacheBytesAsync(string cacheKey)
+		{
+			// check memory cache first
+			byte[]? result = (byte[])_memoryCache.Get(cacheKey);
+			_logger.LogDebug($"[MsIdWeb] MemoryCache: Read {cacheKey} cache size: {result?.Length ?? 0}. ");
+
+			if (result == null)
+			{
+				var measure = await Task.Run(async () =>
```
DavidParks8, line 97: Using Task.Run is usually something that shouldn't be done in an already async function for perf/context switching reasons. It looks like it is only done to measure things. I suggest removing task.run and measuring with StopWatch.StartNew.

src/Microsoft.Identity.Web/TokenCacheProviders/Session/MsalSessionTokenCacheProvider.cs
```diff
@@ -9,119 +9,119 @@
...
+			}
+			finally
+			{
+				_sessionLock.ExitWriteLock();
+			}
+		}
 
-        private readonly ReaderWriterLockSlim _sessionLock = new ReaderWriterLockSlim(LockRecursionPolicy.NoRecursion);
-    }
+		private readonly ReaderWriterLockSlim _sessionLock = new ReaderWriterLockSlim(LockRecursionPolicy.NoRecursion);
```
DavidParks8, line 125: ReaderWriterLockSlim is disposable. Thus MsalSessionTokenCacheProvider should also be disposable.

src/Microsoft.Identity.Web/WebApiExtensions/MicrosoftIdentityWebApiAuthenticationBuilderExtensions.cs
```diff
@@ -16,230 +16,230 @@
...
+					{
+						options.Events = new JwtBearerEvents();
+					}
+
+					// When an access token for our own web API is validated, we add it to MSAL.NET's cache so that it can
+					// be used from the controllers.
+					var tokenValidatedHandler = options.Events.OnTokenValidated;
+					options.Events.OnTokenValidated = async context =>
+					{
+						if (!microsoftIdentityOptions.AllowWebApiToBeAuthorizedByACL && !context!.Principal!.Claims.Any(x => x.Type == ClaimConstants.Scope)
```
DavidParks8, line 225: rather than iterating the claims 4 times, it can all be done in one .Any call.

src/Microsoft.Identity.Web/AzureADB2COpenIDConnectEventHandlers.cs
jennyf19: lol...i had fixed that when doing the ContainsKey fixes.

src/Microsoft.Identity.Web/CertificateManagement/DefaultCertificateLoader.cs
jennyf19: will raise an issue for this.

src/Microsoft.Identity.Web/ClaimsPrincipalFactory.cs
jennyf19: great eye.

src/Microsoft.Identity.Web/CertificateManagement/DefaultCertificateLoader.cs
jennyf19: https://github.com/AzureAD/microsoft-identity-web/issues/1095

src/Microsoft.Identity.Web/MicrosoftIdentityCircuitHandler.cs
jennyf19: https://github.com/AzureAD/microsoft-identity-web/issues/1097

src/Microsoft.Identity.Web/MicrosoftIdentityCircuitHandler.cs
jennyf19: we are given the exception to process here for incremental consent, not sure how we can move this to a try/catch. otherwise, developers would have to do this, and it becomes too complicated.

src/Microsoft.Identity.Web/MicrosoftIdentityCircuitHandler.cs
jennyf19: this is only for blazorwasm scenarios, so we will always have the httpContext.

src/Microsoft.Identity.Web/MicrosoftIdentityCircuitHandler.cs
jennyf19: this is only for blazorwasm...it's a long story.

src/Microsoft.Identity.Web/Resource/OpenIdConnectMiddlewareDiagnostics.cs
jennyf19: developers have to set it explicitly, and i doubt they would use it in release mode. we can make a comment in the wiki, and we could add an if/def too...sometimes we might need the diagnostics in release mode.

src/Microsoft.Identity.Web/Resource/RequiredScopeFilter.cs
jennyf19: so the rest of the controller action is not done. if the scope is not correct, we want to avoid the rest of the controller action from happening. there might be a better way to handle it, will take a look. if you have any suggestions, let us know.

PR #1085, opened 2021-03-24
make cca creation not async

jennyf19:
@DavidParks8 thanks for the suggestion of making cca not async. i think this was done back it was a sample and needed for something else...as things moved to MSAL, we were able to remove that functionality out of id web. @jmprieur

### DIFF
--- a/src/Microsoft.Identity.Web/TokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web/TokenAcquisition.cs
@@ -144,7 +144,7 @@ namespace Microsoft.Identity.Web
 
             try
             {
-                _application = await GetOrBuildConfidentialClientApplicationAsync().ConfigureAwait(false);
+                _application = GetOrBuildConfidentialClientApplication();
 
                 // Do not share the access token with ASP.NET Core otherwise ASP.NET will cache it and will not send the OAuth 2.0 request in
                 // case a further call to AcquireTokenByAuthorizationCodeAsync in the future is required for incremental consent (getting a code requesting more scopes)
@@ -210,7 +210,7 @@ namespace Microsoft.Identity.Web
 
             user = await GetAuthenticatedUserAsync(user).ConfigureAwait(false);
 
-            _application = await GetOrBuildConfidentialClientApplicationAsync().ConfigureAwait(false);
+            _application = GetOrBuildConfidentialClientApplication();
 
             string authority = CreateAuthorityBasedOnTenantIfProvided(_application, tenantId);
 
@@ -263,7 +263,7 @@ namespace Microsoft.Identity.Web
         /// for multi tenant apps or daemons.</param>
         /// <param name="tokenAcquisitionOptions">Options passed-in to create the token acquisition object which calls into MSAL .NET.</param>
         /// <returns>An authentication result for the app itself, based on its scopes.</returns>
-        public async Task<AuthenticationResult> GetAuthenticationResultForAppAsync(
+        public Task<AuthenticationResult> GetAuthenticationResultForAppAsync(
             string scope,
             string? tenant = null,
             TokenAcquisitionOptions? tokenAcquisitionOptions = null)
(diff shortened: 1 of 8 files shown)


src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -421,6 +417,8 @@ public async Task ReplyForbiddenWithWwwAuthenticateHeaderAsync(IEnumerable<strin
             httpResponse.StatusCode = (int)HttpStatusCode.Forbidden;
 
             headers[HeaderNames.WWWAuthenticate] = new StringValues($"{Constants.Bearer} {parameterString}");
+
+            await Task.CompletedTask.ConfigureAwait(false); // we don't want to take a breaking change right now. will be for 2.0
```
DavidParks8, line 421: This is one of those cases where the task should be returned instead of awaited.

src/Microsoft.Identity.Web/TokenCacheProviders/MsalAbstractTokenCacheProvider.cs
```diff
@@ -17,8 +17,7 @@ public abstract class MsalAbstractTokenCacheProvider : IMsalTokenCacheProvider
         /// Initializes the token cache serialization.
         /// </summary>
         /// <param name="tokenCache">Token cache to serialize/deserialize.</param>
-        /// <returns>A <see cref="Task"/> that represents a completed initialization operation.</returns>
-        public Task InitializeAsync(ITokenCache tokenCache)
+        public void Initialize(ITokenCache tokenCache)
```
DavidParks8, line 20: This is technically a breaking change. Is there anything special you need to do to rev the library version?

src/Microsoft.Identity.Web/TokenAcquisition.cs
jennyf19: yes...i talked w/ @jmprieur about this one and we are going to wait until we move to 2.* and take a breaking change. so this will come soon.

henrik-me:
We could start by deprecating and pointing to the updated functions?

henrik-me:
>         private IConfidentialClientApplication? _application;

imo, ahould use the lazy pattern

---
Refers to: src/Microsoft.Identity.Web/TokenAcquisition.cs:36 in 1281902. [](commit_id = 1281902b8991ae9c08c96cb56ac4cadd43bdfb66, deletion_comment = False)

src/Microsoft.Identity.Web/TokenCacheProviders/MsalAbstractTokenCacheProvider.cs
jennyf19: we are going w/henrik's suggestion and doing obsolete attributes so we don't have to take a major version right now.

jennyf19:
yes...good idea, i've fixed that.

---
In reply to: [806008756](https://github.com/AzureAD/microsoft-identity-web/pull/1085#issuecomment-806008756) [](ancestors = 806008756)

jennyf19:
>         private IConfidentialClientApplication? _application;

I'm just fixing this cca async issue, which will enable david to remove some code, i'm not going to work on something from that PR specifically.

---
In reply to: [806012208](https://github.com/AzureAD/microsoft-identity-web/pull/1085#issuecomment-806012208) [](ancestors = 806012208)

---
Refers to: src/Microsoft.Identity.Web/TokenAcquisition.cs:36 in 1281902. [](commit_id = 1281902b8991ae9c08c96cb56ac4cadd43bdfb66, deletion_comment = False)

jmprieur (approved):
LGTM. Thanks @jennyf19 
we'll implement default methods on the interface before releasing (to avoid a breaking change as we've added methods on the interface)

jennyf19:
@jmprieur I'll make a 2.* project board and start a running list of the breaking changes so we don't forget

PR #1088, opened 2021-03-26
initial commit...add support for net462

jennyf19:
@jmprieur i haven't tested manually yet.


jmprieur (approved):
LGTM. Thanks @jennyf19 

Do we want to do a release with just this to help more partners migrating to MSAL?
(and the non-async creation of the CCA)

Or is it better to wait for the performance fixes etc ?

PR #1092, opened 2021-03-26
small clean up in claims principal

jennyf19:
also from david's suggestion


jmprieur (approved):
Thanks @jennyf19 
Nice refactoring!

PR #1094, opened 2021-03-26
fix gov issue again

jennyf19:
one more...that didn't work, but i'll run the release build from here to test

### DIFF
--- a/src/Microsoft.Identity.Web/Microsoft.Identity.Web.csproj
+++ b/src/Microsoft.Identity.Web/Microsoft.Identity.Web.csproj
@@ -94,7 +94,7 @@
     <PackageReference Include="Microsoft.AspNetCore.Authentication.OpenIdConnect" Version="3.1.8" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'net462' ">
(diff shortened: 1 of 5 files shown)


src/Microsoft.Identity.Web/Microsoft.Identity.Web.csproj
```diff
@@ -94,7 +94,7 @@
     <PackageReference Include="Microsoft.AspNetCore.Authentication.OpenIdConnect" Version="3.1.8" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'net462' ">
+  <ItemGroup>
     <PackageReference Include="System.Text.Encodings.Web" Version="4.7.2" />
```
DavidParks8, line 98: This probably shouldn't be included in .net 5 or .net standard, right?

jennyf19: this is for a tsa issue. we got the same issue w/all versions.

jennyf19:
@jmprieur this resolves the tsa issue for now, but need to investigate if there is a better solution. will merge and come back to this.

PR #1098, opened 2021-03-28
second round of suggested changes

jennyf19:


src/Microsoft.Identity.Web/DownstreamWebApiSupport/DownstreamWebApi.cs
```diff
@@ -113,11 +110,7 @@ public class DownstreamWebApi : IDownstreamWebApi
                 user,
                 new StringContent(JsonSerializer.Serialize(input), Encoding.UTF8, "application/json")).ConfigureAwait(false);
 
-            try
-            {
-                response.EnsureSuccessStatusCode();
```
jmprieur, line 118: Can't response.EnsureSuccessStatusCode throw?
it seems it does: 
https://stackoverflow.com/questions/21097730/usage-of-ensuresuccessstatuscode-and-handling-of-httprequestexception-it-throws

src/Microsoft.Identity.Web/Resource/RequiredScopeFilter.cs
```diff
@@ -50,7 +50,7 @@ public void OnAuthorization(AuthorizationFilterContext context)
 
         private void ValidateEffectiveScopes(AuthorizationFilterContext context)
         {
-            if (_effectiveAcceptedScopes == null || !_effectiveAcceptedScopes.Any())
+            if (_effectiveAcceptedScopes == null || _effectiveAcceptedScopes.Length < 0)
```
jmprieur, line 53: probably == 0

jmprieur (commented):
I've 2 questions.

jmprieur (approved):
LGTM
Thanks @jennyf19

src/Microsoft.Identity.Web/DownstreamWebApiSupport/DownstreamWebApi.cs
jennyf19: I reverted back to what we have for now. will take a more indepth look at this later

PR #1104, opened 2021-03-30
implement LoggerMessage for high perf logging

jennyf19:
need to review carefully & test still. creating draft, but feel free to review. - @jmprieur 

For more info see:
[sample](https://github.com/dotnet/AspNetCore.Docs/tree/main/aspnetcore/fundamentals/logging/loggermessage/samples/3.x/LoggerMessageSample)
[asp net engineering guidelines](https://github.com/dotnet/aspnetcore/wiki/Engineering-guidelines#logging-patterns)
[High performance logging doc](https://docs.microsoft.com/en-us/dotnet/core/extensions/high-performance-logging)


src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -637,11 +636,10 @@ private void PrepareAuthorityInstanceForMsal()
...
             catch (MsalUiRequiredException ex)
             {
-                _logger.LogInformation(
-                    string.Format(
-                        CultureInfo.InvariantCulture,
-                        LogMessages.ErrorAcquiringTokenForDownstreamWebApi,
-                        ex.Message));
+                Logger.TokenAcquisitionError(
+                    _logger,
+                    LogMessages.ErrorAcquiringTokenForDownstreamWebApi + ex.Message,
```
jmprieur, line 641: It's probably re-written with the exception, but I guess yes, it does not harm to have the Message as well in the error string

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
```diff
@@ -114,9 +119,9 @@ protected override async Task<byte[]> ReadCacheBytesAsync(string cacheKey)
                         Size = result?.Length,
                     };
 
-                    _logger.LogDebug($"[MsIdWeb] Back propagate from Distributed to Memory, cache size: {result?.Length ?? 0}. ");
+                    Log.BackPropagateL2toL1(_logger, result?.Length ?? 0, null);
```
jmprieur, line 122: Instead of `result?.Length ?? 0` we could use `memoryCacheEntryOptions.Size` as it's already computed

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
```diff
@@ -114,9 +119,9 @@ protected override async Task<byte[]> ReadCacheBytesAsync(string cacheKey)
...
                     _memoryCache.Set(cacheKey, result, memoryCacheEntryOptions);
-                    _logger.LogDebug($"[MsIdWeb] MemoryCache: Count: {_memoryCache.Count}. ");
+                    Log.MemoryCacheCount(_logger, _memoryCacheType, read, _memoryCache.Count, null);
```
jmprieur, line 124: Looking at the logs, when we refresh we display a size of 0.
We could do a bit better by passing to `L2OperationWithRetryOnFailureAsync` below, line 129, the `result` after the cache key, so that the size is displayed.

jmprieur (approved):
LGTM.
Beautiful, and I understand more performant!

- I tested with the redis cache
 - I've left a few suggestions

src/Microsoft.Identity.Web/TokenAcquisition.cs
jennyf19: this is what we were doing before, so tried to keep it the same, but it can be removed.

src/Microsoft.Identity.Web/TokenAcquisition.Logger.cs
```diff
@@ -0,0 +1,34 @@
...
+            /// <summary>
+            /// Logger for handling MSAL exceptions in TokenAcquisition.
+            /// </summary>
+            /// <param name="logger">ILogger.</param>
+            /// <param name="msalErrorMessage">Specific log message from TokenAcquisition.</param>
+            /// <param name="ex">Exception from MSAL.NET.</param>
+            public static void TokenAcquisitionError(
+                ILogger logger,
+                string msalErrorMessage,
+                Exception ex) => s_tokenAcquisitionError(logger, msalErrorMessage, ex);
```
bgavrilMS, line 31: Why do you need an action here? you can just call the `LoggerMessage.Define<string>..` directly

src/Microsoft.Identity.Web/TokenAcquisition.Logger.cs
```diff
@@ -0,0 +1,34 @@
...
+
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Identity.Web
+{
+    /// <summary>
+    /// LoggingMessage class for TokenAcquisition.
+    /// </summary>
+    internal partial class TokenAcquisition
```
bgavrilMS, line 12: Not sure if it's worth splitting this class for Logging.

jennyf19: this is what the docs recommend doing for this type of logging. i've included the links above in the  PR description.

src/Microsoft.Identity.Web/TokenAcquisition.Logger.cs
bgavrilMS: Ok, reading the docs you pointed to :)

jennyf19: following the asp.net core [guidelines](https://github.com/dotnet/aspnetcore/wiki/Engineering-guidelines#logging-patterns), so the log template is compiled once and then we use the delegate to invoke it, when needed.

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.Log.cs
```diff
@@ -0,0 +1,269 @@
...
+{
+    /// <summary>
+    /// LoggingMessage class for MsalDistributedTokenCacheAdapter.
+    /// </summary>
+    public partial class MsalDistributedTokenCacheAdapter
+    {
+        /// <summary>
+        /// LoggingMessage class for MsalDistributedTokenCacheAdapter.
+        /// </summary>
+        private static class Log
```
henrik-me, line 17: >Log [](start = 29, length = 3)

Any reason this is not called logger similar to what is done for tokenacquisition? (same q in the session object)

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.Log.cs
```diff
@@ -0,0 +1,269 @@
...
+            /// <summary>
+            /// Memory cache remove.
+            /// </summary>
+            /// <param name="logger">ILogger.</param>
+            /// <param name="cacheType">Distributed or Memory.</param>
+            /// <param name="operation">Cache operation (Read, Write, etc...).</param>
+            /// <param name="cacheKey">MSAL.NET cache key.</param>
+            /// <param name="cacheSize">Cache size in bytes, or 0 if empty.</param>
+            /// <param name="ex">Exception.</param>
+            public static void MemoryCacheRead(
```
henrik-me, line 82: >MemoryCacheRead [](start = 31, length = 15)

Wondering if these should be next to the definitions instead of separated.

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.Log.cs
```diff
@@ -0,0 +1,269 @@
...
+    {
+        /// <summary>
+        /// LoggingMessage class for MsalDistributedTokenCacheAdapter.
+        /// </summary>
+        private static class Log
+        {
+            private static readonly Action<ILogger, string, string, string, int, bool, Exception?> s_l2CacheState =
+                LoggerMessage.Define<string, string, string, int, bool>(
+                    LogLevel.Debug,
+                    new EventId(2, "DistributedCacheState"),
```
henrik-me, line 22: >2 [](start = 32, length = 1)

All event id's are 2 (or 1 or 3) depending on the class, not sure how maintanable this is across multiple files, it requies search to find all existing used event id's. Perhaps separate class with this definition. (still trying to understand if it's expected to be unique per type or rather unique per message per type).

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.Log.cs
jennyf19: because token aquisition is already has a Log method for the MSAL logs.  i can change them to Logger for consistency.

jennyf19: addressed [here](https://github.com/AzureAD/microsoft-identity-web/pull/1108)

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.Log.cs
jennyf19: addressed [here](https://github.com/AzureAD/microsoft-identity-web/pull/1108) and started discussion in teams.

PR #1106, opened 2021-03-31
remove azuresdk from 462 and 472

jennyf19:


jmprieur (approved):
Good catch.
Thanks @jennyf19

PR #1108, opened 2021-03-31
PR feedback...rename Log -> Logger and use default for EventId

jennyf19:


jennyf19:
or this -> https://github.com/AzureAD/microsoft-identity-web/pull/1111

jmprieur (commented):
I don't understand where default is defined?

jennyf19:
closing, we went w/the defined ids

PR #1109, opened 2021-03-31
Sign Out On B2C Password Reset Flow

Shama-K:
Steps To Reproduce:
1. Run the Sample: [4-2-B2C](https://github.com/Azure-Samples/active-directory-aspnetcore-webapp-openidconnect-v2/tree/master/4-WebApp-your-API/4-2-B2C).
2. Select Reset Password. Once the flow is completed, the user is signed-in automatically.
3. Select Edit Profile. It will take the user to login screen.

With this change: Once password reset flow is completed the user will be redirected to sign-out page.


jennyf19:
thanks @Shama-K is this related to an issue? if not, can you make one w/repro steps? Also, what is the impact w/the [new sign-in/password reset](https://docs.microsoft.com/en-us/azure/active-directory-b2c/add-password-reset-policy?pivots=b2c-user-flow#self-service-password-reset-recommended)? asking because we are working on a 1.9 release, so we could possibly include this, or move it to 1.10...what's the urgency on this? thank you!

Shama-K:
> thanks @Shama-K is this related to an issue? if not, can you make one w/repro steps? Also, what is the impact w/the [new sign-in/password reset](https://docs.microsoft.com/en-us/azure/active-directory-b2c/add-password-reset-policy?pivots=b2c-user-flow#self-service-password-reset-recommended)? asking because we are working on a 1.9 release, so we could possibly include this, or move it to 1.10...what's the urgency on this? thank you!

@jennyf19, I've updated the description with repro steps. Please let me know if more information is required.

jmprieur:
@Shama-K is it because the end user needs to sign-out after resetting password?

Shama-K:
> @Shama-K is it because the end user needs to sign-out after resetting password?

Yes @jmprieur, with current flow user remains signed-in after password reset and that results in inconsistent behavior. For example, edit profile does not work.

jennyf19:
@Shama-K 

> edit profile does not work

what exactly does not work?

Shama-K:
> @Shama-K
> 
> > edit profile does not work
> 
> what exactly does not work?

Click on Edit profile, user needs to sign-in again.
We can put this PR on hold, as we'll test combine SUSI/ResetPwd flow.

derisen:
@jennyf19 @jmprieur the `iss` field in the tokens issued by B2C service contains the name of the policy/user-flow that was used when acquiring them, so for instance if a user signs-in with the sign-up/sign-in user flow, they get a token with `"iss": "login.microsoftonline.com/tfp/b2clogin.com/b2c_1_susi"`. Likewise, when the user signs in automatically after resetting their password, their token looks like `"iss": "login.microsoftonline.com/tfp/b2clogin.com/b2c_1_reset"`. This seems to cause issues like, if the user tries to reset their password in a session after resetting password, B2C service does not recognize session and prompts sign-in again. I'm not sure about the exact mechanism here, whether its a problem with the caching or the B2C service, but in MSAL.js we've been forcing user to sign-out and sign-in again with their new password after the password reset user-flow.

There's now a new combined sign-up/sign-in/password-reset user-flow, which I hadn't had a chance to try yet but it will likely mitigate this issue, so this work might not be needed after all.

jennyf19:
thanks @derisen ...i asked earlier about the new sign-in experience, and that should resolve this issue. the issue is a B2C one, as each policy/userflow, is a separate authorization server, so you're getting tokens from completely different authorization servers, which is why MSAL ends up with multiple tokens in the cache, one per policy/userflow. it also takes awhile for the policies to sync across themselves. have you talked to anyone on the B2C side about this and are you aware of any .NET customer asks for this? 

Also, what's the difference between signing in again when doing edit profile and signing in again immediately after? From a customer perspective, I would rather keep going in the app, and not have to resign in immediately after. The customer may or may not go to another flow to prompt the sign in...just curious here, as this has always been the case with b2c but I don't recall many customer complaints about it, at least for .NET. Is there a way to allow the developer to decide if the user should sign in again immediately or not? Thoughts @jmprieur ?

derisen:
@jennyf19 Ah, ok, actually I remembered now. Indeed the issue is due to multiple tokens in the cache per policy. This was fixed in MSAL.js 1.x with a [workaround](https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/1757), and it resolved switching policies in most cases, but even then we had to sign-out and sign-in the user again (with the sign-up/sign-in policy, not password-reset). The issue [still persists](https://github.com/AzureAD/microsoft-authentication-library-for-js/pull/3237) in MSAL.js 2.x. @tnorling have been carrying the work on multiple policy support, so he should be able to enlighten us.

Shama-K:
As discussed in [another Pull Request](https://github.com/Azure-Samples/active-directory-aspnetcore-webapp-openidconnect-v2/pull/474), sign-out after a password reset is not required.
So, closing this PR.

PR #1111, opened 2021-04-01
add # values to eventIds for logging

jennyf19:


henrik-me (approved):
:shipit:

src/Microsoft.Identity.Web/LoggingEventId.cs
```diff
@@ -0,0 +1,33 @@
...
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Identity.Web
+{
+    /// <summary>
+    /// EventIds for Logging.
+    /// </summary>
+    internal static class LoggingEventId
+    {
+        // DistributedCacheAdapter EventIds 100+
```
henrik-me, line 13: > // DistributedCacheAdapter EventIds 100+ [](start = 7, length = 41)

nit: the other side of the #pragma

jmprieur (approved):
LGTM.
Thanks @jennyf19

src/Microsoft.Identity.Web/LoggingEventId.cs
jennyf19: i'm not sure what you are saying, but my cat has an idea.

PR #1129, opened 2021-04-08
re-add wilson for just .net 5

jennyf19:
tested w/azure functions and the web app calls web API


jmprieur (approved):
LGTM
Thanks @jennyf19 

I did some testing and this seems to work well.

Issue #1137, opened 2021-04-13
ReplyForbiddenWithWwwAuthenticateHeaderAsync raises an obsolete error on v1.9.0


MarcoEnxuto:
Hi, before opening this as a bug, i would like to ask a question before moving forward.
I updated my WebAPI app to this version, and i can't compile because it raises the above error.

Also, the suggestion is a bit vague. What do you mean to "Use initialize instead"? Do you mean in the Startup class?

I took a look on the changes you made recently on the code, but didn't understand what is the path.

Can you clarify more on this? An example? or complete even more this?

The interface has the following.

[Obsolete("IDW10801: Use Initialize instead. See https://aka.ms/ms-id-web/1.9.0. ", true)]
        Task ReplyForbiddenWithWwwAuthenticateHeaderAsync(IEnumerable<string> scopes, MsalUiRequiredException msalServiceException, HttpResponse? httpResponse = null);

Also, if there's another path, your docs should be updated!
Such as this one: https://github.com/AzureAD/microsoft-identity-web/wiki/web-apis#handle-conditional-access

jennyf19:
nice catch, sorry about that. we will do a new release tomorrow (4.14) and have this fixed. thanks for letting us know. @MarcoEnxuto

jmprieur:
Thanks again @MarcoEnxuto for the heads-up.
Microsoft.Identity.Web 1.9.1 just released with this fix

PR #1138, opened 2021-04-14
fix obselete message and make "true" not false.

jennyf19:
#1137


jmprieur (approved):
LGTM
Thanks @jennyf19

Issue #1136, opened 2021-04-13
User.Identity.Name in v.1.9.0 now shows displayname instead of account name


Yomodo:
**Which version of Microsoft Identity Web are you using?**
Microsoft.Identity.Web 1.9.0

Microsoft.Identity.Web 1.8.2: user.Identity.Name contains my account name (UPN): "account@TENANT.onmicrosoft.com"

Microsoft.Identity.Web 1.9.0: user.Identity.Name contains my account display name.

jmprieur:
@Yomodo : do you have repro steps?
- Is it in web apps or web APIs?
- Is it with .NET Core 3.1 or .NET 5.0 or both?

Yomodo:
This is a .NET 5 blazor server web app.

jmprieur:
@Yomodo : could you please check what happens if you force Microsoft.IdentityModel assemblies to be version="6.8.*" ?

```Xlm
    <PackageReference Include="Microsoft.IdentityModel.Logging" Version="6.8.*" />
    <PackageReference Include="Microsoft.IdentityModel.Protocols.OpenIdConnect" Version="6.8.*" />
    <PackageReference Include="System.IdentityModel.Tokens.Jwt" Version="6.8.*" />
```

Yomodo:
These are the used package references:

```
    <PackageReference Include="Humanizer.Core" Version="2.8.26" />
    <PackageReference Include="Microsoft.ApplicationInsights.AspNetCore" Version="2.17.0" />
    <PackageReference Include="Microsoft.Extensions.Logging.AzureAppServices" Version="5.0.5" />
    <PackageReference Include="Microsoft.Identity.Web" Version="1.9.0" />
    <PackageReference Include="Microsoft.Identity.Web.UI" Version="1.9.0" />
    <PackageReference Include="MudBlazor" Version="5.0.7" />
    <PackageReference Include="System.Linq.Async" Version="5.0.0" />
```

jmprieur:
I understand, @Yomodo 
I'm asking if you could add the three package references I provided above to your csproj, in order to see if this still reproes ? I'd like to rule out a change in Microsoft.IdentityModel (as I don't see change in Microsoft.Identity.Web for this)

Yomodo:
1. Removed v1.9.0
2. Added the 3 references
3. Added v 1.8.2
4. Adding v1.9.0 now throws these nuget errors:

Microsoft.Identity.Web 1.9.0:
[...]

PR #1139, opened 2021-04-14
workaround for names claim

jennyf19:
#1136 

thanks for the repro @travisneale ...this workaround should resolve the issue until the next middleware release w/the fix is out.


jmprieur (approved):
LGTM
Thanks @jennyf19

PR #1149, opened 2021-04-19
Fix issue with .NET Core 3.1 project template domain value, not the same as .NET 5.0 / not recognized by msidentity-app-sync

jmprieur:
The ASP.NET 5.0 templates use the following value for the "Domain": "qualified.domain.name"
and this is the value recognized by msidentity-app-sync.

We need to check if the same does not need to be changed in the other web app templates (mvc, webapp, and blazorserver)


jennyf19:
@jmprieur let's merge and we'll test later and make changes as needed. just don't want it sitting around

PR #1150, opened 2021-04-19
fix domain issue for other templates

jennyf19:
I think I found all the discrepancies w/domain, but have not built all the templates yet.


jmprieur (approved):
Thanks @jennyf19 
I have not tested the templates either. 

Do we want to merge your branch into Fix1148, and then test there?

PR #1161, opened 2021-04-26
Value replace by CurrentValue

saddamhossain:
Value does not exist in that palce. CurrentValue working there. Thanks.


jmprieur (commented):
I don't understand this PR, @saddamhossain 
master builds fine? 
Do you have repro steps?

cc: @jennyf19

saddamhossain:
Hi,
Actually it's not working for me in my visual studio and Current Value
works. It's totally fine if master builds correctly.
Thank you so much.

Regards
Saddam Hossain

Issue #1163, opened 2021-04-27
[Bug] IsAppServicesAadAuthenticationEnabled not working with new App Service Authentication flow


B742:
**Which version of Microsoft Identity Web are you using?**
   Microsoft Identity Web 1.9.1 with ASP.NET 5.0

**Where is the issue?**
* Web app
    * [ ] Sign-in users
    * [ ] Sign-in users and call web APIs
* Web API
    * [ ] Protected web APIs (validating tokens)
    * [ ] Protected web APIs (validating scopes)
    * [ ] Protected web APIs call downstream web APIs
* Token cache serialization
     * [ ] In-memory caches
     * [ ] Session caches
     * [ ] Distributed caches
* Other (please describe)
   Identity provider detection

**Is this a new or an existing app?**
   Existing App using Azure App Service and Easy auth. Switched from Authentication (classic) to the new Authentication experience.

**Repro**

Use the following in the Startup.cs:
```csharp
public void ConfigureServices(IServiceCollection services)
{
    ...
    services.AddMicrosoftIdentityWebAppAuthentication(Configuration);
    ...
}
public void Configure(IApplicationBuilder app)
{
    ...
    app.UseAuthentication();
    app.UseAuthorization();
    ...
}
```
Push it to an App Service with the new Authentication experience enabled using AAD. Then try to call the API, you will get an exception saying that you need to provide the ClientId and etc. These values were not needed before the switch to the new experience. 

**Expected behavior**
The library should detect that App Service's Authentication is enabled and not ask for "AzureAd" configuration section.

**Actual behavior**
The ASP.NET 5.0 server throws:
`[Error] Microsoft.AspNetCore.Diagnostics.ExceptionHandlerMiddleware: An exception was thrown attempting to execute the error handler.Microsoft.Extensions.Options.OptionsValidationException: IDW10106: The 'ClientId' option must be provided.at`

It should not.

**Possible solution**
In AppServicesAuthenticationInformation.IsAppServicesAadAuthenticationEnabled, there is 2 conditions to be met.
The first one is still valid when using the new Authentication experience. However, the second one, the string needs to match the variable"AppServicesAuthAzureActiveDirectory" value which is "AzureActiveDirectory". However, using the new experience, the environment variable value is "azureactivedirectory". Since the equal is case-sensitive, it does not work.

**Additional context / logs / screenshots**
When using classic experience:
[...]

PR #1169, opened 2021-04-29
Enable more FxCop analyzers related to string comparison

bgavrilMS:


src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -737,7 +737,8 @@ private void PrepareAuthorityInstanceForMsal()
             {
                 string b2cAuthority = application.Authority.Replace(
                     new Uri(application.Authority).PathAndQuery,
-                    $"/{ClaimConstants.Tfp}/{_microsoftIdentityOptions.Domain}/{userFlow ?? _microsoftIdentityOptions.DefaultUserFlow}");
+                    $"/{ClaimConstants.Tfp}/{_microsoftIdentityOptions.Domain}/{userFlow ?? _microsoftIdentityOptions.DefaultUserFlow}",
```
jmprieur, line 740: what is the point of building an authority which is harder to read when debugging / using Fiddler, given that the STS will understand it anyway?

src/Microsoft.Identity.Web/Extensions.cs
```diff
@@ -17,7 +19,7 @@ public static bool ContainsAny(this string searchFor, params string[] stringColl
         {
             foreach (string str in stringCollection)
             {
-                if (searchFor.Contains(str))
+                if (searchFor.Contains(str, StringComparison.OrdinalIgnoreCase))
```
jmprieur, line 22: do we really want it to be case insensitive in all cases?

jmprieur (commented):
Thanks @bgavrilMS 
there are a couple of questions. I'm not sure why we'd need to be case incentive in some cases?

src/Microsoft.Identity.Web/AzureADB2COpenIDConnectEventHandlers.cs
```diff
@@ -102,8 +102,9 @@ private string BuildIssuerAddress(RedirectContext context, string? defaultUserFl
         {
             if (!_userFlowToIssuerAddress.TryGetValue(userFlow, out var issuerAddress))
             {
-                issuerAddress = context.ProtocolMessage.IssuerAddress.ToLowerInvariant()
-                    .Replace($"/{defaultUserFlow?.ToLowerInvariant()}/", $"/{userFlow.ToLowerInvariant()}/");
+                issuerAddress = context.ProtocolMessage.IssuerAddress
+                    .Replace($"/{defaultUserFlow}/", $"/{userFlow}/", StringComparison.OrdinalIgnoreCase);
```
jennyf19, line 106: would two things with different casing match in the "replace"? i didn't think Replace worked w/casing, so the switch to lower needs to happen before the replace.

src/Microsoft.Identity.Web/Extensions.cs
jennyf19: i'm not sure.

jennyf19:
@bgavrilMS not merging this one right now, we need to do more testing for some of the use cases.

src/Microsoft.Identity.Web/AzureADB2COpenIDConnectEventHandlers.cs
bgavrilMS: I didn't see any unit tests failing. I can take a look at adding some unit tests.

src/Microsoft.Identity.Web/Extensions.cs
bgavrilMS: To avoid bugs like #1163 
The strings we use in our libs are mostly URIs, settings or env variables. We treat them all in a case-insensitive way.

src/Microsoft.Identity.Web/TokenAcquisition.cs
bgavrilMS: I'll have a look.

bgavrilMS: 1. The user could configure an authority with a different casing and the replace will fail. 
2. OrdinalIngnoreCase string operations are about 7 times faster than the case sensitive ones

src/Microsoft.Identity.Web/AzureADB2COpenIDConnectEventHandlers.cs
bgavrilMS: I reverted this logic as I do not understand it / don't have context it in. Generally avoid ToLower, see https://docs.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1308

src/Microsoft.Identity.Web/Extensions.cs
bgavrilMS: https://www.joelonsoftware.com/2003/10/08/the-absolute-minimum-every-software-developer-absolutely-positively-must-know-about-unicode-and-character-sets-no-excuses/

jmprieur (approved):
LGTM
Thanks @bgavrilMS

jennyf19 (approved):

Issue #955, opened 2021-02-11
[Feature Request] Support authentication scheme in all scenarios


jmprieur:
**Which version of Microsoft Identity Web are you using?**
4.16.0

**Repro**

Customers need to be able to implement apps which sign-in users with AAD and B2C, or which are web apps and web APIs. Today, the authentication scheme are supposed to help for this, but they are not taken into account everywhere.

See:
- https://github.com/AzureAD/microsoft-identity-web/issues/173
- https://github.com/AzureAD/microsoft-identity-web/issues/549
- https://github.com/AzureAD/microsoft-identity-web/issues/429
- https://github.com/AzureAD/microsoft-identity-web/issues/958
- https://github.com/AzureAD/microsoft-identity-web/issues/971
- https://github.com/AzureAD/microsoft-identity-web/issues/1126
- https://github.com/AzureAD/microsoft-identity-web/issues/1127
- https://developercommunity2.visualstudio.com/t/I-would-like-to-set-the-AddMicrosoftIden/1339302

This issue might be related: 
- https://github.com/AzureAD/microsoft-identity-web/issues/1164

**Expected behavior**
A fully functional sample of an AAD + B2C web app which also contains web APIS.

**Actual behavior**
See the referenced issues for repro steps

**Possible solution**
See investigation in branch https://github.com/AzureAD/microsoft-identity-web/tree/jmprieur/multipleSchemeInvestigation

jennyf19:
@jmprieur we need to test the confidential client options, i had issues w/the unit tests and had to add `Instance` and `clientSecret`. just adding this as a reminder.

jennyf19:
TODO:
- [ ] Create a wiki page
- [ ] We know PostConfigure() doesn't work - work in progress
- [ ] Check that the regular scenarios still work
- [ ] Run a -preview build

jmprieur:
For Microsoft Graph, I wonder if we need to add a method so that controller actions which are called for a specific auth scheme can pass it, in to the Graph SDK:

`BaseRequestExtensions.WithAuthenticationScheme<T>(this T baseRequest, string authenticationScheme)`

This would have an impact in the Microsoft.Identity.Web.MicrosoftGraph assembly on:
- TokenAcquisitionAuthenticationProviderOption (new AuthenticationScheme property)
- TokenAcquisitionAuthenticationProvider adding an authenticationScheme member, and passing it through to the token acquisition methods (which now have it as a parameter)

[...]

PR #1180, opened 2021-05-06
add scheme to graph call, clean up unused method, add randomness to d…

jennyf19:
…ev ap

@jmprieur i'm trying out some stuff w/the dev app, so it's okay to commit that, as i will probably make more changes once i hear back from asp.net core team on how to handle the sign-out. also, you were right on [this](https://github.com/AzureAD/microsoft-identity-web/issues/955#issuecomment-829424209) LOL 😆 ...so we scratched it a bit too early from the list.

### DIFF
--- a/src/Microsoft.Identity.Web.UI/Areas/MicrosoftIdentity/Controllers/AccountController.cs
+++ b/src/Microsoft.Identity.Web.UI/Areas/MicrosoftIdentity/Controllers/AccountController.cs
@@ -70,7 +70,7 @@ namespace Microsoft.Identity.Web.UI.Areas.MicrosoftIdentity.Controllers
             string policy,
             [FromRoute] string scheme)
         {
-            // scheme ??= OpenIdConnectDefaults.AuthenticationScheme;
+            scheme ??= OpenIdConnectDefaults.AuthenticationScheme;
             Dictionary<string, string?> items = new Dictionary<string, string?>
             {
                 { Constants.Claims, claims },
--- a/src/Microsoft.Identity.Web.UI/Areas/MicrosoftIdentity/Controllers/AccountController.cs
+++ b/src/Microsoft.Identity.Web.UI/Areas/MicrosoftIdentity/Controllers/AccountController.cs
@@ -70,7 +70,7 @@ namespace Microsoft.Identity.Web.UI.Areas.MicrosoftIdentity.Controllers
             string policy,
             [FromRoute] string scheme)
         {
-            //scheme ??= OpenIdConnectDefaults.AuthenticationScheme;
+            // scheme ??= OpenIdConnectDefaults.AuthenticationScheme;
             Dictionary<string, string?> items = new Dictionary<string, string?>
             {
                 { Constants.Claims, claims },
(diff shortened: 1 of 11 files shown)


src/Microsoft.Identity.Web.UI/Areas/MicrosoftIdentity/Controllers/AccountController.cs
```diff
@@ -66,9 +67,10 @@ public IActionResult SignIn([FromRoute] string scheme)
             string loginHint,
             string domainHint,
             string claims,
-            string policy)
+            string policy,
+            [FromRoute] string scheme)
         {
-            string scheme = OpenIdConnectDefaults.AuthenticationScheme;
+            // scheme ??= OpenIdConnectDefaults.AuthenticationScheme;
```
jmprieur, line 73: Good catch. Indeed, I think we'd want to remove this line

tests/MultipleAuthSchemes/Startup.cs
```diff
@@ -44,8 +44,8 @@ public void ConfigureServices(IServiceCollection services)
                     .AddDownstreamWebApi("DownstreamB2CApi", Configuration.GetSection("DownstreamB2CApi"))
                     .AddInMemoryTokenCaches();
 
-            services.Configure<MicrosoftIdentityOptions>(OpenIdConnectDefaults.AuthenticationScheme,
-                                             o => o.ClientSecret = "");
+            //services.Configure<MicrosoftIdentityOptions>(OpenIdConnectDefaults.AuthenticationScheme,
```
jmprieur, line 47: Yes, we want to remove these lines completely :)

jmprieur (approved):
LGTM.
Great work.

Good idea to remove the default scheme and not use OpenIdConnectDefault.AuthenticationScheme in the test app. This is pure now.
and yes :) ... I couldn't figure out how the Graph could work without doing anything when the scheme was not the OpenIdConnectDefault.AuthenticationScheme. ... it didn't :)

tests/MultipleAuthSchemes/Controllers/HomeController.cs
```diff
@@ -10,7 +10,7 @@
 
 namespace mvcwebapp_graph.Controllers
 {
-    [Authorize(AuthenticationSchemes = OpenIdConnectDefaults.AuthenticationScheme)]
+    [Authorize(AuthenticationSchemes = "openid2")]
```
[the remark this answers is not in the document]
jennyf19: fixed.

src/Microsoft.Identity.Web.UI/Areas/MicrosoftIdentity/Controllers/AccountController.cs
jennyf19: i think it's okay, if there is no scheme, it will use the default. we can discuss though.

tests/MultipleAuthSchemes/Startup.cs
jennyf19: i might leave, as it's the dev app, in case we want to test manually like this.

Issue #956, opened 2021-02-11
[Feature Request] Help certificate rotation of client certificates


jmprieur:
**Is your feature request related to a problem? Please describe.**
Today, Microsoft.Identity.Web:
- gets certificates from KeyVault
- helps rotating decrypt certificates (by enabling passing several certificates)

For the client certificates, it also supports send5xc, however, it could help a bit more by fetching from keyvault a newer version of a certificate when the confidential client flow fail because of certificate expiry or invalidation.

**Describe the solution you'd like**
When:
- the certificate proposed by Microsoft.Identity.Web to prove the app identity is rejected by Azure AD (for instance because it was revoked in the app registration, or a newer version is available in KeyVault), Microsoft.Identity.Web should try to refetch the certificate from KeyVault.

**Describe alternatives you've considered**
Subscribe to KeyVault events, but this would require too much configuration from customers

jmprieur:
### What happens: 
1. Create a self signed certificate which will expire soon. For instance if I want the cert to expire today at 4:05 pm. The date/time needs to be adapted to your locale (you might need to wrote "03/15/2021 4:05 pm" in the US

  ```PowerShell
   $cert=New-SelfSignedCertificate -Subject "Expired" -CertStoreLocation "Cert:\CurrentUser\My"  -KeyExportPolicy Exportable -KeySpec Signature -NotAfter "15/03/2021 16:05"
  ```

2. Using the certificate manager, export with private key

3. Add the certificate to the client certificate of the app to test before the certificate expires.

4. After the certificate expires, run the app, and observe that there is an MsalUiRequiredException:
   ` StatusCode: 401
        ResponseBody: {"error":"invalid_client","error_description":"AADSTS700027: Client assertion contains an invalid signature. [Reason - The key was not found., Thumbprint of key used by client: '24C1D15E417EDB6AA9DB1BCF05B3B00479809726'`

wuwei8372:
A few questions:
1) Looks like proposed method will only re-load the certificate when the current certificate is expired or invalid. If a new version of certificate is available and the old version is still valid, could we get the new certificate when calling LoadIfNeeded? (Probably not, right?)

2) Will this proposed method also work for non-aad certificate? For example a client certificate used for service to service call?

[...]

PR #1181, opened 2021-05-06
Jmprieur/cert rot

jmprieur:
Additional processing for https://github.com/AzureAD/microsoft-identity-web/issues/956
Tested with all the cases described in https://github.com/AzureAD/microsoft-identity-web/issues/956#issuecomment-832760201

- Added a boolean to avoid an infinite loop in some of the test cases
- Improved some of the error messages (still hardcoded in the code, though)
- extended to all the flows supported by ITokenAcquisition

### DIFF
--- a/src/Microsoft.Identity.Web/TokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web/TokenAcquisition.cs
@@ -39,6 +39,7 @@ namespace Microsoft.Identity.Web
         ///  Please call GetOrBuildConfidentialClientApplication instead of accessing this field directly.
         /// </summary>
         private IConfidentialClientApplication? _application;
+        private bool retryClientCertificate;
         private readonly IHttpContextAccessor _httpContextAccessor;
         private HttpContext? CurrentHttpContext => _httpContextAccessor.HttpContext;
         private readonly IMsalHttpClientFactory _httpClientFactory;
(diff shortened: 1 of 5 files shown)


src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -170,21 +171,29 @@ internal partial class TokenAcquisition : ITokenAcquisitionInternal
...
             }
+            finally
+            {
+                retryClientCertificate = false;
+            }
+        }
+
+        private bool IsInvalidClientCertificateError(MsalServiceException exMsal)
+        {
+            return !retryClientCertificate && exMsal.ErrorCode == "invalid_client" && exMsal.Message.Contains("AADSTS700027");
```
jennyf19, line 196: >invalid_client [](http://example.com/codeflow?start=67&length=14)

let's make constants for both of these. i can do it or you. not a big deal.

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -170,21 +171,29 @@ internal partial class TokenAcquisition : ITokenAcquisitionInternal
...
                 Logger.TokenAcquisitionError(_logger, LogMessages.ExceptionOccurredWhenAddingAnAccountToTheCacheFromAuthCode, ex);
                 throw;
             }
+            finally
+            {
+                retryClientCertificate = false;
+            }
+        }
+
+        private bool IsInvalidClientCertificateError(MsalServiceException exMsal)
```
jennyf19, line 194: nit: move this to below the public methods (i can do it as well if you prefer)

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -320,7 +342,23 @@ internal partial class TokenAcquisition : ITokenAcquisitionInternal
...
+                return builder.ExecuteAsync();
+            }
+            catch (MsalServiceException exMsal) when (IsInvalidClientCertificateError(exMsal))
+            {
+                DefaultCertificateLoader.ResetCertificates(_microsoftIdentityOptions.ClientCertificates);
+                _application = null;
+
+                // Retry
+                retryClientCertificate = true;
+                return GetAuthenticationResultForAppAsync(scope, tenant, tokenAcquisitionOptions);
```
jennyf19, line 356: >GetAuthenticationResultForAppAsync [](http://example.com/codeflow?start=23&length=34)

we have a different way of doing the logs now...see line 629 (another one i can fix up)

PR #1183, opened 2021-05-07
Configure redirect_uri to use HTTPS instead of detected protocol

WhitWaldo:
I run microservices on Service Fabric using HTTP so as to avoid having to maintain short-lived certificates on the various underlying nodes. The web-facing microservices are exposed to the public internet only through Azure App Gateway in order to handle SSL offloading (and thus provide a single point where the certificates need to be maintained). Thus, any HTTPS request goes App Gateway -> (unwrap SSL) -> send HTTP request to microservice -> reply with HTTP -> (wrap SSL) -> App Gateway replies via HTTPS to caller.

This works fine until I introduce Microsoft Authentication into the picture. Upon attempted authentication, the redirect_uri that's created appears to be based on the protocol configured to be in use by the host, so the non-configurable request to AAD looks like "redirect_uri=http://example.com". AAD app registrations understandably only allow `http://` to be used with localhost and `https://` for all other domains, so this fails with "AADSTS50011: The reply URL specified in the request does not match the reply URLs configured for the application" when I attempt to log into my microservice (even if it's otherwise accessible if the URL were to use HTTPS via App Gateway).

This pull request ([along with this PR](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/2613)) introduces a boolean option to allow the developer to force the use of HTTPS in the ultimately created redirect_uri value so that my service otherwise running HTTP will instead create a value of "redirect_uri=https://example.com" in the AAD request, which _can_ be properly registered with AAD and will otherwise work. This boolean defaults to false so it only impacts users that explicitly set it, so it doesn't introduce any breaking changes.

I'd be happy to make any changes as necessary to accommodate the addition of this function.

### DIFF
--- a/src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
+++ b/src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
@@ -43,6 +43,11 @@ namespace Microsoft.Identity.Web
         public string? ResetPasswordPolicyId { get; set; }
 
         /// <summary>
+        /// Requires that even if the host is detected to use HTTP, the redirect_url value should use HTTPS.
+        /// </summary>
+        public bool ForceHttpsOnRedirectUri { get; set; }
+
+        /// <summary>
         /// Gets the default user flow (which is signUpsignIn).
         /// </summary>
         public string? DefaultUserFlow => SignUpSignInPolicyId;
--- a/src/Microsoft.Identity.Web/TokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web/TokenAcquisition.cs
@@ -416,9 +416,17 @@ namespace Microsoft.Identity.Web
 
             var application = GetOrBuildConfidentialClientApplication();
 
+            // If the configuration indicates that HTTPS should be used regardless of the detected protocol, swap to HTTPS
+            var redirectUri = application.AppConfig.RedirectUri;
+            if (redirectUri.StartsWith("http://", StringComparison.InvariantCulture) &&
+                application.AppConfig.ForceHttpsOnRedirectUrl)
+            {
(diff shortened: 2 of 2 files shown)


jmprieur:
@WhitWaldo: this is not the way to do it.
Please see https://github.com/AzureAD/microsoft-identity-web/wiki/Deploying-Web-apps-to-App-services-as-Linux-containers#issue-with-http-redirect-uri-whereas-https-is-registered

WhitWaldo:
@jmprieur 
The instructions you pointed me to provided instructions for configuring forwarded headers in ASP.NET Core. That's great - in my Startup.cs in the ConfigureServices method, I've got:

`services.Configure<ForwardedHeadersOptions>(opt => opt.ForwardedHeaders = ForwardeddHeaders.Alll });`

And later in the Configure method I've got at the top:

`app.UseForwardedHeaders();`

My application is running on a Service Fabric cluster in Azure, so when I attempt to access the website at https://<subdomain>.innovian.net (via Azure Application Gateway) the initial request flows through fine, but when it attempts to authenticate out to AAD, it still uses a reply URI of http://10.0.0.4 which of course fails as not being registered.

This would be resolved if I could simply set the values somewhere in Identity instead of relying on it to automatically figure out the values. If it would allow me to indicate the reply URI as expecting https://<subdomain>.innovian.net all this would work without an issue, but I'm stuck with this automatic process that's not working.

What course do you recommend taking to move forward from here given that forwarded headers aren't doing the trick?

Tratcher:
Try the troubleshooting steps [here](https://docs.microsoft.com/en-us/aspnet/core/host-and-deploy/proxy-load-balancer?view=aspnetcore-5.0#troubleshoot) to see what headers the proxy is sending you.

PR #1189, opened 2021-05-10
add auth scheme integration tests

jennyf19:


tests/IntegrationTests/IntegrationTestService/Controllers/WeatherForecastController.cs
```diff
@@ -14,7 +14,7 @@
 namespace IntegrationTestService.Controllers
 {
     [ApiController]
-    [Authorize]
+    [Authorize(AuthenticationSchemes = TestConstants.CustomJwtScheme)]
```
jmprieur, line 17: Didn't we want to keep the default scheme? given the other controller has another scheme?

tests/IntegrationTests/IntegrationTestService/Startup.cs
```diff
@@ -28,23 +28,32 @@ public void ConfigureServices(IServiceCollection services)
             KeyVaultSecretsProvider _keyVault = new KeyVaultSecretsProvider();
             string ccaSecret = _keyVault.GetSecret(TestConstants.OBOClientKeyVaultUri).Value;
 
-            var builder = services.AddAuthentication(JwtBearerDefaults.AuthenticationScheme)
-                                  .AddMicrosoftIdentityWebApi(Configuration, subscribeToJwtBearerMiddlewareDiagnosticsEvents: true)
+            services.AddAuthentication()
```
jmprieur, line 31: To have the default scheme, we could call `AddAuthentication(jwtBearerScheme: TestConstants.CustomJwtScheme)` just here too, and then leave the `WeatherForecastController ` controller as it was

tests/MultipleAuthSchemes/Controllers/HomeB2CController.cs
```diff
@@ -24,7 +24,7 @@ public HomeB2CController(ILogger<HomeController> logger, IDownstreamWebApi downs
             ScopeKeySection = "DownstreamB2CApi:Scopes", UserFlow = "b2c_1_susi", AuthenticationScheme = "B2C")]
         public async Task<IActionResult> Index()
         {
-            var value = await _downstreamWebApi.GetForUserAsync<Task>("DownstreamB2CApi", "", null, null, "B2C");
+            var value = await _downstreamWebApi.GetForUserAsync<Task>("DownstreamB2CApi", string.Empty, authenticationScheme:"B2C");
```
jmprieur, line 27: the string.Empty is probably not even necessary? is it?

tests/Microsoft.Identity.Web.Test.Integration/AcquireTokenForUserIntegrationTests.cs
```diff
@@ -81,8 +94,32 @@ public AcquireTokenForUserIntegrationTests(WebApplicationFactory<Startup> factor
...
+                    else
+                    {
+                        services.AddDistributedMemoryCache();
+                        services.AddDistributedTokenCaches();
+                    }
+
+                    services.BuildServiceProvider();
+                });
+            })
+                        .CreateClient(new WebApplicationFactoryClientOptions
```
jmprieur, line 119: There seems to be something funny with spaces here.

jmprieur (approved):
LGTM.
I've left a few suggestions

tests/MultipleAuthSchemes/Controllers/HomeB2CController.cs
jennyf19: there is no default for the relative path.

PR #1191, opened 2021-05-10
Fix Visual Studio authenication timeout errors

benbhall:
Azure.Identity 1.3.0 [fixed some issues causing auth timeouts](https://github.com/Azure/azure-sdk-for-net/blob/master/sdk/identity/Azure.Identity/CHANGELOG.md#130-2020-11-12) when working locally in Visual Studio. Microsoft.Identity.Web 1.9.2 still uses 1.2.3. The reason for the errors is not immediately obvious because of a misleading reference to Azure CLI.

The fix is to upgrade to Azure.Identity 1.3.0.


jmprieur (approved):
LGTM. 
Thanks @benbhall

PR #1192, opened 2021-05-10
cert rotation

jennyf19:
opening so we can review easier...

### DIFF
--- a/src/Microsoft.Identity.Web/CertificateManagement/DefaultCertificateLoader.cs
+++ b/src/Microsoft.Identity.Web/CertificateManagement/DefaultCertificateLoader.cs
@@ -212,9 +212,12 @@ namespace Microsoft.Identity.Web
 
         internal static void ResetCertificates(IEnumerable<CertificateDescription>? clientCertificates)
         {
-            foreach (var cert in clientCertificates)
+            if (clientCertificates != null)
             {
-                cert.Certificate = null;
+                foreach (var cert in clientCertificates)
+                {
+                    cert.Certificate = null;
+                }
             }
         }
 
@@ -279,7 +282,9 @@ namespace Microsoft.Identity.Web
         internal /*for test only*/ static X509Certificate2? LoadFirstCertificate(IEnumerable<CertificateDescription> certificateDescription)
         {
             DefaultCertificateLoader defaultCertificateLoader = new DefaultCertificateLoader();
-            CertificateDescription certDescription = certificateDescription.FirstOrDefault(c => { defaultCertificateLoader.LoadIfNeeded(c);
+            CertificateDescription? certDescription = certificateDescription.FirstOrDefault(c =>
+            {
+                defaultCertificateLoader.LoadIfNeeded(c);
                 return c.Certificate != null;
             });
 
--- a/src/Microsoft.Identity.Web/CertificateManagement/DefaultCertificateLoader.cs
+++ b/src/Microsoft.Identity.Web/CertificateManagement/DefaultCertificateLoader.cs
@@ -102,7 +102,7 @@ namespace Microsoft.Identity.Web
         /// <remarks>This code is inspired by Heath Stewart's code in:
         /// https://github.com/heaths/azsdk-sample-getcert/blob/master/Program.cs#L46-L82.
         /// </remarks>
-        private static X509Certificate2 LoadFromKeyVault(
+        private static X509Certificate2? LoadFromKeyVault(
             string keyVaultUrl,
             string certificateName,
             X509KeyStorageFlags x509KeyStorageFlags)
@@ -115,6 +115,11 @@ namespace Microsoft.Identity.Web
             SecretClient secretClient = new SecretClient(keyVaultUri, credential);
 
             KeyVaultCertificateWithPolicy certificate = certificateClient.GetCertificate(certificateName);
+
+            if (certificate.Properties.NotBefore <= DateTime.UtcNow && certificate.Properties.ExpiresOn <= DateTime.UtcNow)
+            {
+                return null;
+            }
 
             // Return a certificate with only the public key if the private key is not exportable.
             if (certificate.Policy?.Exportable != true)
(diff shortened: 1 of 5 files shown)


src/Microsoft.Identity.Web/CertificateManagement/DefaultCertificateLoader.cs
```diff
@@ -116,6 +116,11 @@ private static X509Certificate2 LoadFromBase64Encoded(string certificateBase64,
 
             KeyVaultCertificateWithPolicy certificate = certificateClient.GetCertificate(certificateName);
 
+            if (certificate.Properties.NotBefore <= DateTime.UtcNow && certificate.Properties.ExpiresOn <= DateTime.UtcNow)
```
bgavrilMS, line 119: A few comments: 

1. NotBefore and ExpiresOn can be null. Not sure why, but I think we should treat that case separately
2. nit; I'm always confused about mixing DateTimeOffset and DateTime.UtcNow. Probably fine. Might be cleaner to use DateTimeOffset.UtcNow
3. I think the condition to accept the cert is : 

NotBefore ---------- now ------------- ExpOn

so the condition to reject the cert is: 

now < NotBefore OR now > ExpOn

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -170,11 +171,24 @@ internal partial class TokenAcquisition : ITokenAcquisitionInternal
 
                 context.HandleCodeRedemption(null, result.IdToken);
             }
+            catch (MsalServiceException exMsal) when (IsInvalidClientCertificateError(exMsal))
+            {
+                DefaultCertificateLoader.ResetCertificates(_microsoftIdentityOptions.ClientCertificates);
```
bgavrilMS, line 176: From what I understand .. the user configures one or more certificate descriptions. Each certificate description can be used to fetch 0 or 1 certificates. 

I think here we only deal with the case where there is a KeyVault certificate description and we try to reload the certificate, using the existing description. 

I am not sure how M.I.W. deals with multiple cert descriptions, e.g. one from KV, one from my local store? Maybe we can chat about this. 

In any case, I think there should be an explicit API on the ClientCertificate class, smth like `Reload`, which may be implemented by some cert types (keyvault) but not others.

src/Microsoft.Identity.Web/CertificateManagement/DefaultCertificateLoader.cs
```diff
@@ -116,6 +116,11 @@ private static X509Certificate2 LoadFromBase64Encoded(string certificateBase64,
...
+            {
+                return null;
```
bgavrilMS, line 121: There seem to be a mix of returning null and throwing exceptions. Would be interesting to try out the non-happy path of configuring MIW with some expired or bad certs and seeing the experience.

src/Microsoft.Identity.Web/TokenAcquisition.cs
jmprieur: Descriptions from other than KV won't have several versions (KeyVault does), but this is not even a goal. 
The cert rotation is still enabled for other descriptions than KeyVault if the developer provides several cert descriptions (one that will expire soon, and another one), whatever the certificate source. The thing with KeyVault is that the same certificate source can have a newer version which extends the expiry of the previous, now expired version. the filepath description could be the same too, and probably the subject name one too from the certificate store.

Why Reload? is it an optimization?

src/Microsoft.Identity.Web/CertificateManagement/DefaultCertificateLoader.cs
jmprieur: Of course this was done, @bgavrilMS !
See https://github.com/AzureAD/microsoft-identity-web/issues/956#issuecomment-832760201 for all the pathological cases which were tested.

src/Microsoft.Identity.Web/TokenAcquisition.cs
bgavrilMS, line 176: I think I get your design here. You reset the cert and the _app, and you force the recreation of the app. Makes sense. 

As a small observation - I noticed that you added quite a few static methods on DefaultCertificateLoader which just pass in the list of `ClientCertificates`. This indicates that you are missing an object, smth like 

```
ICertManager
+ IEnumerable<ClientCertificates>
+ ResetCertificates() 
+ LoadFirstCertificate() 
+ LoadAllCertificates()
```

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -310,7 +337,23 @@ internal partial class TokenAcquisition : ITokenAcquisitionInternal
...
+                _application = null;
+
+                // Retry
+                retryClientCertificate = true;
+                return GetAuthenticationResultForAppAsync(scope, tenant, tokenAcquisitionOptions);
+            }
+            finally
+            {
+                retryClientCertificate = false;
```
bgavrilMS, line 355: How often do we want to retry a failed cert? Is TokenAcquision a long lived service? If it is, when the cert expires, you reload it. The second time the cert expires, you will not reload it. Is this "reload once" the strategy you are aiming for?

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -472,6 +515,11 @@ public async Task RemoveAccountAsync(RedirectContext context)
             }
         }
 
+        private bool IsInvalidClientCertificateError(MsalServiceException exMsal)
+        {
+            return !retryClientCertificate && exMsal.ErrorCode == Constants.InvalidClient && exMsal.Message.Contains(Constants.InvalidKeyError);
```
bgavrilMS, line 520: I would remove the `retryClientCertificate ` check from here and move it up.

src/Microsoft.Identity.Web/TokenAcquisition.cs
jmprieur: yes. @bgavrilMS : let's revisit this when we implement it in MSAL?

jennyf19 (commented):
:shipit:

src/Microsoft.Identity.Web/TokenAcquisition.cs
jmprieur: Although it's not the default, TokenAcquisition can be a singleton.
The reload once is to avoid an infinite loop. It's used in finally blocks with a tail recursion.

PR #1204, opened 2021-05-19
changelog 1.11

jennyf19:


changelog.md
```diff
@@ -1,3 +1,8 @@
+1.11.0
+==========
+### New Features:
+**Microsoft Identity Web now supports multiple authentication schemes**. This means, you can have several authentication schemes in the same ASP.NET Core app. Such as two Azure AD web apps, or an Azure AD app and an Azure AD B2C app, or a web app and a web API. Basically mixing authentication schemes in the same ASP.NET Core app. See the [wiki for details and code samples](https://github.com/AzureAD/microsoft-identity-web/wiki/Multiple-Authentication-Schemes) and related issues: [#549](https://github.com/AzureAD/microsoft-identity-web/issues/549), [#429](https://github.com/AzureAD/microsoft-identity-web/issues/429), [#958](https://github.com/AzureAD/microsoft-identity-web/issues/958), [#1126](https://github.com/AzureAD/microsoft-identity-web/issues/1126), [#971](https://github.com/AzureAD/microsoft-identity-web/issues/971), [#173](https://github.com/AzureAD/microsoft-identity-web/issues/173), [#955](https://github.com/AzureAD/microsoft-identity-web/issues/955), and [#1127](https://github.com/AzureAD/microsoft-identity-web/issues/1127).
```
jmprieur, line 4: "Such as signing-in in a web app with two identity providers" (so several client Ids indeed)

jmprieur (approved):
LGTM.

For web apps,  I wonder if we want to be clearer about "2 web apps" means (2 clientID, and call out the fact we have IDPs).?

On the case of Web app + web API it could be the same ClientID (but OpenIdConnect and JwtBearer for instance)

changelog.md
jennyf19: maybe that should go in the wiki instead? all those details?

jennyf19: agree we can make it clearer. but i can always edit the changelog.

PR #1205, opened 2021-05-19
update msal version and add auth result logging metrics

jennyf19:


jmprieur (approved):
LGTM.
Thanks @jennyf19

PR #1207, opened 2021-05-21
fix cca options w/azure functions

jennyf19:
@jmprieur fix for azure functions when the cca options are not getting populated.

image


jmprieur (approved):
LGTM
Thanks @jennyf19

Issue #1215, opened 2021-05-26
[Bug] MergedOptions UpdateMergedOptionsFromMicrosoftIdentityOptions error


JadynWong:
**Which version of Microsoft Identity Web are you using?**
Note that to get help, you need to run the latest version. 
1.11.0
1.10.0

**Where is the issue?**
* Web app
    * [x] Sign-in users
    * [ ] Sign-in users and call web APIs
* Web API
    * [ ] Protected web APIs (validating tokens)
    * [ ] Protected web APIs (validating scopes)
    * [ ] Protected web APIs call downstream web APIs
* Token cache serialization
     * [x] In-memory caches
     * [ ] Session caches
     * [ ] Distributed caches
* Other (please describe)

**Is this a new or an existing app?**
The app is in development and I have upgraded to a new version of Microsoft Identity Web.

**Repro**

```csharp
context.Services.AddAuthentication()
                .AddMicrosoftIdentityWebApp(
                    configuration: configuration,
                    configSectionName: "AzureAd",
                    openIdConnectScheme: OpenIdConnectDefaults.AuthenticationScheme,
                    cookieScheme: null,
                    subscribeToOpenIdConnectMiddlewareDiagnosticsEvents: false
                );
```

```json
  "AzureAd": {
    "Instance": "https://login.microsoftonline.com/",
    "Domain": "contoso.com",
    "TenantId": "xxxxxxxxxxxx-xxxx-xxxx-xxxxxxxxxxxx",
    "ClientId": "xxxxxxxxxxxx-xxxx-xxxx-xxxxxxxxxxxx",
    "ClientSecret": "xxxxxxxxxxxxxxxxxxxxxxxxxxxxx",
    "CallbackPath": "/signin-oidc",
    "SignedOutCallbackPath ": "/signout-callback-oidc",
    "ResponseType": "code id_token",
    "Scope": [ "email" ]
  },
```

https://github.com/AzureAD/microsoft-identity-web/blob/222e76a60930949e3e4838ece6e5ef49e793f091/src/Microsoft.Identity.Web/WebAppExtensions/MicrosoftIdentityWebAppAuthenticationBuilderExtensions.cs#L275-L291

https://github.com/AzureAD/microsoft-identity-web/blob/222e76a60930949e3e4838ece6e5ef49e793f091/src/Microsoft.Identity.Web/MergedOptions.cs#L184-L192

**Expected behavior**
ResponseType should is  `code id_token`
[...]

PR #1227, opened 2021-05-28
fix for response type in mergedoptions

jennyf19:
#1215


jennyf19:
@jmprieur we need to do more here, fyi. i might wait to merge

jmprieur (approved):
LGTM
Let's do it only for ResponseType.

jennyf19:
fixed.

---
In reply to: [671583673](https://github.com/AzureAD/microsoft-identity-web/pull/1227#issuecomment-671583673) [](http://example.com/codeflow?ancestors=671583673,671583673)

jmprieur (approved):
LGTM
thanks @jennyf19

Issue #1239, opened 2021-06-03
[Feature Request] Add support for CancellationToken


jmprieur:
**Is your feature request related to a problem? Please describe.**
MSAL.NET 4.32.0 brings support for: `CancellationToken` on `GetAccountAsync`, `RemoveAsync` and `TokenCacheNotificationArgs`.

```CSharp
 public abstract class ClientApplicationBase : IClientApplicationBase 
{
  public Task<IAccount> GetAccountAsync(string accountId, CancellationToken cancellationToken=default(CancellationToken));
  public Task RemoveAsync(IAccount account, CancellationToken cancellationToken=default(CancellationToken));
 }
```

```CSharp
    public sealed class TokenCacheNotificationArgs {
        public CancellationToken CancellationToken { get; }
    }
```

**Describe the solution you'd like**
- Have new overrides in MsalAbstractTokenCacheProvider to support the cancellation tokens, implemented in MsalDistributedCacheAdapter. (trying avoiding a breaking change)
- Add a CancellationToken property in [TokenAcquisitionOptions](https://github.com/AzureAD/microsoft-identity-web/blob/master/src/Microsoft.Identity.Web/TokenAcquisitionOptions.cs) and, if available use it in TokenAcquisition in calls to:
  - AcquireTokenXXX methods
  - GetAccountAsync()
  - RemoveAsync

**Additional context**
MSAL.NET 4.32.0 also supports:
- `.WithTenant()` in all flows (which could be used instead of `.WithAuthority`). This would be to remove technical debt
- `WithExtraHttpHeaders` on AcquireTokenXXX

jennyf19:
Fixed in [1.13 release](https://github.com/AzureAD/microsoft-identity-web/releases/tag/1.13.0).

PR #1242, opened 2021-06-05
[wip] add cancellation token support

jennyf19:
#1239

### DIFF
--- a/src/Microsoft.Identity.Web/TokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web/TokenAcquisition.cs
@@ -10,6 +10,7 @@ using System.Net;
 using System.Net.Http;
 using System.Security.Claims;
 using System.Security.Cryptography.X509Certificates;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.AspNetCore.Authentication.JwtBearer;
 using Microsoft.AspNetCore.Authentication.OAuth;
@@ -362,7 +363,7 @@ namespace Microsoft.Identity.Web
 
             try
             {
-                return builder.ExecuteAsync();
+                return builder.ExecuteAsync(tokenAcquisitionOptions != null ? tokenAcquisitionOptions.CancellationToken : CancellationToken.None);
             }
             catch (MsalServiceException exMsal) when (IsInvalidClientCertificateError(exMsal))
             {
@@ -744,7 +745,7 @@ namespace Microsoft.Identity.Web
                         }
                     }
 
-                    return await builder.ExecuteAsync()
+                    return await builder.ExecuteAsync(tokenAcquisitionOptions != null ? tokenAcquisitionOptions.CancellationToken : CancellationToken.None)
                                         .ConfigureAwait(false);
                 }
 
@@ -865,7 +866,7 @@ namespace Microsoft.Identity.Web
                 builder.WithAuthority(authority);
             }
 
-            return builder.ExecuteAsync();
+            return builder.ExecuteAsync(tokenAcquisitionOptions != null ? tokenAcquisitionOptions.CancellationToken : CancellationToken.None);
         }
 
         private static bool AcceptedTokenVersionMismatch(MsalUiRequiredException msalServiceException)
--- a/src/Microsoft.Identity.Web/TokenAcquisitionOptions.cs
+++ b/src/Microsoft.Identity.Web/TokenAcquisitionOptions.cs
@@ -48,7 +48,7 @@ namespace Microsoft.Identity.Web
         public PoPAuthenticationConfiguration? PoPConfiguration { get; set; }
 
         /// <summary>
-        /// todo.
+        /// Cancellation token to be used when calling the token acquisition methods.
         /// </summary>
         public CancellationToken CancellationToken { get; set; } = CancellationToken.None;
 
--- a/src/Microsoft.Identity.Web/TokenAcquisitionOptions.cs
+++ b/src/Microsoft.Identity.Web/TokenAcquisitionOptions.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Threading;
 using Microsoft.Identity.Client.AppConfig;
 
 namespace Microsoft.Identity.Web
@@ -47,6 +48,11 @@ namespace Microsoft.Identity.Web
         public PoPAuthenticationConfiguration? PoPConfiguration { get; set; }
 
         /// <summary>
+        /// todo.
+        /// </summary>
+        public CancellationToken CancellationToken { get; set; } = CancellationToken.None;
+
+        /// <summary>
         /// Clone the options (to be able to override them).
         /// </summary>
         /// <returns>A clone of the options.</returns>
@@ -58,6 +64,8 @@ namespace Microsoft.Identity.Web
                 ExtraQueryParameters = ExtraQueryParameters,
                 ForceRefresh = ForceRefresh,
                 Claims = Claims,
+                PoPConfiguration = PoPConfiguration,
+                CancellationToken = CancellationToken,
             };
         }
     }
--- a/src/Microsoft.Identity.Web/TokenCacheProviders/CacheSerializerHints.cs
+++ b/src/Microsoft.Identity.Web/TokenCacheProviders/CacheSerializerHints.cs
@@ -1,0 +1,18 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Threading;
+
+namespace Microsoft.Identity.Web.TokenCacheProviders
+{
+    /// <summary>
+    /// todo.
+    /// </summary>
+    public class CacheSerializerHints
+    {
+        /// <summary>
+        /// Todo.
+        /// </summary>
+        public CancellationToken CancellationToken { get; set; }
+    }
+}
--- a/src/Microsoft.Identity.Web/TokenCacheProviders/CacheSerializerHints.cs
+++ b/src/Microsoft.Identity.Web/TokenCacheProviders/CacheSerializerHints.cs
@@ -6,12 +6,12 @@ using System.Threading;
 namespace Microsoft.Identity.Web.TokenCacheProviders
 {
     /// <summary>
-    /// todo.
+    /// Set of properties that the token cache serialization implementations might use to optimize the cache.
     /// </summary>
     public class CacheSerializerHints
     {
         /// <summary>
-        /// Todo.
+        /// CancellationToken enabling cooperative cancellation between threads, thread pool, or Task objects.
         /// </summary>
         public CancellationToken CancellationToken { get; set; }
     }
--- a/src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
+++ b/src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
@@ -81,7 +81,7 @@ namespace Microsoft.Identity.Web.TokenCacheProviders.Distributed
         /// from the distributed cache.
         /// </summary>
         /// <param name="cacheKey">Key of the cache to remove.</param>
-        /// <param name="cacheSerializerHints">cacheSerializerHints.</param>
+        /// <param name="cacheSerializerHints">Hints for the cache serialization implementation optimization.</param>
         /// <returns>A <see cref="Task"/> that completes when key removal has completed.</returns>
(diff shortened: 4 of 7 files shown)


src/Microsoft.Identity.Web/TokenAcquisitionOptions.cs
```diff
@@ -46,6 +47,11 @@ public class TokenAcquisitionOptions
         /// </summary>
         public PoPAuthenticationConfiguration? PoPConfiguration { get; set; }
 
+        /// <summary>
+        /// todo.
```
jmprieur, line 51:
```suggestion
        /// Cancellation token to be used when calling the token acquisition methods. 
```

src/Microsoft.Identity.Web/TokenAcquisitionOptions.cs
```diff
@@ -46,6 +47,11 @@ public class TokenAcquisitionOptions
...
+        /// </summary>
+        public CancellationToken CancellationToken { get; set; } = CancellationToken.None;
```
jmprieur, line 53: For a few minutes, I was wondering if we would want to use `CacheSerializerHints` instead of `CancellationToken`
but thinking twice it would not make sense as the properties of the Hint will come from MSAL.NET Token serialization args, not necessarily the developers.

So nothing to change here, but I thought I would explain my thoughts.

src/Microsoft.Identity.Web/TokenCacheProviders/CacheSerializerHints.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Threading;
+
+namespace Microsoft.Identity.Web.TokenCacheProviders
+{
+    /// <summary>
+    /// todo.
```
jmprieur, line 9:
```suggestion
    /// Set of properties that the token cache serialization implementations might use to optimize the cache
    /// when they are capable of it. There is no guaranty that these properties are used (for instance MemoryCache
    /// doesn't support the cancellation token).
```

src/Microsoft.Identity.Web/TokenCacheProviders/CacheSerializerHints.cs
```diff
@@ -0,0 +1,18 @@
...
+    /// </summary>
+    public class CacheSerializerHints
+    {
+        /// <summary>
+        /// Todo.
```
jmprieur, line 14: CancellationToken enabling cooperative cancellation between threads, thread pool, or Task objects. This is used
when calling the token acquisition methods, which then pass-them in to the token cache serializers.

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
```diff
@@ -71,6 +72,18 @@ public partial class MsalDistributedTokenCacheAdapter : MsalAbstractTokenCachePr
...
+        {
+            await RemoveKeyAsync(cacheKey, new CacheSerializerHints()).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Removes a specific token cache, described by its cache key
+        /// from the distributed cache.
+        /// </summary>
+        /// <param name="cacheKey">Key of the cache to remove.</param>
+        /// <param name="cacheSerializerHints">cacheSerializerHints.</param>
```
jmprieur, line 84:
```suggestion
        /// <param name="cacheSerializerHints">Hints for the cache serialization implementation optimization.</param>
```

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
```diff
@@ -91,6 +104,19 @@ protected override async Task RemoveKeyAsync(string cacheKey)
...
+        {
+            return await ReadCacheBytesAsync(cacheKey, new CacheSerializerHints()).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Read a specific token cache, described by its cache key, from the
+        /// distributed cache.
+        /// </summary>
+        /// <param name="cacheKey">Key of the cache item to retrieve.</param>
+        /// <param name="cacheSerializerHints">cacheSerializerHints.</param>
```
jmprieur, line 116:
```suggestion
        /// <param name="cacheSerializerHints">Hints for the cache serialization implementation optimization.</param>
```

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
```diff
@@ -145,6 +174,21 @@ protected override async Task<byte[]> ReadCacheBytesAsync(string cacheKey)
...
+        {
+            await WriteCacheBytesAsync(cacheKey, bytes, new CacheSerializerHints()).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Writes a token cache blob to the serialization cache (by key).
+        /// </summary>
+        /// <param name="cacheKey">Cache key.</param>
+        /// <param name="bytes">blob to write.</param>
+        /// <param name="cacheSerializerHints">cacheSerializerHints.</param>
```
jmprieur, line 186:
```suggestion
        /// <param name="cacheSerializerHints">Hints for the cache serialization implementation optimization.</param>
```

src/Microsoft.Identity.Web/TokenCacheProviders/MsalAbstractTokenCacheProvider.cs
```diff
@@ -105,18 +112,52 @@ public async Task ClearAsync(string homeAccountId)
         /// <returns>A <see cref="Task"/> that represents a completed write operation.</returns>
         protected abstract Task WriteCacheBytesAsync(string cacheKey, byte[] bytes);
 
+        /// <summary>
+        /// Method to be implemented by concrete cache serializers to write the cache bytes.
+        /// </summary>
+        /// <param name="cacheKey">Cache key.</param>
+        /// <param name="bytes">Bytes to write.</param>
+        /// <param name="cacheSerializerHints">cacheSerializerHints.</param>
```
jmprieur, line 120:
```suggestion
        /// <param name="cacheSerializerHints">Hints for the cache serialization implementation optimization.</param>
```

src/Microsoft.Identity.Web/TokenCacheProviders/MsalAbstractTokenCacheProvider.cs
```diff
@@ -105,18 +112,52 @@ public async Task ClearAsync(string homeAccountId)
...
         /// </summary>
         /// <param name="cacheKey">Cache key.</param>
         /// <returns>Read bytes.</returns>
         protected abstract Task<byte[]> ReadCacheBytesAsync(string cacheKey);
 
+        /// <summary>
+        /// Method to be implemented by concrete cache serializers to Read the cache bytes.
+        /// </summary>
+        /// <param name="cacheKey">Cache key.</param>
+        /// <param name="cacheSerializerHints">cacheSerializerHints.</param>
```
jmprieur, line 138:
```suggestion
        /// <param name="cacheSerializerHints">Hints for the cache serialization implementation optimization.</param>
```

src/Microsoft.Identity.Web/TokenCacheProviders/MsalAbstractTokenCacheProvider.cs
```diff
@@ -105,18 +112,52 @@ public async Task ClearAsync(string homeAccountId)
...
         /// </summary>
         /// <param name="cacheKey">Cache key.</param>
         /// <returns>A <see cref="Task"/> that represents a completed remove key operation.</returns>
         protected abstract Task RemoveKeyAsync(string cacheKey);
+
+        /// <summary>
+        /// Method to be implemented by concrete cache serializers to remove an entry from the cache.
+        /// </summary>
+        /// <param name="cacheKey">Cache key.</param>
+        /// <param name="cacheSerializerHints">cacheSerializerHints.</param>
```
jmprieur, line 156:
```suggestion
        /// <param name="cacheSerializerHints">Hints for the cache serialization implementation optimization.</param>
```

src/Microsoft.Identity.Web/TokenCacheProviders/Session/MsalSessionTokenCacheProvider.cs
```diff
@@ -55,7 +55,21 @@ public partial class MsalSessionTokenCacheProvider : MsalAbstractTokenCacheProvi
...
-            await _session.LoadAsync().ConfigureAwait(false);
+            return await ReadCacheBytesAsync(cacheKey, new CacheSerializerHints()).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Read a blob representing the token cache from its key.
+        /// </summary>
+        /// <param name="cacheKey">Key representing the token cache
+        /// (account or app).</param>
+        /// <param name="cacheSerializerHints">cacheSerializerHints.</param>
```
jmprieur, line 66:
```suggestion
        /// <param name="cacheSerializerHints">Hints for the cache serialization implementation optimization.</param>
```

jmprieur (approved):
LGTM. Thanks @jennyf19 
It turns out that the solution of using a `CacheSerializerHints` class looks pretty elegant (I wasn't too sure how nice the programing model would be for the user developer and extended developer, when we discussed it). It's extensible, and the semantics is clear: those are hints that the cache serializers might not honor.

Even if I know this is a draft PR, I've proposed a few suggestions for the XML comments.

src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -362,7 +363,7 @@ internal MergedOptions GetOptions(string authenticationScheme)
 
             try
             {
-                return builder.ExecuteAsync();
+                return builder.ExecuteAsync(tokenAcquisitionOptions != null ? tokenAcquisitionOptions.CancellationToken : CancellationToken.None);
```
bgavrilMS, line 366: The standard pattern for cancellation is that every async method ends with the CancellationToken parameter which defaults to None. It's not expected to be in an options object.

I think this would not be a breaking change to this class?

src/Microsoft.Identity.Web/TokenAcquisitionOptions.cs
bgavrilMS, line 53: As stated above, this is a non-standard way of exposing cancellation token. Best to put it as the last param in every async method on the public API (and internal too, for propagation)

src/Microsoft.Identity.Web/TokenCacheProviders/CacheSerializerHints.cs
```diff
@@ -0,0 +1,18 @@
...
+// Licensed under the MIT License.
+
+using System.Threading;
+
+namespace Microsoft.Identity.Web.TokenCacheProviders
+{
+    /// <summary>
+    /// todo.
+    /// </summary>
+    public class CacheSerializerHints
```
bgavrilMS, line 11: +1 to exposing an object that makes sense for M.I.W., not the lower level object that MSAL exposes (CacheNotificationArgs)

src/Microsoft.Identity.Web/TokenAcquisition.cs
jmprieur: This would be a change in the public interface (ITokenAcquisition) ... hence a breaking change.

src/Microsoft.Identity.Web/TokenCacheProviders/MsalAbstractTokenCacheProvider.cs
```diff
@@ -105,18 +106,52 @@ public async Task ClearAsync(string homeAccountId)
         /// <returns>A <see cref="Task"/> that represents a completed write operation.</returns>
         protected abstract Task WriteCacheBytesAsync(string cacheKey, byte[] bytes);
 
+        /// <summary>
+        /// Method to be implemented by concrete cache serializers to write the cache bytes.
```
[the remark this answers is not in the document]
jennyf19: i will do these myself and commit, but thanks for the suggestions, there are many comments to be added/fixed.

src/Microsoft.Identity.Web/TokenAcquisition.cs
jennyf19: @jmprieur and I synced on this, but to avoid a breaking change, it's going to add a lot more complexity. Basically doubling the API and reducing the developer experience. I think we are leaning towards leaving it as-is and seeing what customers say. When can reconsider at the next major version.

PR #1255, opened 2021-06-10
update to 4.32.1 msal

jennyf19:
tested B2C & web app calls Graph


jmprieur (approved):
LGTM
Thanks @jennyf19

PR #1260, opened 2021-06-10
fix issuer

jennyf19:


src/Microsoft.Identity.Web/Resource/AadIssuerValidator.cs
```diff
@@ -83,6 +102,7 @@ public class AadIssuerValidator
                         IssuerMetadata issuerMetadata =
                             CreateConfigManager(AadAuthority).GetConfigurationAsync().ConfigureAwait(false).GetAwaiter().GetResult();
                         AadIssuerV2 = issuerMetadata.Issuer!;
+
```
jennyf19, line 105:
```suggestion
```

jmprieur (approved):
LGTM
Thanks @jennyf19 

Tested on daemon calls own web API

PR #1264, opened 2021-06-14
Addding authorization options to web APIs.

jmprieur:
Fix for https://github.com/AzureAD/microsoft-identity-web/issues/1265

cc: @hpsin @psignoret

### DIFF
--- a/src/Microsoft.Identity.Web/Constants/ClaimConstants.cs
+++ b/src/Microsoft.Identity.Web/Constants/ClaimConstants.cs
@@ -99,5 +99,15 @@ namespace Microsoft.Identity.Web
         /// Name Identifier ID claim: "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/nameidentifier".
         /// </summary>
         public const string NameIdentifierId = "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/nameidentifier";
+
+        /// <summary>
+        /// Optional claim used to determine the token type.
+        /// </summary>
+        public const string Idtyp = "idtyp";
+
+        /// <summary>
+        /// Optional claim used to determine if the user account is a guest account.
+        /// </summary>
+        public const string Acct = "acct";
     }
 }
--- a/src/Microsoft.Identity.Web/Constants/IDWebErrorMessage.cs
+++ b/src/Microsoft.Identity.Web/Constants/IDWebErrorMessage.cs
@@ -1,6 +1,8 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System.Runtime.Serialization;
+
 namespace Microsoft.Identity.Web
 {
     /// <summary>
@@ -36,6 +38,14 @@ namespace Microsoft.Identity.Web
         public const string BlazorServerUserNotSet = "IDW10206: Using Blazor server but the user was not properly set. ";
         public const string CalledApiScopesAreNull = "IDW10207: The CalledApiScopes cannot be null. ";
         public const string ScopesRequiredToCallMicrosoftGraph = "IDW10208: You need to either pass-in scopes to AddMicrosoftGraph, in the appsettings.json file, or with .WithScopes() on the Graph queries. See https://aka.ms/ms-id-web/microsoftGraph. ";
+        public const string ApiAllowsAppOnlyTokensCantBeComputed = "IDW10209: You set the 'ApiAllowsAppOnlyTokens' configuration to false, but didn't opt-in to the 'idtyp' optional claim in the app registration. See https://aka.ms/ms-id-web/webapi-authorization. ";
+        public const string ApiDoesNotAllowsAppOnlyTokens = "IDW10210: This web API does not allow app-only tokens. ";
+        public const string ApiAllowsGuestAccountsCantBeComputed = "IDW10211: You set the 'ApiAllowsGuestAccounts' configuration to false, but didn't opt-in to the 'acct' optional claim in the app registration. See https://aka.ms/ms-id-web/webapi-authorization. ";
+        public const string ApiDoesNotAllowsGuestAccounts = "IDW10212: This web API does not allow guest accounts. ";
+        public const string ApiAllowedTenantsCantBeComputed = "IDW10213: You set the 'AllowedTenantIds' configuration, but the `tid` or `tenantId` claims are not found. See https://aka.ms/ms-id-web/webapi-authorization. ";
+        public const string ApiTenantNotAllowed = "IDW10214: This web API does not allow this tenant. ";
+        public const string ApiAllowedClientApplicationsCantBeComputed = "IDW10215: You set the 'AllowedClientApplications' configuration, but the `oid` or `objectid` claims are not found. See https://aka.ms/ms-id-web/webapi-authorization. ";
+        public const string ApiClientApplicationNotAllowed = "IDW10216: This web API does not allow this client application. ";
 
         // Token Validation IDW10300 = "IDW10300:"
         public const string IssuerMetadataUrlIsRequired = "IDW10301: Azure AD Issuer metadata address URL is required. ";
--- a/src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
+++ b/src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
@@ -151,7 +151,7 @@ namespace Microsoft.Identity.Web
         public bool ApiAllowsGuestAccounts { get; set; } = true;
 
         /// <summary>
-        /// Applies to web apps and web APIs. Enables to provide the list
+        /// Applies to web apps and web APIs. Enables you to provide the list
         /// of tenant Ids the user of whom are allowed to call the web API
         /// or log-in to the web app.
         /// </summary>
--- a/src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
+++ b/src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
@@ -10,7 +10,7 @@ namespace Microsoft.Identity.Web
     /// <summary>
     /// Options for configuring authentication using Azure Active Directory. It has both AAD and B2C configuration attributes.
     /// </summary>
-    public class MicrosoftIdentityOptions : OpenIdConnectOptions
+    public class MicrosoftIdentityOptions : OpenIdConnectOptions, ITokenAuthorizationOptions
     {
         /// <summary>
         /// Gets or sets the Azure Active Directory instance, e.g. "https://login.microsoftonline.com".
@@ -133,5 +133,34 @@ namespace Microsoft.Identity.Web
         /// See https://docs.microsoft.com/azure/active-directory/managed-identities-azure-resources/how-to-manage-ua-identity-portal.
         /// </summary>
         public string? UserAssignedManagedIdentityClientId { get; set; }
+
+        /// <summary>
+        /// Applies to a web API: does the web API allow app-only tokens?.
+        /// </summary>
+        /// <remarks>
+        /// In the app registration you'll need to opt-in for the idtyp claim.
+        /// </remarks>
+        public bool ApiAllowsAppOnlyTokens { get; set; } = true;
+
+        /// <summary>
+        /// Applies to a web API: does the web API allow guest accounts?.
+        /// </summary>
+        /// <remarks>
+        /// In the app registration for your web API, you'll need to opt-in for the acct claim.
+        /// </remarks>
+        public bool ApiAllowsGuestAccounts { get; set; } = true;
+
+        /// <summary>
+        /// Applies to web apps and web APIs. Enables to provide the list
+        /// of tenant Ids the user of whom are allowed to call the web API
+        /// or log-in to the web app.
+        /// </summary>
+        public IEnumerable<string>? AllowedTenantIds { get; set; }
+
+        /// <summary>
+        /// Applies to web APIs. Enables to provide the list
+        /// of client applications allowed to call this API.
+        /// </summary>
+        public IEnumerable<string>? AllowedClientApplications { get; set; }
     }
 }
--- a/src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
+++ b/src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
@@ -158,7 +158,7 @@ namespace Microsoft.Identity.Web
         public IEnumerable<string>? AllowedTenantIds { get; set; }
 
         /// <summary>
-        /// Applies to web APIs. Enables to provide the list
+        /// Applies to web APIs. Enables you to provide the list
         /// of client applications allowed to call this API.
         /// </summary>
         public IEnumerable<string>? AllowedClientApplications { get; set; }
--- a/src/Microsoft.Identity.Web/Resource/ITokenAuthorizationOptions.cs
+++ b/src/Microsoft.Identity.Web/Resource/ITokenAuthorizationOptions.cs
@@ -1,0 +1,42 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Collections.Generic;
+
+namespace Microsoft.Identity.Web
+{
+    /// <summary>
+    /// Properties related to authorization of access tokens in web APIs.
+    /// </summary>
+    public interface ITokenAuthorizationOptions
+    {
+        /// <summary>
+        /// Applies to a web API: does the web API allow app-only tokens?.
+        /// </summary>
+        /// <remarks>
+        /// In the app registration you'll need to opt-in for the idtyp claim.
+        /// </remarks>
+        public bool ApiAllowsAppOnlyTokens { get; set; }
+
+        /// <summary>
+        /// Applies to a web API: does the web API allow guest accounts?.
+        /// </summary>
+        /// <remarks>
+        /// In the app registration for your web API, you'll need to opt-in for the acct claim.
+        /// </remarks>
+        public bool ApiAllowsGuestAccounts { get; set; }
+
+        /// <summary>
+        /// Applies to web apps and web APIs. Enables to provide the list
+        /// of tenant Ids the user of whom are allowed to call the web API
+        /// or log-in to the web app.
+        /// </summary>
+        public IEnumerable<string>? AllowedTenantIds { get; set; }
+
+        /// <summary>
+        /// Applies to web APIs. Enables to provide the list
+        /// of client applications allowed to call this API.
+        /// </summary>
+        public IEnumerable<string>? AllowedClientApplications { get; set; }
+    }
+}
--- a/src/Microsoft.Identity.Web/Resource/TokenAuthorizationRules.cs
+++ b/src/Microsoft.Identity.Web/Resource/TokenAuthorizationRules.cs
@@ -36,7 +36,7 @@ namespace Microsoft.Identity.Web.Resource
             // Does the web API disallow app only tokens?
             if (!tokenAuthorizationOptions.ApiAllowsAppOnlyTokens)
             {
-                VerifyClaimDoesNotExistOrDoesNotEquals(claimsPrincipal, ClaimConstants.Idtyp, "app", IDWebErrorMessage.ApiAllowsAppOnlyTokensCantBeComputed, IDWebErrorMessage.ApiDoesNotAllowsAppOnlyTokens);
+                VerifyClaimDoesNotExistOrDoesNotEqual(claimsPrincipal, ClaimConstants.Idtyp, "app", IDWebErrorMessage.ApiAllowsAppOnlyTokensCantBeComputed, IDWebErrorMessage.ApiDoesNotAllowsAppOnlyTokens);
             }
 
             // Does the web API disallow guest accounts?
@@ -89,7 +89,7 @@ namespace Microsoft.Identity.Web.Resource
                 throw new UnauthorizedAccessException(messageWhenClaimsNotFound);
             }
 
-            if (string.Equals(claimValue, expectedValue, StringComparison.OrdinalIgnoreCase))
+            if (!string.Equals(claimValue, expectedValue, StringComparison.OrdinalIgnoreCase))
             {
                 throw new UnauthorizedAccessException(messageWhenUnexpectedClaimValue);
             }
@@ -103,7 +103,7 @@ namespace Microsoft.Identity.Web.Resource
         /// <param name="expectedValue">Expected value.</param>
         /// <param name="messageWhenClaimsNotFound">Error message thrown when the claim is not found.</param>
         /// <param name="messageWhenUnexpectedClaimValue">Error message thrown when claim value is not the expected one.</param>
-        private static void VerifyClaimDoesNotExistOrDoesNotEquals(
+        private static void VerifyClaimDoesNotExistOrDoesNotEqual(
             ClaimsPrincipal claimsPrincipal,
             string claimName,
             string expectedValue,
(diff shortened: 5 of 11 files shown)


src/Microsoft.Identity.Web/Resource/TokenAuthorizationRules.cs
```diff
@@ -0,0 +1,150 @@
...
+            // Does the web API disallow app only tokens?
+            if (!tokenAuthorizationOptions.ApiAllowsAppOnlyTokens)
+            {
+                VerifyClaimDoesNotExistOrDoesNotEquals(claimsPrincipal, ClaimConstants.Idtyp, "app", IDWebErrorMessage.ApiAllowsAppOnlyTokensCantBeComputed, IDWebErrorMessage.ApiDoesNotAllowsAppOnlyTokens);
+            }
+
+            // Does the web API disallow guest accounts?
+            if (!tokenAuthorizationOptions.ApiAllowsGuestAccounts)
+            {
+                VerifyClaimEquals(claimsPrincipal, ClaimConstants.Acct, "0", IDWebErrorMessage.ApiAllowsGuestAccountsCantBeComputed, IDWebErrorMessage.ApiDoesNotAllowsGuestAccounts);
```
hpsin, line 45: Validated, correct value is indeed 0.

src/Microsoft.Identity.Web/Resource/TokenAuthorizationRules.cs
```diff
@@ -0,0 +1,150 @@
...
+            {
+                VerifyClaimIsInAcceptedValues(
+                    claimsPrincipal,
+                    new[] { ClaimConstants.Tid, ClaimConstants.TenantId },
+                    tokenAuthorizationOptions.AllowedTenantIds,
+                    IDWebErrorMessage.ApiAllowedTenantsCantBeComputed,
+                    IDWebErrorMessage.ApiTenantNotAllowed);
+            }
+
+            // Does the web API demand specific clients (oid)
```
hpsin, line 59: specific clients or users

src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
```diff
@@ -133,5 +133,34 @@ internal bool HasClientCredentials
...
         /// </summary>
         public string? UserAssignedManagedIdentityClientId { get; set; }
+
+        /// <summary>
+        /// Applies to a web API: does the web API allow app-only tokens?.
+        /// </summary>
+        /// <remarks>
+        /// In the app registration you'll need to opt-in for the idtyp claim.
+        /// </remarks>
+        public bool ApiAllowsAppOnlyTokens { get; set; } = true;
```
yordan-msft, line 143: should not be needed for authz. If app roles are used, the role could be assigned to any security principal

src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
```diff
@@ -133,5 +133,34 @@ internal bool HasClientCredentials
...
+
+        /// <summary>
+        /// Applies to a web API: does the web API allow guest accounts?.
+        /// </summary>
+        /// <remarks>
+        /// In the app registration for your web API, you'll need to opt-in for the acct claim.
+        /// </remarks>
+        public bool ApiAllowsGuestAccounts { get; set; } = true;
```
yordan-msft, line 151: guests are identified with a WID in the token. There should be no need to handle them explicitly, Authorize on the WID, or lack of it.

src/Microsoft.Identity.Web/Resource/TokenAuthorizationRules.cs
```diff
@@ -0,0 +1,150 @@
...
+            if (tokenAuthorizationOptions.AllowedClientApplications != null)
```
hpsin, line 60: This will accept both user and SP OIDs. We _should_ use OID for app only token valdation, but for user validation we should be using roles first, then groups, then oid.

src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
```diff
@@ -133,5 +133,34 @@ internal bool HasClientCredentials
...
+
+        /// <summary>
+        /// Applies to web apps and web APIs. Enables to provide the list
+        /// of tenant Ids the user of whom are allowed to call the web API
+        /// or log-in to the web app.
+        /// </summary>
+        public IEnumerable<string>? AllowedTenantIds { get; set; }
```
yordan-msft, line 158: It is either single or MT app. Consider including the tenant ID in policy definition instead of offering a method.

src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
```diff
@@ -133,5 +133,34 @@ internal bool HasClientCredentials
...
+
+        /// <summary>
+        /// Applies to web APIs. Enables to provide the list
+        /// of client applications allowed to call this API.
+        /// </summary>
+        public IEnumerable<string>? AllowedClientApplications { get; set; }
```
yordan-msft, line 164: Since we recommend against using appids directly for authZ, I would not offer such method.

src/Microsoft.Identity.Web/Constants/ClaimConstants.cs
```diff
@@ -99,5 +99,15 @@ public static class ClaimConstants
...
         /// </summary>
         public const string NameIdentifierId = "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/nameidentifier";
+
+        /// <summary>
+        /// Optional claim used to determine the token type.
+        /// </summary>
+        public const string Idtyp = "idtyp";
+
+        /// <summary>
+        /// Optional claim used to determine if the user account is a guest account.
```
yordan-msft, line 109: we should not need this. There is corresponding WID claim. Check with Maxim.

src/Microsoft.Identity.Web/Resource/ITokenAuthorizationOptions.cs
```diff
@@ -0,0 +1,42 @@
...
+// Licensed under the MIT License.
+
+using System.Collections.Generic;
+
+namespace Microsoft.Identity.Web
+{
+    /// <summary>
+    /// Properties related to authorization of access tokens in web APIs.
+    /// </summary>
+    public interface ITokenAuthorizationOptions
```
yordan-msft, line 11: can we consider passing in authorization policy.

src/Microsoft.Identity.Web/Constants/ClaimConstants.cs
hpsin: That wids claim is both undocumented and flaky.

src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
hpsin: This is an oid check

src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
hpsin: WIDS claim is unreliable.  We want to do as much as we can using only claims in the token, and sometimes you ne4ed to call Graph for wids and groups.

src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
jmprieur: By default, Microsoft.Identity.Web verifies that a token contains at least roles or scopes, or the web API configuration claims that it's protected by ACLs.
Then the developer can add/should add the verification of roles on each controller or even controller action. But that's work.

@yordan-msft : The role could be assigned to any security principal. How does it relate to asking app-only scopes?
 @hpsin : do we want to have this option?

src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
jmprieur: @hpsin : should it be `AllowedClientOids`? And in that case would it be more for the web APIs called with a user access token?

src/Microsoft.Identity.Web/Resource/ITokenAuthorizationOptions.cs
jmprieur: @yordan-msft : you mean only for ASP.NET Core?
The mechanism we propose here also works for ASP.NET classic (OWIN), and any other framework.

src/Microsoft.Identity.Web/Resource/TokenAuthorizationRules.cs
jmprieur: Ok. @hpsin we'll add a condition. 
Do we still use it for app-only tokens (See Yordan's comment https://github.com/AzureAD/microsoft-identity-web/pull/1264/files#r651153866

src/Microsoft.Identity.Web/WebApiExtensions/MicrosoftIdentityWebApiAuthenticationBuilderExtensions.cs
```diff
@@ -241,6 +242,9 @@ public static partial class MicrosoftIdentityWebApiAuthenticationBuilderExtensio
                             throw new UnauthorizedAccessException(IDWebErrorMessage.NeitherScopeOrRolesClaimFoundInToken);
                         }
 
+                        // Validates the token authorization rules.
+                        TokenAuthorizationRules.ValidateApiAuthorization(mergedOptions, context!.Principal!);
```
jmprieur, line 246: This is all the code that would need to be added to an OnTokenValidated event in an ASP.NET Core web api that wouldn't use Ms.Id.Web, or in an ASP.NET classic app. The first member is a TokenAuthorizationRules (So 4 properties), and the second a ClaimsPrincipal, which is always available after token validation)

src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
```diff
@@ -133,5 +133,34 @@ internal bool HasClientCredentials
...
+        /// <summary>
+        /// Applies to a web API: does the web API allow guest accounts?.
+        /// </summary>
+        /// <remarks>
+        /// In the app registration for your web API, you'll need to opt-in for the acct claim.
+        /// </remarks>
+        public bool ApiAllowsGuestAccounts { get; set; } = true;
+
+        /// <summary>
+        /// Applies to web apps and web APIs. Enables to provide the list
```
jennyf19, line 154:
```suggestion
        /// Applies to web apps and web APIs. Enables you to provide the list
```

src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
```diff
@@ -133,5 +133,34 @@ internal bool HasClientCredentials
...
+        /// of tenant Ids the user of whom are allowed to call the web API
```
jennyf19, line 155:
```suggestion
        /// of tenant Ids allowed to call the web API
```

src/Microsoft.Identity.Web/Constants/IDWebErrorMessage.cs
```diff
@@ -36,6 +38,14 @@ internal static class IDWebErrorMessage
         public const string BlazorServerUserNotSet = "IDW10206: Using Blazor server but the user was not properly set. ";
         public const string CalledApiScopesAreNull = "IDW10207: The CalledApiScopes cannot be null. ";
         public const string ScopesRequiredToCallMicrosoftGraph = "IDW10208: You need to either pass-in scopes to AddMicrosoftGraph, in the appsettings.json file, or with .WithScopes() on the Graph queries. See https://aka.ms/ms-id-web/microsoftGraph. ";
+        public const string ApiAllowsAppOnlyTokensCantBeComputed = "IDW10209: You set the 'ApiAllowsAppOnlyTokens' configuration to false, but didn't opt-in to the 'idtyp' optional claim in the app registration. See https://aka.ms/ms-id-web/webapi-authorization. ";
+        public const string ApiDoesNotAllowsAppOnlyTokens = "IDW10210: This web API does not allow app-only tokens. ";
+        public const string ApiAllowsGuestAccountsCantBeComputed = "IDW10211: You set the 'ApiAllowsGuestAccounts' configuration to false, but didn't opt-in to the 'acct' optional claim in the app registration. See https://aka.ms/ms-id-web/webapi-authorization. ";
```
jennyf19, line 43: what is the resolution for this, if i get this error?

src/Microsoft.Identity.Web/Constants/IDWebErrorMessage.cs
```diff
@@ -36,6 +38,14 @@ internal static class IDWebErrorMessage
...
+        public const string ApiDoesNotAllowsGuestAccounts = "IDW10212: This web API does not allow guest accounts. ";
```
jennyf19, line 44: same as above, what action do i need to take now?

src/Microsoft.Identity.Web/Constants/IDWebErrorMessage.cs
```diff
@@ -36,6 +38,14 @@ internal static class IDWebErrorMessage
...
+        public const string ApiAllowedTenantsCantBeComputed = "IDW10213: You set the 'AllowedTenantIds' configuration, but the `tid` or `tenantId` claims are not found. See https://aka.ms/ms-id-web/webapi-authorization. ";
+        public const string ApiTenantNotAllowed = "IDW10214: This web API does not allow this tenant. ";
+        public const string ApiAllowedClientApplicationsCantBeComputed = "IDW10215: You set the 'AllowedClientApplications' configuration, but the `oid` or `objectid` claims are not found. See https://aka.ms/ms-id-web/webapi-authorization. ";
+        public const string ApiClientApplicationNotAllowed = "IDW10216: This web API does not allow this client application. ";
```
jennyf19, line 48: same, what action should i take? how did i get here?

src/Microsoft.Identity.Web/Resource/TokenAuthorizationRules.cs
```diff
@@ -0,0 +1,150 @@
...
+
+            if (claimsPrincipal is null)
+            {
+                throw new ArgumentNullException(nameof(claimsPrincipal));
+            }
+
+            // Does the web API disallow app only tokens?
+            if (!tokenAuthorizationOptions.ApiAllowsAppOnlyTokens)
+            {
+                VerifyClaimDoesNotExistOrDoesNotEquals(claimsPrincipal, ClaimConstants.Idtyp, "app", IDWebErrorMessage.ApiAllowsAppOnlyTokensCantBeComputed, IDWebErrorMessage.ApiDoesNotAllowsAppOnlyTokens);
```
jennyf19, line 39: >s [](http://example.com/codeflow?start=53&length=1)

nit: remove "s"

src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
yordan-msft: we should talk it over. Is oid check really needed?

src/Microsoft.Identity.Web/Resource/ITokenAuthorizationOptions.cs
yordan-msft: as opposed to what? Is there a mechanism that does not work for other frameworks? Let's talk it over.

src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
hpsin: This is a good "easy win" for apps that aren't able to turn on app role assignment yet.  It provides defense in depth as well.

src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
hpsin: both app only tokens and user+app otkens have an oid.  It's not a particularly usable setup,. so consider removing it entirely.

src/Microsoft.Identity.Web/Resource/TokenAuthorizationRules.cs
psignoret: Should this value (and others, such as "app") maybe be defined in a constant?

src/Microsoft.Identity.Web/Constants/IDWebErrorMessage.cs
hpsin: App needs to opt in to the acct claim in the optional claims section of the app reg

src/Microsoft.Identity.Web/Constants/IDWebErrorMessage.cs
hpsin: no action needed, this is a desired outcome.  "Go away" basically.

src/Microsoft.Identity.Web/Resource/TokenAuthorizationRules.cs
```diff
@@ -0,0 +1,150 @@
...
+
+        /// <summary>
+        /// Verify the value of a claim.
+        /// </summary>
+        /// <param name="claimsPrincipal">Claims.</param>
+        /// <param name="possibleClaimNames">Names (types) of the claims to verify.</param>
+        /// <param name="acceptedValues">Expected values.</param>
+        /// <param name="messageWhenClaimsNotFound">Error message thrown when the claim is not found.</param>
+        /// <param name="messageWhenUnexpectedClaimValue">Error message thrown when claim value is not the expected one.</param>
+        private static void VerifyClaimIsInAcceptedValues(
```
psignoret, line 131: Won't this throw an exception when the claim _does_ contain an expected value?

src/Microsoft.Identity.Web/Resource/TokenAuthorizationRules.cs
```diff
@@ -0,0 +1,150 @@
...
+            ClaimsPrincipal claimsPrincipal,
+            string claimName,
+            string expectedValue,
+            string messageWhenClaimsNotFound,
+            string messageWhenUnexpectedClaimValue)
+        {
+            string? claimValue = ClaimsPrincipalExtensions.GetClaimValue(claimsPrincipal, claimName);
+            if (claimValue != null)
+            {
+                if (!string.Equals(claimValue, expectedValue, StringComparison.OrdinalIgnoreCase))
```
psignoret, line 116: This should throw if the value _is_ equal to `expectedValue`, right? (Also, that parameter should probably be named something else, since here we're checking that the value is _not_ this?)

src/Microsoft.Identity.Web/Resource/TokenAuthorizationRules.cs
```diff
@@ -0,0 +1,150 @@
...
+
+        /// <summary>
+        /// Verify the value of a claim.
+        /// </summary>
+        /// <param name="claimsPrincipal">Claims.</param>
+        /// <param name="claimName">Name (type) of the claim to verify.</param>
+        /// <param name="expectedValue">Expected value.</param>
+        /// <param name="messageWhenClaimsNotFound">Error message thrown when the claim is not found.</param>
+        /// <param name="messageWhenUnexpectedClaimValue">Error message thrown when claim value is not the expected one.</param>
+        private static void VerifyClaimEquals(
```
psignoret, line 79: This throws an exception fi the claim _does_ equal the expected value. It should throw if it does _not_, right?

psignoret (changes requested):
I might be misreading, but it looks like all the claim value checks are backwards.

src/Microsoft.Identity.Web/Constants/IDWebErrorMessage.cs
hpsin: As above.  This isn't' an error to fix, so reading between the lines perhaps it shouldn't be an exception, but a response?

src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
jmprieur: you mean an ASP.NET Core policy, @yordan-msft ?

src/Microsoft.Identity.Web/Resource/ITokenAuthorizationOptions.cs
jmprieur: What's the data structure for an authorization policy, @yordan-msft ?

src/Microsoft.Identity.Web/Resource/TokenAuthorizationRules.cs
jmprieur: Do we still agree to remove this altogether, @hpsin?

src/Microsoft.Identity.Web/Resource/TokenAuthorizationRules.cs
jmprieur: Yes. (this is a draft PR for a proof of concept for the moment). Good catch, @psignoret

src/Microsoft.Identity.Web/Resource/TokenAuthorizationRules.cs
```diff
@@ -0,0 +1,150 @@
...
+
+        /// <summary>
+        /// Verify the value of a claim.
+        /// </summary>
+        /// <param name="claimsPrincipal">Claims.</param>
+        /// <param name="claimName">Name (type) of the claim to verify.</param>
+        /// <param name="expectedValue">Expected value.</param>
+        /// <param name="messageWhenClaimsNotFound">Error message thrown when the claim is not found.</param>
+        /// <param name="messageWhenUnexpectedClaimValue">Error message thrown when claim value is not the expected one.</param>
+        private static void VerifyClaimDoesNotExistOrDoesNotEquals(
```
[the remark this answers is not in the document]
jmprieur: Actually, I'm not sure. If we want to have verification that we don't accept app-only tokens, the only way to do it is:
- either the claim does not exist (as it's not generated if the token is not an app-only token)
- or it exists and its value is "app"
Or am I missing something, @psignoret @hpsin ?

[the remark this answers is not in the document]
jmprieur: If we wanted to demand an app-only token, then we'd need to have the claim and it's value should be app.

tests/WebAppCallsWebApiCallsGraph/TodoListService/appsettings.json
```diff
@@ -7,22 +7,17 @@
...
                 "Container": "",
                 "ReferenceOrValue": ""
             }
-        ]
+        ],
+        "Audience": "a4c2469b-cf84-4145-8f5f-cb7bacf814bc",
+        "ApiAllowsAppOnlyTokens": "false",
+        "ApiAllowsGuestAccounts": "false",
+        "AllowedTenantIds": [ "7f58f645-c190-4ce5-9de4-e2b7acd2a6ab" ],
+        "AllowedClientApplications": [ "86699d80-dd21-476a-bcd1-7c1a3d471f75" ]
```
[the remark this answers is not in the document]
jmprieur: No, it's just a proof of concept, not tested yet. This should be the object ID, not the app ID.

src/Microsoft.Identity.Web/Resource/TokenAuthorizationRules.cs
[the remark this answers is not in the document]
psignoret: This method is used when we _don't_ want to allow app-only tokens (because `ApiAllowsAppOnlyTokens` is `false`). To know for sure, we need to verify that the `idtyp` claim is present, AND we need it's value to _not_ be "app". So, we should throw an exception if the claim is not present, OR if it's present and the value is "app".

src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
yordan-msft: I disagree with app having to reason about app only or types. AuthZ is done on identifiers. If app does authZ on OIDs and groups, the principal can be anything. There is no need to exclude app only. 
Seems like we are adding this option as workaround for apps doing something wrong. If this is the case, call out what this something is.

src/Microsoft.Identity.Web/Resource/TokenAuthorizationRules.cs
jmprieur: Thanks @psignoret. Fixed

src/Microsoft.Identity.Web/Resource/TokenAuthorizationRules.cs
jmprieur: Fixed

jennyf19:
@jmprieur this seems stale, so closing

jmprieur:
thanks @jennyf19 
yes it was a precursor to the AuthZ module

PR #1266, opened 2021-06-15
Jennyf/fix legacy

jennyf19:
Fixes a regression that was introduced with the Multi-scheme work where the `LegacyCacheCompatibilityEnabled` value was taken from the ConfidentialClientApplicationOptions (default true), instead of the MicrososftIdentityOptions (default false)


jmprieur (approved):
Thanks @jennyf19 
LGTM

jennyf19:
https://github.com/AzureAD/microsoft-identity-web/issues/1268

PR #1270, opened 2021-06-16
update dev app to test obo rt

jennyf19:


tests/WebAppCallsWebApiCallsGraph/TodoListService/Controllers/Callback.cs
```diff
@@ -0,0 +1,69 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
jmprieur, line 1: You might want to rename this file to CallbackController.cs

tests/WebAppCallsWebApiCallsGraph/TodoListService/Controllers/Callback.cs
```diff
@@ -0,0 +1,69 @@
...
+
+namespace TodoListService.Controllers
+{
+    [Authorize]
+    [Route("api/[controller]")]
+    //[RequiredScope("access_as_user")] 
+    public class CallbackController : Controller
+    {
+        private readonly ITokenAcquisition _tokenAcquisition; // do not remove
+        // The Web API will only accept tokens 1) for users, and 2) having the access_as_user scope for this API
```
jmprieur, line 28: We can remove this comment. This controller allows anonymous.

tests/WebAppCallsWebApiCallsGraph/TodoListService/Controllers/LongRunningProcessContext.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    public class LongRunningProcessContext : IDisposable
+    {
+        JwtSecurityToken _savedToken;
+        HttpContext _httpContext;
+
+        internal LongRunningProcessContext(HttpContext httpContext, JwtSecurityToken tokenToUse)
+        {
+            _httpContext = httpContext;
+            _savedToken = httpContext.Items["JwtSecurityTokenUsedToCallWebAPI"] as JwtSecurityToken;
```
jmprieur, line 15: Too bad we cannot use (as they are internal).
https://github.com/AzureAD/microsoft-identity-web/blob/800b833ddf788032ae885807dbbceb7581db8852/src/Microsoft.Identity.Web/HttpContextExtensions.cs#L17

and

https://github.com/AzureAD/microsoft-identity-web/blob/800b833ddf788032ae885807dbbceb7581db8852/src/Microsoft.Identity.Web/HttpContextExtensions.cs#L31

We also have an issue where the customers would like us to Store the incoming token somewhere.

tests/WebAppCallsWebApiCallsGraph/TodoListService/Controllers/LongRunningProcessContextFactory.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// Get a key associated with the current incoming token
+        /// </summary>
+        /// <param name="httpContext">Http context in which the controller action is running</param>
+        /// <returns>A unique string repesenting the incoming token to the web API. This
+        /// key will be used in the future to retrieve the incoming token even if it has expired therefore
+        /// enabling getting an OBO token.</returns>
+        public string CreateKey(HttpContext httpContext)
+        {
+            JwtSecurityToken token = httpContext.Items["JwtSecurityTokenUsedToCallWebAPI"] as JwtSecurityToken;
+            string key = "1";
```
jmprieur, line 33: you might want to have something else here.
"1" was nice to test, but probably have one per user would be nice?

Maybe key = `httpContext.GetMsalId()` or `Guid.NewGuid().string()`

jmprieur (approved):
LGTM.
Thanks!

tests/WebAppCallsWebApiCallsGraph/TodoListService/Controllers/Callback.cs
jennyf19: done.

tests/WebAppCallsWebApiCallsGraph/TodoListService/Controllers/Callback.cs
jennyf19: done.

tests/WebAppCallsWebApiCallsGraph/TodoListService/Controllers/LongRunningProcessContext.cs
jennyf19: agree...should we open an issue?

tests/WebAppCallsWebApiCallsGraph/TodoListService/Controllers/LongRunningProcessContextFactory.cs
jennyf19: yep. thanks.

PR #1272, opened 2021-06-18
Adding more logging to the long running sample

jmprieur:
Logging every minute
More Logging in the callback:
- display the  URL to call the callback
- display the expiry of the OBO token


jennyf19 (approved):
cool.

Issue #1131, opened 2021-04-09
Add support for DelegatingHandler


qetza:
**Is your feature request related to a problem? Please describe.**
Currently for using Microsoft.Identity.Web when calling APIs you have to use `IDownstreamWebApi` or do it manually through  `ITokenAcquisition`. This can be complicated/time consuming for existing applications.

**Describe the solution you'd like**
It would be a good solution to provide a DelegatingHandler which would use `ITokenAcquisition` to get a token and inject it in the Authorization HTTP headers. This way you only have to update your Startup class where you register your HTTP clients.
This can also benefit for teams that prefer managing all HTTP connection transversal topics through DelegatingHandler (Polly...).

**Describe alternatives you've considered**

**Additional context**
I've already discuss this with Jean-Marc and as I have a basic implementation done I will send PR.

jmprieur:
The options to call an API would then because:
- If you want to use Microsoft Graph, inject `GraphServiceClient`
- If you want to call a REST API handling Json and don’t have an SDK, use `IDownstreamWebApi`
- If you are using an SDK:
  - If this is an Azure SDK, use the `TokenAcquisitionTokenCredential` which was introduced in 1.9
  - If this is another SDK that takes an Http delegate handler, then use a (new class to be provided) `TokenAcquisitionDelegateHandler`
  - If this is another SDK that requires a token, just use `ITokenAcquisition`.

qetza:
Any idea when the PR will be included and released?

jennyf19:
Included in [1.15 release](https://github.com/AzureAD/microsoft-identity-web/releases/tag/1.15.0)

PR #1276, opened 2021-06-20
Add message handlers for downstream api auth

qetza:
Authentication message handlers to inject authorization header using token acquisition.


qetza:
This is for the feature request #1131

src/Microsoft.Identity.Web/DownstreamWebApiSupport/MicrosoftIdentityAuthenticationMessageHandlerOptions.cs
```diff
@@ -0,0 +1,88 @@
...
+// Licensed under the MIT License.
+
+using System;
+
+namespace Microsoft.Identity.Web
+{
+    /// <summary>
+    /// Options passed-in to Microsoft Identity message handlers.
+    /// </summary>
+    public class MicrosoftIdentityAuthenticationMessageHandlerOptions : ICloneable
```
jmprieur, line 11: @qetza why not use the TokenAcquisitionOptions directly (maybe inheriting from them) ?

src/Microsoft.Identity.Web/DownstreamWebApiSupport/MicrosoftIdentityUserAuthenticationMessageHandler.cs
```diff
@@ -0,0 +1,71 @@
...
+            var authResult = await TokenAcquisition.GetAuthenticationResultForUserAsync(
+                options.GetScopes(),
+                authenticationScheme: options.AuthenticationScheme,
+                tenantId: options.Tenant,
+                userFlow: userflow,
+                tokenAcquisitionOptions: options.TokenAcquisitionOptions)
+                .ConfigureAwait(false);
+
+            // add authorization header
+            request.Headers.Add(
```
jmprieur, line 64: Are we sure the request won't already have an Authorization header? 
(would we want to use `[ ]` ?

jmprieur (commented):
Thanks @qetza 
This looks good to me.

How would you use the message handler?
Would you have a sample app? unit tests? 

Thanks!

qetza:
Hi @jmprieur,
I will add some unit tests and copy an existing downstream webapi sample and rewrite it with using the message handler instead of the DownstreamWebApi

src/Microsoft.Identity.Web/DownstreamWebApiSupport/MicrosoftIdentityAuthenticationMessageHandlerOptions.cs
qetza: @jmprieur, i think you mean reusing `DownstreamWebApiOptions`. This class effectively is a subset of `DownstreamWebApiOptions`, i could make `DownstreamWebApiOptions` inherit from this one and add its specific properties.

src/Microsoft.Identity.Web/DownstreamWebApiSupport/MicrosoftIdentityUserAuthenticationMessageHandler.cs
qetza: Indexer property `[ ]` is not supported on `HttpHeaders` but i will add a test to remove an existing headers if already present.

qetza:
Hi @jmprieur,
I've refactor the code to have base options class between downstream and delegating handlers, fix the issue with existing header and added some unit tests.
For the sample part as they are in another repo i'll do a PR there later.

jmprieur:
Thanks @qetza

qetza:
I've rebased the code on the lastest version of `master` to resolve merge conflict.

jmprieur:
Sorry @qetza : I missed that you had added tests.
We'll review. thanks for the head-up.

src/Microsoft.Identity.Web/DownstreamWebApiSupport/MicrosoftIdentityAppAuthenticationMessageHandler.cs
```diff
@@ -0,0 +1,64 @@
...
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Identity.Web
+{
+    /// <summary>
+    /// A DelegatingHandler implementation that add an authorization header with a token for the application.
```
jennyf19, line 14: >add [](http://example.com/codeflow?start=48&length=3)

nit: adds

jennyf19 (approved):
:shipit:

jennyf19:
thanks @qetza i, too, missed the updates here. Could you resolve the merge conflict and then we will merge it? thanks so much.

qetza:
@jennyf19, i've rebased the branch and fixed the merge conflict (on `Microsoft.Identity.Web.xml` as usual)

NWessel:
@qetza Could you provide a link to a repo where you use it?
I really want to try it out right now :)

qetza:
Hi @NWessel,
I've submitted a pull request on the sample repo with an example of using the `MicrosoftIdentityUserAuthenticationMessageHandler` class.

You can see the code here:
https://github.com/qetza/active-directory-aspnetcore-webapp-openidconnect-v2/blob/142a315a36293e25ba5e4fd42640a1a624519d13/3-WebApp-multi-APIs/Startup.cs#L49

NWessel:
Just checked it out, looks like what i would expect it to be ^^
Probably could make the scope optional as well, and take the base url of the request and add /.default

So baseUrl + "/.default"
If no scope is provided.

What is the service name for ? I'm not sure I quite understand the usecase :)

qetza:
You can always create your own extension method without a scope and do the magic in there :smile:

The service name is used internally to link and retrieve the options if you have multiple clients which needs different options.

PR #1281, opened 2021-06-21
update msal 4.33

jennyf19:
tested web app calls web api and b2c.


jmprieur (approved):
Thanks @jennyf19 
LGTM

PR #1284, opened 2021-06-22
remove "s" in "caches"

jennyf19:
@jmprieur idk if i like "caches" anymore...maybe just "cache"? your thoughts?


jmprieur (approved):
Lgtm
Thanks @jennyf19

We might want to add in comments that this adds both the app and user token cache?

Issue #1296, opened 2021-07-03
[Bug] Additional claims are not being requested


DevEnable:
**Which version of Microsoft Identity Web are you using?**
Issue first identified in 1.12 but still present in 1.14.

**Where is the issue?**
* Web app
    * [x] Sign-in users
    * [ ] Sign-in users and call web APIs
* Web API
    * [ ] Protected web APIs (validating tokens)
    * [ ] Protected web APIs (validating scopes)
    * [ ] Protected web APIs call downstream web APIs
* Token cache serialization
     * [ ] In-memory caches
     * [ ] Session caches
     * [ ] Distributed caches
* Other (please describe)

**Is this a new or an existing app?**
The app is in production and I have upgraded to a new version of Microsoft Identity Web.

**Repro**
Configure the scopes (either in code or in a config file) to add an additional scope (e.g. email) beyond the default openid and profile scopes. 

The expected email claim will not be returned and when looking at the request using a web proxy (e.g. Fiddler) the scope will not be part of the request.

I have tried this with both the AddMicrosoftIdentityWebAppAuthentication and AddMicrosoftIdentityWebApp methods.

Example config:
```json
"AzureAd": {
    "Instance": "https://login.microsoftonline.com/",
    "Domain": "[domain].onmicrosoft.com",
    "TenantId": "organizations",
    "ClientId": "xxx",
    "ClientSecret": "xxx",
    "CallbackPath": "/signin-oidc",
    "SignedOutCallbackPath ": "/signout-callback-oidc",
    "Scope": [
      "openid",
      "profile",
      "email"
    ]
  },
```

```csharp
authServices.AddMicrosoftIdentityWebApp(options =>
            {
                _configuration.Bind("AzureAd", options);
            }, openIdConnectScheme: ExternalProviders.AzureActiveDirectory, cookieScheme: ExternalProviders.GetCookieScheme(ExternalProviders.AzureActiveDirectory));
```

**Expected behavior**
The additional email scope should be part of the authorise request as a query string parameter in the request to login.microsoftonline.com.

**Actual behavior**
Only the openid and profile scopes are provided, which was verified by using Fiddler. Therefore, the email claim is not provided in a response.

E.g. The request is: https://login.microsoftonline.com/organizations/oauth2/v2.0/authorize?client_id=xx&redirect_uri=xxx&response_type=id_token&scope=openid%20profile&response_mode=form_post&nonce=xx&client_info=1 etc...

**Possible solution**
[...]

PR #1297, opened 2021-07-05
fix for #1296

jennyf19:


jmprieur (approved):
LGTM.
Thanks @jennyf19 

I'll have a look tomorrow at the other collections

src/Microsoft.Identity.Web/MergedOptions.cs
```diff
@@ -236,6 +237,14 @@ internal static void UpdateMergedOptionsFromMicrosoftIdentityOptions(MicrosoftId
 
             mergedOptions.UseTokenLifetime = microsoftIdentityOptions.UseTokenLifetime;
             mergedOptions._confidentialClientApplicationOptions = null;
+
+            foreach (var scope in microsoftIdentityOptions.Scope)
```
jmprieur, line 241: @jennyf19 
Thinking more, we need to decide if we want to add to the configuration or replace the configuration.

For instance, if by default, the scopes as set by ASP.NET Core are "openid profile offline_access", but a customer doesn't want to have offline_access and profile, do we want to be able to say

```CSharp
services.Configure<MicrosoftIdentityOptions>(OpenIdConnectDefault.AuthenticationScheme, options =>
{
 options.Scopes.Clear();
 options.Scopes.AddRange(new [] {"openid email" })
});
```

as with this current proposed implementation, we would get "openid profile offline_access email"
whereas the developer might intend only "openid email" 

This also applies to any collection in MergedOptions.

jmprieur (approved):
LGTM 
thanks @jennyf19

Issue #1310, opened 2021-07-12
[Bug] Bearer error="invalid_token", error_description="The issuer '(null)' is invalid" in v1.14.0


ThomasBarnekow:
**Which version of Microsoft Identity Web are you using?**
v1.14.0

**Where is the issue?**
* Web app
    * [ ] Sign-in users
    * [ ] Sign-in users and call web APIs
* Web API
    * [X] Protected web APIs (validating tokens)
    * [ ] Protected web APIs (validating scopes)
    * [ ] Protected web APIs call downstream web APIs
* Token cache serialization
     * [ ] In-memory caches
     * [ ] Session caches
     * [ ] Distributed caches
* Other (please describe)

**Is this a new or an existing app?**
This is an app under active development for which I have successfully used v1.12.0.

**Repro**
As there is no change in my code, the repro would be to update a working Web API from v1.12.0 to v1.14.0 and observe that it suddenly fails to validate tokens generated by both Postman v8.7.0 and MSAL for Angular:
```
"@azure/msal-angular": "^2.0.0",
"@azure/msal-browser": "^2.14.2",
```

On the server side, I am using ASP.NET Core 5.0 with the following configuration:

**appsettings.json**
```
"AzureAd": {
  "Instance": "https://login.microsoftonline.com/",
  "ClientId": "[guid]",
  "ClientSecret": "[removed]",
  "TenantId": "organizations"
},
```

**Startup.ConfigureServices(IServiceCollection services)**
```csharp
services.AddAuthentication(JwtBearerDefaults.AuthenticationScheme)
    .AddMicrosoftIdentityWebApi(Configuration.GetSection("AzureAd"))
    .EnableTokenAcquisitionToCallDownstreamApi()
    .AddMicrosoftGraph(Configuration.GetSection("Graph"))
    .AddInMemoryTokenCaches();

services.AddAuthorization(options =>
{
    options.AddPolicy("MinimumScope", policy => policy.RequireClaim("scp", "access_as_user"));
});
```

**Startup.Configure(IApplicationBuilder app, IWebHostEnvironment env)**
```csharp
public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
{
    if (env.IsDevelopment())
    {
        app.UseDeveloperExceptionPage();
    }

    app.UseSerilogRequestLogging();

    app.UseHttpsRedirection();
    app.UseRouting();

    // The UseCors call must be placed between UseRouting and UseAuthentication.
[...]

Issue #1290, opened 2021-06-23
[Bug] Cannot authenticate with managed identity to multi-tenant web api.


henriblMSFT:
**Which version of Microsoft Identity Web are you using?**
Microsoft Identity Web 1.14.0
Microsoft.Identity.Web 1.13.1
Does not repro with version 1.12.0

**Where is the issue?**
* Web API
    * [X] Protected web APIs (validating tokens)

**Is this a new or an existing app?**
a. The app is in production and I have upgraded to a new version of Microsoft Identity Web.

**Repro**
I've observed the issue attempting to authenticate with an azure managed service identity. The issue also reproduces when authenticating as a user with the MSAL library.

```csharp
services
  .AddAuthentication(JwtBearerDefaults.AuthenticationScheme)
  .AddMicrosoftIdentityWebApi(config);
```

appsettings.json
```json
"AzureAd": {
  "Instance": "https://login.microsoftonline.com/",
  "ClientId": "84846a08-b110-44f0-92af-a353b2e94d30",
  "TenantId": "organizations"
},
```

Here are the claims for the managed identity (note that all guids have been changed)
```json
{
  "typ": "JWT",
  "alg": "RS256",
  "x5t": "nOo3ZDrODXEK1jKWhXslHR_KXEg",
  "kid": "nOo3ZDrODXEK1jKWhXslHR_KXEg"
}.{
  "aud": "api://84846a08-b110-44f0-92af-a353b2e94d30",
  "iss": "https://sts.windows.net/7a797836-d1e3-4249-a68d-f2cb8e4f58a5/",
  "iat": 1624403649,
  "nbf": 1624403649,
  "exp": 1624490349,
  "aio": "E2ZgYOAxb73SVu/30LmxqEVvndJ/AA==",
  "appid": "a6af1a81-a127-4797-90eb-2bfa39d4b61e",
  "appidacr": "2",
  "idp": "https://sts.windows.net/7a797836-d1e3-4249-a68d-f2cb8e4f58a5/",
  "oid": "45f9fc44-a2d0-45f4-b123-144298ea2c4b",
  "rh": "0.ARIA6q7FzcUVtk2wefyt0lBdwjt-PftAYDFHvRw_czfWeosSAAA.",
  "roles": [
    "ReadWrite.All"
  ],
  "sub": "45f9fc44-a2d0-45f4-b123-144298ea2c4b",
[...]

PR #1312, opened 2021-07-13
update aadissuer for organizations

jennyf19:
#1290 and #1310


jmprieur (approved):
Awesome, @jennyf19

I've tested with:
-  v1.0 tokens, tenanted and organizations authority

- v2.0  tokens , tenanted, oorgnaiz

Issue #1304, opened 2021-07-08
[Feature Request] Use the new `SuggestedCacheKeyExpiry` in MSAL.NET 4.34.0 to propose eviction in the app token cache 


jmprieur:
**Is your feature request related to a problem? Please describe.**
MSAL.NET 4.34.0 proposes a new property `SuggestedCacheExpiry ` on `TokenCacheNotificationArgs `:

```CSharp
    public sealed class TokenCacheNotificationArgs {
        public Nullable<DateTimeOffset> SuggestedCacheExpiry { get; }
    }
```

It will be null most of the time, except for AcquireTokenForClient (app token cache)
Use this property when it's not null, to optimize the cache eviction. We could at least do it in the in-memory cache.

**Describe the solution you'd like**

- [ ] Add the SuggestedCacheExpiry property to the [CacheSerializerHints](https://github.com/AzureAD/microsoft-identity-web/blob/b348c484aa0a86c460f99e36a8775ac6bc240269/src/Microsoft.Identity.Web/TokenCacheProviders/CacheSerializerHints.cs)
- [ ] Update the `MsalAbstractTokenCacheProvider.CreateHintsFromArgs` method to add this new hint: https://github.com/AzureAD/microsoft-identity-web/blob/d126768ed1e5be62e882d8d705ddc84373f266ae/src/Microsoft.Identity.Web/TokenCacheProviders/MsalAbstractTokenCacheProvider.cs#L69

**Describe alternatives you've considered**
Leave it to the distributed cache serializers to perform the eviction based on developers settings

jennyf19:
Included in [1.14.1 release](https://github.com/AzureAD/microsoft-identity-web/releases/tag/1.14.1)

PR #1317, opened 2021-07-14
use cache expiry

jennyf19:
#1304 
tested w/web app calls web API w/redis, so the cache key is null always.


jmprieur (approved):
LGTM
Tested with https://github.com/Azure-Samples/active-directory-dotnet-v1-to-v2 with both MemoryCache and DistributedMemoryCache

Issue #1044, opened 2021-03-04
[Feature Request] Encryption strategy for Distributed token cache


jmprieur:
**Is your feature request related to a problem? Please describe.**
Today the distributed token cache does not encrypt the blob for caches.

**Describe the solution you'd like**
Provide an optional way to enable apps developer to provide/choose an encryption strategy

jmprieur:
One possibility: https://docs.microsoft.com/en-us/aspnet/core/security/data-protection/introduction

Configuration: https://docs.microsoft.com/en-us/aspnet/core/security/data-protection/configuration/overview?view=aspnetcore-5.0

for instance:
```CSharp
   // In startup.cs / or when adding cache
   services.AddDataProtection() 
      .PersistKeysToAzureBlobStorage(dataProtectionConnectionString, dataProtectionContainerName, dataProtectionKeyFile)                
      .ProtectKeysWithAzureKeyVault(new Uri(dataProtectionKeyKeyUri), cred)                
      .SetApplicationName(AuthenticationConstants.ApplicationName);

    // In the controller etc ...
    var dataProtectionProvider = serviceProvider.GetService<IDataProtectionProvider>
```

jmprieur:
Proposed experience.

Customers who want to benefit from cache encryption will add a few lines in their startup.cs:

```CSharp
   services.AddDataProtection() 
      .PersistKeysToAzureBlobStorage(dataProtectionConnectionString, dataProtectionContainerName, dataProtectionKeyFile)                
      .ProtectKeysWithAzureKeyVault(new Uri(dataProtectionKeyKeyUri), cred)                
      .SetApplicationName(AuthenticationConstants.ApplicationName);
```

Proposed design:

- [ ] Inject `IServiceProvider` in the `MsalAbstractTokenCacheProvider` constructor. Have a null default value to avoid a breaking change in existing token caches deriving from `MsalAbstractTokenCacheProvider`  and not provided by us
- [ ] Add a member `dataProtectionProvider` of type `IDataProtectionProvider` computed from the service provider argument when it's not null
- [ ] Before writing, if `dataProtectionProvider` is not null, encrypt the cache

  https://github.com/AzureAD/microsoft-identity-web/blob/cc518815943b3c05be328dd3af88288f4fe8b8c5/src/Microsoft.Identity.Web/TokenCacheProviders/MsalAbstractTokenCacheProvider.cs#L59

[...]

PR #1323, opened 2021-07-19
provide option to encrypt distributed cache

jennyf19:
Fix for https://github.com/AzureAD/microsoft-identity-web/issues/1044

### DIFF
--- a/src/Microsoft.Identity.Web/Microsoft.Identity.Web.csproj
+++ b/src/Microsoft.Identity.Web/Microsoft.Identity.Web.csproj
@@ -120,6 +120,7 @@
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="5.0.0" />
     <PackageReference Include="Microsoft.Extensions.Logging" Version="5.0.0" />
     <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="5.0.0" />
+    <PackageReference Include="Microsoft.Owin.Security" Version="4.2.0" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/Microsoft.Identity.Web/Microsoft.Identity.Web.csproj
+++ b/src/Microsoft.Identity.Web/Microsoft.Identity.Web.csproj
@@ -120,7 +120,7 @@
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="5.0.0" />
     <PackageReference Include="Microsoft.Extensions.Logging" Version="5.0.0" />
     <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="5.0.0" />
-    <PackageReference Include="Microsoft.Owin.Security" Version="4.2.0" />
+    <PackageReference Include="Microsoft.AspNetCore.DataProtection" Version="5.0.8" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
+++ b/src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
@@ -39,10 +39,14 @@ namespace Microsoft.Identity.Web.TokenCacheProviders.Distributed
         /// <param name="distributedCache">Distributed cache instance to use.</param>
         /// <param name="distributedCacheOptions">Options for the token cache.</param>
         /// <param name="logger">MsalDistributedTokenCacheAdapter logger.</param>
+        /// <param name="serviceProvider">Service provider. Can be null, in which case the token cache
+        /// will not be encrypted. See https://aka.ms/ms-id-web/token-cache-encryption.</param>
         public MsalDistributedTokenCacheAdapter(
                                             IDistributedCache distributedCache,
                                             IOptions<MsalDistributedTokenCacheAdapterOptions> distributedCacheOptions,
-                                            ILogger<MsalDistributedTokenCacheAdapter> logger)
+                                            ILogger<MsalDistributedTokenCacheAdapter> logger,
+                                            IServiceProvider? serviceProvider = null)
+            : base(GetServiceProvider(distributedCacheOptions, serviceProvider))
         {
             if (distributedCacheOptions == null)
             {
@@ -63,6 +67,19 @@ namespace Microsoft.Identity.Web.TokenCacheProviders.Distributed
 
                 _expirationTime = TimeSpan.FromMilliseconds(_distributedCacheOptions.AbsoluteExpirationRelativeToNow.Value.TotalMilliseconds * _distributedCacheOptions.L1ExpirationTimeRatio);
             }
+        }
+
+        private static IServiceProvider? GetServiceProvider(
+            IOptions<MsalDistributedTokenCacheAdapterOptions> distributedCacheOptions,
+            IServiceProvider? serviceProvider)
+        {
+            if (distributedCacheOptions == null)
+            {
+                throw new ArgumentNullException(nameof(distributedCacheOptions));
+            }
+
+            // returns serviceProvider if Encrypt is true, otherwise null.
+            return distributedCacheOptions.Value.Encrypt ? serviceProvider : null;
         }
 
         /// <summary>
--- a/src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
+++ b/src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
@@ -4,6 +4,7 @@
 using System;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.AspNetCore.DataProtection;
 using Microsoft.Extensions.Caching.Distributed;
 using Microsoft.Extensions.Caching.Memory;
 using Microsoft.Extensions.Logging;
@@ -27,6 +28,7 @@ namespace Microsoft.Identity.Web.TokenCacheProviders.Distributed
         private readonly TimeSpan? _expirationTime;
         private readonly string _distributedCacheType = "DistributedCache"; // for logging
         private readonly string _memoryCacheType = "MemoryCache"; // for logging
+        private const string DefaultPurpose = "msal_cache";
 
         /// <summary>
         /// MSAL distributed token cache options.
@@ -46,7 +48,7 @@ namespace Microsoft.Identity.Web.TokenCacheProviders.Distributed
                                             IOptions<MsalDistributedTokenCacheAdapterOptions> distributedCacheOptions,
                                             ILogger<MsalDistributedTokenCacheAdapter> logger,
                                             IServiceProvider? serviceProvider = null)
-            : base(GetServiceProvider(distributedCacheOptions, serviceProvider))
+            : base(GetDataProtector(distributedCacheOptions, serviceProvider))
         {
             if (distributedCacheOptions == null)
             {
@@ -69,7 +71,7 @@ namespace Microsoft.Identity.Web.TokenCacheProviders.Distributed
             }
         }
 
-        private static IServiceProvider? GetServiceProvider(
+        private static IDataProtector? GetDataProtector(
             IOptions<MsalDistributedTokenCacheAdapterOptions> distributedCacheOptions,
             IServiceProvider? serviceProvider)
         {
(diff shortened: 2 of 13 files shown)


src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
```diff
@@ -65,6 +68,18 @@ public partial class MsalDistributedTokenCacheAdapter : MsalAbstractTokenCachePr
             }
         }
 
+        private static IServiceProvider? GetServiceProvider(
```
jmprieur, line 71: Maybe add a little comment to say that the method returns the service provider if the MsalDistributedTokenCacheAdapterOptions.Encrypt property is true, and otherwise null?

jmprieur (approved):
Thanks @jennyf19 
Awesome feature.

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
jennyf19: sure.

jmprieur (approved):
LGTM
Thanks @jennyf19

src/Microsoft.Identity.Web/Microsoft.Identity.Web.csproj
```diff
@@ -120,6 +120,7 @@
     <PackageReference Include="Microsoft.Extensions.Logging.Abstractions" Version="5.0.0" />
     <PackageReference Include="Microsoft.Extensions.Logging" Version="5.0.0" />
     <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="5.0.0" />
+    <PackageReference Include="Microsoft.Owin.Security" Version="4.2.0" />
```
rymeskar, line 123: Is this 'legacy' OWIN dependency required?

rymeskar: By any chance, did you mean the `IDataProtector` from [Microsoft.AspNetCore.DataProtection](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.dataprotection.idataprotector?view=aspnetcore-5.0) namespace?

rymeskar: Net462 and 472 should theoretically still have the Asp.Net Core IDataProtector [available](https://www.nuget.org/packages/Microsoft.AspNetCore.DataProtection.Abstractions/2.0.0).

jmprieur: @rymeskar: It's only for .NET 462 and .NET 472 (ASP.NET classic). If you are on .NET Core it's not added (see the conditional reference)
See src/Microsoft.Identity.Web/TokenCacheProviders/MsalAbstractTokenCacheProvider.cs:

```CSharp
#if DOTNET_472 || DOTNET_462
using IDataProtectionProvider = Microsoft.Owin.Security.DataProtection.IDataProtectionProvider;
using IDataProtector = Microsoft.Owin.Security.DataProtection.IDataProtector;
#else
using IDataProtectionProvider = Microsoft.AspNetCore.DataProtection.IDataProtectionProvider;
using IDataProtector = Microsoft.AspNetCore.DataProtection.IDataProtector;
#endif
```

src/Microsoft.Identity.Web/TokenCacheProviders/MsalAbstractTokenCacheProvider.cs
```diff
@@ -13,6 +20,28 @@ namespace Microsoft.Identity.Web.TokenCacheProviders
     /// <seealso cref="Microsoft.Identity.Web.TokenCacheProviders.IMsalTokenCacheProvider" />
     public abstract class MsalAbstractTokenCacheProvider : IMsalTokenCacheProvider
     {
+        private readonly IServiceProvider? _serviceProvider;
```
rymeskar, line 23: does the service provider actually need to be a class field or can it be used just in the constructor?

src/Microsoft.Identity.Web/TokenCacheProviders/MsalAbstractTokenCacheProvider.cs
```diff
@@ -13,6 +20,28 @@ namespace Microsoft.Identity.Web.TokenCacheProviders
...
+        private readonly IDataProtector? _protector;
+        private readonly string _defaultPurpose = "msal_cache";
```
rymeskar, line 25: const?

src/Microsoft.Identity.Web/TokenCacheProviders/MsalAbstractTokenCacheProvider.cs
```diff
@@ -66,15 +95,36 @@ private async Task OnAfterAccessAsync(TokenCacheNotificationArgs args)
...
         private static CacheSerializerHints CreateHintsFromArgs(TokenCacheNotificationArgs args) => new CacheSerializerHints { CancellationToken = args.CancellationToken, SuggestedCacheExpiry = args.SuggestedCacheExpiry };
 
         private async Task OnBeforeAccessAsync(TokenCacheNotificationArgs args)
         {
             if (!string.IsNullOrEmpty(args.SuggestedCacheKey))
             {
                 byte[] tokenCacheBytes = await ReadCacheBytesAsync(args.SuggestedCacheKey, CreateHintsFromArgs(args)).ConfigureAwait(false);
-                args.TokenCache.DeserializeMsalV3(tokenCacheBytes, shouldClearExistingCache: true);
+
+                args.TokenCache.DeserializeMsalV3(UnprotectBytes(tokenCacheBytes), shouldClearExistingCache: true);
```
rymeskar, line 116: What's the intended behavior/exception during failure to Unprotect the bytes? Should it be the same as during failure to retrieve the cache key, or should there be a more strict failure?

src/Microsoft.Identity.Web/TokenCacheProviders/MsalAbstractTokenCacheProvider.cs
```diff
@@ -13,6 +20,28 @@ namespace Microsoft.Identity.Web.TokenCacheProviders
...
+        /// </summary>
+        /// <param name="serviceProvider">Service provider.</param>
+        protected MsalAbstractTokenCacheProvider(IServiceProvider? serviceProvider = null)
+        {
+            _serviceProvider = serviceProvider;
+            if (_serviceProvider != null)
+            {
+                IDataProtectionProvider? dataProtectionProvider = _serviceProvider.GetService(typeof(IDataProtectionProvider)) as IDataProtectionProvider;
+#if DOTNET_472 || DOTNET_462
+                _protector = dataProtectionProvider?.Create(_defaultPurpose);
```
rymeskar, line 38: Curious: why not require the IDataProtector directly instead of the IServiceProvider? 

Context:
deciding whether to use data protection based on nullity of ServiceProvider does not IMHO show clear intent. (also not documented on the xmldoc)

src/Microsoft.Identity.Web/Microsoft.Identity.Web.csproj
rymeskar: My thinking here was that on Net462 and Net472, consumers can still depend on both classic ASP.NET but also on ASP.NET Core (up to version 2)

By depending on the lightweight (no dependencies) [Microsoft.AspNetCore.DataProtection.Abstractions](https://www.nuget.org/packages/Microsoft.AspNetCore.DataProtection.Abstractions/2.2.0) package, we are actually more versatile for the consumers. 

As for ASP.NET Core, everything is straightforward through the app platform. And for ASP.NET, consumers have two options: 
* implement the [IDataProtector](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.dataprotection.idataprotector?view=aspnetcore-2.0) interface with their API of choice on the given platform.
* depend on [Microsoft.AspNetCore.DataProtection](https://www.nuget.org/packages/Microsoft.AspNetCore.DataProtection/2.0.0). And thus intentionally bring in the dependencies from that package.

src/Microsoft.Identity.Web/TokenCacheProviders/MsalAbstractTokenCacheProvider.cs
jennyf19: correct. this was still in draft and was needed from an earlier implementation. but good catch and thank you!

src/Microsoft.Identity.Web/TokenCacheProviders/MsalAbstractTokenCacheProvider.cs
jennyf19: yes, have updated the xml comment.

src/Microsoft.Identity.Web/TokenCacheProviders/MsalAbstractTokenCacheProvider.cs
jennyf19: fixed.

src/Microsoft.Identity.Web/TokenCacheProviders/MsalAbstractTokenCacheProvider.cs
jennyf19: fixed. great catch. thank you

tests/Microsoft.Identity.Web.Test/CacheEncryptionTests.cs
```diff
@@ -0,0 +1,205 @@
...
+                   services.Configure<MsalDistributedTokenCacheAdapterOptions>(o =>
+                   {
+                       o.Encrypt = isEncrypted;
+                   });
+               });
+            _provider = hostBuilder.Build().Services;
+        }
+
+        private static TokenCacheNotificationArgs InstantiateTokenCacheNotificationArgs(TestTokenCache tokenCache)
+        {
```
jennyf19, line 87: @jmprieur thanks for the fancy method. :)

src/Microsoft.Identity.Web/TokenCacheProviders/Distributed/MsalDistributedTokenCacheAdapter.cs
```diff
@@ -65,6 +69,19 @@ public partial class MsalDistributedTokenCacheAdapter : MsalAbstractTokenCachePr
...
+            IOptions<MsalDistributedTokenCacheAdapterOptions> distributedCacheOptions,
+            IServiceProvider? serviceProvider)
+        {
+            if (distributedCacheOptions == null)
+            {
+                throw new ArgumentNullException(nameof(distributedCacheOptions));
+            }
+
+            // returns serviceProvider if Encrypt is true, otherwise null.
+            return distributedCacheOptions.Value.Encrypt ? serviceProvider : null;
```
bgavrilMS, line 82: The method name does not reflect the logic in comment at line 81

src/Microsoft.Identity.Web/TokenCacheProviders/MsalAbstractTokenCacheProvider.cs
```diff
@@ -13,6 +21,27 @@ namespace Microsoft.Identity.Web.TokenCacheProviders
     /// <seealso cref="Microsoft.Identity.Web.TokenCacheProviders.IMsalTokenCacheProvider" />
     public abstract class MsalAbstractTokenCacheProvider : IMsalTokenCacheProvider
     {
+        private readonly IDataProtector? _protector;
+        private const string DefaultPurpose = "msal_cache";
+
+        /// <summary>
+        /// Constructor.
+        /// </summary>
+        /// <param name="serviceProvider">Service provider. Can be null, in which case the token cache
```
bgavrilMS, line 30: There is an option to turn on data encryption `bool Encrypt`. I think it might be cleaner to: 

- always pass in a service provider, which could be null in some token cache implementations.
- check the Encrypt flag. If it's set and a IDataProtectionProvider exists- use it. If it doesn't exist - throw a meaningful exception.

src/Microsoft.Identity.Web/TokenCacheProviders/Session/MsalSessionTokenCacheProvider.cs
```diff
@@ -41,6 +42,7 @@ public partial class MsalSessionTokenCacheProvider : MsalAbstractTokenCacheProvi
         public MsalSessionTokenCacheProvider(
             ISession session,
             ILogger<MsalSessionTokenCacheProvider> logger)
+            : base(null)
```
bgavrilMS, line 45: An good exception should be thrown if this TokenCache impl is used and the Encyrpt flag is set to true.

src/Microsoft.Identity.Web/TokenCacheProviders/MsalAbstractTokenCacheProvider.cs
jennyf19: fixed.

src/Microsoft.Identity.Web/Microsoft.Identity.Web.csproj
jennyf19: @jmprieur provided a PR which uses your suggested. great to delete so much code.

src/Microsoft.Identity.Web/TokenCacheProviders/Session/MsalSessionTokenCacheProvider.cs
jennyf19: we don't use it for session cache.

PR #1344, opened 2021-07-26
add EnableCacheSynchronization

jennyf19:


src/Microsoft.Identity.Web/MergedOptions.cs
```diff
@@ -43,6 +43,7 @@ public ConfidentialClientApplicationOptions ConfidentialClientApplicationOptions
         public bool IsDefaultPlatformLoggingEnabled { get; set; }
         public LogLevel LogLevel { get; set; }
         public string? RedirectUri { get; set; }
+        public bool EnableCacheSynchronization { get; set; }
```
henrik-me, line 46: >EnableCacheSynchronization [](http://example.com/codeflow?start=20&length=26)

Name is a bit confusing imo, but given that's what we named it in MSAL.

henrik-me (approved):
:shipit:

PR #1365, opened 2021-07-30
Fixes a KeyNotFoundException when calling WithAppOnly(), WithScopes()

ThreeSevenths:
Using Microsoft Graph Client, calling WithAppOnly() or WithScopes() results in a KeyNotFoundException. In SetParameter(), the key is typeof(AuthenticationHandlerOption).ToString() which results in the name being set as "Microsoft.Graph.AuthenticationHandlerOption" while the value is actually present as "AuthenticationHandlerOption". This change retrieves the Name property of the Type which will always return only the class name which will be consistent with the name that is present in the dictionary.


jennyf19 (approved):
thanks so much :shipit:

jennyf19:
Included in [1.16 release](https://github.com/AzureAD/microsoft-identity-web/releases/tag/1.16.0) 

thanks again for the contribution

Issue #929, opened 2021-02-05
[Bug] HttpContext.GetTokenAsync("access_token") returns always null when using EnableTokenAcquisitionToCallDownstreamApi()


berhir:
**Which version of Microsoft Identity Web are you using?**
Microsoft.Identity.Web 1.5.1

**Where is the issue?**
* Web app
    * [ ] Sign-in users
    * [ ] Sign-in users and call web APIs
* Web API
    * [ ] Protected web APIs (validating tokens)
    * [ ] Protected web APIs (validating scopes)
    * [x] Protected web APIs call downstream web APIs
* Token cache serialization
     * [ ] In-memory caches
     * [ ] Session caches
     * [ ] Distributed caches
* Other (please describe)

**Is this a new or an existing app?**
This is a new app 

**Repro**

```csharp
public void ConfigureServices(IServiceCollection services)
{
    // Adds Microsoft Identity platform (AAD v2.0) support to protect this API
    services.AddAuthentication(JwtBearerDefaults.AuthenticationScheme)
                .AddMicrosoftIdentityWebApi(Configuration)
                .EnableTokenAcquisitionToCallDownstreamApi()
                .AddInMemoryTokenCaches();
    // ...
}

public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
{
    // ...

    app.Use(async (context, next) =>
    {
        var token = await context.GetTokenAsync("access_token");
        // token is always null here

        await next();
    });

    // ...
}
```

**Expected behavior**
When [JwtBearerOptions.SaveToken](https://docs.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.authentication.jwtbearer.jwtbeareroptions.savetoken?view=aspnetcore-5.0) is true, ` context.GetTokenAsync("access_token")` returns the access token used in the request.

**Actual behavior**
` context.GetTokenAsync("access_token")` returns always null when `EnableTokenAcquisitionToCallDownstreamApi()` is used.
When I remove `EnableTokenAcquisitionToCallDownstreamApi()`, I get the access token as expected.

**Possible solution**
I debugged the code, and the problem is in `MicrosoftIdentityWebApiAuthenticationBuilder.cs`:

```csharp
options.Events.OnTokenValidated = async context =>
{
    await onTokenValidatedHandler(context).ConfigureAwait(false);
    context.HttpContext.StoreTokenUsedToCallWebAPI(context.SecurityToken as JwtSecurityToken);
    context.Success();
[...]

PR #1387, opened 2021-08-16
Fix for issue 929

ryanfalzon:
Removed context.Success() so that token is also available in ASP.NET Core.


jennyf19:
thanks for the contribution, will need to do some testing on our end before we merge.

jennyf19:
See note here: https://github.com/AzureAD/microsoft-identity-web/issues/929#issuecomment-900785987

jennyf19:
thank you for the PR @ryanfalzon 

much appreciated. this will be in the 1.16 release later this week.

Issue #1390, opened 2021-08-17
[Bug] In .Net framework, when recreating the CCA each time one wants to acquire a token, the cache is not hit


jmprieur:
**Which version of Microsoft Identity Web are you using?**
1.15.1

**Where is the issue?**
* Token cache serialization in .Net framework
     * [x ] In-memory caches
     * [x ] Distributed caches

The issue only affects the TokenCache confidential client extensions used to add token cache support in ASP.net MVC, or .NET Core, but without fully using Id.Web. It only happens when you re create the confidential client application each time you want to get a token

**Repro**

With the in memory and distributed in memory cache, the cache is not hit, because a new InMemory cache is instantiated each time.

The following code
```CSharp
// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT License.

using Microsoft.Extensions.DependencyInjection;
using Microsoft.Extensions.Logging;
using Microsoft.Identity.Client;
using Microsoft.Identity.Web;
using System;
using System.Collections.Generic;
using System.Threading.Tasks;

namespace ConfidentialClientTokenCache
{
    /// <summary>
    /// The goal of this little sample is to show how you can use MSAL token cache adapters
    /// in confidential client applications both in .NET Core, or .NET Framework 4.7.2
    /// Note that if you write a .NET Core web app or web API, Microsoft recommends that you use the
    /// IDownstreamApi or ITokenAcquisition interfaces directly (instead of MSAL)
    /// </summary>
    static class Program
    {
        static async Task<AuthenticationResult> CreateAppAndGetToken(CertificateDescription certDescription, int cacheType)
        {
            string clientId = "6af093f3-b445-4b7a-beae-046864468ad6";
            string tenant = "msidentitysamplestesting.onmicrosoft.com";
            string[] scopes = new[] { "api://8206b76f-586e-4098-b1e5-598c1aa3e2a1/.default" };

            // Create the confidential client application
            IConfidentialClientApplication app;
            app = ConfidentialClientApplicationBuilder.Create(clientId)
                // Alternatively to the certificate you can use .WithClientSecret(clientSecret)
                .WithCertificate(certDescription.Certificate)
                .WithTenantId(tenant)
                .Build();

            if (cacheType == 0)
            {
                // In memory token caches (App and User caches)
                app.AddInMemoryTokenCache();
            }

            // Or

            // Distributed token caches (App and User caches)
            // Add one of the below: SQL, Redis, CosmosDb
            else
[...]

PR #1400, opened 2021-08-20
Add test infra and test for #1390

bgavrilMS:


src/Microsoft.Identity.Web/TokenCacheProviders/TokenCacheExtensions.cs
```diff
@@ -81,20 +81,23 @@ public static class TokenCacheExtensions
...
+            var serviceProvider = s_serviceProviderFromAction.GetOrAdd(initializeCaches.Method, (m) =>
             {
-                IHostBuilder hostBuilder = Host.CreateDefaultBuilder()
-                    .ConfigureLogging(logger => { })
-                    .ConfigureServices(services =>
-                    {
-                        initializeCaches(services);
-                        services.AddDataProtection();
-                    });
+                lock (s_serviceProviderFromAction)
```
bgavrilMS, line 87: After looking more closely at the fix for #1390 , I realize that there is still a race condition in between the 2 dictionary operations (which, by themselves, are thread safe).

src/Microsoft.Identity.Web/Util/Base64UrlHelpers.cs
```diff
@@ -2,10 +2,9 @@
 // Licensed under the MIT License.
 
 using System;
-using System.Globalization;
 using System.Text;
 
-namespace Microsoft.Identity.Web
+namespace Microsoft.Identity.Web.Util
```
bgavrilMS, line 7: I needed to move Base64UrlHelpers to be accessed from .NET classic.

tests/Microsoft.Identity.Web.Test.Common/Mocks/MockHttpClientFactory.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Linq;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Runtime.InteropServices;
+using Microsoft.Identity.Client;
+using Xunit;
+
+namespace Microsoft.Identity.Web.Test.Common.Mocks
+{
+    public class MockHttpClientFactory : IMsalHttpClientFactory, IDisposable
```
bgavrilMS, line 15: This is a variation of the HTTP mocking we do in MSAL. It uses public APIs only. It will avoid writing expensive integration tests.

bgavrilMS: Mock* objects do not need to be reviewed, they are mostly copied from MSAL.

tests/Microsoft.Identity.Web.Test/CacheExtensionsTests.cs
```diff
@@ -23,6 +26,81 @@ public void InMemoryCacheExtensionsTests()
...
+
+            result = await CreateAppAndGetTokenAsync(CacheType.InMemory, false, false).ConfigureAwait(false);
+            Assert.Equal(TokenSource.Cache, result.AuthenticationResultMetadata.TokenSource);
+
+            result = await CreateAppAndGetTokenAsync(CacheType.DistributedInMemory, true, false).ConfigureAwait(false);
+            Assert.Equal(TokenSource.IdentityProvider, result.AuthenticationResultMetadata.TokenSource);
+
+            result = await CreateAppAndGetTokenAsync(CacheType.DistributedInMemory, false, false).ConfigureAwait(false);
+            Assert.Equal(TokenSource.Cache, result.AuthenticationResultMetadata.TokenSource);
+        }
```
bgavrilMS, line 45: We could do this as an integration test instead ...

jennyf19 (approved):
thanks bogdan, appreciate the mock addition.

Issue #1348, opened 2021-07-27
[Bug] NullReferenceException when invoking ITokenAcquisition.GetAccessTokenForAppAsync


gravity00:
**Which version of Microsoft Identity Web are you using?**
Microsoft Identity Web 1.15.1
Microsoft Identity Web MicrosoftGraph 1.15.1

**Where is the issue?**
* Web app
    * [ ] Sign-in users
    * [ ] Sign-in users and call web APIs
* Web API
    * [ ] Protected web APIs (validating tokens)
    * [ ] Protected web APIs (validating scopes)
    * [X] Protected web APIs call downstream web APIs
* Token cache serialization
     * [ ] In-memory caches
     * [ ] Session caches
     * [ ] Distributed caches
* Other (please describe)

**Is this a new or an existing app?**
This is a new app.

**Repro**

```csharp
using System;
using System.Linq;
using System.Threading.Tasks;
using Microsoft.AspNetCore.Authentication.JwtBearer;
using Microsoft.AspNetCore.Builder;
using Microsoft.AspNetCore.Hosting;
using Microsoft.AspNetCore.Mvc.Authorization;
using Microsoft.Extensions.DependencyInjection;
using Microsoft.Extensions.Hosting;
using Microsoft.Graph;
using Microsoft.Identity.Web;

namespace AzureActiveDirectoryTests
{
    public class Startup
    {
        private const string Instance = "https://login.microsoftonline.com/";
        private const string TenantId = "70d...";
        private const string Domain = "...";
        private const string ClientId = "276...";
        private const string ClientSecret = "ykLSY_...";

        public void ConfigureServices(IServiceCollection services)
        {
            services
                .AddAuthentication(JwtBearerDefaults.AuthenticationScheme)
                .AddMicrosoftIdentityWebApi(o =>
                {
                    o.Events = new JwtBearerEvents
                    {
                        OnTokenValidated = OnTokenValidated
                    };
                }, o =>
                {
                    o.Instance = Instance;
                    o.TenantId = TenantId;
                    o.Domain = Domain;
                    o.ClientId = ClientId;
                    o.ClientSecret = ClientSecret;
                    o.AllowWebApiToBeAuthorizedByACL = true;
                })
                .EnableTokenAcquisitionToCallDownstreamApi(o =>
                {
                    o.Instance = Instance;
                    o.TenantId = TenantId;
                    o.ClientId = ClientId;
                    o.ClientSecret = ClientSecret;
                })
[...]

PR #1425, opened 2021-09-03
fix for anonymous controller

jennyf19:
#1348


jmprieur (approved):
LGTM.
Very nice integration test!

Maybe it would be good to add a wiki article about how un-protected APIs (anonymous APIs) can still acquire tokens on their behalf (the code of this test)

Issue #17, opened 2020-02-18
Investigate if AuthorizeForScopes could use a Policy instead of an exception filter


jennyf19:
**Why?**
Policies work independently of ASP.NET, or even Http. They work with SignalR..
The recommendation of the ASP.NET Core team is to not embed rules in a filters
Policies can be parametrized. 

**What?**
AuthorizeForScopes might be be a Policy 
(or not, this might not work, in which case Barry wants to know)

**Resources**
=> GitHub.com/blowdart=>old workshop.

copied from https://github.com/Azure-Samples/active-directory-aspnetcore-webapp-openidconnect-v2/issues/292

jmprieur:
See https://docs.microsoft.com/en-us/aspnet/core/security/authorization/iauthorizationpolicyprovider?view=aspnetcore-3.1

DamianEdwards:
This is more important now with the advent of Minimal APIs in ASP.NET Core 6.0, i.e. a way to author web APIs without using MVC controllers. As the current attribute is implemented as an MVC action filter, it can't be used with APIs authored in this new way.

Ideally the attribute should be reimplemented as endpoint metadata that is read by a configured auth policy. Whether the policy is configured directly in the project using the existing authz APIs or there is a new helper added to MIW to enable doing this more elegantly, or perhaps even by default, is something that should be investigated.

jmprieur:
@DamianEdwards AuthorizeForScopes is not used in web APIs, only in web apps.
But we got the point.

DamianEdwards:
Ah sorry for confusion, I thought this was the issue that was pointed to when we were discussing tracking the work to move the filter attribute to a policy-based approach. Where is that issue? 😄

jmprieur:
@DamianEdwards : this was my point yesterday when we asked Barry to create an issue for the Scopes filter ;-). We don't have such an issue yet.

DamianEdwards:
It's not this one? #1434

jmprieur:
My understanding is that #1434 is what Barry advises to apply to `[AuthorizeForScopes]`, not to `[RequiredScopes]` ?

jennyf19:
closing as not possible. Needs to be filter

PR #1446, opened 2021-09-15
use policy instead of filter

jennyf19:
#17


src/Microsoft.Identity.Web/WebApiExtensions/RequiredScopePolicyProvider.cs
```diff
@@ -0,0 +1,63 @@
...
+        public Task<AuthorizationPolicy> GetDefaultPolicyAsync()
+        {
+#pragma warning disable CS8625 // Cannot convert null literal to non-nullable reference type.
+            return Task.FromResult<AuthorizationPolicy>(null);
+#pragma warning restore CS8625 // Cannot convert null literal to non-nullable reference type.
+        }
+
+        public Task<AuthorizationPolicy> GetFallbackPolicyAsync()
+        {
+#pragma warning disable CS8625 // Cannot convert null literal to non-nullable reference type.
```
jmprieur, line 31: This pragma might need to be conditional. I seems that in .net 5.0 and higher, the return type of the method is `Task<AuthorizationPolicy?>`

src/Microsoft.Identity.Web/WebApiExtensions/RequiredScopePolicyProvider.cs
```diff
@@ -0,0 +1,63 @@
...
+        public const string PolicyPrefix = "RequiredScope(";
+
+        public Task<AuthorizationPolicy> GetDefaultPolicyAsync()
+        {
+#pragma warning disable CS8625 // Cannot convert null literal to non-nullable reference type.
+            return Task.FromResult<AuthorizationPolicy>(null);
+#pragma warning restore CS8625 // Cannot convert null literal to non-nullable reference type.
+        }
+
+        public Task<AuthorizationPolicy> GetFallbackPolicyAsync()
```
jmprieur, line 29: `Task<AuthorizationPolicy?>` in .net 5.0

src/Microsoft.Identity.Web/WebApiExtensions/RequiredScopePolicyProvider.cs
```diff
@@ -0,0 +1,63 @@
...
+        {
+#pragma warning disable CS8625 // Cannot convert null literal to non-nullable reference type.
+            return Task.FromResult<AuthorizationPolicy>(null);
+#pragma warning restore CS8625 // Cannot convert null literal to non-nullable reference type.
+        }
+
+        // Policies are looked up by string name, so expect 'parameters' (like age)
+        // to be embedded in the policy names. This is abstracted away from developers
+        // by the more strongly-typed attributes derived from AuthorizeAttribute
+        public Task<AuthorizationPolicy> GetPolicyAsync(string policyName)
```
jmprieur, line 39: Same remarks Task`<AuthorizationPolicy?>` in .NET 5.0

src/Microsoft.Identity.Web/WebApiExtensions/ScopeAuthorizationRequirement.cs
```diff
@@ -0,0 +1,44 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Security.Claims;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Authorization;
+
+namespace Microsoft.Identity.Web
+{
+    /// <summary>
+    /// Implements an <see cref="IAuthorizationHandler"/> and <see cref="IAuthorizationRequirement"/>
```
jmprieur, line 14: now only implements the <see cref="IAuthorizationRequirement"/>, no longer the handler which you separated.

tests/Microsoft.Identity.Web.Test/Resource/ScopesRequiredAuthorizeAttributeTests.cs
```diff
@@ -14,8 +14,8 @@
 
 namespace Microsoft.Identity.Web.Test.Resource
 {
-    [RequiredScope(RequiredScopesConfigurationKey="AzureAd:Scopes")]
-    [RequiredScope(ScopeOne)]
+    [RequiredScopeAttribute(RequiredScopesConfigurationKey="AzureAd:Scopes")]
```
jmprieur, line 17: Attribute is not needed here. Just   `[RequiredScope(RequiredScopesConfigurationKey="AzureAd:Scopes")]`

tests/WebAppCallsWebApiCallsGraph/TodoListService/Startup.cs
```diff
@@ -67,6 +68,22 @@ public void ConfigureServices(IServiceCollection services)
             services.AddControllers();
             services.AddSingleton<ILongRunningProcessContextFactory, LongRunningProcessContextFactory>();
 
+            services.AddAuthorization();
+            //services.AddAuthorization(options =>
```
jmprieur, line 72: Let's keep this code until we write the documentation of all the possibilities:
- extension method on policy builder
- requirement
- attribute.

jmprieur (approved):
LGTM
A few comments to improve, and the CA issues to fix.

src/Microsoft.Identity.Web/WebApiExtensions/HasScopeHandler.cs
```diff
@@ -0,0 +1,43 @@
...
+            {
+                Claim? scopeClaim = context.User.FindFirst(ClaimConstants.Scp);
+
+                // Fallback to Scope claim name
+                if (scopeClaim == null)
+                {
+                    scopeClaim = context.User.FindFirst(ClaimConstants.Scope);
+                }
+
+                if (scopeClaim != null && scopeClaim.Value.Split(' ').Intersect(requirement.AllowedValues).Any())
```
chwarr, line 34: `Enumerable.Intersect` creates and mutates a [new `HashSet<>`](https://github.com/dotnet/runtime/blob/ebba1d4acb7abea5ba15e1f7f69d1d1311465d16/src/libraries/System.Linq/src/System/Linq/Intersect.cs#L76-L87) every time it is invoked. Could the allowed scopes be pre-processed to a `HashSet<>` in the `ScopeAuthorizationRequirement` ctor and then looped over here?

src/Microsoft.Identity.Web/WebApiExtensions/ScopeAuthorizationRequirement.cs
```diff
@@ -0,0 +1,44 @@
...
+        public IEnumerable<string>? AllowedValues { get; }
+
+        /// <inheritdoc />
+        public override string ToString()
+        {
+            var value = (AllowedValues == null || !AllowedValues.Any())
+                ? string.Empty
+                : $" and `{ClaimConstants.Scp}` or `{ClaimConstants.Scope}` is one of the following values: ({string.Join("|", AllowedValues)})";
+
+            return $"{nameof(ScopeAuthorizationRequirement)}:Scope={value}";
```
chwarr, line 41: Let `AllowedValues` be ["one", "two"].

The result of this will be:

"ScopeAuthorizationRequirement:Scope= and `Scp` or `Scope` is one of the following values: (one|two)"

Am I executing that right in my head? Is that the expected stringification of this type?

src/Microsoft.Identity.Web/WebApiExtensions/RequiredScopePolicyProvider.cs
```diff
@@ -0,0 +1,89 @@
...
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Authentication.JwtBearer;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+
+namespace Microsoft.Identity.Web
+{
+    internal class RequiredScopePolicyProvider : IAuthorizationPolicyProvider
```
davidfowl, line 15: I don't think you need to replace the policy provider, this is a global singleton and you should be able to expose your extension method without replacing this implementation.

src/Microsoft.Identity.Web/WebApiExtensions/RequiredScopePolicyProvider.cs
```diff
@@ -0,0 +1,89 @@
...
+            return Task.FromResult<AuthorizationPolicy>(null);
+#pragma warning restore CS8625 // Cannot convert null literal to non-nullable reference type.
+        }
+
+        // Policies are looked up by string name, so expect 'parameters' (like age)
+        // to be embedded in the policy names. This is abstracted away from developers
+        // by the more strongly-typed attributes derived from AuthorizeAttribute
+        public Task<AuthorizationPolicy> GetPolicyAsync(string policyName)
+        {
+            if (policyName.StartsWith(PolicyPrefix, StringComparison.OrdinalIgnoreCase) && policyName.IndexOf('|', StringComparison.OrdinalIgnoreCase) > 0)
```
davidfowl, line 67: So this is the part of this PR I don't think we should be doing. Using the policy name as anything other than a simple name is messy. It should just be indirection to the policy. Any extra information should come from the attribute endpoint directly (aka the resource).

tests/WebAppCallsWebApiCallsGraph/TodoListService/Startup.cs
```diff
@@ -67,6 +68,22 @@ public void ConfigureServices(IServiceCollection services)
...
+            //        options.AddPolicy("foo", policyBuilder => 
+            //        {
+            //            // Extension metbod to read config, and call the code that needs to be be done.
+            //            //policyBuilder.RequireClaim("scp", "value");
+            //            //policyBuilder.RequireAssertion(context => { context.Succeed(); });
+            //            policyBuilder.RequireScope("access_as_user");
```
davidfowl, line 78: This is super nice and exactly what I thought this was look like!

tests/WebAppCallsWebApiCallsGraph/TodoListService/Controllers/TodoListController.cs
```diff
@@ -16,9 +16,14 @@
...
 {
-    [Authorize]
+    /* equivalent 
+    [Authorize(Policy = "RequiredScope(|AzureAd:Scope")]
+    [Authorize(Policy = "RequiredScope(User.Read")]
+    */
+    //[Authorize(Policy = "foo")]
     [Route("api/[controller]")]
-    [RequiredScope(RequiredScopesConfigurationKey = "AzureAd:Scopes")]
+    [RequiredScope("access_as_user")]
```
davidfowl, line 25: We should discuss how we can mean this work in a cleaner way.

davidfowl (changes requested):
A couple of things I think we should handle before merging:

- Removing the policy provider, this shouldn't be required.
- Using your custom attribute as metadata instead of using the policy name as a mini programming language.
- Merging the requirement and the IAuthorizationHandler (this is a neat trick AssertionRequirement uses)

cc @HaoK

tests/WebAppCallsWebApiCallsGraph/TodoListService/Controllers/TodoListController.cs
jennyf19: I'll set something up for early next week. Thanks so much for all the great feedback.

src/Microsoft.Identity.Web/WebApiExtensions/RequiredScopePolicyProvider.cs
jmprieur: can we still ensure that the existing applications work without changing the code for the authorization?
Today with the filters, there is not even a call to `services.AddAuthorization()` in the templates ...

davidfowl: Yes, the gist shows how to enable code to enable your attributes using a default policy.

jennyf19:
closing, have another PR out shortly.

Issue #1034, opened 2021-03-01
Make Microsoft.Identity.Web independent of the UI strings


Ponant:
Hello,
Currently, it seems that the `Microsoft.Identity.Web` is somewhat dependent on a hard coded path for the error, such as here (I did not check elsewhere yet)
https://github.com/AzureAD/microsoft-identity-web/blob/master/src/Microsoft.Identity.Web/AzureADB2COpenIDConnectEventHandlers.cs

This makes it ok when used in conjunction with the UI package, but not immediate to customize otherwise. I was hitting this issue on Azure B2C by triggering errors on purpose, and I was not using the `Microsoft.Identity.Web.UI` package.

I think it will be great to move this kind of string out of the class to a configuration that can be set in `AddMicrosoftIdentityWebApp`.

Does this makes sense?

jmprieur:
Would you think that this could go in MicrosoftIdentityOptions, @Ponant  ?

Ponant:
I am not familiar enough with this package but I just checked out `MicrosoftIdentityOptions` and it seems this is where developers would enter the azure b2c configuration, which for instance will end up in appsettings.json. Why not. I was imagining something à la .Net Core, such as when we configure the cookie.Name, e.g. `cookieOptions.LogoutPath = new PathString("/logout")`, you would have then 

`options.ErrorPath = new PathString("/error");`

with a default value for `"/MicrosoftIdentity/Account/Error"`
and the same for other strings such as the resetpassword path.

In any case I think it will be clean to do so, currently there is a sort of implicit assumption that you will use the UI package along. I saw a few issues around this where it was suggested to trigger the `OnRemoteFailure` events or to override the RCL pages by creating the page structure in the project. Overriding the rcl might create issues in the future if the overriding behavior is changed in razor pages (I don't expect it to be the case, but who knows). So, the general strategy is to to let the user say where he would like errors and resets path to be and the library would just do that, otherwise would default to the paths in the UI lib.

I am happy to test it if need be.

Ponant:
Hi @jmprieur , I come back to you to see if you are interested in a PR on this issue. Looking at the code and your suggestion, it seems that putting the strings in `MicrosoftIdentityOptions` is good option.
If you are OK I could PR this.
it will be tested in web app Razor Pages I have on Azure B2C.
Let me know

Ponant:
There is also this dependency on the UI route which could/should be fixed
[...]

PR #1452, opened 2021-09-20
Decouple Web lib from UI

Ponant:
Fixes https://github.com/AzureAD/microsoft-identity-web/issues/1034


src/Microsoft.Identity.Web/AzureADB2COpenIDConnectEventHandlers.cs
```diff
@@ -92,7 +92,7 @@ public Task OnRemoteFailure(RemoteFailureContext context)
             {
                 _errorAccessor.SetMessage(context.HttpContext, message);
 
-                context.Response.Redirect($"{context.Request.PathBase}/MicrosoftIdentity/Account/Error");
+                context.Response.Redirect($"{context.Request.PathBase}{Options.ErrorPath}");
```
jmprieur, line 95: same thing.

src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
```diff
@@ -133,5 +134,19 @@ internal bool HasClientCredentials
         /// See https://docs.microsoft.com/azure/active-directory/managed-identities-azure-resources/how-to-manage-ua-identity-portal.
         /// </summary>
         public string? UserAssignedManagedIdentityClientId { get; set; }
+
+        /// <summary>
+        /// Sets the ResetPassword route path.
```
jmprieur, line 139:
```suggestion
        /// ResetPassword path within the application's base path.
```

(proposing to remove the 'Sets' as it's set/get. Adding precision on what the path is.

src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
```diff
@@ -133,5 +134,19 @@ internal bool HasClientCredentials
...
+        /// Defaults to /MicrosoftIdentity/Account/ResetPassword,
+        /// which is the value used by Microsoft.Identity.Web.UI.
+        /// </summary>
+        public PathString ResetPasswordPath { get; set; } = new PathString("/MicrosoftIdentity/Account/ResetPassword");
+
+        /// <summary>
+        /// Sets the Error route path.
```
jmprieur, line 146:
```suggestion
        /// Error path within the application's base path.
```

jmprieur (approved):
LGTM
Thanks @Ponant 

Great improvement!

Ponant:
You are welcome @jmprieur.
Notice that the only thing I did not attempt to change is the 
````cshsharp
internal const string BlazorChallengeUri = "MicrosoftIdentity/Account/Challenge?redirectUri=";
```` 
Which lives in the Web library. So I believe it will be great to find a way to have `MicrosoftIdentity/Account/Challenge` set by the user as part of the configuration, so any suggestion for a future PR are welcomed.

src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
jennyf19: @Ponant are you able to update the comment as per @jmprieur 's suggestion? thx.

Ponant:
@jennyf19 , done

jennyf19:
@Ponant can you remove the Microsoft.Identity.Web.xml file from the PR? I'll then merge to another branch so i can pull it locally and test.

Ponant:
@jennyf19 , is it OK now?

jennyf19:
sorry, i meant just not push it up, so there is no conflict. sorry about the confusion. i'll resolve it.

jennyf19:
@Ponant in `MicrosoftIdentityConsentAndConditionalAccessHandler` you can get the IOptionsMonitor and the Scheme from the ServiceProvider, which is passed into the constructor already, and then do something [like this](https://github.com/AzureAD/microsoft-identity-web/blob/master/src/Microsoft.Identity.Web/TokenAcquisition.cs#L610).

do you want to propose a new PR?

Ponant:
@jennyf19 , I can look into that when I have time, just giv eit some time. How is this url triggered? I guess I need a blazor wasm to invoke this api, but at which stage? Obvisouly, I did not dig into that but I can look at it during weekends, I do not expect it to be difficult.

PR #1456, opened 2021-09-24
update to use policies 

jennyf19:


src/Microsoft.Identity.Web/Policy/ScopeAuthorizationHandler.cs
```diff
@@ -0,0 +1,92 @@
...
+            }
+
+            Claim? scopeClaim = context.User.FindFirst(ClaimConstants.Scp) ?? context.User.FindFirst(ClaimConstants.Scope);
+
+            if (scopeClaim is null)
+            {
+                return Task.CompletedTask;
+            }
+
+            if (scopeClaim != null && scopeClaim.Value.Split(' ').Intersect(scopes).Any())
```
chwarr, line 83: Is there a way to preprocess the scopes once per endpoint + config "version" (via the reload token?) into something like a `HashSet<>` so that the `Intersect` call can be avoided? `Intersect` allocates and mutates a `HashSet<>` every time it it called.

src/Microsoft.Identity.Web/Policy/ScopeAuthorizationRequirement.cs
```diff
@@ -0,0 +1,46 @@
...
+        public string RequiredScopesConfigurationKey { get; set; }
+
+        /// <inheritdoc />
+        public override string ToString()
+        {
+            var value = (AllowedValues == null || !AllowedValues.Any())
+                ? string.Empty
+                : $" and `{ClaimConstants.Scp}` or `{ClaimConstants.Scope}` is one of the following values: ({string.Join("|", AllowedValues)})";
+
+            return $"{nameof(ScopeAuthorizationRequirement)}:Scope={value}";
```
chwarr, line 43: Same comment from the previous incarnation of this code. The stringification looks really weird to me.

For a requirement with the allowed values of "foo" or "bar", I think this code will produce a string like:

"ScopeAuthorizationRequirement:Scope= and \`Scp\` or \`Scope\` is one of the following values: (foo|bar)"

I'd expect something more like

"ScopeAuthorizationRequirement:AllowedValues=foo|bar;RequiredScopesConfigurationKey=<null>"

jennyf19: I thought this was for displaying errors. is that not the case?

src/Microsoft.Identity.Web/Policy/ScopeAuthorizationHandler.cs
jennyf19: what is a reload token? can you point us to a doc?

src/Microsoft.Identity.Web/Policy/ScopeAuthorizationRequirement.cs
chwarr: 🤷‍♂️ I have no idea if authorization policies use the requirement's `ToString` to render error messages. 😀 If they do, do they do some sort of parsing? Because otherwise it will still be a weird error message...

src/Microsoft.Identity.Web/Policy/ScopeAuthorizationHandler.cs
```diff
@@ -0,0 +1,92 @@
...
+            {
+                HttpContext httpContext => httpContext.GetEndpoint(),
+                Endpoint ep => ep,
+                _ => null,
+            };
+
+            var data = endpoint?.Metadata.GetMetadata<IAuthRequiredScopeMetadata>();
+
+            IEnumerable<string>? scopes = null;
+            if (requirement?.RequiredScopesConfigurationKey != null)
```
chwarr, line 59: What is the desired behavior if _both_ the config key and a set of allowed values are passed to the `ctor` for `ScopeAuthorizationRequirement`? It looks to me like the config will win completely and the ctor value will be ignored. Is that what's desired?

src/Microsoft.Identity.Web/Policy/ScopeAuthorizationHandler.cs
```diff
@@ -0,0 +1,92 @@
...
+            {
+                scopes = _configuration.GetValue<string>(requirement?.RequiredScopesConfigurationKey)?.Split(' ');
+            }
+
+            if (scopes == null)
```
chwarr, line 64: Minor: inconsistent use of `== null` and `is null`. Intentional?

src/Microsoft.Identity.Web/Policy/ScopeAuthorizationHandler.cs
chwarr: If you allow me to rephrase my more fundamental underlying point: consider pre-computing as much as is practical to minimize the work done on the critical path of handling a request.

----------------------------

A reload token is a way be notified if the underlying `IConfiguration` has changed. One can be retrieved from [`IConfiguration.GetReloadToken`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.iconfiguration.getreloadtoken?view=dotnet-plat-ext-5.0) and a callback can be registered using [`IChangeToken.RegisterChangeCallback`](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.primitives.ichangetoken.registerchangecallback?view=dotnet-plat-ext-5.0#Microsoft_Extensions_Primitives_IChangeToken_RegisterChangeCallback_System_Action_System_Object__System_Object_).

Now, I have no idea if that's how a policy is supposed to monitor for config changes or not... I don't know this part of ASP.NET Core well.

-------------------------

⚠ WARNING ⚠<br>
Rampant speculation below here

_Briefly_ skimming the [Policy-based authorization in ASP.NET Core](https://docs.microsoft.com/en-us/aspnet/core/security/authorization/policies?view=aspnetcore-5.0) documentation, I _feels_ like the `ScopeAuthorizationRequirement` type is where the pre-processed data structures can live. (I don't know how endpoint metadata comes into play here... Maybe there need to be two places the pre-processed data is cached: per-endpoint if there's any endpoint customization and in the requirement itself?)

Here's a _possible_, **_untested_** implementation of caching the merged allowed values in the requirement. I've marked places where I'm unsure of things with "???". The existing `ScopeAuthorizationRequirement` didn't have access to an `IConfiguration` instance, so I added a helper internal method `GetMergedAllowedValues` to bind it to config at runtime. If the requirement could take a `IConfiguration` instance directly, the code would be a _ton_ simplier. But, I don't know how policies are supposed to be implemented. I'm winging this after skimming 5 minutes. :-) Intended use in the handler:

```
HashSet<string>? allowedScopes = requirement?.GetMergedAllowedValues(_configuration);
```

And the revised requirement:

```csharp
#nullable enable

{
    using System;
    using System.Linq;
    using System.Threading;

    using Microsoft.AspNetCore.Authorization;
    using Microsoft.Extensions.Configuration;

    public class ScopeAuthorizationRequirement : IAuthorizationRequirement, IDisposable
    {
        private readonly List<string>? _allowedValues;

        private MergedAllowedValuesHelper? _mergedAllowedValuesHelper;

        public ScopeAuthorizationRequirement(IEnumerable<string>? allowedValues = null)
        {
            if (allowedValues is not null)
            {
                // make a copy, in case the passed IEnumerable can only be enumerated once
                _allowedValues = allowedValues.ToList();
            }
        }

        public ICollection<string>? AllowedValues => _allowedValues;

        public string? RequiredScopesConfigurationKey { get; init; }

        public void Dispose()
        {
            MergedAllowedValuesHelper? helper = Interlocked.Exchange(ref _mergedAllowedValuesHelper, null);
            helper?.Dispose();
        }

        internal HashSet<string> GetMergedAllowedValues(IConfiguration config)
        {
            // ??? Can this be replaced with a Lazy<MergedAllowedValuesHelper> somehow?

            if (_mergedAllowedValuesHelper is not null)
            {
                return _mergedAllowedValuesHelper.MergedAllowedValues;
            }

            var newMergedAllowedValuesHelper = new MergedAllowedValuesHelper(this, config);

            // Attempt to publish our instances as THE helper.
            MergedAllowedValuesHelper? oldMergedAllowedValuesHelper = Interlocked.CompareExchange(ref _mergedAllowedValuesHelper, newMergedAllowedValuesHelper, null);

            if (oldMergedAllowedValuesHelper is not null)
            {
                // We lost the race to publish. Clean up the junk we created
                // and return the value from the instance that was published.
                newMergedAllowedValuesHelper.Dispose();
                return oldMergedAllowedValuesHelper.MergedAllowedValues;
            }
            else
            {
                // We were first to publish. Return our value. That is also
                // what will be returned from now on.
                return newMergedAllowedValuesHelper.MergedAllowedValues;
            }
        }

        /// <summary>
        /// Handles merging the allowed values specificed on a given <see
        /// cref="ScopeAuthorizationRequirement"/> with any from config and
        /// monitoring for changes to that config.
        /// </summary>
        private class MergedAllowedValuesHelper : IDisposable
        {
            private readonly ScopeAuthorizationRequirement _parent;
            private readonly IConfiguration _config;
            private HashSet<string> _mergedAllowedValues;
            private IDisposable? _configChangeCallbackUnregisterer;

            public MergedAllowedValuesHelper(ScopeAuthorizationRequirement parent, IConfiguration config)
            {
                _parent = parent;
                _config = config;

                _mergedAllowedValues = MakeMergedAllowedValues();

                if (!string.IsNullOrEmpty(_parent.RequiredScopesConfigurationKey))
                {
                    // ??? There's a race between the call to
                    // MakeMergedAllowedValues, which may consult config, and
                    // the installation of this change callback. I don't know
                    // how to resolve that...
                    _configChangeCallbackUnregisterer = _config.GetReloadToken().RegisterChangeCallback(OnConfigChanged, this);
                }
            }

            public HashSet<string> MergedAllowedValues => _mergedAllowedValues;

            public void Dispose()
            {
                IDisposable? unregisterer = Interlocked.Exchange(ref _configChangeCallbackUnregisterer, null);
                unregisterer?.Dispose();
            }

            private static void OnConfigChanged(object thisAsObject)
            {
                ((MergedAllowedValuesHelper)thisAsObject).OnConfigChanged();
            }

            private HashSet<string> MakeMergedAllowedValues()
            {
                // ??? I assumed that config and ctor values are supposed to be merged, but that's not what the prior code did. Adjust as needed.
                string[] valuesFromConfig = Array.Empty<string>();

                if (!string.IsNullOrEmpty(_parent.RequiredScopesConfigurationKey))
                {
                    string? requiredScopesConfigValue = _config.GetValue<string>(_parent.RequiredScopesConfigurationKey);
                    if (requiredScopesConfigValue is not null)
                    {
                        valuesFromConfig = requiredScopesConfigValue.Split(' ');
                    }
                }

                var merged = new HashSet<string>((_parent._allowedValues?.Count ?? 0) + valuesFromConfig.Length);

                if (_parent._allowedValues is not null)
                {
                    foreach (string value in _parent._allowedValues)
                    {
                        merged.Add(value);
                    }
                }

                foreach (string value in valuesFromConfig)
                {
                    merged.Add(value);
                }

                return merged;
            }

            private void OnConfigChanged()
            {
                HashSet<string> newMergedAllowedValues = MakeMergedAllowedValues();

                // Atomically publish the new merged set.
                //
                // ??? This assumes that there will never be more than one
                // configuration changed callback in flight at once. I _think_
                // IConfiguration guarantees this...
                Interlocked.Exchange(ref _mergedAllowedValues, newMergedAllowedValues);
            }
        }
    }
}

```

src/Microsoft.Identity.Web/Policy/ScopeAuthorizationHandler.cs
```diff
@@ -0,0 +1,92 @@
...
+            {
+                scopes = requirement?.AllowedValues ?? data?.AcceptedScope;
+            }
+
+            // Can't determine what to do without scope metadata, so proceed
+            if (scopes is null)
+            {
+                //context.Succeed(requirement);
```
davidfowl, line 72: Don't you want this to succeed since it's always on?

jennyf19: When testing, it would proceed to the web API, but we don't want it to do that if the scopes are null. I'll need to take a look at this specific case again. thanks for pointing it out.

Tratcher:
@haok

src/Microsoft.Identity.Web/Resource/RequiredScopeFilter.cs
```diff
@@ -1,141 +0,0 @@
...
-            }
-
-            if (user == null || userClaims == null || !userClaims.Any())
-            {
-                lock (httpContext)
-                {
-                    httpContext.Response.StatusCode = (int)HttpStatusCode.Unauthorized;
-                }
-
-                throw new UnauthorizedAccessException(IDWebErrorMessage.UnauthenticatedUser);
```
Tratcher, line 77: I'm glad to see this gone.

Tratcher (approved):

src/Microsoft.Identity.Web/Policy/ScopeAuthorizationHandler.cs
jennyf19: @davidfowl So I get here when I have `[Authorize]` on the controller, but no `[RequiredScope()]` and have not defined a policy in the startup.cs. So I assume this means, we have to have an authenticated user, but no scope check? So succeed? Is that correct? What if there were another requirement in Startup.cs would the whole authorization succeed?

HaoK: > What if there were another requirement in Startup.cs would the whole authorization succeed?

There's no way to bypass every requirement being succeeded, although any handler COULD do that if they wanted to just iterate over all requirements and call Succeed on each, but this is just succeeding your requirement here.

src/Microsoft.Identity.Web/Policy/ScopeAuthorizationHandler.cs
HaoK: Yeah, I'd avoid going down change tokens and reloadable config unless you really have a need for that.

HaoK (approved):

jennyf19:
thank you @HaoK & @Tratcher

src/Microsoft.Identity.Web/Policy/ScopeAuthorizationHandler.cs
HaoK: To David's point, if you just fall thru and don't call Succeed on this requirement, that will have the effect of blocking all requests without scope metadata

src/Microsoft.Identity.Web/Policy/ScopeAuthorizationHandler.cs
jennyf19: ack. Thanks for the perspectives.

src/Microsoft.Identity.Web/Policy/ScopeAuthorizationHandler.cs
jennyf19: got it, thanks @HaoK

PR #1465, opened 2021-09-30
fix tsa issues

jennyf19:
fixing the TSA issues...I think we can throw if the requirement is null, like asp.net core does [here](https://github.com/aspnet/Security/blob/26d27d871b7992022c082dc207e3d126e1d9d278/src/Microsoft.AspNetCore.Authorization/AuthorizationHandlerContext.cs#L31) and i'm not sure how `*.Remove(requirement)` will handle the [null value](https://github.com/aspnet/Security/blob/26d27d871b7992022c082dc207e3d126e1d9d278/src/Microsoft.AspNetCore.Authorization/AuthorizationHandlerContext.cs#L95).

### DIFF
--- a/src/Microsoft.Identity.Web/Policy/ScopeAuthorizationHandler.cs
+++ b/src/Microsoft.Identity.Web/Policy/ScopeAuthorizationHandler.cs
@@ -33,15 +33,20 @@ namespace Microsoft.Identity.Web
         ///  Makes a decision if authorization is allowed based on a specific requirement.
         /// </summary>
         /// <param name="context">AuthorizationHandlerContext.</param>
-        /// <param name="requirement">Scope requirement.</param>
+        /// <param name="requirement">Scope authorization requirement.</param>
         /// <returns>Task.</returns>
         protected override Task HandleRequirementAsync(
             AuthorizationHandlerContext context,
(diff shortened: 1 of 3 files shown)


src/Microsoft.Identity.Web/Policy/ScopeAuthorizationHandler.cs
```diff
@@ -33,17 +33,22 @@ public ScopeAuthorizationHandler(IConfiguration configuration)
...
         /// <param name="context">AuthorizationHandlerContext.</param>
-        /// <param name="requirement">Scope requirement.</param>
+        /// <param name="requirement">Scope authorization requirement.</param>
         /// <returns>Task.</returns>
         protected override Task HandleRequirementAsync(
             AuthorizationHandlerContext context,
             ScopeAuthorizationRequirement requirement)
         {
-            if (context == null)
+            if (context is null)
```
jmprieur, line 42: really? what is the difference?

jmprieur (approved):
LGTM. Thanks @jennyf19 

In addition to this fixes, in src/Microsoft.Identity.Web/Policy/ScopeAuthorizationRequirement.cs, it seems that the static analysis wants the `RequiredScopesConfigurationKey` property to be nullable?

PR #1470, opened 2021-10-02
Use macro in csproj to define IdentityModel version

brentschmaltz:
This will speed up testing against local builds of IdentityModel.
Ideally we would include something like this file here: [dependencies.props](https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/blob/dev/build/dependencies.props)

This will move us forward...


jennyf19:
This is awesome. Thanks!

jennyf19 (approved):
:shipit:

PR #1493, opened 2021-10-18
updates to id web to support authentication handlers than JwtBearer

jennyf19:
Have tested with release build, in addition to the migration samples from @jmprieur 

- [x]  WebApp-OIDC
  - [x] MyOrg 
  - [x]  B2C
- [x]  WebApp-graph
   - [x]  Call-MSGraph
- [x]  WebApp-your-API
  - [x]  MyOrg
  - [x]  B2C


jennyf19:
I did a release build last night which we've already been testing with, so not sure if another one is needed at this point.

---
In reply to: [782580128](https://github.com/AzureAD/microsoft-identity-web/pull/1493#issuecomment-782580128) [](http://example.com/codeflow?ancestors=782580128,782580128)

Issue #905, opened 2021-01-22
[Bug] Enable token decryption certificates rotation


jmprieur:
**Which version of Microsoft Identity Web are you using?**
1.5.1

**Where is the issue?**
* Web API
    * [x ] Protected web APIs (validating tokens)

**Is this a new or an existing app?**
New or existing

**Repro**
Today, Microsoft.Identity.Web allows the developers to provide a collection of token decryption certificates (to be able to rotate certificates), but only the fist one is used. See [this code](https://github.com/AzureAD/microsoft-identity-web/blob/9bc1a37c2c52175aa514e577b2fb966f25469e6b/src/Microsoft.Identity.Web/WebApiExtensions/MicrosoftIdentityWebApiAuthenticationBuilderExtensions.cs#L208-L209)

**Expected behavior**
Developers need to be able to rotate their certificates, and therefore all the certs provided in the TokenDecryptionCertificates property

**Actual behavior**
Only the first certificate is loaded and used for the token decryption

**Possible solution**
Use `TokenValidationParameters.TokenDecryptionKeys` instead of `TokenValidationParameters.TokenDecryptionKey`.  We'll also need to have a new method [DefaultCertificateLoader.LoadAllCertificates]() in addition to the existing method [DefaultCertificateLoader.LoadFirstCertificate](https://github.com/AzureAD/microsoft-identity-web/blob/9bc1a37c2c52175aa514e577b2fb966f25469e6b/src/Microsoft.Identity.Web/CertificateManagement/DefaultCertificateLoader.cs#L227), with the same signature, but returning a collection of `X509Certificate2`. 

Note we don't want to remove the existing method (LoadFirstCertificate) as it will still be used for client certificates, for which the rotation is helped by the use of sendX5c.

brentschmaltz:
@jmprieur We need to be careful when using certificates.
Some users have seen issues when specifying the certificate. 
A more performant pattern would be to obtain the RSA object from the certificate once on startup and pass an RSASecurityKey (s) instantiated with the RSA object. Our libraries will not dispose of the RSA as they did not create it.

jennyf19:
[...]

PR #1511, opened 2021-11-02
Remove Outdated Comment on TokenDecryptionCertificates

michaelwstark:
There is a comment on the public class MicrosoftIdentityOptions.cs which indicates only the first certificate in the set is considered.  Reading through the code as well as looking over historical PRs, it appears this limitation was resolved.

https://github.com/AzureAD/microsoft-identity-web/issues/905

Correcting comment to save others from the same investigation.


jennyf19:
Great catch. Thank you @michaelwstark

jennyf19:
@michaelwstark part of [1.20 release](https://github.com/AzureAD/microsoft-identity-web/releases/tag/1.20.0)

PR #1514, opened 2021-11-03
Initial cache little additions and perf improvements

rymeskar:
* More safeguards for cache failures
* Perf improvements on sliding time window refresh
* Adding further extension method on ICCA
* Adding cache key provider


src/Microsoft.Identity.Web.TokenCache/IMsalTokenCacheProvider.cs
```diff
@@ -26,6 +26,13 @@ public interface IMsalTokenCacheProvider
         /// <param name="tokenCache">Token cache for which to initialize the serialization.</param>
         void Initialize(ITokenCache tokenCache);
 
+        /// <summary>
+        /// Safely initializes a token cache (which can be a user token cache or an app token cache).
+        /// </summary>
+        /// <param name="tokenCache">Token cache for which to initialize the serialization.</param>
+        /// <remarks>The safe terminology denotes to not propagating exceptions from caching layer to service, instead fallbacking to eSTS call.</remarks>
+        void SafeInitialize(ITokenCache tokenCache);
```
bgavrilMS, line 34: I had to read the implementation to understand what this does. I would name it more along the lines of "no exceptions"

src/Microsoft.Identity.Web.TokenCache/IMsalTokenCacheProvider.cs
```diff
@@ -26,6 +26,13 @@ public interface IMsalTokenCacheProvider
         /// <param name="tokenCache">Token cache for which to initialize the serialization.</param>
         void Initialize(ITokenCache tokenCache);
 
+        /// <summary>
```
bgavrilMS, line 29: Adding a method to an interface sadly breaks backward compat. 

Solutions: 

1. https://devblogs.microsoft.com/dotnet/default-implementations-in-interfaces/ (doesn't work on .net classic though)
2. extension methods
3. options object

In this case, could the safety flag be passed via the constructor of `MsalAbstractTokenCacheProvider` instead?

jennyf19:
@rymeskar I didn't think this was ready for review yet, as we had discussed. I assume you'll create a branch, not a fork, and advise when you are ready.

PR #1523, opened 2021-11-05
Exercising the MSAL long running API

jmprieur:


src/Microsoft.Identity.Web/TokenAcquisitionOptions.cs
```diff
@@ -52,6 +52,22 @@ public class TokenAcquisitionOptions
         /// </summary>
         public CancellationToken CancellationToken { get; set; } = CancellationToken.None;
 
+        /// <summary>
+        /// Key used for long running web APIs that need to call downstream web
+        /// APIs on behalf of the user. Can be null, if you are not developping a long
```
jennyf19, line 57:
```suggestion
        /// APIs on behalf of the user. Can be null, if you are not developing a long
```

jennyf19 (approved):
:shipit:

jennyf19:
wonder if we can automate a test for this?

Issue #1324, opened 2021-07-19
[Bug] Bearer error="invalid_token", error_description="The issuer '(null)' is invalid" in v1.14.1


throck95:
**Which version of Microsoft Identity Web are you using?**
v1.14.1

**Where is the issue?**
* Web app
    * [ ] Sign-in users
    * [ ] Sign-in users and call web APIs
* Web API
    * [X] Protected web APIs (validating tokens)
    * [ ] Protected web APIs (validating scopes)
    * [ ] Protected web APIs call downstream web APIs
* Token cache serialization
     * [ ] In-memory caches
     * [ ] Session caches
     * [ ] Distributed caches
* Other (please describe)

**Is this a new or an existing app?**
This is an app under active development and live in a production system for which I have successfully used v1.12.0.

**Repro**
Similar to Thomas Barnekow in #1310, I have made no code changes within my application. I branched from main and updated from v1.12.0 to v1.14.1. Following this, the API starts failing to validate tokens generated by Azure AD via MSAL.

Server side, I am using .NET 5 with the following configuration:

**appsettings.json**

```csharp
"AzureAdB2C": {
    "Instance": "https://login.microsoftonline.com/",
    "TenantId": "[tenant_guid]",
    "Domain": "gravitydiagnosticsb2c.onmicrosoft.com",
    "ClientId": "[guid]",
    "AllowWebApiToBeAuthorizedByACL": true
```

My API utilizes the token for authentication and then routes authentication through a database for role assignments. As such, the ACL bypass is needed.

**Startup.ConfigureServices(IServiceCollection services)**

```csharp
var tenantId = Configuration.GetSection("AzureAdB2C:TenantId").Value;
var metadataAddress = $"https://login.microsoftonline.com/{tenantId}/.well-known/openid-configuration";
services.AddAuthentication(JwtBearerDefaults.AuthenticationScheme)
     .AddMicrosoftIdentityWebApi(options =>
     {
          Configuration.Bind("AzureAdB2C", options);

          options.TokenValidationParameters.NameClaimType = "name";
          options.MetadataAddress = metadataAddress;
     },
     options => { Configuration.Bind("AzureAdB2C", options); });
```

**Startup.Configure(IApplicationBuilder app, IWebHostEnvironment env, IApiVersionDescriptionProvider provider)**
```csharp
public void Configure(IApplicationBuilder app, IWebHostEnvironment env, IApiVersionDescriptionProvider provider)
{
[...]

PR #1552, opened 2021-12-02
wilson 6.15

jennyf19:


jmprieur (approved):
LGTM
Thanks @jennyf19 

this could even fix https://github.com/AzureAD/microsoft-identity-web/issues/1324 
(which we could not repro)

PR #1558, opened 2021-12-12
fix:  SignOut callback URL breaks with custom dotnet areas

pj-evans:
We were having an issue with the signout function, upon looking into the code, it seems that the hardcoded callback URL is setup off of root /Account/SignedOut omitting the area /MicrosoftIdentity/.

This caused our app to append the /SignOut off of the last route config action setup... As we renamed/commented them out the action would just append off of the next final action.

Our instincts would be to declare the full area/controller/action here as: /MicrosoftIdentity/Account/SignedOut into the Callback path here, however, we did see other coding consistancy utilizing the ~ home root.  Unsure of which to do (nor are we sure ~ will work here) we went with keeping coding consistant (as it looked like it could have been possible to have this be a simple type-o omission).

Leaving out the declared area is grabbing the current area and sending the request to somewhere that simply doesn't exist.

Please feel free to update as we are pretty new to all of this and are unaware of understanding how to actually test this fix locally before sending a PR.  And please feel free to send an email/message with any questions.

Thank you, have a great day,
--PJ


jmprieur:
Thanks @pj-evans 
We'll test and follow-up

iarekk:
Thank you for taking the time to investigate this and open a PR, @pj-evans -- we appreciate the contribution!

After reviewing the current codebase, `Url.Page("/Account/SignedOut", ...)` resolves correctly within the `MicrosoftIdentity` area because the controller is decorated with `[Area("MicrosoftIdentity")]`, and ASP.NET Core's `IUrlHelper.Page()` respects the ambient area context when resolving absolute page paths.

The `~/` prefix unfortunately doesn't have special meaning in `Url.Page()` (it's a `Url.Content()` convention), so this change could produce unexpected results.

We're going to close this one, but if you (or anyone) encounters sign-out redirect issues in a specific routing configuration, please open a new issue with repro steps and we'll be happy to look into it.

Issue #1566, opened 2021-12-22
[Bug] Eviction is not set on L2 token cache nodes 


bgavrilMS:
**Which version of Microsoft Identity Web are you using?**
1.21.1

**Where is the issue?**
* Web app
    * [ ] Sign-in users
    * [ ] Sign-in users and call web APIs
* Web API
    * [ ] Protected web APIs (validating tokens)
    * [ ] Protected web APIs (validating scopes)
    * [ ] Protected web APIs call downstream web APIs
* Token cache serialization
     * [ ] In-memory caches
     * [ ] Session caches
     * [x] Distributed caches
* Other (please describe)

L2 nodes should be evicted when the underlying tokens are known to expire. MSAL offers this information and IdWeb uses it for L1 eviction, but not for L2. MSAL will offer this information for **client_credential** and **OBO**. 

To fix, I propose: 

- if app developer has set L2 eviction themselves via the options, use it
- if they have not, use MSAL suggested eviction, where available.

bgavrilMS:
PS: I created the "token acquisition" label for bugs that should be handled by the MSAL team. Token cache bugs fall into this category.

jennyf19:
Included in the [1.22.0 release](https://github.com/AzureAD/microsoft-identity-web/releases/tag/1.22.00

PR #1572, opened 2021-12-31
fix l2 cache expiry

jennyf19:
ignore branch name, was from another investigation.

have tested w/redis cache manually as well.

#1566

### DIFF
--- a/src/Microsoft.Identity.Web.TokenCache/Distributed/MsalDistributedTokenCacheAdapter.cs
+++ b/src/Microsoft.Identity.Web.TokenCache/Distributed/MsalDistributedTokenCacheAdapter.cs
@@ -233,22 +233,14 @@ namespace Microsoft.Identity.Web.TokenCacheProviders.Distributed
         {
             string write = "Write";
 
-            TimeSpan? cacheExpiry = null;
-            if (cacheSerializerHints != null && cacheSerializerHints?.SuggestedCacheExpiry != null)
-            {
-                cacheExpiry = cacheSerializerHints.SuggestedCacheExpiry.Value.UtcDateTime - DateTime.UtcNow;
-                if (cacheExpiry < TimeSpan.Zero)
-                {
-                    cacheExpiry = TimeSpan.FromMilliseconds(1);
-                    Logger.MemoryCacheNegativeExpiry(_logger, _memoryCacheType, write, null);
-                }
-            }
+            DateTimeOffset? cacheExpiry = cacheSerializerHints?.SuggestedCacheExpiry;
 
             if (_memoryCache != null)
             {
                 MemoryCacheEntryOptions memoryCacheEntryOptions = new MemoryCacheEntryOptions()
                 {
-                    AbsoluteExpirationRelativeToNow = cacheExpiry ?? _expirationTime,
+                    AbsoluteExpiration = cacheExpiry ?? _distributedCacheOptions.AbsoluteExpiration,
+                    AbsoluteExpirationRelativeToNow = _expirationTime,
                     Size = bytes?.Length,
                 };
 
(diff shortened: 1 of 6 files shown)


jmprieur (approved):
LGTM
Thanks @jennyf19 

Great to see the usage of the `AbsoluteExpiration` which is useful when we have the suggested expiry indeed (as this is an absolute date time offset). We still won't encourage customers to use the Absolute expiration in the configuration settings as it makes little sense to say. All my cache items expire tomorrow at midnight.

And I see it's 100% covered in UT.
Great job @jennyf19

src/Microsoft.Identity.Web.TokenCache/Distributed/MsalDistributedTokenCacheAdapter.cs
```diff
@@ -258,12 +250,25 @@ protected override async Task WriteCacheBytesAsync(string cacheKey, byte[] bytes
...
 
+            if ((cacheExpiry != null && _distributedCacheOptions.AbsoluteExpiration != null && _distributedCacheOptions.AbsoluteExpiration < cacheExpiry)
+               || (cacheExpiry == null && _distributedCacheOptions.AbsoluteExpiration != null))
+            {
+                cacheExpiry = _distributedCacheOptions.AbsoluteExpiration;
+            }
+
+            DistributedCacheEntryOptions distributedCacheEntryOptions = new DistributedCacheEntryOptions()
+            {
+                AbsoluteExpiration = cacheExpiry,
```
bgavrilMS, line 261: Do you know which one of these expirations will win if they are all set? Is it the one closer to DateTime.Now ?

tests/Microsoft.Identity.Web.Test/L1L2CacheTests.cs
```diff
@@ -42,38 +42,114 @@ public async Task WriteCache_WritesInL1L2_TestAsync()
...
+            await CreateL1L2TestWithAbsoluteOptions(1.5).ConfigureAwait(false);
+        }
+
+        [Fact]
+        public async Task WriteCacheL1L2_PositiveExpiryAndAbsoluteOptionsLessThanSuggestedExpiry_TestAsync()
+        {
+            await CreateL1L2TestWithAbsoluteOptions(.5).ConfigureAwait(false);
+        }
+
+        private async Task CreateL1L2TestWithAbsoluteOptions(double time)
```
bgavrilMS, line 109: Pls use TimeSpan here so that the code is clearer.

src/Microsoft.Identity.Web.TokenCache/Distributed/MsalDistributedTokenCacheAdapter.cs
```diff
@@ -233,22 +233,14 @@ protected override async Task WriteCacheBytesAsync(string cacheKey, byte[] bytes
         {
             string write = "Write";
```
bgavrilMS, line 234: not related: since this is in the critical path for perf / mem usage, avoid creating new objects. In this case, just use a constant.

tests/Microsoft.Identity.Web.Test/TestDistributedCache.cs
```diff
@@ -10,13 +10,23 @@ namespace Microsoft.Identity.Web.Test
 {
     public class TestDistributedCache : IDistributedCache
     {
-        public readonly ConcurrentDictionary<string, byte[]> dict = new ConcurrentDictionary<string, byte[]>();
+        internal readonly ConcurrentDictionary<string, Entry> _dict = new ConcurrentDictionary<string, Entry>();
```
bgavrilMS, line 13: Make it a property if it's public / internal. Or create a "Count" method and keep this private.

tests/Microsoft.Identity.Web.Test/L1L2CacheTests.cs
```diff
@@ -42,38 +42,114 @@ public async Task WriteCache_WritesInL1L2_TestAsync()
...
+                Assert.Equal(suggestedExpiry, options.AbsoluteExpiration.Value);
+            }
+
+            absoluteOptions.Value.AbsoluteExpiration = null;
+            await _testCacheAdapter.TestReadCacheBytesAsync(DefaultCacheKey).ConfigureAwait(false);
+            Assert.Equal(1, _testCacheAdapter._memoryCache.Count);
+        }
+
+        private async Task CreateL1L2TestWithSerializerHints(
+            System.DateTimeOffset dateTimeOffset,
```
bgavrilMS, line 136: msalSuggestedCacheExpiry

src/Microsoft.Identity.Web.TokenCache/Distributed/MsalDistributedTokenCacheAdapter.cs
bgavrilMS: Same comment for the memory cache.

tests/Microsoft.Identity.Web.Test/L1L2CacheTests.cs
```diff
@@ -42,38 +42,114 @@ public async Task WriteCache_WritesInL1L2_TestAsync()
...
             await _testCacheAdapter.TestWriteCacheBytesAsync(DefaultCacheKey, cache).ConfigureAwait(false);
 
             // Assert
             Assert.Equal(1, _testCacheAdapter._memoryCache.Count);
-            Assert.Single(L2Cache.dict);
+            Assert.Single(L2Cache._dict);
         }
 
         [Fact]
         public async Task WriteCache_NegativeExpiry_TestAsync()
```
bgavrilMS, line 56: MSAL should never ask you to persist something with negative expiry, this would be a bug. Not sure what the best behavior is in this case: 

- throw an exception
- write the value to the cache, which semantically means "DELETE key" and log an error. If this happens often, it affects the cache hit ratio.

Thoughts?

PR #1573, opened 2022-01-01
Client assertions

### LINKED ISSUE #203: [Feature Request] Support client assertions
Microsoft.Identity.Web supports client secrets and client certificates
Do we need to support client assertions?

jmprieur:
Fixes #203


src/Microsoft.Identity.Web/ClientAssertionDescription.cs
```diff
@@ -0,0 +1,61 @@
...
+        private ClientAssertion? _clientAssertion;
+
+        /// <summary>
+        /// Get the signed assertion (and refreshes it if needed).
+        /// </summary>
+        /// <param name="cancellationToken">Cancellation token.</param>
+        /// <returns>The signed assertion.</returns>
+        public async Task<string> GetSignedAssertion(CancellationToken cancellationToken)
+        {
+            if (_clientAssertion == null || (Expiry != null && DateTime.Now > Expiry))
```
jennyf19, line 42: >DateTime.Now [](http://example.com/codeflow?start=63&length=12)

should this be UTC?

jennyf19:
>             {

do we need to check for an assertion as well know? I believe we added this to fail early before getting to MSAL.

---
Refers to: src/Microsoft.Identity.Web/MergedOptionsValidation.cs:54 in 651d78b. [](commit_id = 651d78b5f84d01422140268383ffb6398b256625, deletion_comment = False)

src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
```diff
@@ -147,5 +148,20 @@ internal bool HasClientCredentials
...
+        /// <example>
+        /// In Startup.cs, after AddMicrosoftIdentityWebXX, add another instruction:
+        /// <code>
+        /// services.ConfigureOptions&lt;MicrosoftIdentityOptions&gt;(OpenIdConnectDefaults.AuthenticationScheme,
+        /// options => {
+        ///   options.ClientAssertionDescription = GetClientAssertionProvider();
+        /// });
+        /// </code>
+        /// </example>
+        public ClientAssertionDescription? ClientAssertionDescription { get; set; }
```
jennyf19, line 165: nit: i would move this section to below the part on certs

tests/Microsoft.Identity.Web.Test/Certificates/CertificatesTests.cs
```diff
@@ -52,7 +53,30 @@
 
             // Act & Assert
             // Should not throw
-            MergedOptionsValidation.ValidateEitherClientCertificateOrClientSecret(options.ClientSecret, microsoftIdentityOptions.ClientCertificates);
+            MergedOptionsValidation.ValidateEitherClientCertificateOrClientSecret(options.ClientSecret, microsoftIdentityOptions.ClientCertificates, null);
+        }
+
+        public void ValidateCredentialType_ClientAssertion(string base64Encoded)
```
jennyf19, line 59: This needs to have `[Fact]`

tests/Microsoft.Identity.Web.Test/Certificates/CertificatesTests.cs
```diff
@@ -52,7 +53,30 @@
...
+                    CertificateDescription.FromBase64Encoded(base64Encoded);
+
+            MicrosoftIdentityOptions microsoftIdentityOptions = new MicrosoftIdentityOptions
+            {
+                Authority = TestConstants.AuthorityCommonTenant,
+                ClientId = TestConstants.ConfidentialClientId,
+                ClientAssertionDescription = new ClientAssertionDescription((_) => { return Task.FromResult(new ClientAssertion("clientAssertion", DateTime.Now.AddSeconds(20))); }),
+            };
+
+            ConfidentialClientApplicationOptions options = new ConfidentialClientApplicationOptions
```
jennyf19, line 72: maybe this is copy/paste from another test?

tests/WebAppCallsMicrosoftGraph/appsettings.json
```diff
@@ -1,14 +1,11 @@
 {
     "AzureAd": {
         "Instance": "https://login.microsoftonline.com/",
-        "Domain": "msidentitytesting.onmicrosofonline.com",
-        "TenantId": "7f58f645-c190-4ce5-9de4-e2b7acd2a6ab",
-        "ClientId": "56c9a633-236e-45ee-9af1-a53d9811fbd6",
-        // To call an API
-        "ClientSecret": "secret-goes-here",
+        "TenantId": "776c277d-e57f-4982-a3df-589bac9fadc6",
```
jennyf19, line 4: can you make a new section for testing this so we don't lose our current app creds?

tests/WebAppCallsMicrosoftGraph/Startup.cs
```diff
@@ -74,6 +91,21 @@ public void ConfigureServices(IServiceCollection services)
             }).AddMicrosoftIdentityUI();
         }
 
+        /// <summary>
+        /// Prototype of certificate-less authentication using a signed assertion
+        /// acquired with MSI (federated identity)
+        /// </summary>
+        /// <returns></returns>
+        private async Task<ClientAssertion> GetSignedAssertionFromMsi(CancellationToken cancellationToken)
+        {
```
jennyf19, line 100: UserAssignedClientId is available in MsIdentityOptions.

jennyf19, line 100: Could this be moved to the product code?

tests/WebAppCallsMicrosoftGraph/Pages/Index.cshtml.cs
```diff
@@ -38,7 +38,9 @@ public async Task OnGet()
                 }
                 ViewData["name"] = user.DisplayName;
 
-                var graphData = await _downstreamWebApi.CallWebApiForUserAsync("GraphBeta");
+                var graphData = await _downstreamWebApi.CallWebApiForUserAsync(
```
jennyf19, line 41: let's comment this out and re-add what we had before

src/Microsoft.Identity.Web/ClientAssertionDescription.cs
jmprieur: Thanks, @jennyf19. I'll use `DateTimeOffset.Now` instead, now that I understand what this is :-)

src/Microsoft.Identity.Web/MicrosoftIdentityOptions.cs
jmprieur: yes. I'll move it. I was trying to not change the order of the fields in the data structure to have a binary compatibility of the data structure. Old outdated reflexes of a time where C was ruling the world.

tests/WebAppCallsMicrosoftGraph/Startup.cs
jmprieur: Thanks for the reminder. Fixed this

jmprieur: This is a question for the virtual team. I'll check with Yordan

tests/WebAppCallsMicrosoftGraph/appsettings.json
jmprieur: done/ Thanks for the suggestion

jmprieur (commented):
Thanks @jennyf19 
I think I addressed all

tests/WebAppCallsMicrosoftGraph/Pages/Index.cshtml.cs
jmprieur: Done. Thanks @jennyf19

jennyf19:
>             {

nice. deleting code!

---
In reply to: [1003661334](https://github.com/AzureAD/microsoft-identity-web/pull/1573#issuecomment-1003661334) [](http://example.com/codeflow?ancestors=1003661334,1003661334)

---
Refers to: src/Microsoft.Identity.Web/MergedOptionsValidation.cs:54 in 651d78b. [](commit_id = 651d78b5f84d01422140268383ffb6398b256625, deletion_comment = False)

jennyf19:
>             {

Plus delete the corresponding error messages.

---
In reply to: [1003666999](https://github.com/AzureAD/microsoft-identity-web/pull/1573#issuecomment-1003666999) [](http://example.com/codeflow?ancestors=1003666999,1003661334,1003661334)

---
Refers to: src/Microsoft.Identity.Web/MergedOptionsValidation.cs:54 in 651d78b. [](commit_id = 651d78b5f84d01422140268383ffb6398b256625, deletion_comment = False)

jennyf19 (approved):
:shipit:

src/Microsoft.Identity.Web/ClientAssertionDescription.cs
```diff
@@ -0,0 +1,62 @@
...
+    /// See https://aka.ms/ms-id-web/client-assertions.
+    /// </summary>
+    public class ClientAssertionDescription
+    {
+        /// <summary>
+        /// Constructor of a ClientAssertionDescription.
+        /// </summary>
+        /// <param name="clientAssertionProvider">delegate providing the client assertion
+        /// when it is necessary.</param>
+        public ClientAssertionDescription(Func<CancellationToken, Task<ClientAssertion>> clientAssertionProvider)
```
bgavrilMS, line 21: Having played around with client assertion recently, I think this "func" mechanism is not very good. I am thinking of making the following changes in MSAL: 

- all customers to specify client assertion on the request (i.e. AcquireToken*). 
- move away from FUNC to interface (this is semantically equivalent but interfaces are easier to reason with)
- interface will have a single method with 2 input params instead of 1:

`Task<string> GetClientAssertionAsync(string audience, CancellationToken token)`

This is because computing the audience (which is the token endpoint string) is quite complicated, but MSAL already knows it at the moment it needs the assertion. 

Thoughts?

src/Microsoft.Identity.Web/ClientAssertionDescription.cs
```diff
@@ -0,0 +1,62 @@
...
+        private ClientAssertion? _clientAssertion;
+
+        /// <summary>
+        /// Get the signed assertion (and refreshes it if needed).
+        /// </summary>
+        /// <param name="cancellationToken">Cancellation token.</param>
+        /// <returns>The signed assertion.</returns>
+        public async Task<string> GetSignedAssertion(CancellationToken cancellationToken)
+        {
+            if (_clientAssertion == null || (Expiry != null && DateTimeOffset.Now > Expiry))
```
bgavrilMS, line 43: I don't think this is going to work in a multi-tenant scenario. We had the same problem in MSAL when we were caching them. 

Scenario: 

1. Get app token for tenant T1 
-> assertion A with `aud`: `l.m.o/T1/oauth/v2.0/token` is generated by invoking the FUNC
2. Get app token for tenant T2
-> assertion A is reused but has bad `aud`

jmprieur: This will work in the case for which the feature was provided (cert-less using Managed identity tokens, as the client assertion only depends on the application), but I get the point for other usages.

How did you solve it in MSAL.NET, @bgavrilMS ?
It's not an option to not cache it for MSI, BTW, as MSI will throttle the app after 5 requests for seconds.

bgavrilMS: We disabled assertion caching and rely on token caching only. 

Can you cache the MSI assertion only instead? Is it the same assertion for tokens to all tenants?

`credential.GetTokenAsync(
                    new TokenRequestContext(new[] { "api://AzureADTokenExchange/.default" }, null),
                    cancellationToken);`

Alternatively, let's get this feature into MSAL instead so that you can rely on token caching?

tools/GenerateMergeOptionsMethods/GenerateMergeOptionsMethods.csproj
```diff
@@ -7,6 +7,8 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <Compile Include="..\..\src\Microsoft.Identity.Web\ClientAssertion.cs" Link="OptionClasses\ClientAssertion.cs" />
```
bgavrilMS, line 10: This is cool

Issue #1047, opened 2021-03-05
[Feature Request] Distributed cache is eventually consistent (does not wait for L2)


jmprieur:
**Eventual consistency** is a programming model in which programmers make the assumption that over a long enough period of time and without changes to a system, the current version of a particular program will eventually distribute until every replica of the program is consistent.

**Is your feature request related to a problem? Please describe.**
Today the Distributed token cache adapter awaits for asynchronous operations on the underlying `IDistributedCache`

**Describe the solution you'd like**
- Don't wait for write and remove operations.
- This might require managing an event list of operations.

**Describe alternatives you've considered**
Task.Run() => { try/catch to do the operation} was tried out. See https://github.com/AzureAD/microsoft-identity-web/pull/1043

**Additional context**
See also Wilson's cache: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/1605/files

jennyf19:
Released in [2.5.0](https://github.com/AzureAD/microsoft-identity-web/releases/tag/2.5.0)

PR #1574, opened 2022-01-03
add fire and forget to L2 when L1 is in use

jennyf19:
#1047 

in one sample, difference in write in one case was 10 ms vs 106 ms before fix.

### DIFF
--- a/src/Microsoft.Identity.Web.TokenCache/Distributed/MsalDistributedTokenCacheAdapter.cs
+++ b/src/Microsoft.Identity.Web.TokenCache/Distributed/MsalDistributedTokenCacheAdapter.cs
@@ -263,14 +263,29 @@ namespace Microsoft.Identity.Web.TokenCacheProviders.Distributed
                 SlidingExpiration = _distributedCacheOptions.SlidingExpiration,
             };
 
-            await L2OperationWithRetryOnFailureAsync(
-            write,
-            (cacheKey) => _distributedCache.SetAsync(
-                cacheKey,
-                bytes,
-                distributedCacheEntryOptions,
-                cacheSerializerHints?.CancellationToken ?? CancellationToken.None),
-            cacheKey).Measure().ConfigureAwait(false);
+            if (_distributedCacheOptions.DisableL1Cache)
+            {
+                await L2OperationWithRetryOnFailureAsync(
+                    write,
+                    (cacheKey) => _distributedCache.SetAsync(
+                        cacheKey,
+                        bytes,
+                        distributedCacheEntryOptions,
+                        cacheSerializerHints?.CancellationToken ?? CancellationToken.None),
+                    cacheKey).Measure().ConfigureAwait(false);
+            }
+            else
+            {
+               _ = Task.Run(async () => await
+                L2OperationWithRetryOnFailureAsync(
+                write,
+                (cacheKey) => _distributedCache.SetAsync(
+                    cacheKey,
+                    bytes,
+                    distributedCacheEntryOptions,
+                    cacheSerializerHints?.CancellationToken ?? CancellationToken.None),
+                cacheKey).Measure().ConfigureAwait(false));
+            }
         }
 
         private async Task L2OperationWithRetryOnFailureAsync(
--- a/src/Microsoft.Identity.Web.TokenCache/Distributed/MsalDistributedTokenCacheAdapter.cs
+++ b/src/Microsoft.Identity.Web.TokenCache/Distributed/MsalDistributedTokenCacheAdapter.cs
@@ -114,7 +114,7 @@ namespace Microsoft.Identity.Web.TokenCacheProviders.Distributed
         /// <returns>A <see cref="Task"/> that completes when key removal has completed.</returns>
         protected override async Task RemoveKeyAsync(string cacheKey, CacheSerializerHints cacheSerializerHints)
         {
-            string remove = "Remove";
+            const string remove = "Remove";
 
             if (_memoryCache != null)
             {
@@ -151,7 +151,7 @@ namespace Microsoft.Identity.Web.TokenCacheProviders.Distributed
         /// (account or app).</returns>
         protected override async Task<byte[]> ReadCacheBytesAsync(string cacheKey, CacheSerializerHints cacheSerializerHints)
         {
-            string read = "Read";
+            const string read = "Read";
             byte[]? result = null;
 
             if (_memoryCache != null)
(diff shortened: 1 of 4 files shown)


src/Microsoft.Identity.Web.TokenCache/Distributed/MsalDistributedTokenCacheAdapter.cs
```diff
@@ -263,14 +263,29 @@ protected override async Task WriteCacheBytesAsync(string cacheKey, byte[] bytes
...
 
-            await L2OperationWithRetryOnFailureAsync(
-            write,
-            (cacheKey) => _distributedCache.SetAsync(
-                cacheKey,
-                bytes,
-                distributedCacheEntryOptions,
-                cacheSerializerHints?.CancellationToken ?? CancellationToken.None),
-            cacheKey).Measure().ConfigureAwait(false);
+            if (_distributedCacheOptions.DisableL1Cache)
```
jmprieur, line 266: Would we want to have an option to disable the L2 fire and forget behavior?

jmprieur (approved):
Lgtm
Thanks @jennyf19

src/Microsoft.Identity.Web.TokenCache/Distributed/MsalDistributedTokenCacheAdapter.cs
```diff
@@ -263,14 +263,29 @@ protected override async Task WriteCacheBytesAsync(string cacheKey, byte[] bytes
...
+            {
+               _ = Task.Run(async () => await
+                L2OperationWithRetryOnFailureAsync(
+                write,
+                (cacheKey) => _distributedCache.SetAsync(
+                    cacheKey,
+                    bytes,
+                    distributedCacheEntryOptions,
+                    cacheSerializerHints?.CancellationToken ?? CancellationToken.None),
+                cacheKey).Measure().ConfigureAwait(false));
```
bgavrilMS, line 287: Why is Measure being called twice? It seems to be called once here and once in `L2OperationWithRetryOnFailureAsync`. Does it still work in fire and forget mode?

tests/Microsoft.Identity.Web.Test/L1L2CacheTests.cs
```diff
@@ -149,6 +151,7 @@ private async Task CreateL1L2TestWithAbsoluteOptions(double time)
 
             // Assert
             Assert.Equal(memoryCacheExpectedCount, _testCacheAdapter._memoryCache.Count);
+            await Task.Delay(10).ConfigureAwait(false); // needed for L2 fire&forget
```
bgavrilMS, line 154: This is not a reliable way to test. Typically what happens is that you'll get failures on the CI build every now and then and have increase the delay. As we add more tests, this logic gets copied and the timer gets 

To test "fire and forget" in this case it would be simplest to inject a custom IDistributedCache which internally uses a semaphore. Then the test can wait on the semaphore and `IDistributedCache.Set` will release the sempahore.

Here's an test that does this with ManualResetEvent instead of semaphore: https://github.com/AzureAD/microsoft-authentication-extensions-for-dotnet/blob/master/tests/Microsoft.Identity.Client.Extensions.Msal.UnitTests/MsalCacheHelperTests.cs#L174

src/Microsoft.Identity.Web.TokenCache/Distributed/MsalDistributedTokenCacheAdapter.cs
jennyf19: added. thanks.

tests/Microsoft.Identity.Web.Test/L1L2CacheTests.cs
jennyf19: Yes, I know it's not a good way to test. That's why I had it marked as a draft until I had time to fix the tests. Thanks for the advice. Have fixed.

src/Microsoft.Identity.Web.TokenCache/Distributed/MsalDistributedTokenCacheAdapter.cs
jennyf19: yes it still works based on my testing, compared with before.

jmprieur (approved):
LGTM
Thanks for adding the compatibility flag i suggested, and for applying Bogdan's recommendations.

jmprieur:
@jennyf19 : we have p0
I suggest merging without waiting for the MSAL crew, and starting a release build.

src/Microsoft.Identity.Web.TokenCache/Distributed/MsalDistributedTokenCacheAdapterOptions.cs
```diff
@@ -53,5 +53,13 @@ public class MsalDistributedTokenCacheAdapterOptions : DistributedCacheEntryOpti
         /// </summary>
         /// The default is <c>false.</c>
         public bool DisableL1Cache { get; set; }
+
+        /// <summary>
+        /// Enable writing to the L2 cache to be async (fire and forget).
+        /// This improves performance as the MSAL.NET will not have to wait
+        /// for the write to complete.
+        /// </summary>
+        /// The default is <c>false.</c>
```
bgavrilMS, line 62: Should we default to true here instead as per original proposal?

bgavrilMS (approved):
LGTM, just one question about the default value for the new option.

SameerK-MSFT (approved):

src/Microsoft.Identity.Web.TokenCache/Distributed/MsalDistributedTokenCacheAdapterOptions.cs
jmprieur: @bgavrilMS @SameerK-MSFT:
I disagree, and I've asked @jennyf19 to set it to `false` for now. 
I want first to be proven that this won't break any of the 1000+ partners we migrated already. So I propose that:
1. We have things tested by a set of partners (including R9 / Office) who will set this to true explicitly.
2. When we see that things go will we'll change the default to `true` and that will benefit to everybody.

tests/Microsoft.Identity.Web.Test/TestDistributedCache.cs
```diff
@@ -11,6 +11,7 @@ namespace Microsoft.Identity.Web.Test
     public class TestDistributedCache : IDistributedCache
     {
         internal readonly ConcurrentDictionary<string, Entry> _dict = new ConcurrentDictionary<string, Entry>();
+        internal static ManualResetEventSlim resetEvent { get; set; } = new ManualResetEventSlim(initialState: false);
```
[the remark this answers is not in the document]
jennyf19: fixed. thanks Sameer.

src/Microsoft.Identity.Web.TokenCache/Distributed/MsalDistributedTokenCacheAdapterOptions.cs
bgavrilMS: Sounds reasonable.

Issue #1598, opened 2022-01-17
[Bug] Unable to use user assigned identity to load key vault certificates


admsteck:
The original feature requested by #1007 and added in v1.8.0 appears to no longer be working.

**Which version of Microsoft Identity Web are you using?**
1.21.1

**Where is the issue?**
Protected web APIs call downstream web APIs

**Is this a new or an existing app?**
The app is in production and I have upgraded to a new version of Microsoft Identity Web.

**Repro**

```csharp
                services
                    .AddAuthentication(JwtBearerDefaults.AuthenticationScheme)
                    .AddMicrosoftIdentityWebApi(options => { }, options =>
                    {
                        options.Instance = azureAdConfig.GetValue<string>("Instance");
                        options.TenantId = azureAdConfig.GetValue<string>("TenantId");
                        options.ClientId = azureAdConfig.GetValue<string>("ClientId");
                        options.UserAssignedManagedIdentityClientId = Configuration.GetValue<string>("IdentityClientId");
                        options.ClientCertificates = new CertificateDescription[]
                        {
                            CertificateDescription.FromKeyVault(keyVaultUri, "DownstreamServiceCert")
                        };
                    })
                    .EnableTokenAcquisitionToCallDownstreamApi((_) => { })
                    .AddDownstreamWebApi("DownstreamService", Configuration.GetSection("DownstreamService"))
                    // Enabling the in-memory token cache will allow the JWT access tokens to be cached and used (until a token expires)
                    .AddInMemoryTokenCaches();

```

**Expected behavior**
DefaultCertificateLoader should use the UserAssignedManagedIdentityClientId value provided when loading certificates from a key vault.

**Actual behavior**
The DefaultCertificateLoader.UserAssignedManagedIdentityClientId property used to be set in the TokenAcquisition constructor but does not appear to be set anymore.  DefaultCertificateLoader falls back to using system assigned identity even if a user assigned identity is provided.

**Work around**
Added the following line to our startup.cs.
```csharp
DefaultCertificateLoader.UserAssignedManagedIdentityClientId = Configuration.GetValue<string>("IdentityClientId");
```

jennyf19:
Fixed in [1.22.2 release](https://github.com/AzureAD/microsoft-identity-web/releases/tag/1.22.2).

PR #1606, opened 2022-01-27
fix user assigned managed identity

jennyf19:
Fix for #1598 
The regression happened when implementing the multiple-authentication schemes.


jmprieur (approved):
LGTM
Thanks @jennyf19

PR #1613, opened 2022-01-30
fix for handling event handlers in web APIs

jennyf19:
Fixes https://github.com/AzureAD/microsoft-identity-web/issues/1615
Fixes an issue that when called from an event handler from the web API, the configuration for JwtBearer was not known


jmprieur (approved):
LGTM
Thanks @jennyf19

PR #1614, opened 2022-01-30
Remove beta pckage StyleCop.Analyzers

jennyf19:
Removing the StyleCop.Analyzers beta has it has a dependency named Unstable


jmprieur (approved):
LGTM
Thanks @jennyf19

Issue #1506, opened 2021-10-29
Client Secret not being read from correct environment variable when using app service authentication v2


jwevansgam:
**Which version of Microsoft Identity Web are you using?**
Microsoft Identity Web 1.18.0

**Where is the issue?**
* Web app
    * [ ] Sign-in users
    * [x] Sign-in users and call web APIs
* Web API
    * [ ] Protected web APIs (validating tokens)
    * [ ] Protected web APIs (validating scopes)
    * [ ] Protected web APIs call downstream web APIs
* Token cache serialization
     * [ ] In-memory caches
     * [ ] Session caches
     * [ ] Distributed caches
* Other (please describe)

**Is this a new or an existing app?**
c. This is a new app or an experiment.

**Repro**

```csharp

services.AddMicrosoftIdentityWebAppAuthentication(configuration)
    .EnableTokenAcquisitionToCallDownstreamApi()
    .AddInMemoryTokenCaches();

```

**Expected behavior**
I am running a ASP.NET Core web app and trying to get an access token so I can call a downstream api on behalf of the user. The Web App and Api is hosted in Azure App Service and both are secured using app service authentication (Easy Auth).

When the web app is hosted in Azure App Service I would expect the secret to be read by first reading the environment variable called WEBSITE_AUTH_CLIENT_SECRET_SETTING_NAME and then reading the actual secret from an environment variable with this setting name. By default when app service authentication is enabled the client secret is actually stored in an environment variable called MICROSOFT_PROVIDER_AUTHENTICATION_SECRET.

**Actual behavior**
The class AppServicesAuthenticationTokenAcquisition in Microsoft.Identity.Web is reading the client secret from a variable called WEBSITE_AUTH_CLIENT_SECRET but the actual client secret is stored in a variable called MICROSOFT_PROVIDER_AUTHENTICATION_SECRET so the client secret is not found.

**Possible solution**
![image](https://user-images.githubusercontent.com/43034431/139462896-bf475953-2990-4f7a-80e9-38af56904f40.png)

**Additional context / logs / screenshots**
It is not possible to create an app setting in the Azure App Service called WEBSITE_AUTH_CLIENT_SECRET as I get this error and I can create other app settings.
![image](https://user-images.githubusercontent.com/43034431/139463237-505692c4-05a6-4a45-ad79-2c052a2c8756.png)

Add any other context about the problem here, such as logs and screenshots.

[...]

PR #1621, opened 2022-02-04
fix settingName for easy auth

jennyf19:
#1506


jmprieur (approved):
LGTM
Thanks @jennyf19

PR #1630, opened 2022-02-14
remove test for checking suggested cache key. not needed anymore. let…

jennyf19:
… msal do the right thing.


jennyf19:
> LGTM Are these left over from the time when the suggested cache key was provided in Id.Web ?

Yes, I believe so. @bgavrilMS and I agree that this check is no longer needed. Customers using POP + L2 cache would hit a cache miss today when updating, as new cache key for pop is `{popKid}{clientID}+"_" ...`

jennyf19:
@bgavrilMS going to merge, as we have already discussed this offline.

bgavrilMS:
LGTM

PR #1642, opened 2022-02-25
update to msal4.42

jennyf19:
tested w/web app calls web api calls graph

### DIFF
--- a/tools/GenerateMergeOptionsMethods/GenerateMergeOptionsMethods.csproj
+++ b/tools/GenerateMergeOptionsMethods/GenerateMergeOptionsMethods.csproj
@@ -32,7 +32,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Identity.Client" Version="4.41.0" />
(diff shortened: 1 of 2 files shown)


tools/GenerateMergeOptionsMethods/GenerateMergeOptionsMethods.csproj
```diff
@@ -32,7 +32,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Identity.Client" Version="4.41.0" />
+    <PackageReference Include="Microsoft.Identity.Client" Version="4.40.0" />
```
gladjohn, line 35: moving  back to 4.40?

jennyf19: lol. nice catch. VS had an error and it reverted back to 4.40 and I didn't notice. I have re-tested

gladjohn (approved):
looks good. Thanks @jennyf19

PR #1644, opened 2022-02-26
When deserialization fails, throw a more descriptive error.

### LINKED ISSUE #1643: [Bug] Exception when cache deserialization fails
1.23.0

**Actual behavior**
Related to MSAL issue [3162](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3162).
When distributed cache serialization and encryption is enabled, in certain cases (like when encryption keys are changed), MSAL will fail to deserialize an already serialized cache entry and throw an exception.

**Repro**
1. Use [WebAppCallsMicrosoftGraph](https://github.com/AzureAD/microsoft-identity-web/tree/master/tests/WebAppCallsMicrosoftGraph).
1. Add distributed token caches, data protection (with data protection discriminator), and encryption.
```csharp
...
.AddDistributedTokenCaches();
services.AddDataProtection(o =>
{
    o.ApplicationDiscriminator = "DataProtectionApplicationDiscriminator12";
});

services.Configure<MsalDistributedTokenCacheAdapterOptions>(options =>
{
    options.Encrypt = true;
});
```
3. Set up some distributed cache like SQL or Redis.
4. Login once, so that encrypted token is in the L2 cache. (Don't sign out, it will delete the cached token.)
5. Change the data protection discriminator to something else.
6. Login with the same account again.
7. Data protection will fail decrypting the cache entry, and pass the encrypted entry to MSAL which will fail at deserialization.

**Expected behavior**
Cache is cleared, code execution continues.

pmaytak:
Fixes #1643.

**Changes**
- Clear the cache instead of throwing an exception when deserialization fails.
- Added logging related code for `MsalAbstractTokenCacheProvider`.

**Testing**
Tested with [WebAppCallsMicrosoftGraph](https://github.com/AzureAD/microsoft-identity-web/tree/master/tests/WebAppCallsMicrosoftGraph). All unit tests pass.

### DIFF
--- a/src/Microsoft.Identity.Web.TokenCache/MsalAbstractTokenCacheProvider.Logger.cs
+++ b/src/Microsoft.Identity.Web.TokenCache/MsalAbstractTokenCacheProvider.Logger.cs
@@ -16,26 +16,29 @@ namespace Microsoft.Identity.Web.TokenCacheProviders
         /// </summary>
         private static class Logger
         {
-            private static readonly Action<ILogger, string, string, Exception> s_cacheDeserializationError =
-                LoggerMessage.Define<string, string>(
+            private static readonly Action<ILogger, string, bool, string, Exception> s_cacheDeserializationError =
+                LoggerMessage.Define<string, bool, string>(
                     LogLevel.Warning,
                     LoggingEventId.DistributedCacheConnectionError,
-                    "[MsIdWeb] Unable to deserialize cache entry. Cache key : {CacheKey} Error message: {ErrorMessage} ");
+                    "[MsIdWeb] Unable to deserialize cache entry. Cache key : {CacheKey}. Encryption enabled: {EncryptionEnabled}. Error message: {ErrorMessage} ");
 
             /// <summary>
             /// Cache deserialization error.
             /// </summary>
             /// <param name="logger">ILogger.</param>
             /// <param name="cacheKey">MSAL.NET cache key.</param>
+            /// <param name="encryptionEnabled">Whether cache is encrypted or not.</param>
             /// <param name="errorMessage">Error message.</param>
             /// <param name="ex">Exception.</param>
             public static void CacheDeserializationError(
                 ILogger logger,
                 string cacheKey,
+                bool encryptionEnabled,
                 string errorMessage,
                 Exception ex) => s_cacheDeserializationError(
                     logger,
                     cacheKey,
+                    encryptionEnabled,
                     errorMessage,
                     ex);
         }
--- a/src/Microsoft.Identity.Web.TokenCache/MsalAbstractTokenCacheProvider.Logger.cs
+++ b/src/Microsoft.Identity.Web.TokenCache/MsalAbstractTokenCacheProvider.Logger.cs
@@ -1,0 +1,43 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using Microsoft.Extensions.Logging;
+
+namespace Microsoft.Identity.Web.TokenCacheProviders
+{
+    /// <summary>
+    /// LoggingMessage class for MsalAbstractTokenCacheProvider.
+    /// </summary>
+    public abstract partial class MsalAbstractTokenCacheProvider
+    {
+        /// <summary>
+        /// LoggingMessage class for MsalAbstractTokenCacheProvider.
+        /// </summary>
+        private static class Logger
+        {
+            private static readonly Action<ILogger, string, string, Exception> s_cacheDeserializationError =
+                LoggerMessage.Define<string, string>(
+                    LogLevel.Warning,
+                    LoggingEventId.DistributedCacheConnectionError,
+                    "[MsIdWeb] Unable to deserialize cache entry. Cache key : {CacheKey} Error message: {ErrorMessage} ");
+
+            /// <summary>
+            /// Cache deserialization error.
+            /// </summary>
+            /// <param name="logger">ILogger.</param>
+            /// <param name="cacheKey">MSAL.NET cache key.</param>
+            /// <param name="errorMessage">Error message.</param>
+            /// <param name="ex">Exception.</param>
+            public static void CacheDeserializationError(
+                ILogger logger,
+                string cacheKey,
+                string errorMessage,
+                Exception ex) => s_cacheDeserializationError(
+                    logger,
+                    cacheKey,
+                    errorMessage,
+                    ex);
+        }
+    }
+}
--- a/src/Microsoft.Identity.Web.TokenCache/MsalAbstractTokenCacheProvider.cs
+++ b/src/Microsoft.Identity.Web.TokenCache/MsalAbstractTokenCacheProvider.cs
@@ -116,18 +116,23 @@ namespace Microsoft.Identity.Web.TokenCacheProviders
                 {
                     args.TokenCache.DeserializeMsalV3(UnprotectBytes(tokenCacheBytes), shouldClearExistingCache: true);
                 }
-                catch (Exception ex)
+                catch (MsalClientException exception)
                 {
-                    // clears the cache
-                    args.TokenCache.DeserializeMsalV3(null, shouldClearExistingCache: true);
                     if (_logger != null)
                     {
                         Logger.CacheDeserializationError(
                           _logger,
                           args.SuggestedCacheKey,
-                          ex.Message,
-                          ex);
+                          _protector != null,
+                          exception.Message,
+                          exception);
                     }
+                    // Adding a better message specifically for JSON parsing error
+                    if (exception.ErrorCode == MsalError.JsonParseError)
+                    {
+                        throw new MsalClientException(MsalError.JsonParseError, TokenCacheErrorMessage.ExceptionDeserializingCache, exception);
+                    }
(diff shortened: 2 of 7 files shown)


src/Microsoft.Identity.Web.TokenCache/MsalAbstractTokenCacheProvider.Logger.cs
```diff
@@ -0,0 +1,43 @@
...
+    public abstract partial class MsalAbstractTokenCacheProvider
+    {
+        /// <summary>
+        /// LoggingMessage class for MsalAbstractTokenCacheProvider.
+        /// </summary>
+        private static class Logger
+        {
+            private static readonly Action<ILogger, string, string, Exception> s_cacheDeserializationError =
+                LoggerMessage.Define<string, string>(
+                    LogLevel.Warning,
```
pmaytak, line 21: Although this is an exception, setting log level to Warning, since the execution is continued.

src/Microsoft.Identity.Web.TokenCache/MsalAbstractTokenCacheProvider.cs
```diff
@@ -98,7 +112,23 @@ private async Task OnBeforeAccessAsync(TokenCacheNotificationArgs args)
...
 
-                args.TokenCache.DeserializeMsalV3(UnprotectBytes(tokenCacheBytes), shouldClearExistingCache: true);
+                try
+                {
+                    args.TokenCache.DeserializeMsalV3(UnprotectBytes(tokenCacheBytes), shouldClearExistingCache: true);
+                }
+                catch (Exception ex)
+                {
+                    // clears the cache
+                    args.TokenCache.DeserializeMsalV3(null, shouldClearExistingCache: true);
```
jennyf19, line 122: What if you have several instances of a web app or web API, and they all share the cache and sharing the decryption keys. One has misconfigured keys and clears the cache, but this will impact the others using the cache. Is this okay?

neha-bhargava (approved):

src/Microsoft.Identity.Web.TokenCache/MsalAbstractTokenCacheProvider.cs
pmaytak: Yes, that could happen. If you have a lot of machines which use their own keys, then each machine will retrieve token from AAD and keep overwriting the cache entry. Thinking further, Line 122 doesn't even matter much, as even without explicitly clearing the internal cache, it will still be empty, which I think I saw in testing.
Example scenario:
1. Id.Web instantiates MSAL, internal cache is empty.
2. This callback is called to get the cache entry from external cache, but fails to decrypt it.
3. Exception is caught, cache is still empty. (Clearing the cache here doesn't matter).
4. MSAL doesn't find token in internal cache. Retrieves one from AAD.
5. New token is serialized with the same cache key, encrypted, and overwriting the existing cache entry.

I think @bgavrilMS suggested adding something like an encryption key identifier to the cache key.

src/Microsoft.Identity.Web.TokenCache/MsalAbstractTokenCacheProvider.Logger.cs
```diff
@@ -0,0 +1,43 @@
...
+                    LoggingEventId.DistributedCacheConnectionError,
+                    "[MsIdWeb] Unable to deserialize cache entry. Cache key : {CacheKey} Error message: {ErrorMessage} ");
```
bgavrilMS, line 23: Can you add an indication if encryption is being used?

src/Microsoft.Identity.Web.TokenCache/MsalAbstractTokenCacheProvider.cs
bgavrilMS: This is a great question. On desktop apps, we do indeed delete the cache if we cannot decrypt it. This is done in order to not crash the app. Works great for long lived apps, but it **silently hides errors.**

So any solution where we hide exception is a compromise between: 

- do not crash the app 
- do not hide the error because it can lead to unexpected behavior 

For L2 cache `id.web.TokenCache` already made a choice to not crash the app. This error occurs lower down the stack though, i.e. we are able to read data from L2, but we cannot decrypt this data. 

My 2c on this, after reading a bunch about DataProtection API (see my next comment for details), is to not hide this error. We should however improve the devex, by re-throwing a more meaningful exception.

Thoughts @jennyf19 and @pmaytak ? CC @henrik-me and @rymeskar as well for an opinion.

bgavrilMS: TLDR what the customer should try to do when configuring DataProtection API:

```csharp

.UnprotectKeysWithAnyCertificate(
        new X509Certificate2("certificate_1.pfx", builder.Configuration["CertificatePassword_1"]),
        new X509Certificate2("certificate_2.pfx", builder.Configuration["CertificatePassword_2"]));

```

src/Microsoft.Identity.Web.TokenCache/MsalAbstractTokenCacheProvider.cs
```diff
@@ -98,7 +112,23 @@ private async Task OnBeforeAccessAsync(TokenCacheNotificationArgs args)
...
+                    if (_logger != null)
+                    {
+                        Logger.CacheDeserializationError(
```
bgavrilMS, line 125: DataProtection API is a complex system that deals with encryption key rotation and key distribution among all the machines involved in the API. This is all taken care of for you, which is pretty amazing. Please see https://docs.microsoft.com/en-us/aspnet/core/security/data-protection/implementation/key-management?view=aspnetcore-6.0 for details.

But keys also need to be encrypted at rest. This is where it gets tricky.

The customer who logged this bug uses certificates for key encryption at rest: 

```csharp
builder.Services.AddDataProtection()
    .PersistKeysToBlobStorage(...)
    .ProtectKeysWithCertificate(cert)
```

This works, but when it comes time to rotate that cert, rotation needs to be handled explicitly by the app developer: 

```csharp
builder.Services.AddDataProtection()
    .PersistKeysToBlobStorage(...)
    .ProtectKeysWithCertificate(new_cert)
    .UnprotectKeysWithAnyCertificate(new_cert, old_cert);
```

This is described here: https://docs.microsoft.com/en-us/aspnet/core/security/data-protection/configuration/overview?view=aspnetcore-6.0#unprotectkeyswithanycertificate

bgavrilMS: @pmaytak - pls ask the app developer who raised this issue to try this out.

src/Microsoft.Identity.Web.TokenCache/MsalAbstractTokenCacheProvider.cs
```diff
@@ -98,7 +112,28 @@ private async Task OnBeforeAccessAsync(TokenCacheNotificationArgs args)
...
+                          _logger,
+                          args.SuggestedCacheKey,
+                          _protector != null,
+                          exception.Message,
+                          exception);
+                    }
+                    // Adding a better message specifically for JSON parsing error
+                    if (exception.ErrorCode == MsalError.JsonParseError)
```
pmaytak, line 131: Catching just the parsing error, since this is the specific case where more info needs to be added.

src/Microsoft.Identity.Web.TokenCache/MsalAbstractTokenCacheProvider.cs
pmaytak: Made the changes here: 779904c45dbedd08585f3949e2279b6d3de1532d

> Maybe add to the [retry mechanism](https://github.com/AzureAD/microsoft-identity-web/blob/bd48335ced3fdff85d83cf4e19f8c5fcc04cc982/src/Microsoft.Identity.Web.TokenCache/Distributed/MsalDistributedTokenCacheAdapter.cs#L321)?

The retry happens before the call to deserialize. We don't know if the deserialization failed at that time.

bgavrilMS (approved):

src/Microsoft.Identity.Web.TokenCache/TokenCacheErrorMessage.cs
```diff
@@ -6,5 +6,6 @@ namespace Microsoft.Identity.Web
     internal static class TokenCacheErrorMessage
     {
         public const string InitializeAsyncIsObsolete = "IDW10801: Use Initialize instead. See https://aka.ms/ms-id-web/1.9.0. ";
+        public const string ExceptionDeserializingCache = "IDW10802: Exception occurred while deserializing token cache. See https://aka.ms/msal-net-token-cache-serialization general guidance and https://aka.ms/ms-id-web/token-cache-troubleshooting for token cache troubleshooting information.";
```
jennyf19, line 9: Is there an action the developer needs to take when they see this error message? It's not clear in the aka.ms links what the action should be when receiving this message.

jennyf19 (approved):
one comment on the exception message. :shipit:

src/Microsoft.Identity.Web.TokenCache/TokenCacheErrorMessage.cs
pmaytak: Updated now.

PR #1648, opened 2022-03-09
Fix null ref when calling from an anonymous controller on a web app

jennyf19:


src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -386,7 +386,13 @@ private void LogAuthResult(AuthenticationResult? authenticationResult)
             if (string.IsNullOrEmpty(mergedOptions.Instance))
             {
                 var mergedOptionsMonitor = _serviceProvider.GetService<IOptionsMonitor<JwtBearerOptions>>();
-                mergedOptionsMonitor?.Get(JwtBearerDefaults.AuthenticationScheme);
```
jmprieur, line 389: I wonder why we had JwtBearerDefaults.AuthenticationScheme.
Probably because it was for a background process on in a web API, and it that case the default scheme was JwtBearerOptions. but indeed, it's better to use `authenticationScheme`

jmprieur (approved):
LGTM
Thanks @jennyf19

src/Microsoft.Identity.Web/TokenAcquisition.cs
jennyf19: Yes, that's what I remember.

PR #1663, opened 2022-03-21
[TokenCache] Downgrade dependency version for Microsoft.Extensions on net472

### LINKED ISSUE #1662: [Bug] Dependencies in Net Framework TokenCache package break Kestrel on Net Framework
Microsoft.Identity.Web.TokenCache all versions

**Where is the issue?**
* Token cache nuget dependencies

Existing app, trying to migrate to .Net Core via older Asp.Net Core release.

**Repro**
Install Microsoft.Identity.Web.TokenCache on a service using net472, and aspnetcore 2.1 or 2.2

**Expected behavior**
Installing the nuget package should not break aspnetcore

**Actual behavior**
Microsoft.Identity.Web.TokenCache pulls in packages from Microsoft.Extensions 5.0.0.
A change to Microsoft.Extensions.Primitives in this release breaks the Asp.Net Core versions that run on .Net Framework, essentially the Kestrel worker thread crashes.

**Possible solution**
Depend on a 3.1.x version of Microsoft.Extensions for Net Framework.

**Additional context / logs / screenshots / link to code**
https://github.com/dotnet/aspnetcore/issues/28112 describes the issue in more detail. For projects stuck on .Net Framework for some reason or other, and those trying to modernize their codebases piecewise, this dependency issue is pretty frustrating.

stephenjust:
Fixes #1662

Microsoft.Extensions.* 5.0 removed a component which breaks Kestrel when running on Net Framework.

I'm only adjusting the net472 target because net462 users typically wouldn't be using Kestrel due to some challenges with netstandard packages.

### DIFF
--- a/src/Microsoft.Identity.Web.TokenCache/Microsoft.Identity.Web.TokenCache.csproj
+++ b/src/Microsoft.Identity.Web.TokenCache/Microsoft.Identity.Web.TokenCache.csproj
@@ -82,14 +82,14 @@
     <DocumentationFile>Microsoft.Identity.Web.xml</DocumentationFile>
   </PropertyGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'net472' Or '$(TargetFramework)' == 'net462' Or '$(TargetFramework)' == 'net5.0'">
+  <ItemGroup Condition="'$(TargetFramework)' == 'net462' Or '$(TargetFramework)' == 'net5.0'">
     <PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="5.0.0" />
     <PackageReference Include="Microsoft.Extensions.Logging" Version="5.0.0" />
     <PackageReference Include="Microsoft.AspNetCore.DataProtection" Version="5.0.8" />
     <PackageReference Include="System.Text.Encodings.Web" Version="5.0.1" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0' or '$(TargetFramework)' == 'net472'">
     <PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="2.1.0" />
     <PackageReference Include="Microsoft.Extensions.Logging" Version="2.1.0" />
     <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="2.1.0" />


src/Microsoft.Identity.Web.TokenCache/Microsoft.Identity.Web.TokenCache.csproj
```diff
@@ -82,14 +82,14 @@
...
-  <ItemGroup Condition="'$(TargetFramework)' == 'net472' Or '$(TargetFramework)' == 'net462' Or '$(TargetFramework)' == 'net5.0'">
+  <ItemGroup Condition="'$(TargetFramework)' == 'net462' Or '$(TargetFramework)' == 'net5.0'">
     <PackageReference Include="Microsoft.Extensions.Caching.Memory" Version="5.0.0" />
     <PackageReference Include="Microsoft.Extensions.Logging" Version="5.0.0" />
     <PackageReference Include="Microsoft.AspNetCore.DataProtection" Version="5.0.8" />
     <PackageReference Include="System.Text.Encodings.Web" Version="5.0.1" />
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0' or '$(TargetFramework)' == 'net472'">
```
bgavrilMS, line 92: @jennyf19 @jmprieur @pmaytak - this package has no conditional compilation at all, so the only reason we have all these frameworks it to be able to have different dependencies, i.e. different versions of dependencies.

In particular, net462 and net472 seem identical?

If we get net4*2 to have the same dependencies as netstandard, we might as well eliminate net4*2 completely.

bgavrilMS: I don't remember the details of the investigation into the target framework, but I see that most packages try to simplify things into supporting only `netstandard2.0`.

stephenjust: I will note that while I was trying to assemble a simple repro of this issue, a project on net462 used the netstandard dependency set for some reason, while net472 used the expected net472 dependencies. I didn't look into it too deeply though, my org's projects are all net472.

jmprieur (approved):
LGTM
I agree that net472 brings netstandard20
Thanks @stephenjust 

cc: @GeoK

jayesh-a-shah:
Can you please make this change for net462 as well?

bgavrilMS (approved):

jmprieur:
> Can you please make this change for net462 as well?

What would be the reason, @jayesh-a-shah ? We carefully chose the dependencies, and unless there is a blocker with repro steps, I'm not going to accept that they are changed.

@stephenjust @jayesh-a-shah out of curiosity, did you try to bump-up these extensions to version 8?

jmprieur:
what does this mean @bgavrilMS ?

bgavrilMS:
> what does this mean @bgavrilMS ?

Id.Web no longer targets the 5.x version of Microsoft.Extensions.*

PR #1672, opened 2022-03-28
update mergedOptions

jennyf19:


src/Microsoft.Identity.Web/Constants/IDWebErrorMessage.cs
```diff
@@ -50,6 +50,7 @@ internal static class IDWebErrorMessage
         public const string ExceptionAcquiringTokenForConfidentialClient = "IDW10501: Exception acquiring token for a confidential client. ";
         public const string MicrosoftIdentityWebChallengeUserException = "IDW10502: An MsalUiRequiredException was thrown due to a challenge for the user. " +
            "See https://aka.ms/ms-id-web/ca_incremental-consent. ";
+        public const string ProvidedAuthenticationSchemeIsIncorrect = "IDW10503: Cannot determine the cloud Instance. The provided authentication scheme was {0}. Microsoft.Identity.Web inferred {1} as the authentication scheme. Available authentication schemes are {2}. See https://aka.ms/id-web/authSchemes. ";
```
jmprieur, line 53: I'd propose we add quotes around the parameters. Sometimes the auth scheme is empty. This will make it more visible. For instance:

image


```suggestion
        public const string ProvidedAuthenticationSchemeIsIncorrect = "IDW10503: Cannot determine the cloud Instance. The provided authentication scheme was '{0}'. Microsoft.Identity.Web inferred '{1}' as the authentication scheme. Available authentication schemes are '{2}'. See https://aka.ms/id-web/authSchemes. ";
```

jmprieur (approved):
Awesome, @jennyf19 
This is a great supportability item.

Issue #1667, opened 2022-03-23
[Feature Request] Improved support for inheriting / customizing MicrosoftIdentity*AuthenticationHandler


MattKotsenas:
# Is your feature request related to a problem?
My specific scenario is likely niche, but I think the capability would be broadly applicable. For context, in my case I have a web API that makes protected downstream web API calls using the `AddMicrosoftIdentityAppAuthenticationHandler` route. I chose this method over `IDownstreamApi` because I have a set of HttpClient delegating handlers (like retry and backoff policies) that vary per HttpClient type.

My specific scenario is that my downstream API has a mix of authenticated and anonymous endpoints, and the `AzureAd` config provided to each instance of my web API may be different and may have been _misconfigured by the user_. That means that if the user provides a bad configuration, all downstream API calls will fail, even the anonymous ones, because there's no `try...catch` around token acquisition in `MicrosoftIdentityAppAuthenticationMessageHandler.cs` (see https://github.com/AzureAD/microsoft-identity-web/blob/2f465726cf62401b474243ef1805f494e95f7401/src/Microsoft.Identity.Web/DownstreamWebApiSupport/MicrosoftIdentityAppAuthenticationMessageHandler.cs#L44-L49) and thus failing to acquire a token is an uncaught exception.

# Describe the workaround you have today
Today I work around this issue by copying the entirety of `MicrosoftIdentityAppAuthenticationMessageHandler.cs` into my library to add the `try...catch`, and then additionally I need to create a new override of `AddMicrosoftIdentityAppAuthenticationHandler` with the implementation copied so that I can add my "safe" handler as the HTTP message handler (see https://github.com/AzureAD/microsoft-identity-web/blob/2f465726cf62401b474243ef1805f494e95f7401/src/Microsoft.Identity.Web/DownstreamWebApiSupport/MicrosoftIdentityAuthenticationMessageHandlerHttpClientBuilderExtensions.cs#L97).

# Describe the solution you'd like
Depending on the level of support that makes sense for a scenario like this, I propose a tiered approach:

## Tier 1: Register and retrieve the MicrosoftIdentity*AuthenticationHandler classes from DI

If instead of calling `new` directly on the handler classes in the extension methods, if the handlers were registered as dependencies and resolved from the `ServiceCollection`, I could inherit from the handler to do my custom work, which would eliminate the need to also duplicate the extension method. Bonus points if the authentication handlers gain an interface.

[...]

PR #1674, opened 2022-03-28
Add `IMicrosoftIdentityAuthenticationDelegatingHandlerFactory` to allow for overriding handlers

MattKotsenas:
Currently, it's very difficult to customize the behavior of the
`DelegatingHandler` classes that add the auth token, as the
`DownstreamWebApiExtensions` `new()` up the handlers directly.

Ideally I would register the handlers in the `IServiceCollection`
directly, and then resolve it from the container, however in the case of
multiple downstream APIs we need to resolve the correct named
configuration from the `IOptionsMonitor`, and I'm not aware of a way to
do that with just the dependency injection abstraction APIs.

As a result, I've instead introduced a public factory interface that
takes the configuration name, which then
allows for creating the handler from the factory class. Users that want
to customize the handlers can then register their own
`IMicrosoftIdentityAuthenticationDelegatingHandlerFactory` 
that creates their own handlers.


jennyf19:
@qetza Do you mind reviewing this as well? thanks.

src/Microsoft.Identity.Web/DownstreamWebApiSupport/MicrosoftIdentityAuthenticationDelegatingHandlerFactory.cs
```diff
@@ -0,0 +1,73 @@
...
+        /// The <see cref="IServiceProvider"/> to resolve dependencies from.
+        /// </param>
+        /// <param name="serviceName">
+        /// Name of the service describing the downstream web API. Used to
+        /// retrieve the appropriate config section.
+        /// </param>
+        /// <returns>
+        /// The <see cref="DelegatingHandler"/>.
+        /// </returns>
+        DelegatingHandler CreateAppHandler(IServiceProvider serviceProvider, string? serviceName);
```
qetza, line 31: The `IServiceProvider` should not be given as a parameter of the method but managed by the class implementing the interface (maybe some implementation doesn't need an IServiceProvider)

src/Microsoft.Identity.Web/DownstreamWebApiSupport/MicrosoftIdentityAuthenticationDelegatingHandlerFactory.cs
```diff
@@ -0,0 +1,73 @@
...
+        /// <param name="serviceProvider"></param>
+        /// The <see cref="IServiceProvider"/> to resolve dependencies from.
+        /// <param name="serviceName">
+        /// Name of the service describing the downstream web API. Used to
+        /// retrieve the appropriate config section.
+        /// </param>
+        /// <returns>
+        /// The <see cref="DelegatingHandler"/>.
+        /// </returns>
+        DelegatingHandler CreateUserHandler(IServiceProvider serviceProvider, string? serviceName);
```
qetza, line 46: Same as above 😄

src/Microsoft.Identity.Web/DownstreamWebApiSupport/MicrosoftIdentityAuthenticationDelegatingHandlerFactory.cs
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    /// <summary>
+    /// The default implementation of <see cref="IMicrosoftIdentityAuthenticationDelegatingHandlerFactory"/>.
+    /// </summary>
+    public class MicrosoftIdentityAuthenticationDelegatingHandlerFactory : IMicrosoftIdentityAuthenticationDelegatingHandlerFactory
```
qetza, line 52: As this is the default implementation and will be managed by the extension methods the class could be `internal` and named `DefaultMicrosoftIdentityAuthenticationDelegatingHandlerFactory`

qetza, line 52: The class should get the `IServiceProvider` as a dependency injection in the constructor and use it internally.

src/Microsoft.Identity.Web/DownstreamWebApiSupport/MicrosoftIdentityAuthenticationMessageHandlerHttpClientBuilderExtensions.cs
```diff
@@ -32,13 +31,12 @@ public static class MicrosoftIdentityAuthenticationMessageHandlerHttpClientBuild
             }
 
             builder.Services.Configure<MicrosoftIdentityAuthenticationMessageHandlerOptions>(serviceName, configuration);
+            builder.Services.AddSingleton<IMicrosoftIdentityAuthenticationDelegatingHandlerFactory, MicrosoftIdentityAuthenticationDelegatingHandlerFactory>();
```
qetza, line 34: You should use `TryAddSingleton` on `IMicrosoftIdentityAuthenticationDelegatingHandlerFactory` so that a user can configure his own implementation before and we will not override it.

As this is done in all extension methods maybe put this in a `AddMicrosoftIdentityAuthenticationHandlerCore` private method called by all other extension methods so that common configuration is shared.

qetza (changes requested):
Hi @MattKotsenas,
This is a good idea, i've added some comment.

src/Microsoft.Identity.Web/DownstreamWebApiSupport/MicrosoftIdentityAuthenticationDelegatingHandlerFactory.cs
MattKotsenas: Good idea! I removed `IServiceProvider` from the interface's signature and added it for constructor injection in the default implementation.

src/Microsoft.Identity.Web/DownstreamWebApiSupport/MicrosoftIdentityAuthenticationDelegatingHandlerFactory.cs
MattKotsenas: 👍

MattKotsenas:
Thanks for all the feedback @qetza! I believe I've addressed all your comments. For some reason clicking the "re-request review" button doesn't seem to be doing anything for me, so I'm tagging you here to take another look. Thanks again!

jennyf19 (approved):

src/Microsoft.Identity.Web/DownstreamWebApiSupport/IMicrosoftIdentityAuthenticationDelegatingHandlerFactory.cs
```diff
@@ -0,0 +1,40 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Net.Http;
+
+namespace Microsoft.Identity.Web
+{
+    /// <summary>
+    /// Interface to a class that provides the <see cref="DelegatingHandler"/> that
+    /// addes an authorization header with a token for the application.
```
jmprieur, line 10:
```suggestion
    /// adds an authorization header with a token for the application.
```

jmprieur (approved):
LGTM

src/Microsoft.Identity.Web/DownstreamWebApiSupport/IMicrosoftIdentityAuthenticationDelegatingHandlerFactory.cs
MattKotsenas: Fixed!

jennyf19:
#1667

PR #1675, opened 2022-03-28
Add abstract method to MicrosoftIdentityAuthenticationBaseMessageHandler so subclasses can more easily override

MattKotsenas:
Add an abstract method `GetToken()` on
`MicrosoftIdentityAuthenticationBaseMessageHandler` in order
to more easily let inheriting implementations to override token
options or acquisition behavior.

### DIFF
--- a/src/Microsoft.Identity.Web/DownstreamWebApiSupport/MicrosoftIdentityAppAuthenticationMessageHandler.cs
+++ b/src/Microsoft.Identity.Web/DownstreamWebApiSupport/MicrosoftIdentityAppAuthenticationMessageHandler.cs
@@ -7,6 +7,7 @@ using System.Threading;
 using System.Threading.Tasks;
 
 using Microsoft.Extensions.Options;
+using Microsoft.Identity.Client;
 
 namespace Microsoft.Identity.Web
 {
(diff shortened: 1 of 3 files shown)


src/Microsoft.Identity.Web/DownstreamWebApiSupport/MicrosoftIdentityAppAuthenticationMessageHandler.cs
```diff
@@ -29,6 +30,22 @@ public class MicrosoftIdentityAppAuthenticationMessageHandler : MicrosoftIdentit
...
+            if (options == null)
+            {
+                throw new ArgumentNullException(nameof(options));
+            }
+
+            return await TokenAcquisition.GetAuthenticationResultForAppAsync(
+                options.Scopes!,
+                options.AuthenticationScheme,
+                options.Tenant,
+                options.TokenAcquisitionOptions)
```
jennyf19, line 45: The user flow is no longer included.

MattKotsenas: Sorry, can you expand on the issue? In the app case, I don't see the user flow in the old code either. I do include the user flow in the "ForUser" case.

If I'm missing something, please let me know. Thanks!

jennyf19:
I will need to spend more time looking at this one. Thanks for the PR.

jennyf19:
I will need to spend more time looking at this one. Thanks for the PR.

Issue #1677, opened 2022-03-29
Wrong parameter description.


krispenner:
https://github.com/AzureAD/microsoft-identity-web/blob/93e06894eff32179f5bbc1f37375eac6a12f7e2c/src/Microsoft.Identity.Web/WebAppExtensions/MicrosoftIdentityWebAppServiceCollectionExtensions.cs#L30

Change this:
```cs
/// (by default OpenIdConnectDefaults.AuthenticationScheme).</param> 
```

To this:
```cs
/// (by default CookieAuthenticationDefaults.AuthenticationScheme).</param> 
```

jennyf19:
Thanks @krispenner

jennyf19:
Thanks @krispenner

jennyf19:
Released in [1.24.0](https://github.com/AzureAD/microsoft-identity-web/releases/tag/1.24.0).

PR #1679, opened 2022-03-31
fix param name

jennyf19:
#1677


jmprieur (approved):
LGTM
Thanks @jennyf19

PR #1692, opened 2022-04-07
simplify string comparison

jennyf19:


jmprieur (approved):
LGTM
Thanks @jennyf19

src/Microsoft.Identity.Web/AuthorizeForScopesAttribute.cs
```diff
@@ -169,8 +169,8 @@ public override void OnException(ExceptionContext context)
 
         private static bool IsAjaxRequest(HttpRequest request)
         {
-            return string.Equals(request.Query[Constants.XRequestedWith], Constants.XmlHttpRequest, StringComparison.Ordinal) ||
-                string.Equals(request.Headers[Constants.XRequestedWith], Constants.XmlHttpRequest, StringComparison.Ordinal);
+            return request.Query[Constants.XRequestedWith] == Constants.XmlHttpRequest ||
```
bgavrilMS, line 172: The `==` operator is equivalent to a culture sensitive comparison, which is slower than ordinal.

PR #1704, opened 2022-04-13
Fix error message to LocalMachine and not CurrentMachine

jennyf19:


jmprieur (approved):
LGTM
Thanks @jennyf19

PR #1711, opened 2022-04-19
fix for 1707

### LINKED ISSUE #1707: [Bug] ScopeAuthorizationHandler only checks first scope claim
1.23.1

**Where is the issue?**
* Web API
    * [x ] Protected web APIs (validating tokens)

This is a new app or an experiment.

**Context**

Using Okta, scopes are separated into multiple claims. Imo, this is the correct representation once parsed and in the claims collection (not space delimited).

When the ScopeAuthorizationHandler runs, it only checks for the first claim. If the first claim is one of the claims listed in the `RequiredScope` attribute, then the request is authorized. If the first claim is not in the `RequiredScope` attribute, then it is not authorized. 

**Expected behavior**
All scopes (in all scope claims) should be tested such that, if a scope in any of the claims is listed in the `RequiredScope` attribute, then the request is authorized; otherwise, it should not be authorized.

**Actual behavior**
Only the first claim is tested. So, the authorization becomes based on the claim order which is incorrect.

**Possible solution**
Replace [this](https://github.com/AzureAD/microsoft-identity-web/blob/88e6a4ea58df80416b7808e7714e3001670f1626/src/Microsoft.Identity.Web/Policy/ScopeAuthorizationHandler.cs#L84-L95)

with
```csharp
          var scopeClaims = context.User.FindAll(ClaimConstants.Scp)
              .Union(context.User.FindAll(ClaimConstants.Scope))
              .ToList();

          if (!scopeClaims.Any()) {
              return Task.CompletedTask;
          }

          var hasScope = scopeClaims.SelectMany(s => s.Value.Split(' ')).Intersect(scopes).Any();

          if (hasScope) {
              context.Succeed(requirement);
              return Task.CompletedTask;
          }
```

jennyf19:
Fixes #1707

### DIFF
--- a/src/Microsoft.Identity.Web/Policy/ScopeAuthorizationHandler.cs
+++ b/src/Microsoft.Identity.Web/Policy/ScopeAuthorizationHandler.cs
@@ -81,14 +81,18 @@ namespace Microsoft.Identity.Web
                 return Task.CompletedTask;
             }
 
-            Claim? scopeClaim = context.User.FindFirst(ClaimConstants.Scp) ?? context.User.FindFirst(ClaimConstants.Scope);
+            var scopeClaims = context.User.FindAll(ClaimConstants.Scp)
+              .Union(context.User.FindAll(ClaimConstants.Scope))
+              .ToList();
 
-            if (scopeClaim is null)
+            if (!scopeClaims.Any())
             {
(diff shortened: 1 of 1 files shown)


jmprieur (approved):
LGTM
Thanks @jennyf19

PR #1715, opened 2022-04-20
add log message for auth scheme

jennyf19:


jmprieur (approved):
LGTM
Thanks @jennyf19

PR #1717, opened 2022-04-22
Return TokenValidatedContext.Fail instead of throwing UnauthorizedAccessException in case of missing Roles / Scopes

### LINKED ISSUE #1716: [Bug] Middleware throws System.UnauthorizedAccessException, causes 500 error for users
1.22.3 (but from code inspection will also repro on the latest version)

**Where is the issue?**
* Web app
    * [ ] Sign-in users
    * [X] Sign-in users and call web APIs
* Web API
    * [X] Protected web APIs (validating tokens)
    * [X] Protected web APIs (validating scopes)
    * [ ] Protected web APIs call downstream web APIs
* Token cache serialization
     * [ ] In-memory caches
     * [ ] Session caches
     * [ ] Distributed caches

This is an existing app that we're building; I haven't upgraded Microsoft Identity Web, but started seeing this issue.

**Repro**
There isn't an obvious code repro here, in our particular case we believe it was from trying to run our app without yet granting admin consent for the required roles.

**Expected behavior**
Like with other bad tokens or configuration issues (like a missing `tid` or `tenantId` claim) I would expect the middleware pipeline to return a 401 Unauthorized

**Actual behavior**
The `JwtBearerHandler` will throw an `UnauthorizedAccessException` that looks like this:

```    
System.UnauthorizedAccessException: IDW10201: Neither scope or roles claim was found in the bearer token.
  at Microsoft.Identity.Web.MicrosoftIdentityWebApiAuthenticationBuilderExtensions.<>c__DisplayClass3_1.<<AddMicrosoftIdentityWebApiImplementation>b__1>d.MoveNext()
  at Microsoft.AspNetCore.Authentication.JwtBearer.JwtBearerHandler.HandleAuthenticateAsync()
  at Microsoft.AspNetCore.Authentication.JwtBearer.JwtBearerHandler.HandleAuthenticateAsync()
  at Microsoft.AspNetCore.Authentication.AuthenticationHandler`1.AuthenticateAsync()
  at Microsoft.AspNetCore.Authentication.AuthenticationService.AuthenticateAsync(HttpContext context, String scheme)
  at Microsoft.AspNetCore.Authentication.AuthenticationMiddleware.Invoke(HttpContext context)
``` 
which results in an unhandled exception in the auth middleware pipeline, which in turn results in a 500 error for the user.

**Possible solution**
A workaround for users hitting this issue today is to add the following into your `IServiceCollection`:

```csharp
services.Configure<JwtBearerOptions>(scheme, options =>
{
    options.Events ??= new JwtBearerEvents();

    var tokenValidatedHandler = options.Events.OnTokenValidated;
    options.Events.OnTokenValidated = async context =>
    {
        try
        {
            await tokenValidatedHandler(context).ConfigureAwait(false);
        }
        catch (UnauthorizedAccessException e)
        {
            context.Fail(e);
        }
    };
});
```
This converts the unhandled exception into a handled auth failure that automatically does the appropriate Challenge response (at the expense of an exception throw / catch in a possible hot path).

Since other configuration issues (as mentioned in the "Expected Behavior" section) result in 401s, I would expect this issue to behave similarly.

I'll post a PR shortly with a proposed solution.

MattKotsenas:
Today, if a token is valid, but both:

1. Has neither Scopes nor Roles claims
2. Not authorized via ACL

Then the `JwtBearerHandler` will throw an `UnauthorizedAccessException` that looks like this:

```
System.UnauthorizedAccessException: IDW10201: Neither scope or roles claim was found in the bearer token.
  at Microsoft.Identity.Web.MicrosoftIdentityWebApiAuthenticationBuilderExtensions.<>c__DisplayClass3_1.<<AddMicrosoftIdentityWebApiImplementation>b__1>d.MoveNext()
  at Microsoft.AspNetCore.Authentication.JwtBearer.JwtBearerHandler.HandleAuthenticateAsync()
  at Microsoft.AspNetCore.Authentication.JwtBearer.JwtBearerHandler.HandleAuthenticateAsync()
  at Microsoft.AspNetCore.Authentication.AuthenticationHandler`1.AuthenticateAsync()
  at Microsoft.AspNetCore.Authentication.AuthenticationService.AuthenticateAsync(HttpContext context, String scheme)
  at Microsoft.AspNetCore.Authentication.AuthenticationMiddleware.Invoke(HttpContext context)
```

which by default results in an unhandled exception in the auth middleware pipeline, which in turn results in a 500 error for the user.
    
While one could make the argument that a 500 is appropriate in this scenario, as the most likely cause is a misconfiguration on the AAD side, similar types of misconfiguration, such as a missing `tid` or `tenantId` claim results in a 401 Unauthorized error.

In order to make this change testable, I did the refactor in three steps, each in its own commit, to make the process easier to follow.

First, I moved the Roles / Scopes validation logic from the registration codepath into a new internal helper method, so that it could be called directly from unit tests. I used the `JwtBearerMiddlewareDiagnosticsTests` as a template.

Second, to simplify the new helper, I moved the `AllowWebApiToBeAuthorizedByACL` check out of the token validation codepath, and instead returned it to the registration codepath. This is both a minor performance win for ACL cases (because it skips the event callback when it isn't needed) and slightly reduces complexity of the helper (by making the conditional simpler), at the expense of making the ACL case difficult to test again. If there's desire to test the ACL case, we can make follow up changes.

Lastly, with the logic in a helper and tests written to validate the behavior, I replace the exception with a call to `TokenValidatedContext.Fail(message)`.

Fixes #1716

### DIFF
--- a/src/Microsoft.Identity.Web/WebApiExtensions/MicrosoftIdentityWebApiAuthenticationBuilderExtensions.cs
+++ b/src/Microsoft.Identity.Web/WebApiExtensions/MicrosoftIdentityWebApiAuthenticationBuilderExtensions.cs
@@ -232,7 +232,11 @@ namespace Microsoft.Identity.Web
 
                     // When an access token for our own web API is validated, we add it to MSAL.NET's cache so that it can
                     // be used from the controllers.
-                    ChainOnTokenValidatedEventForClaimsValidation(options.Events, jwtBearerScheme, mergedOptions.AllowWebApiToBeAuthorizedByACL);
+
+                    if (!mergedOptions.AllowWebApiToBeAuthorizedByACL)
+                    {
+                        ChainOnTokenValidatedEventForClaimsValidation(options.Events, jwtBearerScheme);
+                    }
 
                     if (subscribeToJwtBearerMiddlewareDiagnosticsEvents)
                     {
@@ -250,14 +254,12 @@ namespace Microsoft.Identity.Web
         /// </summary>
         /// <param name="events">The <see cref="JwtBearerEvents"/> object to modify.</param>
         /// <param name="jwtBearerScheme">The JWT bearer scheme name to be used. By default it uses "Bearer".</param>
-        /// <param name="allowWebApiToBeAuthorizedByACL">If <see langword="true"/>, tokens are not required to have Scopes or Roles.</param>
-        internal static void ChainOnTokenValidatedEventForClaimsValidation(JwtBearerEvents events, string jwtBearerScheme, bool allowWebApiToBeAuthorizedByACL)
+        internal static void ChainOnTokenValidatedEventForClaimsValidation(JwtBearerEvents events, string jwtBearerScheme)
         {
             var tokenValidatedHandler = events.OnTokenValidated;
             events.OnTokenValidated = async context =>
             {
-                if (!allowWebApiToBeAuthorizedByACL
-                    && !context!.Principal!.Claims.Any(x => x.Type == ClaimConstants.Scope
+                if (!context!.Principal!.Claims.Any(x => x.Type == ClaimConstants.Scope
                         || x.Type == ClaimConstants.Scp
                         || x.Type == ClaimConstants.Roles
                         || x.Type == ClaimConstants.Role))
--- a/src/Microsoft.Identity.Web/WebApiExtensions/MicrosoftIdentityWebApiAuthenticationBuilderExtensions.cs
+++ b/src/Microsoft.Identity.Web/WebApiExtensions/MicrosoftIdentityWebApiAuthenticationBuilderExtensions.cs
@@ -264,8 +264,7 @@ namespace Microsoft.Identity.Web
                         || x.Type == ClaimConstants.Roles
                         || x.Type == ClaimConstants.Role))
                 {
-                    throw new UnauthorizedAccessException(string.Format(CultureInfo.InvariantCulture,
-                        IDWebErrorMessage.NeitherScopeOrRolesClaimFoundInToken, jwtBearerScheme));
+                    context.Fail(string.Format(CultureInfo.InvariantCulture, IDWebErrorMessage.NeitherScopeOrRolesClaimFoundInToken, jwtBearerScheme));
                 }
 
                 await tokenValidatedHandler(context).ConfigureAwait(false);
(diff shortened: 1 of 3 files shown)


MattKotsenas:
This _could_ be a one-line change to switch the `throw` statement to `context.Fail()`, but I wanted to provide test coverage of the change, which necessitated a small refactor. If you have thoughts or feelings about better ways to refactor to both make this change and provide test coverage, please let me know!

MattKotsenas:
Additionally, I searched back through the history to see if there was an intentional change to make this an exception, and I wasn't able to see one back to the "initial commit" 6a393a981604e02ac588ea7a58ddb3c13f303a6b.

From browsing, it appears that the exception was to be symmetric with APIs like `HttpContext.ValidateAppRole()` and `HttpContext.VerifyUserHasAnyAcceptedScope()`, however those APIs are "user mode" APIs for manual validation inside a daemon app or a controller body, where an exception flow is an appropriate way to ensure processing is interrupted. In this case the exception will always (as far as I see?) be from inside the auth middleware pipeline, where exceptions are discouraged and instead we should return failure to the pipeline.

If I missing any motivation for why the API is this way, please let me know. Thanks!

src/Microsoft.Identity.Web/WebApiExtensions/MicrosoftIdentityWebApiAuthenticationBuilderExtensions.cs
```diff
@@ -256,5 +246,29 @@ public static partial class MicrosoftIdentityWebApiAuthenticationBuilderExtensio
                     }
                 });
         }
+
+        /// <summary>
+        /// In order to ensure that the Web API only accepts tokens from tenants where it has been consented and provisioned, a token that is missing
+        /// both Roles and Scopes claims should be rejected. To enforce that rule, add an event handler to the beginning of the
```
jennyf19, line 252: >both [](http://example.com/codeflow?start=12&length=4)

has neither Roles or Scopes...

jennyf19 (approved):
:shipit:

jennyf19:
Thanks so much for the contribution @MattKotsenas Really appreciate it.

src/Microsoft.Identity.Web/WebApiExtensions/MicrosoftIdentityWebApiAuthenticationBuilderExtensions.cs
MattKotsenas: Fixed!

PR #1734, opened 2022-05-12
add cancellationToken param to DownstreamWebApi methods

h3rmanj:
Addresses #1739 

Add optional cancellationToken param to DownstreamWebApi methods and extension methods allowing consumers the possibility to cancel the operations.

For now I passed the cancellationToken down to methods supporting it. I'm not sure if it would be better to call `ThrowIfCancellationRequested` ourselves between `await` calls.

### DIFF
--- a/src/Microsoft.Identity.Web/DownstreamWebApiSupport/DownstreamWebApi.cs
+++ b/src/Microsoft.Identity.Web/DownstreamWebApiSupport/DownstreamWebApi.cs
@@ -59,6 +59,13 @@ namespace Microsoft.Identity.Web
                 throw new ArgumentException(IDWebErrorMessage.ScopesNotConfiguredInConfigurationOrViaDelegate);
             }
 
+            if (effectiveOptions.TokenAcquisitionOptions == null)
+            {
+                effectiveOptions.TokenAcquisitionOptions = new TokenAcquisitionOptions();
+            }
+
+            effectiveOptions.TokenAcquisitionOptions.CancellationToken = cancellationToken;
+
             MicrosoftIdentityOptions microsoftIdentityOptions = _microsoftIdentityOptionsMonitor
                 .Get(_tokenAcquisition.GetEffectiveAuthenticationScheme(authenticationScheme));
 
@@ -74,16 +81,6 @@ namespace Microsoft.Identity.Web
             else
             {
                 userflow = effectiveOptions.UserFlow;
-            }
-
-            if (cancellationToken != default)
-            {
-                if (effectiveOptions.TokenAcquisitionOptions == null)
-                {
-                    effectiveOptions.TokenAcquisitionOptions = new TokenAcquisitionOptions();
-                }
-
-                effectiveOptions.TokenAcquisitionOptions.CancellationToken = cancellationToken;
             }
 
             AuthenticationResult authResult = await _tokenAcquisition.GetAuthenticationResultForUserAsync(
@@ -173,19 +170,17 @@ namespace Microsoft.Identity.Web
                 throw new ArgumentException(IDWebErrorMessage.ScopesNotConfiguredInConfigurationOrViaDelegate);
             }
 
+            if (effectiveOptions.TokenAcquisitionOptions == null)
+            {
+                effectiveOptions.TokenAcquisitionOptions = new TokenAcquisitionOptions();
+            }
+
+            effectiveOptions.TokenAcquisitionOptions.CancellationToken = cancellationToken;
+
             string apiUrl = effectiveOptions.GetApiUrl();
 
             CreateProofOfPossessionConfiguration(effectiveOptions, apiUrl);
 
-            if (cancellationToken != default)
-            {
-                if (effectiveOptions.TokenAcquisitionOptions == null)
-                {
-                    effectiveOptions.TokenAcquisitionOptions = new TokenAcquisitionOptions();
-                }
-
-                effectiveOptions.TokenAcquisitionOptions.CancellationToken = cancellationToken;
-            }
 
             AuthenticationResult authResult = await _tokenAcquisition.GetAuthenticationResultForAppAsync(
                 effectiveOptions.Scopes,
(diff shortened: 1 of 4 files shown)


src/Microsoft.Identity.Web/DownstreamWebApiSupport/DownstreamWebApi.cs
```diff
@@ -106,28 +118,37 @@ public class DownstreamWebApi : IDownstreamWebApi
...
                 response.EnsureSuccessStatusCode();
             }
             catch
             {
+#if DOTNET_50_AND_ABOVE
+                string error = await response.Content.ReadAsStringAsync(cancellationToken).ConfigureAwait(false);
+#else
                 string error = await response.Content.ReadAsStringAsync().ConfigureAwait(false);
-
+#endif
```
h3rmanj, line 143: Not quite sure about this `#if` stuff. I've never used it before but I think it's right? `DOTNET_50_AND_ABOVE` is defined in the `Microsoft.Identity.Web.csproj`.
The `ReadAsStringAsync` method didn't take a `cancellationToken`, so I assumed it was because it targeted multiple frameworks. The [5.0 branch of `System.Net.Http`](https://github.com/dotnet/runtime/blob/release/5.0/src/libraries/System.Net.Http/src/System/Net/Http/HttpContent.cs#L155) includes a `HttpContent.ReadAsStringAsync` that takes a `cancellationToken`.

src/Microsoft.Identity.Web/DownstreamWebApiSupport/IDownstreamWebApi.cs
```diff
@@ -28,19 +29,22 @@ public interface IDownstreamWebApi
...
         /// <see cref="HttpMethod.Patch"/>, <see cref="HttpMethod.Post"/>, <see cref="HttpMethod.Put"/>.</param>
+        /// <param name="cancellationToken">The cancellation token to cancel the operation.</param>
         /// <returns>An <see cref="HttpResponseMessage"/> that the application will process.</returns>
         public Task<HttpResponseMessage> CallWebApiForUserAsync(
             string serviceName,
             Action<DownstreamWebApiOptions>? calledDownstreamWebApiOptionsOverride = null,
             ClaimsPrincipal? user = null,
-            StringContent? content = null)
+            StringContent? content = null,
+            CancellationToken cancellationToken = default)
```
bgavrilMS, line 39: Is this not a breaking change? If I have an implementation for IDownstreamApi (e.g.for test), it will be broken by this change?

src/Microsoft.Identity.Web/DownstreamWebApiSupport/DownstreamWebApi.cs
```diff
@@ -74,6 +76,16 @@ public class DownstreamWebApi : IDownstreamWebApi
                 userflow = effectiveOptions.UserFlow;
             }
 
+            if (cancellationToken != default)
```
bgavrilMS, line 79: I don't think this check is needed. Passing along a default cancellation token is ok.

src/Microsoft.Identity.Web/DownstreamWebApiSupport/DownstreamWebApi.cs
bgavrilMS: Yeah, this seems to be the way forward, the API was only added in .NET 5 https://stackoverflow.com/questions/25219287/net-httpclient-cancel-readasstringasync

`DOTNET_50_AND_ABOVE` is defined implicity by .NET

src/Microsoft.Identity.Web/DownstreamWebApiSupport/IDownstreamWebApi.cs
h3rmanj: Ah, you're right. I only thought of consumers using the interface through the `AddDownstreamWebApi`, not of implementors. I guess it will have to wait for the next major?

bgavrilMS: I'll let the package owners (@jmprieur  and @jennyf19 ) answer that, but ideally we should not make interface changes. Alternatives are: 

- extension methods
- default interface methods https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/default-interface-methods

h3rmanj: I tried having a look at this;
**Extension methods**
I don't think this would work, as this needs to happen through the implementation to be of use.

**`ICancellableDownstreamWebApi : IDownstreamWebApi`**
Could technically work, but it would have to be registered twice with DI, and would be too clunky as a consumer imo.

**Default Interface methods**
I thought this would work. I tried having default interface methods for the cancellationToken versions of the methods point to the ones without, and do it in reverse in the implementation. That way it would be opt-in for implementors to accept cancellationTokens, and the default implementation would accept it.
The problem was that having these two definitions would make the many existing calls ambiguous
```csharp
// Simplified interface
Task CallForUserAsync(string url, string? scheme = null);
Task CallForUserAsync(string url, string? scheme = null, CancellationToken cancellationToken = default);

// This call would be ambigous
api.CallForUserAsync("/endpoint");
```
I could probably fix it by having the CancellationToken not be an optional param, but then I would have to order it before the optional params. I don't know if it's an actual rule, but everywhere I've seen a cancellationToken, it's appeared last, so I assume it's preferred to have it as a last param.
Either way, there would be some code complexity added, with default interface methods calling each other, and the implementation the other way.

**TL:DR**
To not break for `IDownstreamWebApi` implementors, there would either have to be some code complexity added to this project, or it would be clunky for consumers to use (having two interfaces to choose from). IMO it's not worth the effort needed, to not break for implementors, even though it would be desired.

I'll leave it up to the project owners how they'll want to do it, if they want this feature.

jmprieur:
@jennyf19. Let's take this in IDownstreamRestApi
(@h3rmanj : this will be a new version of IDownstreamWebApi with a better API than that one which as grown organically)

MauRiEEZZZ:
@jmprieur 
Is there any information about the timeline we can expect IDownstreamRestApi you mentioned?

h3rmanj:
Implemented in 2.5.0

PR #1750, opened 2022-06-07
fix: include StatusCode in HttpRequestException for .NET 5

h3rmanj:
A `StatusCode` property was introduced on `HttpRequestException` in .NET 5.  
Populate it when throwing our own `HttpRequestException` in `DownstreamWebApi` when targeting supported frameworks.


bgavrilMS (approved):

jennyf19 (approved):

jennyf19:
Thanks @h3rmanj

PR #1760, opened 2022-06-15
Make LoggingOptions internal

jennyf19:
I was going to run a release build and noticed the LoggingOptions is public, I don't think it needs to be, so marking as internal.


jmprieur (approved):
LGTM
Thanks Jenny

PR #1769, opened 2022-06-18
Jennyf/issue template

jennyf19:
[JS](https://github.com/AzureAD/microsoft-authentication-library-for-js/issues/new?assignees=&labels=bug-unconfirmed%2Cquestion&template=bug_report.yml) team is using yml instead of markdown for their bug report and it looks pretty nice. What do you think, @jmprieur ?


jmprieur (approved):
LGTM
Thanks @jennyf19 

Super cool!

PR #1782, opened 2022-06-24
Jennyf/newtonsoft

jennyf19:


jmprieur (approved):
LGTM
Thanks very much for addressing the consequences of the newtonsoft issues so quickly, @jennyf19.

PR #1810, opened 2022-07-18
Baseline to have ID.Web.MicrosoftGraph reference Id.Web.TokenAcqusition (and not Id.Web)

jmprieur:
- In AuthenticationOptions and MicrosoftAuthenticationOptions, move properties were they should have been in the first place (when AAD specific)
- Move LoggingOptions to Ms.Id.Web.TokenAcquisition
- MicrosoftIdentityBaseAuthenticationBuilder and MicrosoftIdentityAppCallingWebApiAuthenticationBuilder move to Ms.Id.Web.TokenAcquisition
- Session methods of MicrosoftIdentityAppCallingWebApiAuthenticationBuilder move to an extension method
- TokenAcquisitionAspNetCore inherits from TokenAcquisition instead of being a partial class implementation (Could move to another assembly if we choose to)
- AddTokenAcquisition now register ITokenAcquirer (And only this one in the case of framework others than .NET Core)


src/Microsoft.Identity.Web/Microsoft.Identity.Web.csproj
```diff
@@ -25,8 +25,8 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.IdentityModel.LoggingExtensions" Version="$(IdentityModelVersion)" />
-    <PackageReference Include="Microsoft.IdentityModel.Logging" Version="$(IdentityModelVersion)" />
+    <!--<PackageReference Include="Microsoft.IdentityModel.LoggingExtensions" Version="$(IdentityModelVersion)" />
```
jennyf19, line 28: Should these be removed or uncommented?

src/Microsoft.Identity.Web/WebAppExtensions/MicrosoftIdentityAppCallingWebApiAuthenticationBuilderExt.cs
```diff
@@ -0,0 +1,65 @@
...
+using Microsoft.Identity.Web.TokenCacheProviders.InMemory;
+using Microsoft.Identity.Web.TokenCacheProviders.Session;
+
+namespace Microsoft.Identity.Web
+{
+    /// <summary>
+    /// Authentication builder returned by the EnableTokenAcquisitionToCallDownstreamApi methods
+    /// enabling you to decide token cache implementations.
+    /// </summary>
+    public static class MicrosoftIdentityAppCallsWebApiAuthenticationBuilderExtension
```
jennyf19, line 24: >MicrosoftIdentityAppCallsWebApiAuthenticationBuilderExtension [](http://example.com/codeflow?start=24&length=61)

before shipping we might need a better name

src/Microsoft.Identity.Web.TokenAcquisition/Microsoft.Identity.Web.TokenAcquisition.csproj
```diff
@@ -17,10 +17,12 @@
     <PackageReference Include="Microsoft.AspNetCore.Authentication.JwtBearer" Version="$(MicrosoftAspNetCoreAuthenticationJwtBearerVersion)" />
     <PackageReference Include="Microsoft.AspNetCore.Authentication.OpenIdConnect" Version="$(MicrosoftAspNetCoreAuthenticationOpenIdConnectVersion)" />
     <PackageReference Include="Microsoft.IdentityModel.Protocols.OpenIdConnect" Version="$(IdentityModelVersion)" />
+    <PackageReference Include="Microsoft.Extensions.Http" Version="3.1.3" />
```
jennyf19, line 20: This should move to Directory.Build.props

src/Microsoft.Identity.Web.TokenAcquisition/ServiceCollectionExtensions.cs
```diff
@@ -62,26 +62,31 @@ public static class ServiceCollectionExtensions
...
                 services.AddHttpContextAccessor();
+
+                services.AddScoped<ITokenAcquisition, TokenAcquisitionAspNetCore>();
+                services.AddScoped(s => (ITokenAcquirer)s.GetRequiredService<ITokenAcquisition>());
+
                 services.AddScoped<ITokenAcquisitionHost, TokenAcquisitionAspnetCoreHost>();
                 services.AddScoped(s => (ITokenAcquisitionInternal)s.GetRequiredService<ITokenAcquisition>());
 #else
                 services.AddScoped<ITokenAcquisitionHost, PlainDotNetTokenAcquisitionHost>();
+                services.AddSingleton<ITokenAcquirer, TokenAcquisition>();
```
jennyf19, line 89: >AddSingleton [](http://example.com/codeflow?start=25&length=12)

AddScoped?

src/Microsoft.Identity.Web.TokenAcquisition.Abstractions/ApplicationOptions/MicrosoftAuthenticationOptions.cs
```diff
@@ -76,7 +78,9 @@ internal bool IsB2C
         {
             get => !string.IsNullOrWhiteSpace(DefaultUserFlow);
         }
+        #endregion
```
jennyf19, line 81: I don't think we need the regions necessarily, but if so, please add the region name at the end of the region : ex `#endregion B2C`

jennyf19 (approved):
:shipit:

src/Microsoft.Identity.Web/Microsoft.Identity.Web.csproj
jmprieur: thanks. Done!

PR #1814, opened 2022-07-19
Create a mix of ScopesRequiredHttpContextExtensions and RolesRequired…

BluMichele:
…HttpContextExtensions

### DIFF
--- a/src/Microsoft.Identity.Web/Resource/RolesOrScopesRequiredHttpContextExtensions.cs
+++ b/src/Microsoft.Identity.Web/Resource/RolesOrScopesRequiredHttpContextExtensions.cs
@@ -1,0 +1,94 @@
+
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using System.Security.Claims;
+using global::Microsoft.AspNetCore.Http;
+using global::Microsoft.Identity.Web;
+using Bludata.Blu.Web.Security.Entities;
+using Bludata.Blu.Web.Security.Interface;
+
+namespace Bludata.Blu.Web.Security.Utils;
+
+/// <summary>
+/// Extension class providing the extension methods for <see cref="HttpContent"/> that
+/// can be used in web APIs to validate scopes and roles in controller actions.
+/// </summary>
+public static class RolesOrScopesRequiredHttpContextExtensions
+{
+    /// <summary>
+    /// When applied to an <see cref="HttpContext"/>, verifies that the user authenticated in the
+    /// web API has any of the accepted scopes or roles.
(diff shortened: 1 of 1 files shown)


src/Microsoft.Identity.Web/Resource/RolesOrScopesRequiredHttpContextExtensions.cs
```diff
@@ -0,0 +1,94 @@
...
+using System.Linq;
+using System.Net;
+using System.Net.Http;
+using System.Security.Claims;
+using global::Microsoft.AspNetCore.Http;
+using global::Microsoft.Identity.Web;
+using Bludata.Blu.Web.Security.Entities;
+using Bludata.Blu.Web.Security.Interface;
+
+namespace Bludata.Blu.Web.Security.Utils;
```
jennyf19, line 14: namespace needs to be updated and using removed.
please add license info like we have on other classes

jennyf19:
thanks for this @BluMichele. We have something similar on the attributes and will need to test once you have updated the file. It will take us sometime to circle back on this, just an FYI.

src/Microsoft.Identity.Web/Resource/RolesOrScopesRequiredHttpContextExtensions.cs
BluMichele: I did it, thanks for letting me know. Hope it will be useful.

src/Microsoft.Identity.Web/Resource/RolesOrScopesRequiredHttpContextExtensions.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        var message = string.Format(CultureInfo.InvariantCulture,
+                        $"The 'scope' or 'scp' claim does not contain scopes '{0}' nor " +
+                        $"the 'roles' or 'role' claim does not contain roles '{1}'",
+                        new string[] { string.Join(",", acceptedScopes), string.Join(",", acceptedRoles) });
+
+        lock (context)
+        {
+            context.Response.StatusCode = (int)HttpStatusCode.Forbidden;
+            context.Response.WriteAsync(message);
```
jmprieur, line 85: We should rather set that in an ErrorDescription header or www-Authenticate header, maybe (and not WriteAsync / CompleteAsync())

BluMichele: You might be right, I quite much copied the code style of https://github.com/AzureAD/microsoft-identity-web/blob/master/src/Microsoft.Identity.Web/Resource/ScopesRequiredHttpContextExtensions.cs

Issue #1835, opened 2022-08-10
Pod Identity signed assertions are supported  by Ms.Id.Web.TokenAcquisition


jmprieur:
## Why
When running inside a container, confidential client applications can use Pod Identity to get a signed assertion in lieu of client credentials. In practice, this is a file that is shared with the container, and that is read periodically.

## What

By configuration, define the file in which Pod identity will save the signed assertion, so that the application running in the container can access it.

```JSon
"AzureAd" : {
   "ClientCredentials": [
      {
       "SourceType" : "SignedAssertionFilePath",
       "SignedAssertionFileDiskPath " : "/somePathOnDisk"
      }
    ]
}
```

# How

- [x] Expose `PodIdentityClientAssertion` in certificateless, in addition to `ManagedIdentityClientAssertion`
     - [x] ClientAssertionDescription should become public, and probably renamed BaseClientAssertion?
          - [x] its constructor is no longer needed. The ClientAssertionProvider property should be internal.
          - [x] In GetSignedAssertion, check that ClientAssertionProvider is not null. If it's null, throw an exception that "ClientAssertionProvider must be initialized in the constructor of the derived classes"
     - [x] Have ManagedIdentityClientAssertion inherit from BaseClientAssertion.. Its default constructor should now set ClientAssertionProvider to GetSignedAssertionFromFederatedTokenProvider. No need for ClientAssertionProvider, _clientAssertion, GetSignedAssertion and Expiry any longer (they are already in the base class)=> delete the last four members of the ManagedIdentityClientAssertion class
     - [x] Add a new class `PodIdentityClientAssertion` inheriting from BaseClientAssertion, and inspired by `ManagedIdentityClientAssertion ` but which gets the signed assertion from a file (and the file can be referenced in the `AZURE_FEDERATED_TOKEN_FILE` environment variable. Something like the following (no guarantee it builds)
 
          ```c#
               internal class PodIdentityClientAssertion : BaseAssertionDescription
               {
                  /// <summary>
                  /// Gets a signed assertion from PodIdentity. The file is provided
                  /// by an environment variable ("AZURE_FEDERATED_TOKEN_FILE")
                  /// See https://aka.ms/ms-id-web/certificateless.
                  /// </summary>
                  public PodIdentityClientAssertion()
                  {
                      _filePath = Environment.GetEnvironmentVariable("AZURE_FEDERATED_TOKEN_FILE");
[...]

PR #1837, opened 2022-08-12
Adding support for PodIdentity credentials

jmprieur:
Contributes to fixing https://github.com/AzureAD/microsoft-identity-web/issues/1835 (part I: read and cache the assertion)

- Did the factorization proposed in the issue
- Updated TokenAcquirerTest as it was using the implicit usings, but this did not work across source control (Visual Studio tooling issue)

image

Remains to do:
- In TokenAcquisition, do a similar processing as for ManagedIdentityClientAssertion, but for PodIdentityClientAssertion.

### DIFF
--- a/TokenAcquirerTests/TokenAcquirer.cs
+++ b/TokenAcquirerTests/TokenAcquirer.cs
@@ -1,10 +1,12 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System.Threading.Tasks;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Graph;
 using Microsoft.Identity.Web;
 using Microsoft.Identity.Web.TokenCacheProviders.InMemory;
+using Xunit;
 
 namespace TokenAcquirerTests
 {
--- a/TokenAcquirerTests/TokenAcquirerTests.csproj
+++ b/TokenAcquirerTests/TokenAcquirerTests.csproj
@@ -2,7 +2,6 @@
 
   <PropertyGroup>
     <TargetFrameworks>netcoreapp3.1; net6.0</TargetFrameworks>
-    <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
 
     <IsPackable>false</IsPackable>
--- a/src/Microsoft.Identity.Web.Certificateless/Certificateless.cd
+++ b/src/Microsoft.Identity.Web.Certificateless/Certificateless.cd
@@ -14,11 +14,11 @@
       <FileName>ClientAssertion.cs</FileName>
     </TypeIdentifier>
   </Class>
-  <Class Name="Microsoft.Identity.Web.ClientAssertionDescription">
+  <Class Name="Microsoft.Identity.Web.ClientAssertionProviderBase">
     <Position X="7.25" Y="3.25" Width="3.5" />
     <TypeIdentifier>
       <HashCode>AAAAAAAAABAAAAAAAAAQAAAAAAAAIBAAAAAAAAAAAAA=</HashCode>
-      <FileName>ClientAssertionDescription.cs</FileName>
+      <FileName>ClientAssertionProviderBase.cs</FileName>
     </TypeIdentifier>
     <ShowAsAssociation>
       <Field Name="_clientAssertion" />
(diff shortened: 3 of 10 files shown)


TokenAcquirerTests/TokenAcquirer.cs
```diff
@@ -1,10 +1,12 @@
 // Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System.Threading.Tasks;
```
jmprieur, line 4: This was using the Implicit usings (which involve some tooling magic, which did not work in my VS when I got the rel/v2 branch from the source control)

TokenAcquirerTests/TokenAcquirerTests.csproj
```diff
@@ -2,7 +2,6 @@
 
   <PropertyGroup>
     <TargetFrameworks>netcoreapp3.1; net6.0</TargetFrameworks>
-    <ImplicitUsings>enable</ImplicitUsings>
```
jmprieur, line 5: Disabled the magic

src/Microsoft.Identity.Web.Certificateless/Certificateless.cd
```diff
@@ -0,0 +1,50 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
```
jmprieur, line 1: It's the class diagram

image

src/Microsoft.Identity.Web.Certificateless/ClientAssertionDescription.cs
```diff
@@ -11,22 +11,12 @@ namespace Microsoft.Identity.Web
     /// Description of a client assertion in the application configuration.
     /// See https://aka.ms/ms-id-web/client-assertions.
     /// </summary>
-    internal class ClientAssertionDescription
+    public class ClientAssertionDescription
```
jmprieur, line 14: I have not renamed it. Maybe we want to rename it to BaseClientAssertion?

src/Microsoft.Identity.Web.Certificateless/ClientAssertionDescription.cs
```diff
@@ -11,22 +11,12 @@ namespace Microsoft.Identity.Web
...
     {
         /// <summary>
-        /// Constructor of a ClientAssertionDescription.
+        /// delegate to get the client assertion from a provider.
         /// </summary>
-        /// <param name="clientAssertionProvider">delegate providing the client assertion
-        /// when it is necessary.</param>
-        public ClientAssertionDescription(Func<CancellationToken, Task<ClientAssertion>> clientAssertionProvider)
```
jmprieur, line 21: Could not initialize the delegate.

src/Microsoft.Identity.Web.Certificateless/ManagedIdentityClientAssertion.cs
```diff
@@ -47,41 +48,5 @@ private async Task<ClientAssertion> GetSignedAssertionFromFederatedTokenProvider
                 cancellationToken).ConfigureAwait(false);
             return new ClientAssertion(result.Token, result.ExpiresOn);
         }
-
```
jmprieur, line 50: Moved these to the base class (as they are also used for the PodIdentityClientAssertion)

src/Microsoft.Identity.Web.Certificateless/PodIdentityClientAssertion.cs
```diff
@@ -0,0 +1,55 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
jmprieur, line 1: The only new file

tests/Microsoft.Identity.Web.Test/ClientAssertionTests.cs
```diff
@@ -15,14 +15,16 @@ public class ClientAssertionTests
         public async Task TestClientAssertion()
         {
             int n = 0;
-            ClientAssertionDescription clientAssertionDescription = new ClientAssertionDescription(
-                cancellationToken =>
+            ClientAssertionDescription clientAssertionDescription = new ClientAssertionDescription()
+            {
+                ClientAssertionProvider = (cancellationToken =>
```
jmprieur, line 20: The ClientAssertionDescription class was internal and used to take the delegate in the constructor, but the delegate cannot be initialized form the derived classes, so I changed it to be a protected member, and updated the test

src/Microsoft.Identity.Web.Certificateless/ClientAssertionDescription.cs
jennyf19: makes sense

jennyf19 (approved):
:shipit:

jennyf19:
thanks for picking this up. :ship-it:

shueybubbles:
@jmprieur Am I reading this correctly?
1. Existing apps using Managed Identity auth that want to work in a pod scenario will need to switch to the new PodIdentity. If so, is there a particular reason why managed identity is not aware of pod identity? It seems like a fairly big burden for apps to have to be aware of this new model which is pretty much a configuration change as opposed to a fundamental auth change.
2. The "Default" credential model will NOT include Pod identity in its lookup. Where would that change have to be made?

PR #1854, opened 2022-08-25
update to wilson 6.22.0

jennyf19:


TimHannMSFT (approved):
Seems like the title should be "update to wilson 6.22.1" but otherwise looks good

PR #1855, opened 2022-08-26
Routing log always messages from MSAL to critical in logger extensions

trwalke:
I noticed that the log messages for log.always are not included in the switch statement in the code. Since the log always information is very important for diagnosing some issues, I added it.

### DIFF
--- a/src/Microsoft.Identity.Web/TokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web/TokenAcquisition.cs
@@ -1093,6 +1093,9 @@ namespace Microsoft.Identity.Web
         {
             switch (level)
             {
+                case Client.LogLevel.Always:
+                    _logger.LogCritical(message);
+                    break;
                 case Client.LogLevel.Error:
                     _logger.LogError(message);
                     break;


src/Microsoft.Identity.Web/TokenAcquisition.cs
```diff
@@ -1093,6 +1093,9 @@ private static bool AcceptedTokenVersionMismatch(MsalUiRequiredException msalSer
         {
             switch (level)
             {
+                case Client.LogLevel.Always:
```
trwalke, line 1096: Since the logging extensions have no concept of "Always", I set it to log at the lowest logging level which is critical.

jmprieur (approved):
LGTM
Thanks @trwalke

jennyf19:
Thanks @trwalke . we are working on a v2 right now, so I might add this there instead of into master. I had asked Bogdan to wait on this until we were done. Sorry for that.

jennyf19:
https://github.com/AzureAD/microsoft-identity-web/pull/1856

PR #1865, opened 2022-09-05
Jennyf/fix texts

jennyf19:


jennyf19:
wrong branch but i already merged to rel/v2

PR #1866, opened 2022-09-05
fix tests

jennyf19:


src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquirerFactory.cs
```diff
@@ -114,10 +114,11 @@ private IConfiguration ReadConfiguration()
         /// <returns>Returns the base path for configuration files</returns>
         protected virtual string DefineConfiguration(IConfigurationBuilder builder)
         {
-            return Path.GetDirectoryName(Assembly.GetEntryAssembly()!.Location)!;
+            Assembly assembly = Assembly.GetEntryAssembly() ?? Assembly.GetExecutingAssembly();
```
jennyf19, line 117: common issue when running tests

PR #1868, opened 2022-09-06
Moving the performance testing to another tenant

jmprieur:
contosolob.onmicrosoft.com as the previous one
is no longer accessible to us

Fixing the Powershell script that probably never worked


jennyf19 (approved):
:shipit: Thanks for updating this!

PR #1877, opened 2022-09-11
Provides a default implementation for IAuthorizationHeaderProvider

jmprieur:
Provides a default implementation for IAuthorizationHeaderProvider and a service to implement it.

Fixes IAuthorizationHeaderProvider to have both verbs (for User and for Apps) Update the daemon sample


jennyf19 (approved):
:shipit:

PR #1888, opened 2022-09-18
Fix the MISE scenario 

jmprieur:
We overlooked this mis-usage of the MergedOptions after adding the PostConfiguration handlers.


jennyf19 (approved):
thank you!

jennyf19 (approved):

PR #1904, opened 2022-10-08
Fix tests, deprecate net5.0 and add net6.0 everywhere

jennyf19:


jmprieur (approved):
LGTM
Thanks @jennyf19

PR #1913, opened 2022-10-13
Added support for certificates with private keys.

brentschmaltz:
This PR allows for loading X509Certificates with private keys by passing the password in CertificateDescription.CertificatePassword.


jmprieur (approved):
LGTM
Thanks @brentschmaltz

what will happen for credscans?

brentschmaltz:
@jmprieur  we have two options.
We can add a credscan exclusion file OR add a comment to the commit ****Credscan ByPass**** (or something similar).

TimHannMSFT (approved):
Thanks @brentschmaltz, makes sense to me. only concern would be this allows folks to be less safe in how they're handling secrets (though I suppose nothing stops them from putting an encoded cert in their config and have it simply not work). As you rightly pointed out offline though, this is no different than how symmetric keys are handled, so I think this is the right call.

With this fix at least, there's not an esoteric error message.

jennyf19 (approved):
:shipit: 

you will probably need to add the file to this exclusion file for cred scan: https://github.com/AzureAD/microsoft-identity-web/blob/master/build/credscan-exclusion.json

jmprieur:
Abandoning, as it's merged in rel/v2

PR #1919, opened 2022-10-18
update to 1.0.0-preview abstractions

jennyf19:


jmprieur:
@jennyf19 : there is no code to change? (due to the renaming)

PR #1923, opened 2022-10-19
Support for private keys in certificates.

brentschmaltz:
Support loading certificates with private keys from Base64UrlEncoded and pfx representations.
Fix bug where loading from path was using wrong value.


jennyf19 (approved):
Thanks @brentschmaltz

PR #1924, opened 2022-10-21
Suggestions from R9 analyzers - round 1

jennyf19:


src/Microsoft.Identity.Web.TokenAcquisition/AspNetCore/TokenAcquisition-AspnetCore.cs
```diff
@@ -142,17 +140,9 @@ internal class TokenAcquisitionAspNetCore : TokenAcquisition, ITokenAcquisition,
         public async Task AddAccountToCacheFromAuthorizationCodeAsync(
             AuthorizationCodeReceivedContext context,
             IEnumerable<string> scopes,
-            string authenticationScheme /*= OpenIdConnectDefaults.AuthenticationScheme*/)
+            string authenticationScheme = OpenIdConnectDefaults.AuthenticationScheme /*= OpenIdConnectDefaults.AuthenticationScheme*/)
```
jmprieur, line 143: I'm not sure about this one. Let's just take a note shouldn't prevent you from merging

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquirerFactory-GetTokenAcquirers.cs
```diff
@@ -112,6 +107,14 @@ public ITokenAcquirer GetTokenAcquirer(string authenticationScheme = "")
             }
             return acquirer;
         }
+        private void CheckServiceProviderNotNull()
```
jmprieur, line 110: I wonder about the performance impact. Is it an r9 rule?

jennyf19: They do something else for exceptions around parameter checks, which I will do next.

src/Microsoft.Identity.Web.TokenAcquisition/AspNetCore/TokenAcquisition-AspnetCore.cs
jennyf19: The interface defines the default.

PR #1926, opened 2022-10-21
revert the removal of the null param for authentication scheme and ad…

jennyf19:
…d the tests to rel/v2

removing broke some tests, so adding it back.

### DIFF
--- a/.github/workflows/dotnetcore.yml
+++ b/.github/workflows/dotnetcore.yml
@@ -5,7 +5,7 @@ on:
     branches:
       - "master"
       - "rel/v2"
-  pull-request:
+  pull_request:
     branches:
       - "master"
       - "rel/v2"
--- a/.github/workflows/dotnetcore.yml
+++ b/.github/workflows/dotnetcore.yml
@@ -6,8 +6,12 @@ on:
       - "master"
       - "rel/v2"
   pull-request:
+    branches:
       - "master"
       - "rel/v2"
+    paths:
+      - ".github/workflows/dotnetcore.yml"
+
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


.github/workflows/dotnetcore.yml
```diff
@@ -1,6 +1,16 @@
 name: .NET Core
 
-on: [push]
+on:
+  push:
+    branches:
+      - "master"
+      - "rel/v2"
```
jmprieur, line 7: ❤️

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -641,7 +647,7 @@ private IConfidentialClientApplication BuildConfidentialClientApplication(Merged
                     else
                     {
                         string? sessionKey = tokenAcquisitionOptions!.LongRunningWebApiSessionKey;
-                        if (sessionKey == TokenAcquisitionOptions.LongRunningWebApiSessionKeyAuto)
+                        if (sessionKey == Abstractions.AcquireTokenOptions.LongRunningWebApiSessionKeyAuto)
```
jmprieur, line 650: Is it something we missed?

jmprieur (approved):
LGTM
Thanks @jennyf19

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
jennyf19: i don't think so, it was a suggestion from VS

PR #1927, opened 2022-10-21
using high-perf logging for the newly added log messages

jennyf19:
and clean up a few existing log formats


src/Microsoft.Identity.Web.TokenCache/Distributed/MsalDistributedTokenCacheAdapter.Logger.cs
```diff
@@ -78,20 +78,18 @@ private static class Logger
...
-                int cacheSize,
-                Exception? ex) => s_l1CacheRead(
+                int cacheSize) => s_l1CacheRead(
                     logger,
                     cacheType,
                     operation,
                     cacheKey,
                     cacheSize,
-                    ex);
+                    default!);
```
jmprieur, line 92: what is it?

jmprieur (approved):
LGTM
Thanks @jennyf19

src/Microsoft.Identity.Web.TokenCache/Distributed/MsalDistributedTokenCacheAdapter.Logger.cs
jennyf19: It's a default replacement for the exception when an exception is not used in the log message. https://learn.microsoft.com/en-us/dotnet/core/extensions/high-performance-logging

PR #1899, opened 2022-10-03
Bugfix #749  Claim Actions Missing & Not Running

warrenbuckley:
Hello :wave: @jennyf19 
I believe this PR will fix the following issue #749 where ClaimActions are not run if you alter these when setting up Microsoft.Identity.Web in configuration/options.

I have unable to test this easily with my own code project, but with ClaimActions totally missing in the MergedOptions class I would believe this is the problem that myself and others are having.

Look forward to your feedback and input

Thanks,
Warren :)

Issue #749, opened 2020-11-04
[Bug] ClaimActions not being invoked


tonytilbo:
**Which version of Microsoft Identity Web are you using?**
E.g. Microsoft Identity Web 1.2.0

**Where is the issue?**
* Web app
    * [x] Sign-in users
    * [ ] Sign-in users and call web APIs
* Web API
    * [ ] Protected web APIs (validating tokens)
    * [ ] Protected web APIs (validating scopes)
    * [ ] Protected web APIs call downstream web APIs
* Token cache serialization
     * [ ] In-memory caches
     * [ ] Session caches
     * [ ] Distributed caches
* Other (please describe)
This is a new app or an experiment.

**Repro**
We are attempting to use a custom ClaimAction that will do some additional mapping of JWT claims to custom claims on the identity.

```csharp
public class CustomClaimAction : ClaimAction
{
	public CustomClaimAction(string claimType, string valueType): base (claimType, valueType) { }
        public override void Run(JsonElement userData, ClaimsIdentity identity, string issuer) 
                                  => throw new System.NotImplementedException();
}

services.AddAuthentication(JwtBearerDefaults.AuthenticationScheme)
  .AddMicrosoftIdentityWebApi(jwt => { }, options =>
  {
      options.TenantId = Configuration["AzureAd:TenantId"];
      options.ClientId = Configuration["AzureAd:ClientId"];
      options.Domain = Configuration["AzureAd:Domain"];
      options.Instance = Configuration["AzureAd:Instance"];

       options.ClaimActions.Add(new CustomClaimAction (ClaimTypes.Role, ClaimValueTypes.String));
 });
```

**Expected behavior**
I expect the Run method of the CustomClaimAction to be invoked.

**Actual behavior**
Method is not called.

jmprieur:
@tonytilbo thanks for the heads-up.

This property is normally an OpenIdConnect property (for a web app, and acting on the IDToken claims), whereas here you have a web API, but indeed, given it's surfaced in MicrosoftIdentityOptions, we should probably make it work for the Access token claims as well.

Out of curiosity, what is your scenario? what are you trying to achieve?

tonytilbo:
We are looking to check for "groups" claims in the incoming access token and maps these based on some external configuration to a set of roles which would be added to the ClaimsIdentity so we can manage access to the API used role based authorization.

jmprieur:
@tonytilbo. Thanks for the update.
[...]

PR #1930, opened 2022-10-24
Bugfix #749  Claim Actions Missing & Not Running (#1899)

jennyf19:
* Add in missing ClaimActions into MergedOptions

* Add unit test to verify that the merged option for ClaimAction will appear

@jmprieur - I did some local testing


jennyf19:
already approved.

PR #1940, opened 2022-10-26
Add signing Ms.Id.Web.Diagnostics in the release build.

jmprieur:


jennyf19 (approved):
good catch. completely missed this one

Issue #1957, opened 2022-11-08
UpdateMergedOptionsFromMicrosoftIdentityOptions Collection was modified; enumeration operation may not execute.


RbsTechOnline:
### Microsoft.Identity.Web Library

Microsoft.Identity.Web

### Microsoft.Identity.Web version

1.25.5

### Web app

Sign-in users and call web APIs

### Web API

Protected web APIs (validating tokens)

### Token cache serialization

Not Applicable

### Description

Fist call to web api is causing an error an 

`Collection was modified; enumeration operation may not execute`

in 

`internal static void UpdateMergedOptionsFromMicrosoftIdentityOptions(MicrosoftIdentityOptions microsoftIdentityOptions, MergedOptions mergedOptions)`

Seemingly at

`if (!mergedOptions.ClaimActions.Contains(claimAction))`

### Reproduction steps

This is just a preliminary assessment. Can work towards a full repro if needed

### Error message

```
System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
   at System.Collections.Generic.List`1.Enumerator.MoveNextRare()
   at System.Linq.Enumerable.Contains[TSource](IEnumerable`1 source, TSource value, IEqualityComparer`1 comparer)
   at Microsoft.Identity.Web.MergedOptions.UpdateMergedOptionsFromMicrosoftIdentityOptions(MicrosoftIdentityOptions microsoftIdentityOptions, MergedOptions mergedOptions)
   at Microsoft.Identity.Web.MicrosoftIdentityWebApiAuthenticationBuilderExtensions.<>c__DisplayClass3_0.<AddMicrosoftIdentityWebApiImplementation>b__0(JwtBearerOptions options, IServiceProvider serviceProvider, IOptionsMonitor`1 mergedOptionsMonitor, IOptionsMonitor`1 msIdOptionsMonitor, IOptions`1 msIdOptions)
   at Microsoft.Extensions.Options.ConfigureNamedOptions`5.Configure(String name, TOptions options)
   at Microsoft.Extensions.Options.OptionsFactory`1.Create(String name)
   at Microsoft.Extensions.Options.OptionsCache`1.<>c__3`1.<GetOrAdd>b__3_0(String name, ValueTuple`2 arg)
   at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd[TArg](TKey key, Func`3 valueFactory, TArg factoryArgument)
[...]

PR #1960, opened 2022-11-15
fix for claimAction and Scope with .NET 7

jennyf19:
fix for #1957 tested w/ Blazor server .NET 7 using the hot reload shared by a customer w/several browser tabs opened.


jmprieur (approved):
LGTM
Thanks @jennyf19

PR #2019, opened 2022-12-29
Jmprieur/update abstractions 1 0 6 preview

### LINKED ISSUE #1445: [Feature Request] Adding a method on downstreamWebApi to call without an access token
I have an API, some methods are private and need a token and the downstreamWebApi  is great, but it has one public method which is used on the home page which doesn't require to be sigin.
In the service I need to add a http client to do that call.

I would like to call the public endpoint throught the downstreamWebApi  

N/A

**Additional context**
Example:
https://github.com/mspnp/multitenant-saas-guidance/pull/124#discussion_r708515676
https://github.com/mspnp/multitenant-saas-guidance/blob/a99058120927aec5f4180543c9a4d85aaa4c3890/src/Tailspin.Surveys.Web/Services/SurveyService.cs#L70-L76

jmprieur:
- update Ms.Id.abstractions 1 0 6 preview
- Fixes #1445 (as we were there)
- Update the web app calls web API sample.
- Fixing the solution which did no longer propose to run the TodoListService AAD


tests/DevApps/WebAppCallsWebApiCallsGraph/Client/Controllers/TodoListController.cs
```diff
@@ -16,53 +16,59 @@ namespace TodoListClient.Controllers
...
         {
             var value = await _downstreamWebApi.GetForUserAsync<Todo>(
                 ServiceName,
-                $"api/todolist/{id}");
+                options => options.RelativePath = $"api/todolist/{id}");
             return View(value);
         }
 
-        // GET: TodoList/Create
+        // Create an present to the user (no service call)
```
jennyf19, line 46:
```suggestion
        // Create and present to the user (no service call)
```

tests/DevApps/WebAppCallsWebApiCallsGraph/Client/Controllers/TodoListController.cs
```diff
@@ -72,28 +78,25 @@ public async Task<ActionResult> Edit(int id)
...
-     options.HttpMethod = HttpMethod.Patch;
-     options.RelativePath = $"api/todolist/{todo.Id}";
- });
+                options => options.RelativePath = $"api/todolist/{todo.Id}");
+ 
             return RedirectToAction("Index");
         }
 
-        // GET: TodoList/Delete/5
+        // Get the content of the TODO of ID id to present it to the user for deletion
```
jennyf19, line 94:
```suggestion
        // Get the content of the TODO of ID to present it to the user for deletion
```

jennyf19 (approved):

PR #2023, opened 2022-12-30
Changing the min reference for net472 to Microsoft.Extentions

jmprieur:
Changing the min reference for net472 to Microsoft.Extentions 2.x as is done for .NET Standard 2.0.
Keeping net462 as it is as it would otherwise drag the whole framework


jennyf19 (approved):
:shipit:

Issue #2018, opened 2022-12-29
[Feature Request] Consider obsoleting IDownstreamWebApi in rel/v2


jmprieur:
Consider obsoleting IDownstreamWebApi and provide a migration experience.

`[Obsolete("Rather use IDownstreamRestApi in Microsoft.Identity.Abstractions, implemented in Microsoft.Identity.Web.DownstreamRestApi. See aka.ms/id-web-downstream-api-v2")]`

jmprieur:
Fixed in https://github.com/AzureAD/microsoft-identity-web/pull/2037

Issue #1697, opened 2022-04-07
[Bug] 'ClaimsPrincipalFactory.FromTenantIdAndObjectId' populates incorrect ObjectIdentifier claim


julealgon:
**Which version of Microsoft Identity Web are you using?**
Microsoft.Identity.Web v1.23.1

**Where is the issue?**
* Web app
    * [ ] Sign-in users
    * [ ] Sign-in users and call web APIs
* Web API
    * [ ] Protected web APIs (validating tokens)
    * [ ] Protected web APIs (validating scopes)
    * [ ] Protected web APIs call downstream web APIs
* Token cache serialization
     * [ ] In-memory caches
     * [ ] Session caches
     * [ ] Distributed caches
* Other (please describe)
     * [x] Helper methods for instantiating `ClaimsPrincipal` objects

**Is this a new or an existing app?**
This is a new app or an experiment.

**Repro**

```csharp
        var principal = ClaimsPrincipalFactory.FromTenantIdAndObjectId(tenantId: "someTenant", objectId: "someObject");
        var objectId = principal.GetObjectId();
        Assert.Equal("someObject", objectId);
```

**Expected behavior**
The `ClaimsPrincipalFactory.FromTenantIdAndObjectId` method should populate the correct ObjectIdentifier claim (`oid`) when building the principal. It populates the `uid` claim, however.

**Actual behavior**
`ClaimsPrincipalFactory.FromTenantIdAndObjectId` populates the `uid` claim for the object identifier, which is incorrect: it should be using the `oid` claim instead.

**Possible solution**
https://github.com/AzureAD/microsoft-identity-web/blob/bba91c4298411e780c9f082e8e3843eacd8e543c/src/Microsoft.Identity.Web/ClaimsPrincipalFactory.cs#L44

Use `ClaimConstants.Oid` or `ClaimConstants.ObjectId` instead.

**Additional context / logs / screenshots / link to code**

https://github.com/AzureAD/microsoft-identity-web/blob/bba91c4298411e780c9f082e8e3843eacd8e543c/src/Microsoft.Identity.Web/ClaimsPrincipalFactory.cs#L38-L46

jennyf19:
@julealgon We need to give the home tenant and home object ID, especially important for guest accounts. The name of the method is misleading but it's more of an implementation detail with the claims.

jennyf19:
@julealgon Can you give more details on your use case/scenario? What do you do with these claims? Is it related to using MSAL?

[...]

PR #2029, opened 2023-01-03
final updates based on API review and add obsolete message to IDownst…

jennyf19:
…reamWebApi
#2018 
and https://github.com/AzureAD/microsoft-identity-web/issues/1697


jmprieur (approved):
LGTM
Thanks @jennyf19 

We'll have a few devapps to update too, but that's not preventing the release

PR #2036, opened 2023-01-04
Update test apps and integration tests to DownstreamRestApi

jmprieur:
(from DownstreamWebApi)

The only question is, do we want to keep integration test for DownstreamWebApis (and have new ones for DownstreamRestApi), or move the DownstreamWebApis ones to DownstreamRestApi (Which this PR does, in addition to updating the Devapps)


jennyf19 (approved):
:shipit:

PR #2037, opened 2023-01-05
Remove all TSA warnings

jmprieur:
Remove all TSA warnings
Adds EditorBrowsable(Never) attributes for the methods that were obsoleted Auth code flow redemption now also provides the access token (#1771)

### DIFF
--- a/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
@@ -102,7 +102,7 @@ namespace Microsoft.Identity.Web
             _credentialsLoader = credentialsLoader;
         }
 
-        public async Task<string> AddAccountToCacheFromAuthorizationCodeAsync(
+        public async Task<AcquireTokenResult> AddAccountToCacheFromAuthorizationCodeAsync(
             IEnumerable<string> scopes,
             string authCode,
             string authenticationScheme,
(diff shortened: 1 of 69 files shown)


src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -153,7 +153,14 @@ class OAuthConstants
                     _tokenAcquisitionHost.SetSession(Constants.SpaAuthCode, result.SpaAuthCode);
                 }
 
-                return result.IdToken;
+                return new AcquireTokenResult(
```
jennyf19, line 156: nice. this is way better.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -162,7 +169,7 @@ class OAuthConstants
 
                 // Retry
                 _retryClientCertificate = true;
-                await AddAccountToCacheFromAuthorizationCodeAsync(scopes, authCode, authenticationScheme, clientInfo, codeVerifier, userFlow).ConfigureAwait(false);
+                return await AddAccountToCacheFromAuthorizationCodeAsync(scopes, authCode, authenticationScheme, clientInfo, codeVerifier, userFlow).ConfigureAwait(false);
```
jennyf19, line 172: >return [](http://example.com/codeflow?start=16&length=6)

nice catch

jennyf19 (approved):
:shipit:

PR #2043, opened 2023-01-08
add docfx project

jmprieur:


Directory.Build.props
```diff
@@ -20,6 +20,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
+    <TargetFramework>net6.0</TargetFramework>
```
jmprieur, line 23: that's the reason why we don't want to merge the branch yet.
This is because docfx does not understand multiple targe frameworks
Ideally we'd change this dynamically.

jennyf19 (approved):
LGTM

PR #2045, opened 2023-01-16
re-add netcoreapp3.1 target fwk

jennyf19:
- needs testing on .NET 5


jmprieur (approved):
Thanks @jennyf19 
Thanks also for cleaning up the projects that are packageable

PR #2051, opened 2023-01-22
Adding the AzureAd section by default to TokenAcquirerFactory

jmprieur:


jennyf19 (approved):
much better, I think :)

PR #2079, opened 2023-02-11
Making ClientAssertion class public to inherit IDWeb repo on dMSI lib

saitharun-msft:
Making ClientAssertion class public to inherit IDWeb repo on dMSI lib

### DIFF
--- a/src/Microsoft.Identity.Web.Certificateless/ClientAssertion.cs
+++ b/src/Microsoft.Identity.Web.Certificateless/ClientAssertion.cs
@@ -8,7 +8,7 @@ namespace Microsoft.Identity.Web
     /// <summary>
     /// Client assertion.
     /// </summary>
-    internal sealed class ClientAssertion
+    public sealed class ClientAssertion
     {
         /// <summary>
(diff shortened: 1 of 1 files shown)


jennyf19:
This class needs to remain internal because we may change its implementation, especially around the cache.. You can derive from [ClientAssertionProviderBase](https://github.com/AzureAD/microsoft-identity-web/blob/master/src/Microsoft.Identity.Web.Certificateless/ClientAssertionProviderBase.cs).

JoshLozensky:
[ClientAssertionProviderBase](https://github.com/AzureAD/microsoft-identity-web/blob/master/src/Microsoft.Identity.Web.Certificateless/ClientAssertionProviderBase.cs) has an abstract method on line 21 "GetClientAssertion" which requires returning a Task<ClientAssertion>. Without the ability to make a ClientAssertion object, or otherwise altering the coupling here, there is no way to implement the abstract method required to derive ClientAssertionProviderBase.

jennyf19 (approved):
:shipit:

PR #2113, opened 2023-03-17
remove Azure KeyVault package from the lab insfrastructure and remove…

jennyf19:
… unneccessary classes. this is to address a CG issue.


jmprieur (approved):
Lgtm
Thanks @jennyf19 

For later let's plan to.use cert by name

PR #2119, opened 2023-03-17
Update codeql-analysis.yml to use .NET 6

jennyf19:


.github/workflows/codeql-analysis.yml
```diff
@@ -36,13 +36,13 @@ jobs:
     #    and modify them (or add more) to build your code if your project
     #    uses a compiled language
 
-    - name: Setup .NET 5.0.408
-      uses: actions/setup-dotnet@v2.1.0
+    - name: Setup .NET 6.0.301
+      uses: actions/setup-dotnet@v1.7.2
```
jennyf19, line 40:
```suggestion
      uses: actions/setup-dotnet@v2.1.0
```

.github/workflows/codeql-analysis.yml
```diff
@@ -36,13 +36,13 @@ jobs:
     #    and modify them (or add more) to build your code if your project
     #    uses a compiled language
 
-    - name: Setup .NET 5.0.408
+    - name: Setup .NET 6.0.301
       uses: actions/setup-dotnet@v2.1.0
       with:
-        dotnet-version: 5.0.408
+        dotnet-version: 6.0.301
```
jmprieur, line 42: can we use 6.x ?

jmprieur (approved):
LGTM
Thanks @jennyf19

.github/workflows/codeql-analysis.yml
jennyf19: yes, but sometimes it's not happy with a wildcard

PR #2124, opened 2023-03-19
fix codeQL analysis

jennyf19:


.github/workflows/codeql-analysis.yml
```diff
@@ -40,9 +46,21 @@ jobs:
...
+      with:
+        dotnet-version: 7.0.x
+
+   # Install workload
+    - name: Setup wasm-tools
+      run: dotnet workload install wasm-tools
 
     - name: Build with .NET 6
-      run: dotnet build Microsoft.Identity.Web.sln -f net6.0 -p:FROM_GITHUB_ACTION=true --configuration Release
+      run: dotnet test Microsoft.Identity.Web.sln -f net6.0 -p:FROM_GITHUB_ACTION=true --configuration Release --filter "(FullyQualifiedName!~Microsoft.Identity.Web.Test.Integration)&(FullyQualifiedName!~WebAppUiTests)&(FullyQualifiedName!~IntegrationTests)"
```
jmprieur, line 60: Do we need a dotnet test? I would think that dotnet build would be enough for Code QL?

jmprieur (approved):
LGTM
thanks @jennyf19 

It seems to have worked. It found an issue in ToeknAcquirerTests/TokenAcquirer line 206

.github/workflows/codeql-analysis.yml
jennyf19: It's not working with dotnet test. Can revisit later, but at least it's runnign now.

PR #2128, opened 2023-03-19
add a few more tests

jennyf19:


tests/Microsoft.Identity.Web.Test/Base64UrlHelpersTests.cs
```diff
@@ -92,5 +92,28 @@ public void DecodeToBytes_InvalidBase64UrlStringLength_ThrowsException()
...
+
+            // Assert
+            Assert.NotEmpty(result!);
+        }
+
+        [Fact]
+        public void DecodeBytes_TakesNullArgument_ReturnsNull()
+        {
+            // Arrange & Act
+            var result = Base64UrlHelpers.DecodeBytes(null);
```
jmprieur, line 113: Any reason why we don't use Token.Base64UrlEncoder.Decode? (from Wilson)

jmprieur (approved):
LGTM
Thanks @jennyf19

tests/Microsoft.Identity.Web.Test/Base64UrlHelpersTests.cs
jennyf19: Yes, we should. will do in later PR.

PR #2129, opened 2023-03-19
add license info automatically

jennyf19:


jmprieur (approved):
LGTM
Very nice!
Do we want to do the same for Abstractions?

PR #2154, opened 2023-03-27
Move type out of aspnetcore folder

keegan-caruso:
conditional compile for targets


jmprieur:
@keegan-caruso : I would have also expected a change in the csproj?

keegan-caruso:
@jmprieur sorry I don't follow, what change would that be?

jmprieur:
Move the `MicrosoftIdentityAppCallsWebApiAuthenticationBuilder.cs` from the `AspNetCore` folder to the main folder of the project?

keegan-caruso:
I did do that, github UI is just showing it as a rename.

https://github.com/AzureAD/microsoft-identity-web/blob/b9b4c501ad50847bf750c932eb1d12922fcfaa84/src/Microsoft.Identity.Web.TokenAcquisition/MicrosoftIdentityAppCallingWebApiAuthenticationBuilder.cs

Since it is in the root of the project folder, it is getting added to all of the targets now.

jennyf19 (approved):
:shipit:

westin-m (approved):

PR #2156, opened 2023-03-28
Westinm/compute cache key

westin-m:


jennyf19 (approved):
:shipit:

jmprieur (approved):
LGTM
Thanks @westin-m

westin-m:
> Did we want to change that too?
That test is in a different enough scope that including this method there introduces unneeded extra work

PR #2027, opened 2023-01-02
fix: AzureApp RoleClaimType with IdentityProvider

mr-sven:
Adding option to set the default RoleClaimType via AppServicesAuthenticationOptions to use correct roles claim with Microsoft Identity Provider. Fixes #1983 

Can be set via
```csharp
builder.Services.Configure<AppServicesAuthenticationOptions>(AppServicesAuthenticationDefaults.AuthenticationScheme, options =>
{
    options.RoleClaimType = "roles";
});
```

PR #2166, opened 2023-04-02
app services roles

jmprieur:
Fix the roles claims type in apps services
the issuer is v2.0, and it does not go through
Wilson's claimsTypeMapping, so the claims type
will always be "roles"

See also; https://github.com/AzureAD/microsoft-identity-web/pull/2027

### DIFF
--- a/src/Microsoft.Identity.Web/AppServicesAuth/AppServicesAuthenticationInformation.cs
+++ b/src/Microsoft.Identity.Web/AppServicesAuth/AppServicesAuthenticationInformation.cs
@@ -7,7 +7,6 @@ using System.Linq;
 using System.Security.Claims;
 using Microsoft.Extensions.Primitives;
 using Microsoft.IdentityModel.JsonWebTokens;
-using static Microsoft.Identity.Web.AppServicesAuthenticationTokenAcquisition;
 
 namespace Microsoft.Identity.Web
 {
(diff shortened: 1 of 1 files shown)


src/Microsoft.Identity.Web/AppServicesAuth/AppServicesAuthenticationInformation.cs
```diff
@@ -190,7 +189,7 @@ public static bool IsAppServicesAadAuthenticationEnabled
                     jsonWebToken.Claims,
                     idp,
                     isAadV1Token ? Constants.NameClaim : Constants.PreferredUserName,
-                    ClaimsIdentity.DefaultRoleClaimType));
+                    "roles"));
```
jennyf19, line 192: nit: We have a constant in TokenAcquisition for [Roles](https://github.com/AzureAD/microsoft-identity-web/blob/c1135ea1831443182304ad31aeb3562c67e72dfe/src/Microsoft.Identity.Web.TokenAcquisition/ClaimConstants.cs#L71)

jennyf19 (approved):

src/Microsoft.Identity.Web/AppServicesAuth/AppServicesAuthenticationInformation.cs
jmprieur: Thanks. Fixed

PR #2179, opened 2023-04-03
feat: add requirement for denying guest users

h3rmanj:
As a developer, I expect to be able to easily allow all members of a tenant to access my application, without needing to set up roles and manual assignment.

However, Azure AD allows for guest users, that applications like Teams rely on. These guest users will then by default have access to applications that doesn't have any roles, but still only should be accessible by tenant members.

This PR introduces the `DenyGuests` extension to `AuthorizationPolicyBuilder`, which easily lets the developer deny guests users access to their applications with policies.

I think that many enterprise application would want this behavior by default, and without it, the application would be exposed to some vulnerabilities.

//cc @loekensgard @JonasKs

### DIFF
--- a/tests/Microsoft.Identity.Web.Test.Common/TestConstants.cs
+++ b/tests/Microsoft.Identity.Web.Test.Common/TestConstants.cs
@@ -49,7 +49,7 @@ namespace Microsoft.Identity.Web.Test.Common
         public const string AadIssuer = AadInstance + "/" + TenantIdAsGuid + "/v2.0";
         public const string UsGovIssuer = "https://login.microsoftonline.us/" + UsGovTenantId + "/v2.0";
         public const string UsGovTenantId = "72f988bf-86f1-41af-91ab-2d7cd011db47";
-        public const string V1Issuer = "https://sts.windows.net/f645ad92-e38d-4d1a-b510-d1b09a74a8ca/";
+        public const string V1Issuer = "https://sts.windows.net/" + TenantIdAsGuid + "/";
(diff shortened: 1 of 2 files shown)


src/Microsoft.Identity.Web/Policy/DenyGuestsAuthorizationRequirement.cs
```diff
@@ -0,0 +1,15 @@
...
+
+using Microsoft.AspNetCore.Authorization;
+
+namespace Microsoft.Identity.Web
+{
+    /// <summary>
+    /// Implements an <see cref="IAuthorizationRequirement"/>
+    /// which requires the current user to be a member of the tenant.
+    /// </summary>
+    public class DenyGuestsAuthorizationRequirement : IAuthorizationRequirement
```
h3rmanj, line 12: Not locked on the naming of this. I see many requirements are prefixed with `Require`, but `RequireTenantMembers` or `RequireNonGuests` sounds weird and isn't as explicit in what it tries to achieve.

loekensgard:

jmprieur:
thanks @h3rmanj 
Adding more authorization is definitively something we want to do, but we are thinking of doing it a bit differently
cc: @GeoK

h3rmanj:
Mind sharing your thoughts? This is how we've done it internally for now, but we'd love to hear the plans for actual implementation in the SDK or Azure itself.

h3rmanj:
@jmprieur @GeoK

This has been open for some time, considering closing as outdated.
Is there any update on Guest Users having access to all role-less applications by default? Either in the SDK or on the Entra ID platform?

PR #2180, opened 2023-04-03
update IdentityModel version

westin-m:


jennyf19 (approved):
Thanks @westin-m :shipit:

PR #2209, opened 2023-04-24
Integration test update

westin-m:
update test to reflect current users and change condition to avoid future updates


jmprieur (approved):
LGTM
Thanks @westin-m

Issue #1900, opened 2022-10-06
[Feature Request] Log L1/L2 cache details to TokenCacheNotificationArgs


neha-bhargava:
Add L1/L2 cache details to TokenCacheNotificationArgs once the cache details parameter is available from MSAL. This will be available when we release MSAL 4.48.0.

This will involve adding which cache was used L1/L2 to fetch the token. 
MSAL already captures cache latency.

Ref: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/issues/3596
PR to add cache details in MSAL: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/3737

jmprieur:
In 2.13.4

jennyf19:
https://github.com/AzureAD/microsoft-identity-web/releases/tag/2.13.4

PR #2256, opened 2023-05-24
Adding support for MSAL telemetry cache data.

trwalke:
Fix for https://github.com/AzureAD/microsoft-identity-web/issues/1900

Here is the implementation of the telemetry cache details for `TokenCacheNotificationArgs`. 
It appears to me that the `MsalDistributedTokenCacheAdapter` seems to be the correct place to make these changes but I want to confirm this with you @jmprieur and @jennyf19.

- Are there any other locations where I need to implement this logic?
- Are there any expectations for the customer to implement their own versions of `MsalAbstractTokenCacheProvider` or `MsalDistributedTokenCacheAdapter`
- If the above is correct, do you know of anyway to gather this information from the customers implementation? I am trying to avoid adding to much additional api service.


trwalke:
Cant build until after MSAL release.

src/Microsoft.Identity.Web.TokenCache/MsalAbstractTokenCacheProvider.cs
```diff
@@ -219,6 +221,18 @@ protected virtual Task WriteCacheBytesAsync(string cacheKey, byte[] bytes, Cache
...
         }
 
+        /// <summary>
+        /// Method to be overridden by concrete cache serializers to Read the cache bytes.
+        /// </summary>
+        /// <param name="cacheKey">Cache key.</param>
+        /// <param name="cacheSerializerHints">Hints for the cache serialization implementation optimization.</param>
+        /// <param name="telemetryData">Stores details to log to the <see cref="ITelemetryClient"/></param>
+        /// <returns>Read bytes.</returns>
+        protected virtual Task<byte[]?> ReadCacheBytesAsync(string cacheKey, CacheSerializerHints cacheSerializerHints, TelemetryData? telemetryData)
```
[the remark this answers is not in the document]
trwalke: I was thinking of putting it there but it did not seem like the right place based on what I saw in the class. 
"Set of properties that the token cache serialization implementations might use to optimize the cache."

src/Microsoft.Identity.Web.TokenCache/Distributed/MsalDistributedTokenCacheAdapter.cs
```diff
@@ -185,6 +201,11 @@ protected override async Task RemoveKeyAsync(string cacheKey, CacheSerializerHin
                         Logger.BackPropagateL2toL1(_logger, memoryCacheEntryOptions.Size ?? 0);
                         _memoryCache.Set(cacheKey, result, memoryCacheEntryOptions);
                         Logger.MemoryCacheCount(_logger, _memoryCacheType, read, _memoryCache.Count);
+
+                        if (telemetryData != null)
+                        {
+                            telemetryData.CacheLevel = Client.Cache.CacheLevel.L2Cache;
```
jennyf19, line 207: >telemetryData [](http://example.com/codeflow?start=28&length=13)

 I would rather not take this now, as we still have not figured out our client side telemetry story and would prefer to not have API changes which we cannot remove later.

src/Microsoft.Identity.Web.TokenCache/MsalAbstractTokenCacheProvider.cs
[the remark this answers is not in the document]
jennyf19: would rather not add to the public API.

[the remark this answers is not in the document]
bgavrilMS: It's a fair point, I'd add it to the hints. The hints was always designed to be a property bag to allow extensibility without us having to create overloads.

src/Microsoft.Identity.Web.TokenCache/Distributed/MsalDistributedTokenCacheAdapter.cs
bgavrilMS: Are you objecting to how the data is collected or to what data is collected? I agree with the previous comment on the "HOW".

But on the WHY, we have agreed with E&D team to deliver this, in a meeting facilitated by @TimHannMSFT . The spec is [here](https://microsoft-my.sharepoint-df.com/:w:/p/nebharg/EWrRAFHR6VdGvx5i_-uCZiUBROlxzCXZzLdRUJ8ePTKkgw?e=bHx3KO)

src/Microsoft.Identity.Web.TokenCache/Distributed/MsalDistributedTokenCacheAdapter.cs
```diff
@@ -185,6 +188,11 @@ protected override async Task RemoveKeyAsync(string cacheKey, CacheSerializerHin
                         Logger.BackPropagateL2toL1(_logger, memoryCacheEntryOptions.Size ?? 0);
                         _memoryCache.Set(cacheKey, result, memoryCacheEntryOptions);
                         Logger.MemoryCacheCount(_logger, _memoryCacheType, read, _memoryCache.Count);
+
+                        if (telemetryData != null)
```
bgavrilMS, line 192: nit: `telemetryData?.CacheLevel = ...`

src/Microsoft.Identity.Web.TokenCache/Distributed/MsalDistributedTokenCacheAdapter.cs
```diff
@@ -9,6 +9,8 @@
 using Microsoft.Extensions.Caching.Memory;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
+using Microsoft.Identity.Client.TelemetryCore.TelemetryClient;
+using Microsoft.IdentityModel.Abstractions;
 
 namespace Microsoft.Identity.Web.TokenCacheProviders.Distributed
```
bgavrilMS, line 15: You need to also update `MsalMemoryTokenCacheProvider` - the outcome can be only L1.

tests/Microsoft.Identity.Web.Test/L1L2CacheTests.cs
```diff
@@ -202,20 +206,56 @@ public async Task EmptyL1Cache_ReadL2AndSetL1_TestAsync()
...
 
             // Assert
             Assert.NotNull(result);
             Assert.Equal(4, result[0]);
             Assert.Equal(1, _testCacheAdapter._memoryCache.Count);
             Assert.Single(L2Cache._dict);
+            Assert.Equal(CacheLevel.L2Cache, telemetryData.CacheLevel);
+        }
+
+        [Fact]
```
bgavrilMS, line 227: These tests are ok, but higher level tests are always better because they are easier to read and cover the E2E. Have a look at `CacheExtensionsTests`. 

It should be pretty easy to hook up MSAL to a setup with L1+L2 cache and to a setup with only L1 cache. We should make sure we have a test for when L1 cache is disabled (this is one of the options).

bgavrilMS (changes requested):
See comments

src/Microsoft.Identity.Web.TokenCache/Distributed/MsalDistributedTokenCacheAdapter.cs
trwalke: using `telemetryData?` on the left side of the `=` does not work for me.

src/Microsoft.Identity.Web.TokenCache/Distributed/MsalDistributedTokenCacheAdapter.cs
```diff
@@ -185,6 +188,11 @@ protected override async Task RemoveKeyAsync(string cacheKey, CacheSerializerHin
                         Logger.BackPropagateL2toL1(_logger, memoryCacheEntryOptions.Size ?? 0);
                         _memoryCache.Set(cacheKey, result, memoryCacheEntryOptions);
                         Logger.MemoryCacheCount(_logger, _memoryCacheType, read, _memoryCache.Count);
+
+                        if (telemetryData != null)
+                        {
+                            telemetryData.CacheLevel = Client.Cache.CacheLevel.L2Cache;
```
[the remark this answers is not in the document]
trwalke: I put it here because it is in the null check for the result. I figured if you did not get a value from the L2 cache it doesn't really count as a cache hit. Or does it?

I guess I should move the logic out of the null check for the memory cache.

tests/Microsoft.Identity.Web.Test/CacheExtensionsTests.cs
```diff
@@ -33,19 +37,39 @@ public void InMemoryCacheExtensionsTests()
...
+            AssertCacheTelemetry(testTelemetryClient, CacheLevel.L1Cache);
+        }
+
+        [Fact]
+        public async Task CacheExtensions_CcaAlreadyExistsL2_TestsAsync()
+        {
+            AuthenticationResult result;
+            TestTelemetryClient testTelemetryClient = new TestTelemetryClient(TestConstants.ClientId);
+            // new DistributedInMemory serializer with L1 cache disabled
+            result = await CreateAppAndGetTokenAsync(CacheType.DistributedInMemory, testTelemetryClient, addInstanceMock: true, disableL1Cache: true).ConfigureAwait(false);
```
bgavrilMS, line 66: Consider adding variations for:

- `disableL1Cache` is `false. 
- MsalMemoryCacheProvider
- no cache (i.e. use MSAL's internal cache)

bgavrilMS (approved):
Approving code changes, but please consider the PR blocked until we figure out the path forward with telemetry.

jennyf19 (changes requested):
marking as blocked until the telemetry questions are sorted out.

bgavrilMS:
@jennyf19  @jennyf19 - looks like our partners want to use `ITelemetryClient`. Do we need more clarifications from them?

src/Microsoft.Identity.Web.TokenCache/MsalAbstractTokenCacheProvider.cs
```diff
@@ -7,6 +7,8 @@
 using Microsoft.AspNetCore.DataProtection;
 using Microsoft.Extensions.Logging;
 using Microsoft.Identity.Client;
+using Microsoft.Identity.Client.TelemetryCore.TelemetryClient;
+using Microsoft.IdentityModel.Abstractions;
```
jennyf19, line 11: remove unneeded using statements

src/Microsoft.Identity.Web.TokenCache/Distributed/MsalDistributedTokenCacheAdapter.cs
```diff
@@ -9,6 +9,8 @@
 using Microsoft.Extensions.Caching.Memory;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
+using Microsoft.Identity.Client.TelemetryCore.TelemetryClient;
+using Microsoft.IdentityModel.Abstractions;
```
jennyf19, line 13: remove using statements

jennyf19:
> @jennyf19 @jennyf19 - looks like our partners want to use `ITelemetryClient`. Do we need more clarifications from them?

We are seeing much progress with OTel, so, at this point, we would not use ITelemetryClient. Will know more in another week or two.

tests/Microsoft.Identity.Web.Test/TestTelemetryClient.cs
```diff
@@ -0,0 +1,59 @@
...
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.IdentityModel.Abstractions;
+
+namespace Microsoft.Identity.Web.Test
+{
+    internal class TestTelemetryClient : ITelemetryClient
```
bgavrilMS, line 13: We want to deprecate `ITelemetryClient`. Can you please re-write the test to not use this?

src/Microsoft.Identity.Web.TokenCache/CacheSerializerHints.cs
```diff
@@ -3,6 +3,8 @@
 
 using System;
 using System.Threading;
+using Microsoft.Identity.Client.TelemetryCore.TelemetryClient;
+using Microsoft.IdentityModel.Abstractions;
```
jmprieur, line 7: Is it still needed?

jennyf19:
@trwalke do we have an ETA on wrapping this up? We are doing a release early next week. thanks.

trwalke:
> @trwalke do we have an ETA on wrapping this up? We are doing a release early next week. thanks.

Targeting end of week or a day before. I encountered an issue with MSAL tests as a result of some of MSAL's internal static caching which is causing tests to fail on this branch. Go ahead and release. @jennyf19

tests/Microsoft.Identity.Web.Test/TestTelemetryClient.cs
trwalke: I cant seems to find an easy way to get the cache telemetry data without some kind of telemetry client. Maybe it is better to update the tests when the new telemetry client is released. Atleast this is able to validate its functionality for now.

src/Microsoft.Identity.Web.TokenCache/TokenCacheExtensions.cs
```diff
@@ -19,7 +19,8 @@ namespace Microsoft.Identity.Web
     /// </summary>
     public static class TokenCacheExtensions
     {
-        private static readonly ConcurrentDictionary<MethodInfo, IServiceProvider> s_serviceProviderFromAction
+        //internal for testing only
+        internal static readonly ConcurrentDictionary<MethodInfo, IServiceProvider> s_serviceProviderFromAction
```
trwalke, line 23: The MSAL tests need a way to reset this cache because when they run in a batch, they have the potential to collide and tests fail. Enabling tests to reset this fixed this issue.

src/Microsoft.Identity.Web.TokenCache/Distributed/MsalDistributedTokenCacheAdapter.cs
bgavrilMS, line 13: This using should not be needed.

tests/Microsoft.Identity.Web.Test.Common/Mocks/MockHttpMessageHandler.cs
```diff
@@ -41,6 +44,21 @@ protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage reques
...
+            if (uri.AbsoluteUri.Contains("/discovery/instance"))
+            {
+                ReplaceMockHttpMessageHandler(this);
+
+                var responseMessage = new HttpResponseMessage(HttpStatusCode.OK)
+                {
+                    Content = new StringContent(TestConstants.DiscoveryJsonResponse),
+                };
+
+                return new TaskFactory().StartNew(() => responseMessage, cancellationToken);
```
bgavrilMS, line 59: Use `Task.FromResult(responseMessage)`

tests/Microsoft.Identity.Web.Test.Common/Mocks/MockHttpMessageHandler.cs
```diff
@@ -41,6 +44,21 @@ protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage reques
 
             var uri = request.RequestUri;
             Assert.NotNull(uri);
+
+            //Intercept instance discovery requests and serve a response. 
+            //Also, requeue the current mock handler for MSAL's next request.
+            if (uri.AbsoluteUri.Contains("/discovery/instance"))
```
bgavrilMS, line 50: See the comments from GitHubActions around string comparison.

tests/Microsoft.Identity.Web.Test/CacheExtensionsTests.cs
```diff
@@ -206,17 +238,32 @@ private enum CacheType
...
+                            });
+                        }
                     });
                     break;
             }
 
             var result = await confidentialApp.AcquireTokenForClient(new[] { TestConstants.s_scopeForApp })
                 .ExecuteAsync().ConfigureAwait(false);
 
+            tokenHandler.ReplaceMockHttpMessageHandler = null;
```
bgavrilMS, line 256: Please resolve the GitHub Actions warnings. Id.Web uses nullable. Have a look at https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/null-forgiving

bgavrilMS (approved):

tests/Microsoft.Identity.Web.Test/TestTelemetryClient.cs
trwalke: CC @jennyf19  @jmprieur

src/Microsoft.Identity.Web.TokenCache/TokenCacheExtensions.cs
trwalke: CC @jennyf19 @jmprieur

bgavrilMS:
All telemetry questions are sorted out. Cache Level is required data point for MSAL - irrespective of telemetry solution.

`ITelemetryClient` is not being deprecated for now, so it will appear there as well and it's acceptable way to test for this datapoint.

jennyf19:
@bgavrilMS can you make sure someone from the Id Web side of things approves PRs in the future before merging them? Especially ones opened a few months ago. It's hard to keep track of what is ready to review and what is not.
Thanks

cc: @jmprieur @JoshLozensky @westin-m

jennyf19:
> All telemetry questions are sorted out. Cache Level is required data point for MSAL - irrespective of telemetry solution.
> 
> `ITelemetryClient` is not being deprecated for now, so it will appear there as well and it's acceptable way to test for this datapoint.

also, we don't want `ITelemetryClient` that's the solution we are going with.

PR #2292, opened 2023-06-15
fix CVE-2023-29331

jennyf19:


jmprieur (approved):
LGTM
I had not even seen it

PR #2337, opened 2023-07-21
Test fix for AccountID being in x.y format

bgavrilMS:
The ID.Web test in question fails when presented with a DEBUG version of MSAL because it uses an incorrect id.


jmprieur (approved):
Thanks @bgavrilMS

PR #2341, opened 2023-07-24
Update changelog.md

westin-m:


jmprieur (approved):
LGTM
Thanks @westin-m

jennyf19 (approved):
:shipit: 

Don't we also have the .net8 conditional compilation in this one or was that 2.13.?

Issue #2083, opened 2023-02-15
OpenIdConnect events called twice for a single login - when using EnableTokenAcquisitionToCallDownstreamApi 


markharrison:
### Microsoft.Identity.Web Library

Microsoft.Identity.Web

### Microsoft.Identity.Web version

1.26.0

### Web app

Sign-in users

### Web API

Not Applicable

### Token cache serialization

In-memory caches

### Description

When I add 
```
     .EnableTokenAcquisitionToCallDownstreamApi(builder.Configuration.GetValue<string>
                            ("DownstreamApi:Scopes")?.Split(' '))
     .AddInMemoryTokenCaches();
```
to my startup logic - then during my login activity, all my OpenId events are called twice.

Why is this happening?

### Reproduction steps

Use Code A - get events called twice for a single logon

Use Code B - get events called once for a single logon

using Framework : .NET 7.0.3

### Error message

n/a

### Id Web logs

I have log /trace messages when the event is called

Logs A - events appear twice for single login
```
Microsoft.Identity.Web.TokenAcquisition: Information: False MSAL 4.46.0.0 MSAL.NetCore .NET 7.0.3 Microsoft Windows 10.0.22621 [2023-02-15 17:47:14Z - 9d18da67-e9a5-4e35-a36f-5b962176077b] Saving Id Token and Account in cache ...
Microsoft.Identity.Web.TokenAcquisition: Information: False MSAL 4.46.0.0 MSAL.NetCore .NET 7.0.3 Microsoft Windows 10.0.22621 [2023-02-15 17:47:14Z - 9d18da67-e9a5-4e35-a36f-5b962176077b] Saving RT in cache...
Microsoft.Identity.Web.TokenAcquisition: Information: False MSAL 4.46.0.0 MSAL.NetCore .NET 7.0.3 Microsoft Windows 10.0.22621 [2023-02-15 17:47:14Z - 9d18da67-e9a5-4e35-a36f-5b962176077b] 
	=== Token Acquisition finished successfully:
Microsoft.Identity.Web.TokenAcquisition: Information: False MSAL 4.46.0.0 MSAL.NetCore .NET 7.0.3 Microsoft Windows 10.0.22621 [2023-02-15 17:47:14Z - 9d18da67-e9a5-4e35-a36f-5b962176077b]  AT expiration time: 15/02/2023 19:02:24 +00:00, scopes: openid profile User.Read email. source: IdentityProvider
[...]

PR #2350, opened 2023-07-26
Lozensky/fix perf issue CallDownstreamApi

JoshLozensky:
# fix perf issue CallDownstreamApi

The Services.Configure line deleted also occurs in [WebApiAuthenticationBuilder](https://github.com/AzureAD/microsoft-identity-web/blob/48e4a5e9af22439ac9e633707860073bd5d4c91e/src/Microsoft.Identity.Web/WebApiExtensions/MicrosoftIdentityWebApiAuthenticationBuilder.cs#L45C1-L45C98)

## Description
The line deleted in MicrosoftIdentityWebApiAuthenticationBuilderExtensions.cs calls builder.Services.Configure which is also happening in the builder constructor itself (MicrosoftIdentityWebApiAuthenticationBuilder) with the same arguments which duplicated the configuration every time a builder was run using the CallDownstreamApi.

Testing:
- Unit tests all pass
- Succeeded running the DaemonConsoleCallingDownstreamApi sample against the MinimalWebApi sample.

#2083
#2355

### DIFF
--- a/src/Microsoft.Identity.Web/WebApiExtensions/MicrosoftIdentityWebApiAuthenticationBuilderExtensions.cs
+++ b/src/Microsoft.Identity.Web/WebApiExtensions/MicrosoftIdentityWebApiAuthenticationBuilderExtensions.cs
@@ -79,7 +79,7 @@ namespace Microsoft.Identity.Web
         {
             _ = Throws.IfNull(configurationSection);
             _ = Throws.IfNull(builder);
-
+            
             AddMicrosoftIdentityWebApiImplementation(
                 builder,
                 options => configurationSection.Bind(options),
@@ -89,7 +89,7 @@ namespace Microsoft.Identity.Web
 
             return new MicrosoftIdentityWebApiAuthenticationBuilderWithConfiguration(
                 builder.Services,
-                jwtBearerScheme,
+                jwtBearerScheme,// first sends here
                 options => configurationSection.Bind(options),
                 options => configurationSection.Bind(options),
                 configurationSection);
@@ -142,7 +142,6 @@ namespace Microsoft.Identity.Web
             bool subscribeToJwtBearerMiddlewareDiagnosticsEvents)
         {
             builder.AddJwtBearer(jwtBearerScheme, configureJwtBearerOptions);
-            builder.Services.Configure(jwtBearerScheme, configureMicrosoftIdentityOptions);
             builder.Services.AddSingleton<IMergedOptionsStore, MergedOptionsStore>();
 
             builder.Services.AddHttpContextAccessor();
@@ -166,7 +165,7 @@ namespace Microsoft.Identity.Web
 
             // Change the authentication configuration to accommodate the Microsoft identity platform endpoint (v2.0).
             builder.Services.AddOptions<JwtBearerOptions>(jwtBearerScheme)
-                .Configure<IServiceProvider, IMergedOptionsStore, IOptionsMonitor<MicrosoftIdentityOptions>>((
+                .Configure<IServiceProvider, IMergedOptionsStore, IOptionsMonitor<MicrosoftIdentityOptions>>(( //here
                 options,
                 serviceProvider,
                 mergedOptionsMonitor,
--- a/src/Microsoft.Identity.Web/WebApiExtensions/MicrosoftIdentityWebApiAuthenticationBuilderExtensions.cs
+++ b/src/Microsoft.Identity.Web/WebApiExtensions/MicrosoftIdentityWebApiAuthenticationBuilderExtensions.cs
@@ -83,7 +83,6 @@ namespace Microsoft.Identity.Web
             AddMicrosoftIdentityWebApiImplementation(
                 builder,
                 options => configurationSection.Bind(options),
-                options => configurationSection.Bind(options),
                 jwtBearerScheme,
                 subscribeToJwtBearerMiddlewareDiagnosticsEvents);
 
@@ -122,7 +121,6 @@ namespace Microsoft.Identity.Web
             AddMicrosoftIdentityWebApiImplementation(
                 builder,
                 configureJwtBearerOptions,
-                configureMicrosoftIdentityOptions,
                 jwtBearerScheme,
                 subscribeToJwtBearerMiddlewareDiagnosticsEvents);
 
@@ -137,7 +135,6 @@ namespace Microsoft.Identity.Web
         private static void AddMicrosoftIdentityWebApiImplementation(
             AuthenticationBuilder builder,
             Action<JwtBearerOptions> configureJwtBearerOptions,
-            Action<MicrosoftIdentityOptions> configureMicrosoftIdentityOptions,
             string jwtBearerScheme,
             bool subscribeToJwtBearerMiddlewareDiagnosticsEvents)
         {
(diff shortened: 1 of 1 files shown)


src/Microsoft.Identity.Web/WebApiExtensions/MicrosoftIdentityWebApiAuthenticationBuilderExtensions.cs
```diff
@@ -142,7 +142,6 @@ public static class MicrosoftIdentityWebApiAuthenticationBuilderExtensions
             bool subscribeToJwtBearerMiddlewareDiagnosticsEvents)
         {
             builder.AddJwtBearer(jwtBearerScheme, configureJwtBearerOptions);
-            builder.Services.Configure(jwtBearerScheme, configureMicrosoftIdentityOptions);
```
jennyf19, line 145: I don't think this will work because we need the MicrosoftIdentityOptions as well, if they are used.

jennyf19: For example: https://github.com/AzureAD/microsoft-identity-web/blob/master/tests/DevApps/WebAppCallsWebApiCallsGraph/Client/Startup.cs#L79 but on the web API side.

JoshLozensky: Hi @jennyf19, could you help me understand how the MicrosoftIdentityOptions wouldn't be included?  I only see two
code paths and both make a Services.Configure call using the same JwtBearerOptions and MicrosoftIdentityOptions. Full paths I'm following below:

The private function [AddMicrosoftIdentityWebApiImplementation](https://github.com/AzureAD/microsoft-identity-web/blob/48e4a5e9af22439ac9e633707860073bd5d4c91e/src/Microsoft.Identity.Web/WebApiExtensions/MicrosoftIdentityWebApiAuthenticationBuilderExtensions.cs#L137) I deleted the builder.Services.Configure call from is called by the two implementations of the AddMicrosoftIdentityWebApi method in the same file ([1](https://github.com/AzureAD/microsoft-identity-web/blob/48e4a5e9af22439ac9e633707860073bd5d4c91e/src/Microsoft.Identity.Web/WebApiExtensions/MicrosoftIdentityWebApiAuthenticationBuilderExtensions.cs#L74), [2](https://github.com/AzureAD/microsoft-identity-web/blob/48e4a5e9af22439ac9e633707860073bd5d4c91e/src/Microsoft.Identity.Web/WebApiExtensions/MicrosoftIdentityWebApiAuthenticationBuilderExtensions.cs#L111)). These two functions return a new builder, using the same JwtBearerOptions and MicrosoftIdentityOptions used as arguments in the private function whose line I deleted. Those two returned builder objects either are or derive from the [MicrosoftIdentityWebApiAuthenticationBuilder](https://github.com/AzureAD/microsoft-identity-web/blob/48e4a5e9af22439ac9e633707860073bd5d4c91e/src/Microsoft.Identity.Web/WebApiExtensions/MicrosoftIdentityWebApiAuthenticationBuilder.cs#L45) where on line 45 in the constructor a Services.Configure call is made using those same JwtBearerOptions and MicrosoftIdentityOptions arguments.

jennyf19: Thanks for the explanation @JoshLozensky. If you run the dev app which is webAppCallsWebApiCallsGraph, and add this to the service:

`services.Configure<MicrosoftIdentityOptions>(
                                           o => o.ClientSecret = "your_secret-here");`

and comment it out in the [appsettings.json](https://github.com/AzureAD/microsoft-identity-web/blob/master/tests/DevApps/WebAppCallsWebApiCallsGraph/TodoListService/appsettings.json#L7) then run them both as startup projects. I get this error:

An unhandled exception occurred while processing the request.

HttpRequestException: 500 InternalServerError MSAL.NetCore.4.55.0.0.MsalServiceException:
ErrorCode: invalid_client
Microsoft.Identity.Client.MsalServiceException: A configuration issue is preventing authentication - check the error message from the server for details. You can modify the configuration in the application registration portal. See https://aka.ms/msal-net-invalid-client for details. Original exception: AADSTS7000215: Invalid client secret provided. Ensure the secret being sent in the request is the client secret value, not the client secret ID, for a secret added to app 'a4c2469b-cf84-4145-8f5f-cb7bacf814bc'.

As it's not longer picking up the MicrosoftIdentityOptions. Does it repro like this for you?

jennyf19: comment out the clientSecret in the appsettings.json. ^ not sure if that was clear.

jennyf19: Actually, if I have this `services.Configure<MicrosoftIdentityOptions>(JwtBearerDefaults.AuthenticationScheme, o => o.ClientSecret = "your_secret-here");` it seems to work. 

@jmprieur thoughts on this requiring the authScheme now? 

@JoshLozensky were you able to repro the issue attached?

src/Microsoft.Identity.Web/WebApiExtensions/MicrosoftIdentityWebApiAuthenticationBuilderExtensions.cs
```diff
@@ -79,7 +79,7 @@ public static class MicrosoftIdentityWebApiAuthenticationBuilderExtensions
         {
             _ = Throws.IfNull(configurationSection);
             _ = Throws.IfNull(builder);
-
+            
             AddMicrosoftIdentityWebApiImplementation(
                 builder,
                 options => configurationSection.Bind(options),
```
jennyf19, line 85: here and 86, 83, 94, we can remove the lambda:
`options => configurationSection.Bind(options),` -> `configurationSection.Bind(options),`

src/Microsoft.Identity.Web/WebApiExtensions/MicrosoftIdentityWebApiAuthenticationBuilderExtensions.cs
jennyf19: Last thing, do we need to pass this in now? 
`Action<MicrosoftIdentityOptions> configureMicrosoftIdentityOptions,`

If not, please clean that up as part of this PR as well.

JoshLozensky: Hey @jennyf19, I couldn't repro the exact same error in this branch, though I did get the same error when I ran it in the main repo branch. In this branch I did get a similar scenario with the call failing unless JwtBearerDefaults.AuthenticationScheme is added, but it doesn't appear to be related to the code in this PR.

It looks like in both services.Configure<MicrosoftIdentityOptions> scenarios a call to services.AddSingleton<IConfigureOptions>(name, configureOptions) is ultimately made (as seen in the image below). The difference is the name is set to either JwtBearerDefaults.AuthenticationScheme or Options.Options.DefaultName. It seems like the call without the authScheme would only work if the Options.Options.DefaultName was set to the same value as JwtBearerDefaults.AuthenticationScheme regardless of the changes made here.

image

JoshLozensky: I cleaned up the Action<MicrosoftIdentityOptions> configureMicrosoftIdentityOptions, per your other comment.

jennyf19: Thanks @JoshLozensky! It should fail if "similar scenario with the call failing unless JwtBearerDefaults.AuthenticationScheme is added".

jennyf19 (approved):
:shipit:

jmprieur (approved):
LGTM
Thanks @JoshLozensky

Issue #2354, opened 2023-07-29
[Eng excellence] Use Playwright for UI test automation


jennyf19:
Currently, we use Selenium for the UI automation, but it is unreliable and causes slow down in work velocity.

Switch current UI tests over to using https://playwright.dev/ and add more tests (this would be a second increment).

See -> https://playwright.dev/docs/codegen

jennyf19:
done.

jennyf19:
https://github.com/AzureAD/microsoft-identity-web/releases/tag/2.13.4

PR #2370, opened 2023-08-08
Replace Selenium With Playwright

JoshLozensky:
Selenium UI tests replaced with Playwright


.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
+name: Playwright Tests
```
jennyf19, line 1: this is cool. Will this enable these tests to run on the PRs? How will it know which tests to run?

playwright.config.ts
```diff
@@ -0,0 +1,77 @@
+import { defineConfig, devices } from '@playwright/test';
```
jennyf19, line 1: why do we need this?

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
...
+  test:
+    timeout-minutes: 60
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-node@v3
+      with:
+        node-version: 18
+    - name: Install dependencies
+      run: npm ci
```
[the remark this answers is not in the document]
jmprieur: You can even install it from the test itself:
https://playwright.dev/dotnet/docs/browsers#install-browsers-via-api

```Csharp
var exitCode = Microsoft.Playwright.Program.Main(new[] {"install"});
if (exitCode != 0)
{
    throw new Exception($"Playwright exited with code {exitCode}");
}
```

.github/workflows/playwright.yml
JoshLozensky: Playwright can recognize its own tests and has integration with GitHub actions/events so we can definitely have it run the tests in response to things like opening or committing to a PR.

playwright.config.ts
JoshLozensky: This is the way the playwright documentation says to do imports, so I followed it... Not sure why it needs to be this way, I can try to find out if you think it would be valuable.

.github/workflows/playwright.yml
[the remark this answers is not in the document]
JoshLozensky: No particular reason, the yml was auto-generated and that is what it chose, I will try switching it to PowerShell

tests/IntegrationTests/WebAppUiTests/WebAppUiTests.csproj
```diff
@@ -10,6 +10,7 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.Mvc.Testing" Version="$(MicrosoftAspNetCoreMvcTestingVersion)" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="$(MicrosoftNetTestSdkVersion)" />
+    <PackageReference Include="Microsoft.Playwright" Version="1.36.0" />
```
JoshLozensky, line 13: I will add this version to the relevant props file if one exists at the UI tests level

jennyf19:
Make sure to update with master and let VS resolve the merge conflicts.

jennyf19:
Looks like there are still merge conflicts. Please resolve those first.

jennyf19:
Looking great so far. the builds are failing due to the namespaces not being removed. I'll try it locally later today.

jennyf19 (approved):
:shipit: 

Great to see all the code removed too.

westin-m (approved):

tests/IntegrationTests/WebAppUiTests/TestingWebAppLocally.cs
```diff
@@ -71,5 +62,25 @@ public async Task ChallengeUser_SignInSucceedsTestAsync_LocalHttp()
...
+    private Process? StartWebAppLocally()
+    {
+        string uiTestAssemblyLocation = typeof(TestingWebAppLocally).Assembly.Location;
+        // e.g. microsoft-identity-web\tests\IntegrationTests\WebAppUiTests\bin\Debug\net6.0\WebAppUiTests.dll
+        string testedAppLocation = Path.Combine(Path.GetDirectoryName(uiTestAssemblyLocation)!);
+        // e.g. microsoft-identity-web\tests\IntegrationTests\WebAppUiTests\bin\Debug\net6.0
+        string[] segments = testedAppLocation.Split(Path.DirectorySeparatorChar);
+        int numberSegments = segments.Length;
+        int startLastSegments = numberSegments - 3;
+        int endFirstSegments = startLastSegments - 2;
```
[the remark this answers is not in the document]
JoshLozensky: Very fair question, this code predated me but my guess is they want to isolate the path segments to make it easier to run in/on different domains, subdomains, and/or ports.

jennyf19:
#2354

pmaytak:
@JoshLozensky What're the benefits of using Playwright instead of Selenium? Just reliability? Has it been more reliable so far?

JoshLozensky:
Good question @pmaytak, in addition to being more reliable via built-in retry logic that helps prevent flaky tests, it is faster, much easier to implement solutions (especially cross-browser), has more informative debugging tools, and it has a live auto-code tool. The auto-code tool in particular will make upkeep much easier in the case of future changes to the samples by enabling anyone to easily step through the flow and reference the elements of the pages. This will help avoid issues caused by silos of knowledge by reducing the barrier to entry for anyone new to the test code.

PR #2374, opened 2023-08-10
Fix for #2371

### LINKED ISSUE #2371: Allow Downstream API calls to be used outside of ASP.NET Core Hosting environment [Feature Request]
When calling downstream api from a console application allowed schemes are not fetched properly. This is because IAuthenticationSchemeProvider is not registered and aspnet core ITokenAcquisitionHost is used by default.

Ability to inject ITokenAcquisitionHost but I would ideally like idweb to not be dependent on asp.net core dlls, but anything to make it work would suffice.

Taking dependency on asp.net core myself and registering the type.

**Additional context**

jmprieur:
# Fix for #2371 

## Description

In .NET 5+, we now use the DefaultTokenAcquisitionHost (the host for SDK apps) instead of the Asp.NET core one, when the service collection was not initialized by ASP.NET Core (that is the `IWebHostEnvironment` is not present in the collection.

If developers want the ASP.NET Core host, they would need to use the WebApplication.CreateBuilder().Services instead of instanciating a simple service collection.

Fixes #2371

### DIFF
--- a/src/Microsoft.Identity.Web.TokenAcquisition/PrincipalExtensionsForSecurityTokens.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/PrincipalExtensionsForSecurityTokens.cs
@@ -29,7 +29,7 @@ namespace Microsoft.Identity.Web
             {
                 return new JsonWebToken(s);
             }
-            return null;
+            return (o != null) ? new JsonWebToken(o.ToString()) : null;
         }
     }
 }
--- a/src/Microsoft.Identity.Web.TokenAcquisition/PrincipalExtensionsForSecurityTokens.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/PrincipalExtensionsForSecurityTokens.cs
@@ -1,0 +1,35 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Security.Claims;
+using System.Security.Principal;
+using Microsoft.IdentityModel.JsonWebTokens;
+using Microsoft.IdentityModel.Tokens;
+
+namespace Microsoft.Identity.Web
+{
+    /// <summary>
+    /// Extensions to retrive a <see cref="SecurityToken"/> from <see cref="ClaimsPrincipal"/>.
+    /// </summary>
+    public static class PrincipalExtensionsForSecurityTokens
+    {
+        /// <summary>
+        /// Get the <see cref="SecurityToken"/> used to call a protected web API.
+        /// </summary>
+        /// <param name="claimsPrincipal"></param>
+        /// <returns></returns>
+        public static SecurityToken? GetBootstrapToken(this IPrincipal claimsPrincipal)
+        {
+            object? o = (claimsPrincipal?.Identity as ClaimsIdentity)?.BootstrapContext;
+            if (o is SecurityToken securityToken)
+            {
+                return securityToken;
+            }
+            if (o is string s)
+            {
+                return new JsonWebToken(s);
+            }
+            return null;
+        }
+    }
+}
--- a/src/Microsoft.Identity.Web.TokenAcquisition/ServiceCollectionExtensions.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/ServiceCollectionExtensions.cs
@@ -89,7 +89,7 @@ namespace Microsoft.Identity.Web
             // Token acquisition service
             if (isTokenAcquisitionSingleton)
             {
-#if !NETSTANDARD2_0 && !NET462 && !NET472
+#if NETCOREAPP3_1_OR_GREATER
                 // ASP.NET Core
                 services.AddHttpContextAccessor();
                 if (forceSdk)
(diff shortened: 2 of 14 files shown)


src/Microsoft.Identity.Web.TokenAcquisition/ServiceCollectionExtensions.cs
```diff
@@ -41,6 +41,10 @@ public static class ServiceCollectionExtensions
         {
             _ = Throws.IfNull(services);
 
+#if !NETSTANDARD2_0 && !NET462 && !NET472
```
MZOLN, line 44: Are you sure that this is correct? Later below it says that #else is a netfwk.
Netstadrdad2_0 can end up being consumed in any NET Core runtime including 7.0 and i imagine the unreleased 8.0

tests/IntegrationTests/TokenAcquirerTests/TokenAcquirer.cs
```diff
@@ -33,6 +33,16 @@ public TokenAcquirer()
             TokenAcquirerFactory.ResetDefaultInstance(); // Test only
         }
 
+        [Fact]
+        public void TokenAcquirerFactoryDoesNotUseAspNetCoreHost()
```
MZOLN, line 37: This test project only tests NET7.0 and NET8.0 it will never test netstandard, which is part of the change.
Would it make sense to add netstandard as part of the target to the test project?

src/Microsoft.Identity.Web.TokenAcquisition/PrincipalExtensionsForSecurityTokens.cs
```diff
@@ -0,0 +1,35 @@
...
+
+using System.Security.Claims;
+using System.Security.Principal;
+using Microsoft.IdentityModel.JsonWebTokens;
+using Microsoft.IdentityModel.Tokens;
+
+namespace Microsoft.Identity.Web
+{
+    /// <summary>
+    /// Extensions to retrive a <see cref="SecurityToken"/> from <see cref="ClaimsPrincipal"/>.
```
jennyf19, line 12:
```suggestion
    /// Extensions to retrieve a <see cref="SecurityToken"/> from <see cref="ClaimsPrincipal"/>.
```

src/Microsoft.Identity.Web.TokenAcquisition/PrincipalExtensionsForSecurityTokens.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// </summary>
+    public static class PrincipalExtensionsForSecurityTokens
+    {
+        /// <summary>
+        /// Get the <see cref="SecurityToken"/> used to call a protected web API.
+        /// </summary>
+        /// <param name="claimsPrincipal"></param>
+        /// <returns></returns>
```
jennyf19, line 20:
```suggestion
        /// <returns><see cref="SecurityToken"/></returns>
```

tests/Microsoft.Identity.Web.Test/PrincipalExtensionsForSecurityTokensTests.cs
```diff
@@ -0,0 +1,73 @@
+using Microsoft.IdentityModel.JsonWebTokens;
```
jennyf19, line 1:
```suggestion
﻿// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT License.

using Microsoft.IdentityModel.JsonWebTokens;
```

tests/Microsoft.Identity.Web.Test/PrincipalExtensionsForSecurityTokensTests.cs
```diff
@@ -0,0 +1,73 @@
...
+    {
+        [Fact]
+        public void GetBootstrapToken_Returns_Null_When_ClaimsPrincipal_Is_Null()
+        {
+            // Arrange
+            IPrincipal claimsPrincipal = null;
+
+            // Act
+            var result = claimsPrincipal.GetBootstrapToken();
+
```
jennyf19, line 20: there are a few warnings in the test to address

jennyf19 (approved):
:shipit:

src/Microsoft.Identity.Web.TokenAcquisition/ServiceCollectionExtensions.cs
jmprieur: Fixed. Thanks!

tests/IntegrationTests/TokenAcquirerTests/TokenAcquirer.cs
jmprieur: No longer needed with the fix

tests/Microsoft.Identity.Web.Test/PrincipalExtensionsForSecurityTokensTests.cs
jmprieur: Done. Thanks!

tests/IntegrationTests/TokenAcquirerTests/TokenAcquirer.cs
MZOLN: Well we should still test the whole library for netstandard2.0 dont you think? It`s a release target after all.

PR #2412, opened 2023-08-30
Set EnableCacheSynchronization to false by default.

### LINKED ISSUE #2413: [Feature Request] EnableCacheSynchronization should be false by default
In the next version MSAL will set `EnableCacheSynchronization` flag to true by default. (This will help to make singleton cache consistent.) This flag should be false by default in Identity Web to not affect the users. This is similar to the `LegacyCacheCompatibilityEnabled` (#1268).

pmaytak:
## Description
In the next version MSAL will set `EnableCacheSynchronization` flag to true by default. These changes will keep the flag false by default here to not affect the users. This is similar to the `LegacyCacheCompatibilityEnabled` (#1266).

Fixes #2413

### DIFF
--- a/src/Microsoft.Identity.Web.TokenAcquisition/MicrosoftIdentityOptions.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/MicrosoftIdentityOptions.cs
@@ -93,8 +93,22 @@ namespace Microsoft.Identity.Web
         /// Performance improvements when working with MSAL only apps.
         /// Set to true if you have a shared cache with ADAL apps.
         /// </summary>
-        /// The default is <c>false.</c>
+        /// <remarks>
+        /// The default is <c>false</c>.
+        /// </remarks>
         public bool LegacyCacheCompatibilityEnabled { get; set; }
+
+        /// <summary>
+        /// When set to <c>true</c>, MSAL will lock cache access at the ConfidentialClientApplication level, i.e.
+        /// the block of code between BeforeAccessAsync and AfterAccessAsync callbacks will be synchronized. 
+        /// Apps can set this flag to <c>false</c> to enable an optimistic cache locking strategy, which may result in better performance
+        /// at the cost of cache consistency. 
+        /// Setting this flag to <c>false</c> is only recommended for apps which create a new ConfidentialClientApplication per request.
+        /// </summary>
+        /// <remarks>
+        /// The default is <c>false</c>.
+        /// </remarks>
+        public bool EnableCacheSynchronization { get; set; }
 
         /// <summary>
         /// Is considered B2C if the attribute SignUpSignInPolicyId is defined.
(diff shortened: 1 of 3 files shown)


src/Microsoft.Identity.Web.TokenAcquisition/MicrosoftIdentityOptions.cs
```diff
@@ -93,8 +93,22 @@ public class MicrosoftIdentityOptions : OpenIdConnectOptions
...
+        /// When set to <c>true</c>, MSAL will lock cache access at the ConfidentialClientApplication level, i.e.
+        /// the block of code between BeforeAccessAsync and AfterAccessAsync callbacks will be synchronized. 
+        /// Apps can set this flag to <c>false</c> to enable an optimistic cache locking strategy, which may result in better performance
+        /// at the cost of cache consistency. 
+        /// Setting this flag to <c>false</c> is only recommended for apps which create a new ConfidentialClientApplication per request.
+        /// </summary>
+        /// <remarks>
+        /// The default is <c>false</c>.
+        /// </remarks>
+        public bool EnableCacheSynchronization { get; set; }
```
bgavrilMS, line 111: There are cases, for example [here](https://github.com/AzureAD/microsoft-identity-web/blob/master/src/Microsoft.Identity.Web.TokenAcquisition/ServiceCollectionExtensions.cs#L40) where singletons are used. Can we force this setting to `true` there?

I think it's better to err on the side of correctness and not on the side of "perf" (btw, I don't think there is a sensible perf improvement that this makes, historically we created this setting to work around a deadlock). 

If it's too complex to keep track of object lifetime (i.e. some users might use CCA singleton via config), then it's better to set this to `true` as default.

jmprieur: Thanks
MicrosoftIdentityOptions is not the recommended way in 2.x. we recommended customers to use MicrosoftIdentityApplicationOptions.

Is a boolean exposed in Msal?

jennyf19 (changes requested):
This is not the design we want, we will propose an alternative.

src/Microsoft.Identity.Web.TokenAcquisition/MicrosoftIdentityOptions.cs
pmaytak: Here: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/1b56a35f5add14c9bb439d50b3ae95368f5c107d/src/client/Microsoft.Identity.Client/AppConfig/ConfidentialClientApplicationOptions.cs#L46

pmaytak:
> This is not the design we want, we will propose an alternative.

Ok. Let me know how I can help.

bgavrilMS:
@jennyf19 @jmprieur - gentle reminder on this.

jennyf19:
> @jennyf19 @jmprieur - gentle reminder on this.

The implementation is not right, as JM mentioned. We won't have time to circle back on this until next week.

jmprieur:
As agreed with Bogdan, IdWeb will just respect MSAL's default

PR #2423, opened 2023-09-04
[OWIN] Avoid null reference when session is disabled

MatyRi:
# Avoid null reference when session is disabled

## Description

When sessions are disabled in Web.config, Session property is null and the AuthorizationCodeReceived handler throws exception. This change fallbacks to OwinContext.Environment dictionary, because the only use of session is to hand over Client Info from one handler to another in a single request.


jennyf19 (approved):
Thank you for the contribution @MatyRi

jmprieur (approved):
LGTM
Thanks @MatyRi

PR #2424, opened 2023-09-04
[OWIN] Avoid null reference in SecurityTokenValidated with B2C

MatyRi:
# Avoid null reference in SecurityTokenValidated with B2C

## Description

When using B2C, returned claims do not have `preferred_username` and therefore the `.Value` property throws an exception in the SecurityTokenValidated handler. This change fallbacks to just `name` claim if the original is not found and checks for empty value before adding new claim to the identity.

### DIFF
--- a/src/Microsoft.Identity.Web.OWIN/AppBuilderExtension.cs
+++ b/src/Microsoft.Identity.Web.OWIN/AppBuilderExtension.cs
@@ -177,8 +177,14 @@ namespace Microsoft.Identity.Web
                             httpContext.Session.Remove(ClaimConstants.ClientInfo);
                         }
 
-                        string name = context.AuthenticationTicket.Identity.FindFirst("preferred_username").Value;
-                        context.AuthenticationTicket.Identity.AddClaim(new Claim(ClaimTypes.Name, name, string.Empty));
+                        Claim nameClaim = context.AuthenticationTicket.Identity.FindFirst("preferred_username")
+                                          ?? context.AuthenticationTicket.Identity.FindFirst("name");
+
+                        if (!string.IsNullOrEmpty(nameClaim?.Value))
+                        {
+                            context.AuthenticationTicket.Identity.AddClaim(new Claim(ClaimTypes.Name, nameClaim.Value, string.Empty));
+                        }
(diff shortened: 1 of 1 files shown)


src/Microsoft.Identity.Web.OWIN/AppBuilderExtension.cs
```diff
@@ -177,8 +177,14 @@ public static class AppBuilderExtension
...
 
-                        string name = context.AuthenticationTicket.Identity.FindFirst("preferred_username").Value;
-                        context.AuthenticationTicket.Identity.AddClaim(new Claim(ClaimTypes.Name, name, string.Empty));
+                        Claim nameClaim = context.AuthenticationTicket.Identity.FindFirst("preferred_username")
+                                          ?? context.AuthenticationTicket.Identity.FindFirst("name");
+
+                        if (!string.IsNullOrEmpty(nameClaim?.Value))
+                        {
+                            context.AuthenticationTicket.Identity.AddClaim(new Claim(ClaimTypes.Name, nameClaim.Value, string.Empty));
+                        }
```
jennyf19, line 186: Not sure how this will impact MSAL's caching, as I believe, they were requiring `preferred_username`

jennyf19 (approved):
Thanks again!

jmprieur (approved):
LGTM
Thanks, @MatyRi 

I've the same question as @jennyf19: not sure that MSAL.NET will appreciate not having the preferred_username claim

MatyRi:
> LGTM Thanks, @MatyRi
> 
> I've the same question as @jennyf19: not sure that MSAL.NET will appreciate not having the preferred_username claim

I double checked and it seems MSAL fallbacks to UPN if preferred_username is not available. Although all the operations mentioning this claim are on the IdToken and not the Request/Owin Identity. 

In theory the claim is required, but on the other hand the code handles the missing value on multiple places with either UPN or "missing" string (example): https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/src/client/Microsoft.Identity.Client/TokenResponseHelper.cs#L34

Thank you for the approvals!

Issue #2426, opened 2023-09-04
DownstreamAPI does not indicate API failures unless trying to deserialize to an output object


MattCost:
### Microsoft.Identity.Web Library

Microsoft.Identity.Web

### Microsoft.Identity.Web version

2.11.1

### Web app

Sign-in users and call web APIs

### Web API

Protected web APIs call downstream web APIs

### Token cache serialization

Not Applicable

### Description

When using any of the convenience methods in `DownstreamApi.HttpMethods.cs`, if you do not try to deserialize the response into an object, then you will never get the return code from the API.

In the downstream api code, the only place `response.EnsureSuccessStatusCode()` is called is in the private `DeserializeOutput` method.

In order to call an api endpoint that only returns a status code one must call CallApiForUser, and manually set the http method, and then manually check for success on the response. 

It appears that all the convenience methods in HttpMethods are setup to catch the `HttpRequestException` the only thing missing is to call `EnsureSuccessStatusCode` on the response.

### Reproduction steps

Use `DeleteForUserAsync<TOutput>` to call any api that returns 404.
There is no indication that the call failed.

### Relevant code snippets

```csharp
public async Task<IActionResult> OnPostDeleteAsync()
        {
            Logger.LogTrace("Entering OnPostDeleteAsync");
            try
            {
                await API.DeleteForUserAsync("API", string.Empty, options =>{
                    options.RelativePath = $"RideEvents/{Id}";
                });
            }
            catch(Exception ex)
            {
                Logger.LogError(ex, "Exception trying to delete RideEvent Id {Id}", Id);
                PreviousPageAction = "RideEvent/Edit/OnPostDelete";
                PreviousPageErrorMessage = $"Error deleting RideEvent. {ex.Message}";
            }
            return RedirectToPage("Index");
        }
```

### Expected behavior

I expect API.DeleteForUserAsync to throw an HttpResponseException with details of the non-success error code the api returned.

DaleyKD:
Curious. I'm currently on 2.13.3, and when I use:

``` csharp
return _accountsApi.CallApiForAppAsync<AccountModel>(this.ServiceName, optionsAction, cancellationToken);
```

It throws an exception.

In fact, I'm looking and looking, trying to find the proper way to handle 404s (my REST API returns a 404 if the specific account isn't found by ID).

Am I missing the point of your post?  (Any ideas on how to handle 404s? 😄 )

MattCost:
[...]

PR #2427, opened 2023-09-04
api: Check all responses for success

MattCost:
Check all HttpResponses for success before returning to the caller.

Check all API Responses for success

## Description

Update HttpMethods to check the response for success before returning to the user.
Currently only calls that try to de-serialize were getting checked.

Issue #2426


jmprieur (commented):
- This should be done (once) in the DownstreamApi.HttpMethods.tt file, and then the code will be regenerated
- Isn't it a breaking change (today we return null?)

MattCost:
- I can move change to the template file, overlooked that.
- Yes, it is a breaking change. However it "breaks" in a good way (imo), by removing silent failures. If the methods that return output check for success, then all methods should check for success.

jmprieur (approved):
LGTM
Thanks @MattCost

PR #2431, opened 2023-09-05
change value of expected properties to 60

jennyf19:


jmprieur (approved):
LGTM
Thanks @jennyf19 

do we run this test for net6.0 and net7.0? (as if we do, I would expect a condition and a value 60

PR #2438, opened 2023-09-05
Introduces a lock while loading credentials from Credential Source

sruke:
# Introduces a lock while loading credentials from Credential Source

Summary of the changes
Ensure multiple requests don't attempt to load the same credentialDescription concurrently.

## Description
When a customer configures the source of a token decryption certificate in the inbound policy, SAL attempts to load the certificate from the specified source or retrieves the cached certificate if it has been loaded previously. SAL then creates an X509SecurityKey from the certificate and adds it to InboundPolicy.TokenValidationParameters.TokenDecryptionKeys if the key was not already added. This logic is executed for every incoming token validation request, and SAL utilizes ID.Web's implementation of ICredentialLoader to retrieve the certificate.

Until the certificate is cached by ID.Web (i.e., when credentialDescription.CachedValue is set), multiple parallel requests (potentially exceeding 100 incoming requests, as indicated by Keegan) may attempt to load the same certificate from the source. This raises performance concerns, particularly when fetching the certificate involves making an HTTP request (e.g., fetching from KeyVault).

Open Questions:
1. How does ID.Web detect when a certificate has been rotated?
2. Is the ResetCredentials API invoked when a certificate rotation occurs?

Note: Unit tests will be added once the design is approved

### DIFF
--- a/src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
+++ b/src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
@@ -64,7 +64,7 @@ namespace Microsoft.Identity.Web
                 string key = GenerateKey(credentialDescription);
 
                 // Get or create a semaphore for this key
-                SemaphoreSlim semaphore;
+                SemaphoreSlim? semaphore;
                 lock (_semaphoreLock)
                 {
                     if (!_loadingSemaphores.TryGetValue(key, out semaphore))
--- a/src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
+++ b/src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
@@ -3,6 +3,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Threading;
 using System.Threading.Tasks;
 using Microsoft.Extensions.Logging;
 using Microsoft.Identity.Abstractions;
@@ -15,6 +16,9 @@ namespace Microsoft.Identity.Web
     public class DefaultCredentialsLoader : ICredentialsLoader
     {
         ILogger<DefaultCredentialsLoader>? _logger;
+        
+        private readonly Dictionary<string, SemaphoreSlim> _loadingSemaphores = new Dictionary<string, SemaphoreSlim>();
+        private readonly object _semaphoreLock = new object();
 
         /// <summary>
         /// Constructor with a logger
@@ -56,11 +60,45 @@ namespace Microsoft.Identity.Web
 
             if (credentialDescription.CachedValue == null)
             {
-                if (CredentialSourceLoaders.TryGetValue(credentialDescription.SourceType, out ICredentialSourceLoader? loader))
+                // Generate a unique key for the credentialDescription
+                string key = GenerateKey(credentialDescription);
+
+                // Get or create a semaphore for this key
+                SemaphoreSlim semaphore;
+                lock (_semaphoreLock)
                 {
-                    await loader.LoadIfNeededAsync(credentialDescription, parameters);
+                    if (!_loadingSemaphores.TryGetValue(key, out semaphore))
+                    {
+                        semaphore = new SemaphoreSlim(1);
+                        _loadingSemaphores[key] = semaphore;
+                    }
+                }
+
+                // Wait to acquire the semaphore
+                await semaphore.WaitAsync();
+
+                try
+                {
+                    if (credentialDescription.CachedValue == null)
+                    {
+                        if (CredentialSourceLoaders.TryGetValue(credentialDescription.SourceType, out ICredentialSourceLoader? loader))
+                        {
+                            await loader.LoadIfNeededAsync(credentialDescription, parameters);
+                        }
+                    }
+                }
+                finally
+                {
+                    // Release the semaphore
+                    semaphore.Release();
                 }
             }
+        }
+
+        private string GenerateKey(CredentialDescription credentialDescription)
+        {
+            // Generate a unique key for the credentialDescription
+            return $"{credentialDescription.SourceType}_{credentialDescription.ReferenceOrValue}";
         }
 
         /// <inheritdoc/>
--- a/src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
+++ b/src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
-using System;
 using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Threading;
@@ -79,12 +78,6 @@ namespace Microsoft.Identity.Web
                     semaphore.Release();
                 }
             }
-        }
-
-        private string GenerateKey(CredentialDescription credentialDescription)
-        {
-            // Generate a unique key for the credentialDescription
-            return $"{credentialDescription.SourceType}_{credentialDescription.ReferenceOrValue}";
         }
 
         /// <inheritdoc/>
--- a/src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
+++ b/src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
@@ -59,7 +59,7 @@ namespace Microsoft.Identity.Web
             if (credentialDescription.CachedValue == null)
             {
                 // Get or create a semaphore for this credentialDescription
-                var semaphore = _loadingSemaphores.GetOrAdd(credentialDescription.Id, new SemaphoreSlim(1));
+                var semaphore = _loadingSemaphores.GetOrAdd(credentialDescription.Id, (v) => new SemaphoreSlim(1));
 
                 // Wait to acquire the semaphore
                 await semaphore.WaitAsync();
--- a/src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
+++ b/src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT License.
 
 using System;
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Threading;
 using System.Threading.Tasks;
@@ -16,9 +17,7 @@ namespace Microsoft.Identity.Web
     public class DefaultCredentialsLoader : ICredentialsLoader
     {
         ILogger<DefaultCredentialsLoader>? _logger;
-        
-        private readonly Dictionary<string, SemaphoreSlim> _loadingSemaphores = new Dictionary<string, SemaphoreSlim>();
-        private readonly object _semaphoreLock = new object();
+        private readonly ConcurrentDictionary<string, SemaphoreSlim> _loadingSemaphores = new ConcurrentDictionary<string, SemaphoreSlim>();
 
         /// <summary>
         /// Constructor with a logger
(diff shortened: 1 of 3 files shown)


jmprieur:
@sruke:
> How does ID.Web detect when a certificate has been rotated?
IdWeb detects when client certificates were rotated because AAD returns an explicit error

> Is the ResetCredentials API invoked when a certificate rotation occurs?
On this error, IDWeb calls ResetCredentials, and retrys the call, which triggers reloading the certificate. This happens here:
https://github.com/AzureAD/microsoft-identity-web/blob/cc6fc991ba69acbe3be05f5364baff83feeda34e/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs#L172-L178

IdWeb does not do anything with decrypt certificate, as the guidance (with IdWeb) is for service owners to provide several decyrpt credentials, and Wilson tries them all.

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
```diff
@@ -15,6 +16,9 @@ namespace Microsoft.Identity.Web
     public class DefaultCredentialsLoader : ICredentialsLoader
     {
         ILogger<DefaultCredentialsLoader>? _logger;
+        
+        private readonly Dictionary<string, SemaphoreSlim> _loadingSemaphores = new Dictionary<string, SemaphoreSlim>();
+        private readonly object _semaphoreLock = new object();
```
jmprieur, line 21: can't we just do a double lock on the credentialDescription?

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
```diff
@@ -56,13 +60,47 @@ public async Task LoadCredentialsIfNeededAsync(CredentialDescription credentialD
 
             if (credentialDescription.CachedValue == null)
             {
-                if (CredentialSourceLoaders.TryGetValue(credentialDescription.SourceType, out ICredentialSourceLoader? loader))
+                // Generate a unique key for the credentialDescription
+                string key = GenerateKey(credentialDescription);
```
brentschmaltz, line 64: Will the key change?
If not, should it be part of the CredentialDescription so we don't have to create it every time?
We can avoid creating a string for every validation.

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
```diff
@@ -15,6 +16,9 @@ namespace Microsoft.Identity.Web
     public class DefaultCredentialsLoader : ICredentialsLoader
     {
         ILogger<DefaultCredentialsLoader>? _logger;
+        
+        private readonly Dictionary<string, SemaphoreSlim> _loadingSemaphores = new Dictionary<string, SemaphoreSlim>();
```
brentschmaltz, line 20: How does the Dictionary<string, SemaphoreSlim> help us?

sruke: A semaphore is created per CredentialDescription. A thread can safely acquire a semaphore for a specific CredentialDescription. Other threads are free to acquire a semaphore to load the certificate for other CredentialDescriptions. Aim here is to minimize thread blocking.

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
sruke: The Key will not change and can be part of the CredentialDescription class to avoid unnecessary string creation on every request.

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
sruke: Locks are easy to easy but will only allow one request to access a section of the code. Main reason for using a SephamoreSlim over a lock is that while we want only a single request to load a certificate for a specific CredentialDescription we also want to allow other requests to be able to access/load certificates for other CredentialDescriptions.

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
jmprieur: @sruke @brentschmaltz @jennyf19 
What about the following?

```diff
public async Task LoadCredentialsIfNeededAsync(CredentialDescription credentialDescription, CredentialSourceLoaderParameters? parameters = null)
        {
            _ = Throws.IfNull(credentialDescription);

              if (credentialDescription.CachedValue == null)
              {
+             lock(credentialDescription)
+             {
+               if (credentialDescription.CachedValue == null)
+               {
                  if (CredentialSourceLoaders.TryGetValue(credentialDescription.SourceType, 
                                                        out ICredentialSourceLoader? loader))
                  {
                     await loader.LoadIfNeededAsync(credentialDescription, parameters);
                  }
+               }
+            }
            }
        }
```

jennyf19: prefer #2, as we should lock on as little code as possible. 
do we need to check this twice? `credentialDescription.CachedValue == null`

sruke: Checking credentialDescription.CachedValue == null a second time is important. While a request was waiting to acquire a lock to load CertificateA, another request ahead of it could have loaded the certificate, cached it in credentialDescription.CachedValue and released the lock. By checking the value again we're utilizing the cache.

sruke: Like the simplified approach @jmprieur. 

Only concern is that requests that need to work with different CredentialsDescriptions will be blocked waiting for a lock. For example, RequestA acquired a lock to load CertificateA. RequestB needs to load CertificateB and could proceed in parallel but it needs to wait for RequestA.

A SemaphoneSlim would provide more granular locking and could improve performance further. However, if this is not a big concern then the second approach LGTM.

jmprieur: I'm not following, @sruke : the lock is on a given CredentialDescription? not on the collection?
how could it interfere between CredentialDescription ?

sruke: Ah, you're locking in the CredentialDescription object itself. Got it. LGTM, I have no concerns.

jennyf19: what about something like this:

```csharp
private readonly SemaphoreSlim _semaphore = new SemaphoreSlim(1, 1);

public async Task LoadCredentialsIfNeededAsync(CredentialDescription credentialDescription, CredentialSourceLoaderParameters? parameters = null)
{
    _ = Throws.IfNull(credentialDescription);

    await _semaphore.WaitAsync();

    try
    {
        if (credentialDescription.CachedValue == null)
        {
            if (CredentialSourceLoaders.TryGetValue(credentialDescription.SourceType, out ICredentialSourceLoader? loader))
            {
                await loader.LoadIfNeededAsync(credentialDescription, parameters);
            }
        }
    }
    finally
    {
        _semaphore.Release();
    }
}
```

you can add a lock in an async method. The `lock` statement is a synchronous construct, but it can be used in an async method as well. When you use the `lock` statement in an async method, the method will still be async, but the code block inside the `lock` statement will be executed synchronously, ensuring that only one thread can execute it at a time. Probably better to use the semaphore in this case, but not sure of the proposed implementation.

sruke: The implementation in this draft PR is a variation of what is being suggested here. The only difference is that a semaphore will be maintained for each CredentialDescription to concurrently allow requests to load different certificates and reduce the number of blocked threads.

jennyf19: Ok, how about a ConcurrentDictionary then, which allows for thread-safe access without the need for a lock, which can help w/perf and reduce contention. If we can use `GetOrAdd` to an existing semaphore that can help reduce the number of semaphores that need to be created and managed.

for example:
```csharp
private readonly ConcurrentDictionary<string, SemaphoreSlim> _loadingSemaphores = new ConcurrentDictionary<string, SemaphoreSlim>();

....
        // Generate a unique key for the credentialDescription
        string key = GenerateKey(credentialDescription);

        // Get or create a semaphore for this key
        SemaphoreSlim semaphore = _loadingSemaphores.GetOrAdd(key, new SemaphoreSlim(1));

        // Wait to acquire the semaphore
        await semaphore.WaitAsync();
try
{
....
}
finally
        {
            // Release the semaphore
            semaphore.Release();
        }
....
```

TimHannMSFT: `ConcurrentDictionary` makes heavy use of locking internally (this seemed to be a good write up when I skimmed: https://www.red-gate.com/simple-talk/blogs/inside-the-concurrent-collections-concurrentdictionary/)

The scenario would really matter here to say which would be more performant. As a general rule, seems like something like Benchmark.net would be the best tool to answer this question. Even still, to the extent I'm tracking the conversation, `ConcurrentDictionary` seems like a good option given its semantics can be clearer than rolling our own solution based on semaphores. I would imagine the performance improvements granted by creating something bespoke for our purposes would not overwhelm the clearer syntax. Just my 2c.

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
sruke: Simplified the implementation using a ConcurrentDictionary and a SemaphoreSlim.

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
sruke: Thank you for the hint Jenny. Simplified the implementation using a ConcurrentDictionary and a SemaphoreSlim.

jennyf19 (approved):

tests/IntegrationTests/TokenAcquirerTests/TokenAcquirer.cs
```diff
@@ -126,6 +127,46 @@ public async Task AcquireToken_WithFactoryAndAuthorityClientIdCert_ClientCredent
             Assert.False(string.IsNullOrEmpty(result.AccessToken));
         }
 
+        [IgnoreOnAzureDevopsFact]
+        //[Fact]
```
westin-m, line 131: Should this comment be kept if the other attribute is being used?

westin-m (approved):

jmprieur (commented):
LGTM
Thanks @sruke 

I proposed a small tweak.

jmprieur (approved):
LGTM
thanks @sruke 

I proposed a small tweak

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
```diff
@@ -56,9 +58,24 @@ public async Task LoadCredentialsIfNeededAsync(CredentialDescription credentialD
 
             if (credentialDescription.CachedValue == null)
             {
-                if (CredentialSourceLoaders.TryGetValue(credentialDescription.SourceType, out ICredentialSourceLoader? loader))
+                // Get or create a semaphore for this credentialDescription
+                var semaphore = _loadingSemaphores.GetOrAdd(credentialDescription.Id, new SemaphoreSlim(1));
```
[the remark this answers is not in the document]
sruke: Updated.

tests/IntegrationTests/TokenAcquirerTests/TokenAcquirer.cs
sruke: Kept the pattern followed by other tests in the file.

Issue #2444, opened 2023-09-06
[Feature Request] When ESTS sends a 401 because a certificate is revoked due to a rotation, attempt to reload once (like when the certificate has expired)


jmprieur:
**Is your feature request related to a problem? Please describe.**
Today, Microsoft identity web detects that a certificate has an invalid key or has expired (`Constants.InvalidKeyError` or `Constants.SignedAssertionInvalidTimeRange`), and retries to load the certificate from the certificate description.

See:
https://github.com/AzureAD/microsoft-identity-web/blob/767331ff44c30d467d6c3f932d5afd8838d21243/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs#L546-L556

with:
https://github.com/AzureAD/microsoft-identity-web/blob/767331ff44c30d467d6c3f932d5afd8838d21243/src/Microsoft.Identity.Web.TokenAcquisition/Constants.cs#L130-L131

It does not look at the "certificate revoked" event ()

**Describe the solution you'd like**
Amend the `IsInvalidClientCertificateOrSignedAssertionError` method to add a test for AADSTS7000214: Certificate has been revoked.

jennyf19:
https://github.com/AzureAD/microsoft-identity-web/releases/tag/2.13.4

PR #2450, opened 2023-09-07
Add revoked cert condition

westin-m:
Addresses #2444


westin-m:
#2444

jennyf19 (approved):

jmprieur (approved):
LGTM
Thanks, @westin-m 

I checked other possible error codes in https://learn.microsoft.com/en-us/azure/active-directory/develop/reference-error-codes, but didn't find any other that would apply to client certificates (there are a lot of certificate usage in ESTS)

PR #2464, opened 2023-09-11
Add a test for Jwt bearer options length

westin-m:


jmprieur (approved):
LGTM
Thanks @westin-m

jennyf19 (approved):
:shipit:

jmprieur (approved):
LGTM
Thanks @westin-m 

We also need to fix anything here? 
https://github.com/AzureAD/microsoft-identity-web/blob/1c65d4a36e2155af04a4fa3b8a0a530f5aa3c121/src/Microsoft.Identity.Web.TokenAcquisition/AspNetCore/JwtBearerOptionsMerger.cs#L4

What are the new properties? Do they need to be in MergedOptions (I would think not, but it would be great to confirm)

westin-m:
> What are the new properties?

The same new ones in OIDC; TokenHandlers, UseSecurityTokenValidators, and TimeProvider.

PR #2496, opened 2023-10-02
Jmprieur/test cert rotation

### LINKED ISSUE #2458: [Feature Request] Provide an API to query the certificate that is currently in use for authentication
There is currently no way to programmatically query the certificate (or list of certificates) that is being used to authenticate with AAD

An API to determine: 
* the certificate that is currently in use for authentication,
* if a collection of certificates was provided, the list of certificates that will be considered by the id-web SDK
* (potentially) a callback that would allow us to observe changes when the certificate used has changed 

We have considered using using logs for observability purposes, but it is not the same as being able to query directly in code. 
This does not solve the case when the caller needs to pass in or otherwise understand the specific certificate that is being used (e.g., in the case of a list of certs being passed in, but the caller knows that a cert got revoked and needs to act)

**Additional context**
This issue was found because we were trying to use a revoked certificate to access AAD via id-web. This was a gap in our code but we also discovered there was no great way for us to identify the certificate we were using programmatically and/or force a reload of a certificate if needed aside from reinitializing the id-web provider

jmprieur:
# Improving client certificate rotation

## Description

- Adding a ResetCertificates with an override for an enumeration of CredentialDescription
- Adding a way for the apps to observe that the certs are selected or unselected: fixes #2458
- Fix an issue in the rotation of client certificates.
- Adds a test that create a daemon app registration (if not already there). Adds 2 certs (one that expires in 3mins, and one that is valid from in now+2 mins and expires in 10 mins, and add them to the app registration, We wait a few seconds (because the app registration is not immediate)

Then we acquire tokens for 5 mins. After 3 mins, IdWeb switches from the first certificate to the second.

image

### DIFF
--- a/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
@@ -691,8 +691,7 @@ namespace Microsoft.Identity.Web
                     {
                         Action = action,
                         Certificate = app.AppConfig.ClientCredentialCertificate,
-                        CredentialDescription = mergedOptions.ClientCredentials?.FirstOrDefault(c =>
-                            c.Certificate == selectedCertificate)
+                        CredentialDescription = mergedOptions.ClientCredentials?.FirstOrDefault(c => c.Certificate == selectedCertificate)
                     });
                 ;
             }
--- a/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
@@ -10,8 +10,10 @@ using System.IdentityModel.Tokens.Jwt;
 using System.Linq;
 using System.Net.Http;
 using System.Security.Claims;
+using System.Security.Cryptography.X509Certificates;
 using System.Threading;
 using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Microsoft.Identity.Abstractions;
 using Microsoft.Identity.Client;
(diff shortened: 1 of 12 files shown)


jennyf19 (approved):
:shipit:

tests/IntegrationTests/TokenAcquirerTests/TokenAcquirer.cs
```diff
@@ -287,12 +287,10 @@ private static async Task CreateGraphClientAndAssert(TokenAcquirerFactory tokenA
             var serviceProvider = tokenAcquirerFactory.Build();
             GraphServiceClient graphServiceClient = serviceProvider.GetRequiredService<GraphServiceClient>();
             var users = await graphServiceClient.Users
-                .Request()
-                .WithAppOnly()
-                .WithAuthenticationScheme(s_optionName)
-                //     .WithAuthenticationOptions(options => options.ProtocolScheme = "Pop")
-                .GetAsync();
-            Assert.True(users.Count >= 56);
+                .GetAsync(o => o.Options
```
jmprieur, line 290: This needed to change as TokenAcquirerTests need Graph 5X for the CertificateRotationTests, and therefore
this test CreateGraphClientAndAssert now leverages Microsoft.Identity.Web.GraphServiceClient and no longer
Microsoft.Identity.web.MicrosoftGraph

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -654,6 +674,30 @@ private IConfidentialClientApplication BuildConfidentialClientApplication(Merged
...
+            {
+                _certificatesObserver.OnClientCertificateChanged(
+                    new CertificateChangeEventArg()
+                    {
+                        Action = action,
+                        Certificate = app.AppConfig.ClientCredentialCertificate,
+                        CredentialDescription = mergedOptions.ClientCredentials?.FirstOrDefault(c =>
+                            c.Certificate == selectedCertificate)
+                    });
+                ;
```
jennyf19, line 697: nit, can this be moved up?

PR #2499, opened 2023-10-03
Fixes #2410, #2410

### LINKED ISSUE #2410: Error GetAccessTokenForAppAsync Exception thrown: 'System.NullReferenceException' in System.Private.CoreLib.dll

Microsoft.Identity.Web

2.13.3

### Web app

Sign-in application and call web APIs

### Web API

Protected web APIs call downstream web APIs

### Token cache serialization

In-memory caches

### Description

Iam building a console app in dotnet core that consume a protected webapi.  Since I upgraded to 2.13.3 tokenresult thrown an exception "Exception thrown: 'System.NullReferenceException' in System.Private.CoreLib.dll" Object reference not set to an instance of an object.

```csharp
public async Task<string> GetAuthenticationForClientResultAsync(string appId)
    {
        try
        {
            var tokenresult = await _tokenAcquisition.GetAccessTokenForAppAsync(
                $"{appId}/.default"
            );

            return tokenresult;
        }
        catch (Exception ex)
        {
            _consentHandler.HandleException(ex);
        }

        return null;
    }
```

### Reproduction steps

Here is how Iam starting the console app with dependancy injection.  I think it might be the cause of my issue but I dont know enough to explain why.

```csharp
using gcb_libs.Services;
using gcb_taskscheduler;
using gcb_taskschudler.Jobs;
using Microsoft.AspNetCore.Authentication.JwtBearer;
using Microsoft.Extensions.Configuration;
using Microsoft.Extensions.DependencyInjection;
using Microsoft.Extensions.Hosting;
using Microsoft.Identity.Web;
using Quartz;
using System.IO;

var configuration = new ConfigurationBuilder()
    .SetBasePath(Directory.GetCurrentDirectory())
    .AddJsonFile("appsettings.json", true, false)
    .AddEnvironmentVariables()
    .AddUserSecrets<Program>()
    .Build();


var host = Host.CreateDefaultBuilder()
    .ConfigureServices(
        (context, services) =>
        {
            services.AddSingleton<IConfiguration>(configuration);
            services
                .AddAuthentication(JwtBearerDefaults.AuthenticationScheme)
                .AddMicrosoftIdentityWebApp(configuration.GetSection("AzureAd"))
                .EnableTokenAcquisitionToCallDownstreamApi()
                .AddInMemoryTokenCaches();
            services.AddMicrosoftIdentityConsentHandler();
            services.AddScoped<AzureAuthTokenService>();
            services.AddScoped<LogService>();
            services.AddScoped<ScheduledJobService>();
            services.AddHostedService<WorkerService>();
        }
    )
    .Build();

host.Run();
```

### Error message

{"Object reference not set to an instance of an object."}

### Id Web logs

_No response_

### Relevant code snippets

```csharp
tokenresult is always null. 

  var tokenresult = await _tokenAcquisition.GetAccessTokenForAppAsync(
                $"{appId}/.default"
            );
```

### Regression

2.13.2

### Expected behavior

It should return the token

jmprieur:
# Fixes #2410, #2410
Based on customer feedback, change the condition by which a host is used rather than another (DefaultTokenAcquisitionHost vs TokenAcquisitionAspnetCoreHost)

### DIFF
--- a/tests/Microsoft.Identity.Web.Test/TestTokenAcquisitionHost.cs
+++ b/tests/Microsoft.Identity.Web.Test/TestTokenAcquisitionHost.cs
@@ -1,0 +1,51 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Builder;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+using Xunit;
+
+namespace Microsoft.Identity.Web.Test
+{
+    public class TestTokenAcquisitionHost
+    {
+        [Fact]
+        public void TestTokenAcquisitionHostNotAspNetCore()
+        {
+            // When not adding Services.AddAuthentication, the host should be 
+            TokenAcquirerFactory tokenAcquirerFactory = TokenAcquirerFactory.GetDefaultInstance();
+            var host = tokenAcquirerFactory.Services.First(s => s.ServiceType.Name == "Microsoft.Identity.Web.ITokenAcquisitionHost");
+            Assert.True(host.ImplementationType!.FullName == "Microsoft.Identity.Web.Hosts.DefaultTokenAcquisitionHost");
+        }
+
+        [Fact]
+        public void TestTokenAcquisitionAspNetCoreBuilderNoAuth()
+        {
+            // When not adding Services.AddAuthentication, the host should be 
+            var builder = WebApplication.CreateBuilder();
+            builder.Services.AddTokenAcquisition();
+            var host = builder.Services.First(s => s.ServiceType.Name == "ITokenAcquisitionHost");
+            Assert.True(host.ImplementationType!.FullName == "Microsoft.Identity.Web.Hosts.DefaultTokenAcquisitionHost");
+        }
+
+
+        [Fact]
+        public void TestTokenAcquisitionAspNetCoreBuilderAuth()
+        {
+            // When not adding Services.AddAuthentication, the host should be 
+            var builder = WebApplication.CreateBuilder();
+            builder.Services.AddAuthentication()
+                .AddMicrosoftIdentityWebApi(builder.Configuration)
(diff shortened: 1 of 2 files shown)


MZOLN (approved):
LGTM

westin-m (approved):

tests/Microsoft.Identity.Web.Test/TestTokenAcquisitionHost.cs
```diff
@@ -0,0 +1,51 @@
...
+namespace Microsoft.Identity.Web.Test
+{
+    public class TestTokenAcquisitionHost
+    {
+        [Fact]
+        public void TestTokenAcquisitionHostNotAspNetCore()
+        {
+            // When not adding Services.AddAuthentication, the host should be 
+            TokenAcquirerFactory tokenAcquirerFactory = TokenAcquirerFactory.GetDefaultInstance();
+            var host = tokenAcquirerFactory.Services.First(s => s.ServiceType.Name == "Microsoft.Identity.Web.ITokenAcquisitionHost");
```
westin-m, line 23:
```suggestion
            var host = tokenAcquirerFactory.Services.First(s => s.ServiceType.Name == "ITokenAcquisitionHost");
```

westin-m: this test fails on .net6, I have tested that this change fixes it

JoshLozensky (changes requested):
It looks like the build is failing, is that expected?

tests/Microsoft.Identity.Web.Test/TestTokenAcquisitionHost.cs
jmprieur: yes. it needs to use Fullname. my bad

jennyf19 (approved):
nice! let's go! :shipit:

tests/Microsoft.Identity.Web.Test/TestTokenAcquisitionHost.cs
jmprieur: Your fix works too, @westin-m, but there could be ITokenAcquisitionHost in several namespaces (purely hypothetical, though)

kellyyangsong (approved):

JoshLozensky (approved):
build works now

PR #2555, opened 2023-10-26
update bug report with default tags

jennyf19:
using `untriaged,needs attention`


jmprieur (approved):
LGTM
Thanks @jennyf19

PR #2556, opened 2023-10-27
Reorganize UI tests and minor code updates

JoshLozensky:
# Reorganize UI tests

## Description

- Merged UI tests into one project
- Changed inline OS check to instead use the built-in SupportedOSPlatform class
- Added a delay between launching service and client to solve transient issue with client crashing
- Modified path strings to use Path.DirectorySeparatorChar rather than hardcoding it


tests/IntegrationTests/WebAppUiTests/TestingWebAppCallsApiCallsGraphLocally.cs
```diff
@@ -1,139 +1,145 @@
...
+using Microsoft.Identity.Lab.Api;
+using Microsoft.Playwright;
+using Xunit;
+using Xunit.Abstractions;
+using Process = System.Diagnostics.Process;
+
+namespace WebAppUiTests
+#if !FROM_GITHUB_ACTION
+{
+    [CollectionDefinition(nameof(TestingWebAppCallsApiCallsGraphLocally), DisableParallelization = true)] // since this changes environment variables we'd prefer it not run at the same time as other tests
```
westin-m, line 19: Nit: stay consistent with putting comments on their own line, directly above the one to which they refer

jennyf19 (approved):
:shipit:

tests/IntegrationTests/WebAppUiTests/UiTestHelpers.cs
```diff
@@ -235,27 +236,17 @@ public static void KillProcessTrees(Queue<Process> processQueue)
...
-            }
-            else
-            {
-                throw new NotImplementedException("Not implemented for this OS");
-            }
+            ManagementObjectSearcher processSearch = new ManagementObjectSearcher($"Select * From Win32_Process Where ParentProcessID={process.Id}");
+            IList<Process> processList = processSearch.Get()
+                .Cast<ManagementObject>()
+                .Select(mo =>
+                    Process.GetProcessById(Convert.ToInt32(mo["ProcessID"], System.Globalization.CultureInfo.InvariantCulture)))
```
jmprieur, line 246: You might want to do that outside the loop?

jmprieur (approved):
LGTM
Thanks @JoshLozensky

tests/IntegrationTests/WebAppUiTests/UiTestHelpers.cs
JoshLozensky: It is done in the loop because as far as I can tell grandchild processes would have the spawning child process as the ParentProcessID rather than the root parent which is why we need to search each child processes children as well, or am I missing something?

tests/IntegrationTests/WebAppUiTests/TestingWebAppCallsApiCallsGraphLocally.cs
JoshLozensky: altrered

PR #2567, opened 2023-11-03
Make ConfidentialClientApplicationBuilderExtension.WithClientCredentials async

### LINKED ISSUE #2566: Hang in ConfidentialClientApplicationBuilderExtension.WithClientCredentials

Microsoft.Identity.Web.TokenAcquisition

2.15.3

### Web app

Sign-in users and call web APIs

### Web API

Protected web APIs (validating tokens)

### Token cache serialization

In-memory caches

### Description

In production the project that I'm currently working on, has experienced hangs on the sign-in proces of our Blazor Server website. We couldn't pinpoint the error in production, but I've found similar behaviour after starting the website locally a couple of times.
The code is hanging in `ConfidentialClientApplicationBuilderExtension.WithClientCredentials` on the line with 
```
credentialsLoader.LoadCredentialsIfNeededAsync(credential, credentialSourceLoaderParameters).GetAwaiter().GetResult();
```
I've looked at the implementation of the potential credentials-loaders and found that the `KeyVault`-loader requires an async-path (maybe others as well).

I'm preparing a PR to make `ConfidentialClientApplicationBuilderExtension.WithClientCredentialsAsync` and the path upwards also `async`.

While making my change in the code I came across an obsolete message on `ITokenAcquisition.ReplyForbiddenWithWwwAuthenticateHeader` which pointed to [Performance improvements and consequence on async methods in Microsoft.Identity.Web 1.9.0](https://github.com/AzureAD/microsoft-identity-web/wiki/1.9.0#performance-improvements-and-consequence-on-async-methods). So I suspect this bug has been in the code-base for a while.

### Reproduction steps

Sorry, I'm not able to reproduce this consistently.

### Error message

No error message, the application just halts and the browser shows a message that the server took to long to respond after a while.

### Id Web logs

_No response_

### Relevant code snippets

```csharp
Sorry, I can't share the code of this application.
```

### Regression

_No response_

### Expected behavior

The call to `ConfidentialClientApplicationBuilderExtension.WithClientCredentials` to be fully async.

xs4free:
Make calls to `ConfidentialClientApplicationBuilderExtension.WithClientCredentials` fully async.

## Description
Instead of using `GetAwaiter().GetResult()` in the code, I've added an overload `WithClientCredentialsAsync` which calls all credential-loaders asynchronously. This probably fixes our bug in production which hangs our Blazor Server during login.

Fixes #2566


xs4free:

src/Microsoft.Identity.Web.TokenAcquisition/ITokenAcquisition.cs
```diff
@@ -279,11 +281,26 @@ public interface ITokenAcquisition
...
+        /// Used in web APIs (which therefore cannot have an interaction with the user).
+        /// Replies to the client through the HttpResponse by sending a 403 (forbidden) and populating wwwAuthenticateHeaders so that
+        /// the client can trigger an interaction with the user so the user can consent to more scopes.
+        /// </summary>
+        /// <param name="scopes">Scopes to consent to.</param>
+        /// <param name="msalServiceException"><see cref="MsalUiRequiredException"/> triggering the challenge.</param>
+        /// <param name="authenticationScheme">Authentication scheme. If null, will use OpenIdConnectDefault.AuthenticationScheme
+        /// if called from a web app, and JwtBearerDefault.AuthenticationScheme if called from a web API.</param>
+        /// <param name="httpResponse">The <see cref="HttpResponse"/> to update.</param>
+        Task ReplyForbiddenWithWwwAuthenticateHeaderAsync(
```
keegan-caruso, line 299: This is a breaking change in a public interface: https://learn.microsoft.com/en-us/dotnet/standard/library-guidance/breaking-changes

Can you elaborate on why this is needed?

jmprieur: I love the fact that the credential loaders can be called async.
For the WWAuthenticate, indeed, how is it related?

xs4free: Of coarse I can elaborate. The method `ReplyForbiddenWithWwwAuthenticateHeader` uses the internal method `GetOrBuildConfidentialClientApplication`, which calls the private method `BuildConfidentialClientApplication` which calls `builder.WithClientCredentials`. Because I made the last method async, I also think the path with `ReplyForbiddenWithWwwAuthenticateHeader` should be made async. 

The only thing about this change that is breaking is that I removed the `[Obsolete]` attribute from the `ITokenAcquisition.ReplyForbiddenWithWwwAuthenticateHeaderAsync` since that should be the method used instead of the `ITokenAcquisition.ReplyForbiddenWithWwwAuthenticateHeader`. I've kept the sync version, added `[Obsolete]` as a warning and implemented it using a `GetAwaiter().GetResult()`. The consumers of this Interface initially don't need to change anything and just receive a warning the sync-method is obsolete.

If you think I should solve this (`ReplyForbiddenWithWwwAuthenticateHeader`) another way, I would love any input and will gladly implement/change that in my PR.

keegan-caruso: This change adds a method to the interface, ReplyForbiddenWithWwwAuthenticateHeaderAsync that also takes an authenticationScheme, that is a breaking change for us. From the doc I linked: 

```txt
Adding anything to an interface will cause existing types that implement it to fail.
```

xs4free: Thanks for explaining @keegan-caruso. I'm not used to these requirements, because I mainly work on internal software, but thinking it through I understand your perspective. Sorry for not picking it up right after your first message.

I've added another commit to my PR where I remove the changes in the `ITokenAcquisition` interface.
This introduces a new `GetAwaiter().GetResult()` in the implementation of the sync version of `ReplyForbiddenWithWwwAuthenticateHeader`, but users already have the option to use the async version: `ReplyForbiddenWithWwwAuthenticateHeaderAsync`.

What do you prefer I do with the `[Obsolete]`-attributes? Are they correct the way the are now in my PR? Or do you want to redirect users using the sync version `ReplyForbiddenWithWwwAuthenticateHeader` towards the async version `ReplyForbiddenWithWwwAuthenticateHeaderAsync`?

keegan-caruso: not a worry at all 😄 

We will work on adding the breaking api check to automation soon, it will be much clearer for contributors then.

I'm going to defer to @jmprieur on the API / obsolete questions

xs4free: Any update on this PR @keegan-caruso  or @jmprieur ? 

I've been running the code from this branch in production for over a month now, and haven't seen any blocked threads any more (while more users are using our Blazor Server-website). So I'm confident making this path async solved our problem.

jennyf19: We cannot take a breaking change without doing a major version, which we are not quite ready for. Are you willing to keep this branch up to date until we take a 3x (maybe May/June, but really cannot commit to anything right now), otherwise will try to think of another way, such as using an extension method.

xs4free: @jennyf19 thanks for your reply!
Could you please elaborate on what you consider a "breaking change" in my current PR?
I have the feeling that I've removed all things that are considered breaking in my latest commit (no interfaces are changed and no public classes are changed, only an internal class). 
The only question I still have is is removing an `Obsolete`-attribute considered breaking? I could also put that attribute back if this is considered breaking.
If needed I can update this branch until the next major release, but I'm hoping this isn't necessary.

src/Microsoft.Identity.Web.TokenAcquisition/ConfidentialClientApplicationBuilderExtension.cs
```diff
@@ -12,12 +13,24 @@ namespace Microsoft.Identity.Web
...
     {
+        [Obsolete(IDWebErrorMessage.WithClientCredentialsIsObsolete, false)]
         public static ConfidentialClientApplicationBuilder WithClientCredentials(
             this ConfidentialClientApplicationBuilder builder,
             IEnumerable<CredentialDescription> clientCredentials,
             ILogger logger,
             ICredentialsLoader credentialsLoader,
             CredentialSourceLoaderParameters credentialSourceLoaderParameters)
+        {
+            return WithClientCredentialsAsync(builder, clientCredentials, logger, credentialsLoader,
```
jmprieur, line 24: do you think we could also run this one async?

jmprieur (approved):
LGTM
@jennyf19: I think we should take it

src/Microsoft.Identity.Web.TokenAcquisition/ITokenAcquisition.cs
jennyf19: Got it, thanks @xs4free . This won't make the 2.17.1 release (this week), but our next one. I'd like some time to review and will try to prioritize for that by end of week. Appreciate your patience on this.

jennyf19 (approved):
:shipit: 
Thanks for the contribution @xs4free

Issue #2551, opened 2023-10-24
[Feature Request] Improve MSI support in Id.Web


bgavrilMS:
This is a large feature consisting of several work items: 

- [x] Add direct MSI support in Id.Web, based on MSAL https://github.com/AzureAD/microsoft-identity-web/issues/1775 (note: Azure Identity should continue to be used for the MSI connection to AKV for loading certs)
- [x] Update FIC support (Workload Identity) to use MSAL in the first leg instead of Azure Identity (MSI and AKV are currently supported by Id.Web)
- [ ] Ensure MSI and FIC are CAE enabled https://github.com/AzureAD/microsoft-identity-web/issues/2550
- [ ] Update Workload Identity samples to use Id.Web instead of MSAL
- [x] Add a sample for MSI next to the client_credentials sample
- [x] Add a sample for certificatless next to the client_credentials sample (?)

bgavrilMS:
@jmprieur, @jennyf19 - I added here some notes about samples for MSI and FIC. I think we are all in agreement that MSI standalone sample is needed. Not sure how you feel about FIC, but I think we should have one (another chapter in client_credentials)

jennyf19:
@JoshLozensky  for the E2E test, would recommend creating a new test class under TokenAcquirerTests, these have similar E2E tests. You'll want to make sure the test only runs in ADO, because we'll be using the system assigned managed identity, so check this `IgnoreOnAzureDevopsFactAttribute` for an idea on what to do. If you look at the DaemonConsoleCallingDownstreamApi, you'll get an idea of how we can have a very simple test using `TokenAcquirerFactory`, but instead of using `IDownstreamApi` as the service, you will want to use `IAuthorizationHeaderProvider`. Then `api.CreateAuthorizationHeaderForAppAsync()` and use keyvault as the scopes, as when running in ADO, in the MwWilson1ESHostedPool, there is a user assigned managed identity: idwebvalidationbuild - which has access to keyvault, so we will get a token to keyvault. The scopes to use are `"https://vault.azure.net/.default"` and then just check that the result is not null.

jennyf19:
@bgavrilMS - can you give links to the specific samples you want updated:
`Update Workload Identity samples to use Id.Web instead of MSAL`
 `Add a sample for MSI next to the client_credentials sample`
 `Add a sample for certificatless next to the client_credentials sample (?)`
[...]

PR #2610, opened 2023-12-05
Add support for managed identities

### LINKED ISSUE #1775: [Feature Request] Managed Identity support in IDownstreamWebApi / IDownstreamApi
Can I also use Managed Identity (System or user-assigned) for `IDownstreamWebApi`? 
Shouldn't Managed Identity the first option that should be mentioned in the documentation for the users? My proposed order would be `Managed Identity`, `client certificate`, `client secret`.

Why do I first need to call `AddMicrosoftIdentityWebApiAuthentication(IConfiguration)` then `EnableTokenAcquisitionToCallDownstreamApi` and then `AddDownstreamWebApi`  in order to enable `IDownStreamWebApi` usage? In my case a web api (not protected) calls another web api (protected) and I would like to use Managed Identity, so there should be no further configuration needed. My expectation would be something like
```csharp
// For a system-assigned Managed Identity client id does not need to be passed
builder.Services.AddDownstreamWebApi(string managedIdentityClientId = null);
```

See also: https://github.com/MicrosoftDocs/azure-docs/issues/94620

JoshLozensky:
# Add support for managed identities
Added logic to default to managed identity credentials if available.

## Description

In TokenAquisition.GetAuthenticationResultForAppAsync() added a check for managed identity configuration in the tokenAquisitionOptions parameter. If present, default to this authentication method and use system-assigned or user-assigned managed identity per the user's configuration.

Fixes #1775 #2551


src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -358,10 +360,40 @@ private void LogAuthResult(AuthenticationResult? authenticationResult)
                 throw new ArgumentException(IDWebErrorMessage.ClientCredentialTenantShouldBeTenanted, nameof(tenant));
             }
 
+            // If using managed identity 
+            if (tokenAcquisitionOptions != null && tokenAcquisitionOptions.ManagedIdentity != null)
+            {
+                IManagedIdentityApplication miApplication;
+
+                if (tokenAcquisitionOptions.ManagedIdentity.ManagedIdentityType == ManagedIdentityType.UserAssigned &&
+                    !tokenAcquisitionOptions.ManagedIdentity.ClientId.IsNullOrEmpty()
```
jmprieur, line 369: Doesn't it provoke a null reference exception if ClientID is null? How does it work? is it an extension method?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -358,10 +360,40 @@ private void LogAuthResult(AuthenticationResult? authenticationResult)
                 throw new ArgumentException(IDWebErrorMessage.ClientCredentialTenantShouldBeTenanted, nameof(tenant));
             }
 
+            // If using managed identity 
+            if (tokenAcquisitionOptions != null && tokenAcquisitionOptions.ManagedIdentity != null)
+            {
+                IManagedIdentityApplication miApplication;
+
+                if (tokenAcquisitionOptions.ManagedIdentity.ManagedIdentityType == ManagedIdentityType.UserAssigned &&
```
jmprieur, line 368: do you see now why the ManagedIdentityType is not necessary?

tokenAcquisitionOptions.ManagedIdentity != null => it's a managed identity
- tokenAcquisitionOptions.ManagedIdentity.ClientId is not null => it's a user assigned managed identity
- tokenAcquisitionOptions.ManagedIdentity.ClientID is not => it's a system assigned managed identity

jmprieur (commented):
Thanks @JoshLozensky 
It's a good start, and I think you should test it as is.

For a productized version we might not want to re-create the managed identity application each time, but cache it like we for for the MSAL, otherwise there will never be going to be a cache.

@bgavrilMS to confirm. Thanks!

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
JoshLozensky: I think that is easier to get wrong since if you miss the documentation saying system-assigned needs to be null and put in the ClientID of your application i.e. the one that correlates to your system your request will fail which feels like an anti-pattern to me. It is unintuitive that putting in the ClientID of your system application would cause system-assigned managed identity to fail.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
JoshLozensky: The IsNullOrEmpty method is static and built into the String class, among others. It is designed to test for null values so it has no problem being called from a null value.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
jmprieur: To be clear, when I mentionned caching the managed identity application, we'd probably want to use a concurrent dictionary of managed identity by ClientID (the empty client ID representing the system-assigned managed identity)

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -585,10 +608,75 @@ private bool IsInvalidClientCertificateOrSignedAssertionError(MsalServiceExcepti
...
+        {
+            IManagedIdentityApplication? application;
+            string key = GetCacheKeyForManagedId(managedIdentityOptions);
+
+            // Lock the potential write of the dictionary to prevent multiple threads from creating the same application.
+            await _managedIdSemaphore.WaitAsync();
+            try
+            {
+                // Check if the application is already built and not null, if so return it
+                if (_managedIdentityApplicationsByClientId.TryGetValue(key, out application) && application != null)
```
keegan-caruso, line 631: not quite a standard double check lock, is it supposed to be?

If it is, check for the presence before trying to acquire the lock.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -585,10 +608,75 @@ private bool IsInvalidClientCertificateOrSignedAssertionError(MsalServiceExcepti
...
+                }
+    
+                // Build the application
+                application = BuildManagedIdentityApplication(
+                    managedIdentityId,
+                    mergedOptions.ConfidentialClientApplicationOptions.EnablePiiLogging
+                );
+                // Add the application to the cache
+                _managedIdentityApplicationsByClientId.TryAdd(key, application);
+            } finally
```
keegan-caruso, line 654: spacing

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -585,10 +608,75 @@ private bool IsInvalidClientCertificateOrSignedAssertionError(MsalServiceExcepti
...
             }
+            return application;
+        }
 
+        /// <summary>
+        /// Gets a cached ManagedIdentityApplication object or builds a new one if not found.
+        /// </summary>
+        /// <param name="mergedOptions"></param>
+        /// <param name="managedIdentityOptions"></param>
+        /// <returns></returns>
```
keegan-caruso, line 619: quite  a few empty doc comments

tests/Microsoft.Identity.Web.Test/TokenAcquisitionAuthorityTests.cs
```diff
@@ -309,5 +310,86 @@ public void MergeExtraQueryParameters_MergedOptionsNull_Test()
...
+            InitializeTokenAcquisitionObjects();
+            string key = GetCacheKeyForManagedIdReflection(managedIdentityOptions);
+            ConcurrentDictionary<string, IManagedIdentityApplication?> cacheDict = GetManagedIdCacheReflection();
+
+            // Act
+            IManagedIdentityApplication app = 
+                _tokenAcquisition.GetOrBuildManagedIdentityApplication(mergedOptions, managedIdentityOptions).Result;
+
+            // Assert
+            Assert.Equal(app, cacheDict[key]);
```
keegan-caruso, line 360: Not a huge fan of this method with using so much reflection. Why do we have to get the private dictionary? Isn't it sufficient to do something like getting calling the GetOrBuild method twice and checking that they are the same object?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -585,10 +608,75 @@ private bool IsInvalidClientCertificateOrSignedAssertionError(MsalServiceExcepti
...
+        internal async Task<IManagedIdentityApplication> GetOrBuildManagedIdentityApplication(
+            MergedOptions mergedOptions, ManagedIdentityOptions managedIdentityOptions)
+        {
+            IManagedIdentityApplication? application;
+            string key = GetCacheKeyForManagedId(managedIdentityOptions);
+
+            // Lock the potential write of the dictionary to prevent multiple threads from creating the same application.
+            await _managedIdSemaphore.WaitAsync();
```
keegan-caruso, line 627: I'd love to see tests at some level for checking correctness around concurrency.

tests/Microsoft.Identity.Web.Test/TokenAcquisitionAuthorityTests.cs
```diff
@@ -309,5 +310,86 @@ public void MergeExtraQueryParameters_MergedOptionsNull_Test()
...
+                UserAssignedClientId = clientId
+            };
+            BuildTheRequiredServices();
+            MergedOptions mergedOptions = new();
+            InitializeTokenAcquisitionObjects();
+            string key = GetCacheKeyForManagedIdReflection(managedIdentityOptions);
+            ConcurrentDictionary<string, IManagedIdentityApplication?> cacheDict = GetManagedIdCacheReflection();
+
+            // Act
+            IManagedIdentityApplication app =
```
keegan-caruso, line 356: id avoid using .Result even in tests, just make the Test method `async Task`

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
JoshLozensky: Thanks I always forget them because they look like documentation on their own. XD

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
JoshLozensky: Solid point, is there a common way to do this I could use as a starting point?

tests/Microsoft.Identity.Web.Test/TokenAcquisitionAuthorityTests.cs
JoshLozensky: You're completely right, had too much fun playing with reflection in the previous method and didn't stop to think whether I needed it XD

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
JoshLozensky: My initial thought would be to make several threads and set a gate so they all call the method at roughly the same time.

tests/Microsoft.Identity.Web.Test/TokenAcquisitionAuthorityTests.cs
JoshLozensky: Removed it, out of curiosity, what is wrong with using .Result?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
JoshLozensky: Wasn't familiar with that pattern, thanks!

tests/Microsoft.Identity.Web.Test/TokenAcquisitionAuthorityTests.cs
keegan-caruso: It is async over sync.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
keegan-caruso: sounds reasonable

Directory.Build.props
```diff
@@ -86,7 +86,7 @@
     <MicrosoftGraphVersion>4.36.0</MicrosoftGraphVersion>
     <MicrosoftGraphBetaVersion>4.57.0-preview</MicrosoftGraphBetaVersion>
     <MicrosoftExtensionsHttpVersion>3.1.3</MicrosoftExtensionsHttpVersion>
-    <MicrosoftIdentityAbstractions>5.0.0</MicrosoftIdentityAbstractions>    
+    <MicrosoftIdentityAbstractions>5.1.0</MicrosoftIdentityAbstractions>
```
jennyf19, line 89: make sure to update w/main branch often, but with the release on Thursday, a few of these were updated.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -45,11 +46,14 @@ class OAuthConstants
...
 
         private readonly object _applicationSyncObj = new();
+        private readonly SemaphoreSlim _managedIdSemaphore = new(1, 1);
 
         /// <summary>
-        ///  Please call GetOrBuildConfidentialClientApplication instead of accessing this field directly.
+        ///  Please call GetOrBuildConfidentialClientApplication instead of accessing _applicationsByAuthorityClientId directly.
         /// </summary>
         private readonly ConcurrentDictionary<string, IConfidentialClientApplication?> _applicationsByAuthorityClientId = new ConcurrentDictionary<string, IConfidentialClientApplication?>();
+        private readonly ConcurrentDictionary<string, IManagedIdentityApplication?> _managedIdentityApplicationsByClientId = new ConcurrentDictionary<string, IManagedIdentityApplication?>();
```
jennyf19, line 55: >new ConcurrentDictionary [](http://example.com/codeflow?start=125&length=24)

nit - new expression can be simplied

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -585,10 +608,82 @@ private bool IsInvalidClientCertificateOrSignedAssertionError(MsalServiceExcepti
...
+        /// <summary>
+        /// Gets a cached ManagedIdentityApplication object or builds a new one if not found.
+        /// </summary>
+        /// <param name="mergedOptions"></param>
+        /// <param name="managedIdentityOptions"></param>
+        /// <returns></returns>
+        internal async Task<IManagedIdentityApplication> GetOrBuildManagedIdentityApplication(
+            MergedOptions mergedOptions, ManagedIdentityOptions managedIdentityOptions)
+        {
+            IManagedIdentityApplication? application;
```
jennyf19, line 623: this can be declare inline:
`out IManagedIdentityApplication? application`

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -585,10 +608,82 @@ private bool IsInvalidClientCertificateOrSignedAssertionError(MsalServiceExcepti
...
+            }
             return application;
         }
 
+        /// <summary>
+        /// Gets the key value for the ManagedIdentity cache, the default key for system-assigned identity is used if there is no
+        /// clientId for a user-assigned identity specified.
+        /// </summary>
+        /// <param name="managedIdOptions">Holds the clientId for managed identity if none is present</param>
+        /// <returns></returns>
```
jennyf19, line 674: >/// <returns></returns> [](http://example.com/codeflow?start=8&length=23)

nit: complete return xml or remove. we don't have ref doc for private methods, so okay to remove.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -674,6 +768,27 @@ private IConfidentialClientApplication BuildConfidentialClientApplication(Merged
             }
         }
 
+        /// <summary>
+        /// Creates a managed identity client application.
+        /// </summary>
+        /// <param name="managedIdentityId">Indicates if system-assigned or user-assigned managed identity is used</param>
```
jennyf19, line 774: nit: periods at the end

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -674,6 +768,27 @@ private IConfidentialClientApplication BuildConfidentialClientApplication(Merged
...
+        /// <param name="enablePiiLogging">Indicates if logging that may contain personally identifiable information should be enabled</param>
+        /// <returns>A managed identity application </returns>
```
jennyf19, line 776: >  [](http://example.com/codeflow?start=51&length=1)

remove space

tests/Microsoft.Identity.Web.Test/TokenAcquisitionAuthorityTests.cs
```diff
@@ -309,5 +311,88 @@ public void MergeExtraQueryParameters_MergedOptionsNull_Test()
             // Assert
             Assert.Null(mergedDict);
         }
+
+        [Theory]
+        [MemberData(nameof(ManagedIdClientIdCases))]
```
jennyf19, line 316: >[MemberData(nameof(ManagedIdClientIdCases))] [](http://example.com/codeflow?start=8&length=44)

I think this is overcomplicating the test because it's not a complex object. You can just pass a string, like the test on line 146.

tests/Microsoft.Identity.Web.Test/TokenAcquisitionAuthorityTests.cs
```diff
@@ -309,5 +311,88 @@ public void MergeExtraQueryParameters_MergedOptionsNull_Test()
...
+            {
+                Assert.Equal(defaultKey, key);
+            }
+            else
+            {
+                Assert.Equal(clientId, key);
+            }
+        }
+
+
```
jennyf19, line 342: nit: rmv line

tests/Microsoft.Identity.Web.Test/TokenAcquisitionAuthorityTests.cs
```diff
@@ -309,5 +311,88 @@ public void MergeExtraQueryParameters_MergedOptionsNull_Test()
...
+            IManagedIdentityApplication app1 = 
+                await _tokenAcquisition.GetOrBuildManagedIdentityApplication(mergedOptions, managedIdentityOptions);
+            IManagedIdentityApplication app2 = 
+                await _tokenAcquisition.GetOrBuildManagedIdentityApplication(mergedOptions, managedIdentityOptions);
+
+            // Assert
+            Assert.Same(app1, app2);
+        }
+
+
```
jennyf19, line 366: nit: rmv line

jennyf19 (changes requested):
:clock1:

jennyf19:
Looking good @JoshLozensky left a comment in one of the issues w/an idea for E2E testing. If you can address the current comments, we can move it out of draft at least. 

---
In reply to: [1766521586](https://github.com/AzureAD/microsoft-identity-web/pull/2610#issuecomment-1766521586) [](http://example.com/codeflow?ancestors=1766521586)

Directory.Build.props
JoshLozensky: done

PR #2616, opened 2023-12-13
IDownstreamApi handles Continuous Access Evaluation (CAE)

### LINKED ISSUE #2550: [Feature Request] IDownstream API should do CAE on its own
Enable CAE in IDownstream API, even if the underlying downstream API does not support it. 

This means: 

- [ ] add "client capabilities" `cp1` value by default. If other client capabilities are added by the user, merge `cp1` into them.
- [ ] if downstream api replies with 401 and with WWWAuthenticate header, parse it and extract headers (exact value of WWWAuthenticate TBD, but expect it to have claims)
- [ ] acquire another token with claims from WWWAuthenticate
- [ ] if another 401 comes in, bail. 
- [ ] update CAE docs to mention ID.Web improvements https://learn.microsoft.com/en-us/entra/identity-platform/claims-challenge?tabs=dotnet

This work should be done for both user scenarios and for S2S scenarios (client credetials and MSI). Priority order is: 

- [ ] client_credentials
- [ ] client_credentials + certificatless - FIC scenario (Leg1, getting token from MSI, is covered by Azure.Identity, Leg2 should be covered by client credentials) (no work needed, but needs testing)
- [ ] MSI (after Id Web is implemented to use MSAL)
- [ ] web site (challenge the user)
- [ ] web api (return 401 + WWW-Authenticate with claims to the client)

Docs: https://learn.microsoft.com/en-us/entra/identity/conditional-access/concept-continuous-access-evaluation

pmaytak:
Completes (partially) #2550

**Changes:**
- Add a retry in `IDownstreamApi` for user and app requests. If the resource returns a 401 error, use claims to get a new token and call the resource again.

As per offline discussion, users will have to add client capabilities, which can be done like so:
```csharp
services.Configure<ConfidentialClientApplicationOptions>(
                                 o => o.ClientCapabilities = new[] { "cp1" });
```

**Tests:**
- Unit tests.
- Integration tests will be part of #2653 when CAE environment is setup in the lab.
- Manual tests.
   - User flow were tested by revoking the test user's session; app flow - by disabling the app's service principal.
   - Tested by calling Graph: `/me` endpoint for user flow, `/users` endpoint for app flow using the [client and service](https://github.com/AzureAD/microsoft-identity-web/tree/master/tests/DevApps/WebAppCallsWebApiCallsGraph) dev apps.

**Test scenarios:**
- Web app, user flow - when the user's session is revoked, Graph does send a 401 error with claims; when claims are used to get a new token, STS returns an error that the user needs to re-login; Identity Web handles the MsalUiRequiredException; after the user signs-in again, the app is redirected to call the downstream API again, which succeeds with the new token.
- Web app, client credentials flow - when the service principal is disabled, Graph does send a 401 error with claims; but when acquiring a new token with claims, STS returns an error that the SP is disabled.
- Web API, OBO flow - CAE doesn't work, not supported. The resource doesn't enforce the policies.

**Manual test steps:**
- Set up dev up [WebAppCallsWebApiCallsGraph](https://github.com/AzureAD/microsoft-identity-web/tree/master/tests/DevApps/WebAppCallsWebApiCallsGraph/Client) with an appropriate tenant config.
- In the appsettings add scopes for Graph for user and app:
```json
    "GraphUser": {
        "Scopes": [ "user.read" ],
        "BaseUrl": "https://graph.microsoft.com/v1.0"
    },
    "GraphApp": {
        "Scopes": [ "https://graph.microsoft.com/.default" ],
        "BaseUrl": "https://graph.microsoft.com/v1.0"
    },
```

**Testing user flow:**
1. In the controller Index action, add calling Graph for user: `await _downstreamApi.GetForUserAsync<TestPoco>("GraphUser", options => options.RelativePath = "me");`
2. Run the client app, sign in with a test user.
3. The token cache is populated with an access and refresh token.
4. Navigate to the above controller action to call _downstreamApi.GetForUser.
5 Internally Id Web uses the user token from the cache and calls Graph with it.
6. Graph should return an HTTP 200 successful response.
7. Revoke the test user's session.
- Go to Azure Portal > Microsoft Entra ID > Users blade > select the test user > click Revoke sessions.
- image
8. Wait about 5 minutes for the changes to propagate to Graph.
9. Navigate to the above controller action again to call _downstreamApi.GetForUser.
10 Internally Id Web will use the cached token and call Graph with it.
11. Graph will now return an HTTP 401 with claims.
12. The retry logic will try to get a new user token, AAD will return an error, Id Web will catch the MsalUiRequiredEx, and ask the user to sign in again.
13. Sign in again with the test user.
14. Id Web will redirect to the controller action again which will call _downstreamApi.GetForUser.
15. Id Web will retrieve the new cached token, and successfully call Graph.

**Testing client credentials:**
1. In the controller Index action, add calling Graph for app: `await _downstreamApi.GetForAppAsync<TestPoco>("GraphApp", options => options.RelativePath = "users");`
2. Run the client app, sign in with a test user.
3. Navigate to the above controller action to call _downstreamApi.GetForApp.
4. Internally Id Web gets an app token from AAD and calls Graph with it.
5. Graph should return an HTTP 200 successful response.
6. Disable the app's service principal.
- Go to Azure Portal > Microsoft Entra ID > Enterprise applications > select this application > Properties blade > change "Enabled for users to sign-in?" to No.
- This will prevent user sign-in and also accessing with app tokens.
- image
7. Wait about 5 minutes for the changes to propagate to Graph.
8. Navigate to the above controller action again to call _downstreamApi.GetForApp.
9. Internally Id Web will use the cached token and call Graph with it.
10. Graph will now return an HTTP 401 with claims.
11. The retry logic will try to get a new user token with claims, AAD will return an error that the service principal is disabled.


src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -599,8 +599,26 @@ private IConfidentialClientApplication BuildConfidentialClientApplication(Merged
 
             try
             {
+                IList<string> AddContinuousAccessEvaluationCapability(IEnumerable<string> clientCapabilities)
+                {
+                    if (clientCapabilities == null)
+                    {
+                        return new[] { Constants.CaeCapability };
+                    }
+                    else
```
bgavrilMS, line 608: else not needed.

tests/Microsoft.Identity.Web.Test.Integration/DownstreamApiForAppCaeTests.cs
```diff
@@ -0,0 +1,134 @@
...
+        {
+            KeyVaultSecretsProvider keyVaultSecretsProvider = new(TestConstants.BuildAutomationKeyVaultName);
+            _ccaSecret = keyVaultSecretsProvider.GetSecretByName(TestConstants.AzureADIdentityDivisionTestAgentSecret).Value;
+
+            // Need the secret before building the services
+            if (!string.IsNullOrEmpty(_ccaSecret))
+            {
+                BuildRequiredServices();
+            }
+            else
```
bgavrilMS, line 41: nit: else not needed.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -599,8 +599,26 @@ private IConfidentialClientApplication BuildConfidentialClientApplication(Merged
...
             {
+                IList<string> AddContinuousAccessEvaluationCapability(IEnumerable<string> clientCapabilities)
+                {
+                    if (clientCapabilities == null)
+                    {
+                        return new[] { Constants.CaeCapability };
+                    }
+                    else
+                    {
+                        var finalClientCapabilities = new List<string>(clientCapabilities);
```
bgavrilMS, line 610: I am not convinced that this logic is well unit tested. Granted that the unit tests here are hard to read, but do you have test for both branches?

bgavrilMS (approved):

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
pmaytak: Why not? 

Currently CAE is the only capability but when there will be more, we'd want to add CAE if some other ones were added already. Also currently someone can add a non-existent capability; so we'd want to add CAE also (although I don't know if STS throws an exception for invalid capabilities).

tests/Microsoft.Identity.Web.Test.Integration/DownstreamApiForAppCaeTests.cs
pmaytak: Copy-pasted from other tests. Will refactor.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
pmaytak: Or are you saying just remove the else, keep the rest?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
pmaytak: Added additional test cases.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -599,8 +599,24 @@ private IConfidentialClientApplication BuildConfidentialClientApplication(Merged
 
             try
             {
+                static IList<string> AddContinuousAccessEvaluationCapability(IEnumerable<string> clientCapabilities)
```
jmprieur, line 602: I don't think it should be done in TokenAcquisition. It should be done in DownstreamApi.

ITokenAcquirer and IAuthorizationHeaderProvider are not CAE capable if the app or library using them is not. DownstreamApi is CAE capable

jmprieur (changes requested):

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
```diff
@@ -302,43 +303,65 @@ internal partial class DownstreamApi : IDownstreamApi
...
             // Downstream API URI
-            string apiUrl = effectiveOptions.GetApiUrl();
+            string apiUrl = effectiveOptions.GetApiUrl();
+
+            using HttpClient client = string.IsNullOrEmpty(serviceName) ? _httpClientFactory.CreateClient() : _httpClientFactory.CreateClient(serviceName);
 
-            // Creation of the Http Request message with the right HTTP Method
-            using HttpRequestMessage httpRequestMessage = new HttpRequestMessage(
-                new HttpMethod(effectiveOptions.HttpMethod),
+            async Task UpdateRequestAsync(HttpRequestMessage request)
```
jennyf19, line 310: >Task [](http://example.com/codeflow?start=18&length=4)

Can you move this to outside the method? Also could probably return a `ValueTask` to save an allocation.

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
```diff
@@ -302,43 +303,65 @@ internal partial class DownstreamApi : IDownstreamApi
...
+            {
+                if (content != null)
+                {
+                    request.Content = content;
+                }
+
+                // Acquire an access token and create an authorization header
+                if (effectiveOptions.Scopes != null && effectiveOptions.Scopes.Any())
```
jennyf19, line 318: please use the previous comment thanks

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
```diff
@@ -302,43 +303,65 @@ internal partial class DownstreamApi : IDownstreamApi
...
+            effectiveOptions.AcquireTokenOptions.Claims = WwwAuthenticateParameters.GetClaimChallengeFromResponseHeaders(downstreamApiResult.Headers);
+
+            using HttpRequestMessage retryHttpRequestMessage = new(
+                new HttpMethod(effectiveOptions.HttpMethod),
+                apiUrl);
+
+            await UpdateRequestAsync(retryHttpRequestMessage);
+
+            return await client.SendAsync(retryHttpRequestMessage, cancellationToken).ConfigureAwait(false);
         }
```
jennyf19, line 365: I think we need the e2e test before merging.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
jennyf19: +1. DownstreamApi parses the WWW-authenticate header, and is CAE capable.

jennyf19 (changes requested):
:clock1:

doesn't seem to be the right approach and would prefer to have E2E test working before we merge.

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
pmaytak: Tested E2E manually. For integration tests, we'll need to wait for the lab to have a CAE test app set up then..

jennyf19: Where can we get details on the E2E test that was done?

pmaytak: In the PR description. Which details are you looking for?

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
pmaytak: Moved. Since this method will probably do an async network call to AAD or L2 cache, ValueTask will probably not help.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
pmaytak: What's the preferred way to pass the CAE flag from DownstreamApi.cs to TokenAcquisition?

pmaytak: Also what about these two scenarios:
- User flow in a web app. If the CAE flag is moved to DownstreamApi level, when a user sign into the site non-CAE token gets cached. Then DownstreamApi.GetTokenForUser will get the cached token before calling the API.
- Singleton token acquisition. If the CCA gets created without the CAE flag, then we're stuck into that configuration.

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
jennyf19: I tried testing this several several months ago and couldn't get it to work. I guess I will need more steps on what exactly to do.

pmaytak: Added more testing steps in the PR description.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
jennyf19: Optimal way forward is for MSAL.NET to expose `WithClientCapabilities` on each request. Then in Id Web, we can use it.

jennyf19: cca requests, specifically

bgavrilMS: We discussed a few options, @pmaytak will follow up with a detailed list: 

- `WithClientCapabilities` was always designed to be cache agnostic, MSAL has no idea if an access token is CAE-enabled or not. 
- We could add this functionality as an opt-in to MSAL, where MSAL differentiates between CAE capable / CAE non-capable tokens. It's not trivial and we'll take a small perf hit too.
- DownstreamAPI is ZTA but TokenAcquisition is not ZTA, so today Id.Web based website samples are not ZTA, but S2S samples are.
- Wondering if we can still make the website samples ZTA / CAE-capable by adding a "CAE capable" flag maybe in `AddMicrosoftIdentityWebApp` ?

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
bgavrilMS: Is this the only blocking comment remaining? Integration testing?

pmaytak:
I've reduced the scope of this PR. It **doesn't** set CAE flag by default; it only retries downstream API requests if the first response was 401 and contains claims.

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
pmaytak: I've reduced the scope of this PR. It doesn't set CAE flag by default; it only retries downstream API requests if the first response was 401 and contains claims. Added more unit tests.

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
```diff
@@ -286,15 +287,45 @@ internal partial class DownstreamApi : IDownstreamApi
...
-            {
-                httpRequestMessage.Content = content;
-            }
 
+            await UpdateRequestAsync(httpRequestMessage, content, effectiveOptions, appToken, user, cancellationToken);
+
+            // Send the HTTP message           
+            var downstreamApiResult = await client.SendAsync(httpRequestMessage, cancellationToken).ConfigureAwait(false);
+
+            effectiveOptions.AcquireTokenOptions.Claims = WwwAuthenticateParameters.GetClaimChallengeFromResponseHeaders(downstreamApiResult.Headers);
```
jmprieur, line 302: Why do we get the Claims if we don't have a 401? can it happen that there is a claims challenge with a 200? What is the scenario?

Apparently we only retry in the case of a 401 below. So I think we should delay extracting the claims until we have tested it's a 401?

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
```diff
@@ -286,15 +287,45 @@ internal partial class DownstreamApi : IDownstreamApi
...
+
+        private async Task UpdateRequestAsync(
+            HttpRequestMessage request,
+            HttpContent? content,
+            DownstreamApiOptions effectiveOptions,
+            bool appToken,
+            ClaimsPrincipal? user,
+            CancellationToken cancellationToken)
+        {
+            request.Content = content;
```
jmprieur, line 327: can we do that if the content is null? (which will happen for HTTP Get for instance?

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
```diff
@@ -286,15 +287,45 @@ internal partial class DownstreamApi : IDownstreamApi
...
+                await UpdateRequestAsync(retryHttpRequestMessage, content, effectiveOptions, appToken, user, cancellationToken);
+
+                return await client.SendAsync(retryHttpRequestMessage, cancellationToken).ConfigureAwait(false);
+            }
+
+            return downstreamApiResult;
+        }
+
+        private async Task UpdateRequestAsync(
+            HttpRequestMessage request,
```
jmprieur, line 320: Not sure why you needed to rename the parameter (from `httpRequestMessage`). This made reviewing your PR more difficult to me. (please keep in mind for the future)

jmprieur (changes requested):
LGTM

There is one question that needs to be solved (do not extract the claims if we don't have a 401)?

jmprieur (approved):
LGTM
thanks @pmaytak

jennyf19 (approved):
:shipit:

PR #2619, opened 2023-12-27
Adding Check Against Injection Attacks

FuPingFranco:
## Description

Add a check in ID.Web to check for the presence of '&' for assertion and sub-assertion in ExtraQueryParameters dictionary.

### DIFF
--- a/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
@@ -783,6 +783,9 @@ namespace Microsoft.Identity.Web
                             // Special case when the OBO inbound token is composite (for instance PFT)
                             if (dict.ContainsKey(assertionConstant) && dict.ContainsKey(subAssertionConstant))
                             {
+
+                                CheckAssertionsForInjectionAttempt(dict[assertionConstant], dict[subAssertionConstant]);
+
                                 builder.OnBeforeTokenRequest((data) =>
                                 {
                                     // Replace the assertion and adds sub_assertion with the values from the extra query parameters
@@ -828,6 +831,20 @@ namespace Microsoft.Identity.Web
                     LogMessages.ErrorAcquiringTokenForDownstreamWebApi + ex.Message,
                     ex);
                 throw;
+            }
+        }
+
+        /// <summary>
+        /// Checks assertion and sub_assertion passed from merging extra query parameters to ensure they do not contain unsupported characters.
+        /// </summary>
+        /// <param name="assertion">The assertion.</param>
+        /// <param name="subAssertion">The sub_assertion.</param>
+        private static void CheckAssertionsForInjectionAttempt(string assertion, string subAssertion)
+        {
+            if (assertion != null || subAssertion != null)
+            {
+                if (assertion.Contains('&')) throw new ArgumentException(IDWebErrorMessage.InvalidAssertion, nameof(assertion));
+                else if (subAssertion.Contains('&')) throw new ArgumentException(IDWebErrorMessage.InvalidSubAssertion, nameof(subAssertion));
             }
         }
 
--- a/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
@@ -841,7 +841,7 @@ namespace Microsoft.Identity.Web
         /// <param name="subAssertion">The sub_assertion.</param>
         private static void CheckAssertionsForInjectionAttempt(string assertion, string subAssertion)
         {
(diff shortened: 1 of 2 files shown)


MZOLN (commented):
This looks like a DID security fix am I correct?
If it is, we should be using whitelisting vs black listing.
Your comment in the function suggests that you perform (paraphrasing) "checks to ensure that assertions do not contain unsupported characters" but the implementation only checks for "&". There could be other characters or representation of characters that cause issue. 

If we are discussing a security feature, the best approach is to whitelist vs blacklist, thus validate if each character in the assertion is in the expected range (for example A-z0-9) vs check if it contains unexpected characters.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -831,6 +834,20 @@ private void NotifyCertificateSelection(MergedOptions mergedOptions, IConfidenti
...
+        /// <summary>
+        /// Checks assertion and sub_assertion passed from merging extra query parameters to ensure they do not contain unsupported characters.
+        /// </summary>
+        /// <param name="assertion">The assertion.</param>
+        /// <param name="subAssertion">The sub_assertion.</param>
+        private static void CheckAssertionsForInjectionAttempt(string assertion, string subAssertion)
+        {
+            if (!assertion.IsNullOrEmpty() || !subAssertion.IsNullOrEmpty())
+            {
+                if (assertion.Contains('&', StringComparison.InvariantCultureIgnoreCase)) throw new ArgumentException(IDWebErrorMessage.InvalidAssertion, nameof(assertion));
```
jmprieur, line 846: Won't you still have a null reference exception if assertion is null, but not subAssertion?
Shouldn't it be?

```csharp
 if (!assertion.IsNullOrEmpty()  && assertion.Contains('&', StringComparison.InvariantCultureIgnoreCase)) throw new ArgumentException(IDWebErrorMessage.InvalidAssertion, nameof(assertion));
if (!subAssertion.IsNullOrEmpty() && (subAssertion.Contains('&', StringComparison.InvariantCultureIgnoreCase)) throw new ArgumentException(IDWebErrorMessage.InvalidSubAssertion, nameof(subAssertion));
```

jmprieur (commented):
I still don't understand the logic in `CheckAssertionsForInjectionAttempt`

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
FuPingFranco: I believe this is taken care of in line 780, 781 and 784 the only reason I even did the check was due to warnings from Nullable checks.

FuPingFranco:
> This looks like a DID security fix am I correct? If it is, we should be using whitelisting vs black listing. Your comment in the function suggests that you perform (paraphrasing) "checks to ensure that assertions do not contain unsupported characters" but the implementation only checks for "&". There could be other characters or representation of characters that cause issue.
> 
> If we are discussing a security feature, the best approach is to whitelist vs blacklist, thus validate if each character in the assertion is in the expected range (for example A-z0-9) vs check if it contains unexpected characters.

Good point, I just did this check based on this PR https://identitydivision.visualstudio.com/DevEx/_git/Microsoft-IdentityModel-S2S/pullrequest/9985?_a=files&path=/src/Microsoft.IdentityModel.S2S.Tokens/TokenManager.cs I'll get more details to ensure I have a more complete solution. Thanks Marcin!

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
jennyf19: What @jmprieur suggests will work. By checking for null before calling the `IsNullOrEmpty()` method, you can ensure that the method is only called on non-null strings

FuPingFranco: Thanks @jmprieur & @jennyf19

FuPingFranco:
> This looks like a DID security fix am I correct? If it is, we should be using whitelisting vs black listing. Your comment in the function suggests that you perform (paraphrasing) "checks to ensure that assertions do not contain unsupported characters" but the implementation only checks for "&". There could be other characters or representation of characters that cause issue.
> 
> If we are discussing a security feature, the best approach is to whitelist vs blacklist, thus validate if each character in the assertion is in the expected range (for example A-z0-9) vs check if it contains unexpected characters.

@MZOLN Could you open a new issues with a bit more details? I would like to close this task with the current defined scope. Thanks Marcin!

jmprieur (approved):

jennyf19 (approved):
:shipit:

jennyf19:
@FuPingFranco please create an issue in Id Web for this for easier tracking and link to the follow-up PR I did as well. thanks. https://github.com/AzureAD/microsoft-identity-web/pull/2639

PR #2633, opened 2024-01-10
Update RELEASES.md to remove mentioning of a package that doesn't exist

henrik-me:
Update RELEASES.md to remove mentioning of a package name that doesn't exist

# Update RELEASES.md to remove mentioning of a package name that doesn't exist

Removed MSALiOS and replaced with MSAL for iOS

Fixes a general documentation ask to remove elements that can sound like real package names.

### DIFF
--- a/RELEASES.md
+++ b/RELEASES.md
@@ -31,7 +31,7 @@ version="[1.1,1.2)"
 When we release a new MINOR version, the previous MINOR version is abandoned.
 
 When we release a new MAJOR version, we will continue to apply bug fixes to the existing features in the previous MAJOR version for up to the 2 year support cycle for Azure.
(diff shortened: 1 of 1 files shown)


henrik-me:
pls. merge if in agreement with the change

jmprieur:
This is a duplicate of https://github.com/AzureAD/microsoft-identity-web/pull/2632, which was proposed earlier.

henrik-me:
Ok the issue to be solved is the same, I suggested a different way that would keep the example for clarification.

jmprieur:
I saw but thought we have a link on semantic versionning already which explains better so I decided to remove the sentence. I cam read yours but mentioning idweb not ios

PR #2635, opened 2024-01-10
change langVer to 12

kellyyangsong:
related to https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/2184

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -72,7 +72,7 @@
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(TargetFramework)' == 'net472' Or '$(TargetFramework)' == 'net462' Or '$(TargetFramework)' == 'netstandard2.0' Or '$(TargetFramework)' == 'netcoreapp3.1'">
-    <LangVersion>11</LangVersion>
+    <LangVersion>12</LangVersion>
   </PropertyGroup>
 
   <PropertyGroup Label="Common dependency versions">


Directory.Build.props
```diff
@@ -72,7 +72,7 @@
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(TargetFramework)' == 'net472' Or '$(TargetFramework)' == 'net462' Or '$(TargetFramework)' == 'netstandard2.0' Or '$(TargetFramework)' == 'netcoreapp3.1'">
-    <LangVersion>11</LangVersion>
+    <LangVersion>12</LangVersion>
```
FuPingFranco, line 75: Do we want to continue to update this manually moving forward or use 'preview' instead to always be up to date?

jmprieur (approved):

Directory.Build.props
jmprieur: With preview don't we risk that a preview compiler produces binaries which are not right? Would they be meant for prod?

FuPingFranco: There's that risk but didn't we used "latest" before which technically had the same risk? also, it does allow us to catch any issues from a new C# version earlier and not later.

kellyyangsong: "latest" did not infer the same risk and maps to pre-11

FuPingFranco (approved):

PR #2649, opened 2024-01-23
Fixed incorrect test CollectionDefinition

JoshLozensky:
The B2CWebAppCallsWebApiLocally CollectionDefinition incorrectly referred to a different UI test class causing it to potentially run in parrallel with the other UI test class causing failures when multiple instances of an app tried to access the same port at the same time.


jmprieur (approved):

kellyyangsong (approved):

jennyf19 (approved):
:shipit: would be great to fix the warnings too

PR #2650, opened 2024-01-24
Lozensky/enable managed identity

### LINKED ISSUE #1775: [Feature Request] Managed Identity support in IDownstreamWebApi / IDownstreamApi

JoshLozensky:
# Add support for managed identities
Added logic to default to managed identity credentials if available.

## Description

In TokenAquisition.GetAuthenticationResultForAppAsync() added a check for managed identity configuration in the tokenAquisitionOptions parameter. If present, default to this authentication method and use system-assigned or user-assigned managed identity per the user's configuration.

Fixes #1775 #2551


jennyf19:
@JoshLozensky I will need some time to take a look at the proposal. ETA is Thursday.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -585,10 +608,81 @@ private bool IsInvalidClientCertificateOrSignedAssertionError(MsalServiceExcepti
...
+        /// <param name="mergedOptions">The configuration options for the app</param>
+        /// <param name="managedIdentityOptions">The configuration specific to managed identity</param>
+        /// <returns>The application object used to request a token with managed identity</returns>
+        internal async Task<IManagedIdentityApplication> GetOrBuildManagedIdentityApplication(
+            MergedOptions mergedOptions, ManagedIdentityOptions managedIdentityOptions)
+        {
+            string key = GetCacheKeyForManagedId(managedIdentityOptions);
+
+            // Check if the application is already built, if so return it without grabbing the key
+            if (_managedIdentityApplicationsByClientId.TryGetValue(key, out IManagedIdentityApplication? application) && application != null)
```
bgavrilMS, line 626: I think the way "TryGetValue" type of APIs work is that they return false if the value cannot be fetched. So no need to check if `applicaiton != null`

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -585,10 +608,81 @@ private bool IsInvalidClientCertificateOrSignedAssertionError(MsalServiceExcepti
...
+    
+                // Build the application
+                application = BuildManagedIdentityApplication(
+                    managedIdentityId,
+                    mergedOptions.ConfidentialClientApplicationOptions.EnablePiiLogging
+                );
 
+                // Add the application to the cache
+                _managedIdentityApplicationsByClientId.TryAdd(key, application);
+            } finally
```
keegan-caruso, line 660: spacing

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -585,10 +608,81 @@ private bool IsInvalidClientCertificateOrSignedAssertionError(MsalServiceExcepti
...
+        /// <returns>A key value for the Managed Identity cache</returns>
+        internal /* for testing */ static string GetCacheKeyForManagedId(ManagedIdentityOptions managedIdOptions) 
+        {
+            if (managedIdOptions.UserAssignedClientId.IsNullOrEmpty())
+            {
+                return SystemAssignedManagedIdentityKey;
+            }
+            else
+            {
+                return managedIdOptions.UserAssignedClientId!;
```
keegan-caruso, line 682: do we know this is not null?

JoshLozensky: Since the if checks this value for NullOrEmpty, the else shouldn't have any way to be null. 

(Other than if someone messes with the object between the if check and dereference but that would mean they implemented incorrectly as this value should come from the appsettings.json config file.)

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisitionOptions.cs
westin-m, line 1: Please put periods at the ends of params and return descriptions

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -585,10 +608,81 @@ private bool IsInvalidClientCertificateOrSignedAssertionError(MsalServiceExcepti
...
             }
+            return application;
+        }
+
+        /// <summary>
+        /// Gets a cached ManagedIdentityApplication object or builds a new one if not found.
+        /// </summary>
+        /// <param name="mergedOptions">The configuration options for the app</param>
+        /// <param name="managedIdentityOptions">The configuration specific to managed identity</param>
+        /// <returns>The application object used to request a token with managed identity</returns>
```
westin-m, line 619: Please put periods at the ends of params and return descriptions

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -585,10 +608,81 @@ private bool IsInvalidClientCertificateOrSignedAssertionError(MsalServiceExcepti
...
             return application;
         }
 
+        /// <summary>
+        /// Gets the key value for the Managed Identity cache, the default key for system-assigned identity is used if there is no
+        /// clientId for a user-assigned identity specified.
+        /// </summary>
+        /// <param name="managedIdOptions">Holds the clientId for managed identity if none is present</param>
+        /// <returns>A key value for the Managed Identity cache</returns>
+        internal /* for testing */ static string GetCacheKeyForManagedId(ManagedIdentityOptions managedIdOptions)
```
westin-m, line 674: Can the testing comment be part of the summary?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
JoshLozensky: Agreed in this case since I control all values that go into the dictionary and don't allow null ones. However, for future reference, TryGetValue will return false only if the key cannot be found. Even if the value is null for a given key it will return true as long as the key is there.

tests/E2E Tests/TokenAcquirerTests/TokenAcquirer.cs
```diff
@@ -299,5 +301,45 @@ private static async Task CreateGraphClientAndAssert(TokenAcquirerFactory tokenA
             Assert.NotNull(result.AccessToken);
         }
     }
+
+    public class AcquireTokenManagedIdentity 
+    {
+        [OnlyOnAzureDevopsFact]
+        //[Fact]
```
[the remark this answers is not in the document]
JoshLozensky: It was the convention on other tests with overidden decorators, I assume for the sake readability. @jennyf19  any thoughts?

tests/Microsoft.Identity.Web.Test/Microsoft.Identity.Web.Test.csproj
```diff
@@ -18,6 +18,7 @@
   </ItemGroup>
 
   <ItemGroup>
+    <PackageReference Include="CommunityToolkit.HighPerformance" Version="8.2.2" />
```
keegan-caruso, line 21: ?

tests/E2E Tests/TokenAcquirerTests/TokenAcquirer.cs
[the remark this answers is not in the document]
jennyf19: It's if you want to run it locally. I would leave it.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -358,10 +363,28 @@ private void LogAuthResult(AuthenticationResult? authenticationResult)
...
+            {
+                try
+                {
+                    IManagedIdentityApplication managedIdApp = await GetOrBuildManagedIdentityApplication(
+                        mergedOptions, 
+                        tokenAcquisitionOptions.ManagedIdentity
+                    );
+                    return await managedIdApp.AcquireTokenForManagedIdentity(scope).ExecuteAsync().ConfigureAwait(false);
+                }
+                catch(Exception ex)
```
jennyf19, line 377: >catch(Exception  [](http://example.com/codeflow?start=16&length=16)

check spacing here
ctrl + kd

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -585,10 +608,82 @@ private bool IsInvalidClientCertificateOrSignedAssertionError(MsalServiceExcepti
...
+        }
+
+        /// <summary>
+        /// Gets a cached ManagedIdentityApplication object or builds a new one if not found.
+        /// </summary>
+        /// <param name="mergedOptions">The configuration options for the app.</param>
+        /// <param name="managedIdentityOptions">The configuration specific to managed identity.</param>
+        /// <returns>The application object used to request a token with managed identity.</returns>
+        internal async Task<IManagedIdentityApplication> GetOrBuildManagedIdentityApplication(
+            MergedOptions mergedOptions, ManagedIdentityOptions managedIdentityOptions)
```
jennyf19, line 621: nit: move this to next line

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -585,10 +608,82 @@ private bool IsInvalidClientCertificateOrSignedAssertionError(MsalServiceExcepti
...
+            }
             return application;
         }
 
+        /// <summary>
+        /// Gets the key value for the Managed Identity cache, the default key for system-assigned identity is used if there is
+        /// no clientId for a user-assigned identity specified. The method is internal rather than private for testing purposes.
+        /// </summary>
+        /// <param name="managedIdOptions">Holds the clientId for managed identity if none is present</param>
+        /// <returns>A key value for the Managed Identity cache</returns>
```
jennyf19, line 674: nit: periods on all these xml comments.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
jennyf19: fun fact that `TryGetValue` is faster  when looking up a specific key in a dictionary compared to `ContainsKey`

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
jennyf19: maybe it's safer to use `string.IsNullOrEmpty(managedIdOptions.UserAssignedClientId)` and remove the `!`

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -674,6 +769,26 @@ private IConfidentialClientApplication BuildConfidentialClientApplication(Merged
             }
         }
 
+        /// <summary>
+        /// Creates a managed identity client application.
+        /// </summary>
+        /// <param name="managedIdentityId">Indicates if system-assigned or user-assigned managed identity is used.</param>
+        /// <param name="enablePiiLogging">Indicates if logging that may contain personally identifiable information should be enabled.</param>
```
jennyf19, line 776: >should be [](http://example.com/codeflow?start=117&length=9)

nit: "is" as in "is enabled" I think makes more sense here

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -674,6 +769,26 @@ private IConfidentialClientApplication BuildConfidentialClientApplication(Merged
...
+        private IManagedIdentityApplication BuildManagedIdentityApplication(ManagedIdentityId managedIdentityId, bool enablePiiLogging)
+        {
+            return ManagedIdentityApplicationBuilder
+                .Create(managedIdentityId)
+                .WithHttpClientFactory(_httpClientFactory)
+                .WithLogging
+                (
+                    Log,
+                    ConvertMicrosoftExtensionsLogLevelToMsal(_logger),
+                    enablePiiLogging: enablePiiLogging
```
jennyf19, line 787: nit: this spacing looks strange to me.

tests/E2E Tests/TokenAcquirerTests/TokenAcquirer.cs
```diff
@@ -299,5 +301,45 @@ private static async Task CreateGraphClientAndAssert(TokenAcquirerFactory tokenA
...
+        public async Task AcquireTokenWithManagedIdentity_UserAssigned()
+        {
```
jennyf19, line 310: I know we don't have this consistently in our tests, but yours in a nice example to have it split up w/
//Arrange
//Act
//Assert

labels on the different parts.

tests/E2E Tests/TokenAcquirerTests/TokenAcquirer.cs
```diff
@@ -299,5 +301,45 @@ private static async Task CreateGraphClientAndAssert(TokenAcquirerFactory tokenA
...
+            const string baseUrl = "https://vault.azure.net";
+            const string clientId = "9c5896db-a74a-4b1a-a259-74c5080a3a6a";
+            TokenAcquirerFactory tokenAcquirerFactory = TokenAcquirerFactory.GetDefaultInstance();
+            _ = tokenAcquirerFactory.Services;
+            IServiceProvider serviceProvider = tokenAcquirerFactory.Build();
+
+            // Get the authorization header provider and add the options to tell it to use Managed Identity
+            IAuthorizationHeaderProvider? api = serviceProvider.GetRequiredService<IAuthorizationHeaderProvider>();
+            Assert.NotNull(api);
+            string result = await api.CreateAuthorizationHeaderForAppAsync(scope, GetAuthHeaderOptions_ManagedId(baseUrl, clientId));
```
jennyf19, line 321: >GetAuthHeaderOptions_ManagedId [](http://example.com/codeflow?start=82&length=30)

when we write up the docs, customers will probably just set the options directly, so keep that in mind (not create another method to call).

tests/Microsoft.Identity.Web.Test/Microsoft.Identity.Web.Test.csproj
jennyf19: agree. this doesn't seem to be needed, as well as the playwright traces...I have several showing up in code flow.

tests/Microsoft.Identity.Web.Test/TokenAcquisitionAuthorityTests.cs
```diff
@@ -13,11 +15,11 @@
 using Microsoft.Extensions.Options;
 using Microsoft.Identity.Abstractions;
 using Microsoft.Identity.Client;
-using Microsoft.Identity.Web.Test.Common;
 using Microsoft.Identity.Web.Test.Common.Mocks;
 using Microsoft.Identity.Web.Test.Common.TestHelpers;
 using Microsoft.Identity.Web.TokenCacheProviders.InMemory;
 using Xunit;
+using TC = Microsoft.Identity.Web.Test.Common.TestConstants;
```
jennyf19, line 22: >TC [](http://example.com/codeflow?start=6&length=2)

this is great, thanks for doing this.

tests/Microsoft.Identity.Web.Test/TokenAcquisitionAuthorityTests.cs
```diff
@@ -84,42 +86,40 @@ public void VerifyCorrectSchemeTests(string scheme)
...
-                    ClientId = TestConstants.ConfidentialClientId,
-                    ClientSecret = TestConstants.ClientSecret,
+                    ClientId = TC.ConfidentialClientId,
+                    ClientSecret = TC.ClientSecret,
                 });
-
-                BuildTheRequiredServices();
             }
 
+            BuildTheRequiredServices();
```
jennyf19, line 122: nice catch.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -358,10 +363,28 @@ private void LogAuthResult(AuthenticationResult? authenticationResult)
...
+                catch(Exception ex)
+                {
+                    Logger.TokenAcquisitionError(_logger, ex.Message, ex);
+                    throw;
+                }
+            }
+
             // Use MSAL to get the right token to call the API
-            var application = GetOrBuildConfidentialClientApplication(mergedOptions);
+            IConfidentialClientApplication application = GetOrBuildConfidentialClientApplication(mergedOptions);
```
jennyf19, line 385: >IConfidentialClientApplication [](http://example.com/codeflow?start=12&length=30)

would revert back to `var` you can save a small allocation by not creating the interface

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -358,10 +363,28 @@ private void LogAuthResult(AuthenticationResult? authenticationResult)
                 throw new ArgumentException(IDWebErrorMessage.ClientCredentialTenantShouldBeTenanted, nameof(tenant));
             }
 
+            // If using managed identity 
+            if (tokenAcquisitionOptions != null && tokenAcquisitionOptions.ManagedIdentity != null)
+            {
+                try
+                {
+                    IManagedIdentityApplication managedIdApp = await GetOrBuildManagedIdentityApplication(
```
jennyf19, line 371: >IManagedIdentityApplication [](http://example.com/codeflow?start=20&length=27)

try to use var instead to save an allocation

tests/Microsoft.Identity.Web.Test/TokenAcquisitionAuthorityTests.cs
```diff
@@ -1,10 +1,12 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System.Collections.Concurrent;
 using System.Collections.Generic;
 using System.Globalization;
 using System.Net.Http;
-using Microsoft.AspNetCore.Authentication;
+using System.Threading;
+using System.Threading.Tasks;
```
jennyf19, line 9: rmv using

tests/Microsoft.Identity.Web.Test/TokenAcquisitionAuthorityTests.cs
```diff
@@ -309,5 +309,109 @@ public void MergeExtraQueryParameters_MergedOptionsNull_Test()
...
+            ManagedIdentityOptions managedIdentityOptions = new()
+            {
+                UserAssignedClientId = clientId
+            };
+            MergedOptions mergedOptions = new();
+            BuildTheRequiredServices();
+            InitializeTokenAcquisitionObjects();
+
+            // Act
+            IManagedIdentityApplication app1 =
```
jennyf19, line 356: >IManagedIdentityApplication [](http://example.com/codeflow?start=12&length=27)

as mentioned earlier, this creates an additional allocation for interface, so just use var...I know it's less reader friendly, but just a good habit for perf. PS> i know this is a test :)

tests/Microsoft.Identity.Web.Test/TokenAcquisitionAuthorityTests.cs
```diff
@@ -309,5 +309,109 @@ public void MergeExtraQueryParameters_MergedOptionsNull_Test()
...
+            threadsDone.Wait();
+            IManagedIdentityApplication testApp = await _tokenAcquisition.GetOrBuildManagedIdentityApplication(mergedOptions, managedIdentityOptions);
+
+            // Assert
+            Assert.True(appsBag.Count == numThreads, "Not all threads put objects in the concurrent bag");
+            foreach (IManagedIdentityApplication app in appsBag)
+            {
+                Assert.Same(testApp, app);
+            }
+        }
```
jennyf19, line 415: >} [](http://example.com/codeflow?start=8&length=1)

really nice tests, Josh.

tests/Microsoft.Identity.Web.Test/TokenAcquisitionAuthorityTests.cs
```diff
@@ -309,5 +309,109 @@ public void MergeExtraQueryParameters_MergedOptionsNull_Test()
...
+            Assert.Same(app1, app2);
+        }
+
+        [Theory]
+        [InlineData("https://localhost:1234")]
+        [InlineData(null)]
+        public async void GetOrBuildManagedIdentity_TestConcurrencyAsync(string? clientId)
+        {
+            // Arrange
+            int numThreads = 20;
```
jennyf19, line 371: > 20 [](http://example.com/codeflow?start=28&length=3)

reason for 20? just curious.

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderProvider.cs
```diff
@@ -49,19 +49,9 @@ public async Task<string> CreateAuthorizationHeaderForAppAsync(string scopes, Au
...
-                ExtraQueryParameters = downstreamApiOptions?.AcquireTokenOptions.ExtraQueryParameters,
-                ForceRefresh = downstreamApiOptions?.AcquireTokenOptions.ForceRefresh ?? false,
-                LongRunningWebApiSessionKey = downstreamApiOptions?.AcquireTokenOptions.LongRunningWebApiSessionKey,
-                Tenant = downstreamApiOptions?.AcquireTokenOptions.Tenant,
-                UserFlow = downstreamApiOptions?.AcquireTokenOptions.UserFlow,
-                PopPublicKey = downstreamApiOptions?.AcquireTokenOptions.PopPublicKey,
-            };
+            TokenAcquisitionOptions tokenAcquisitionOptions = TokenAcquisitionOptions.CloneFromBaseClass(downstreamApiOptions?.AcquireTokenOptions);
+            tokenAcquisitionOptions.CancellationToken = cancellationToken;
+            return tokenAcquisitionOptions;
```
jennyf19, line 54: Why not just add 
`ManagedIdentity = downstreamApiOptions?.AcquireTokenOptions.ManagedIdentity`?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisitionOptions.cs
```diff
@@ -25,6 +26,47 @@ public class TokenAcquisitionOptions : AcquireTokenOptions
...
+            // This is my personal preference as it is most performant since no conversion would be needed. The drawback is I'd
+            // need to replace all instances of TokenAcquisitionOptions with AcquireTokenOptions in the codebase.
+            //      2.1 - I spoke with JM and he said this would be a longer conversation as he wants to deprecate
+            //      TokenAcquisitionOptions entirely and drop the CancellationToken and PoPConfiguration fields as they
+            //      apparently don't belong in AcquireTokenOptions.
+
+            // This is how we have been doing it so far (in the DefaultAuthorizationHeaderProvider class) I put it here to make
+            // it easier to find if we ever add more fields to the base class, but it will still require manual intervention to
+            // add the new fields here.
+            return new TokenAcquisitionOptions
```
jennyf19, line 55: Not quite right, and agree that we will deprecate these options at some point (hopefully soon for .NET 9)

Why not just do `ManagedIdentity = downstreamApiOptions?.AcquireTokenOptions.ManagedIdentity,` with the others?

jennyf19 (changes requested):
:clock1:

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
bgavrilMS: I think var just helps make the code more readable, it doesn't add any perf penalty. The compiler replaces "var" with the type at compile (build) time.

https://stackoverflow.com/questions/41479/use-of-var-keyword-in-c-sharp

Whatever you decide on this, there are settings in VS that help enforce it, see 

image

In MSAL we don't enforce this (we should) but the general opinion is that if the type is obvious, use `var` (e.g. `var x = new Account()`), else use the type.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
bgavrilMS, line 621: nit: for anybody else reviewing this - this type is in https://github.com/AzureAD/microsoft-identity-abstractions-for-dotnet

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -585,10 +608,82 @@ private bool IsInvalidClientCertificateOrSignedAssertionError(MsalServiceExcepti
...
 
+        /// <summary>
+        /// Gets the key value for the Managed Identity cache, the default key for system-assigned identity is used if there is
+        /// no clientId for a user-assigned identity specified. The method is internal rather than private for testing purposes.
+        /// </summary>
+        /// <param name="managedIdOptions">Holds the clientId for managed identity if none is present</param>
+        /// <returns>A key value for the Managed Identity cache</returns>
+        internal static string GetCacheKeyForManagedId(ManagedIdentityOptions managedIdOptions) 
+        {
+            if (managedIdOptions.UserAssignedClientId.IsNullOrEmpty())
```
bgavrilMS, line 677: This is going to be interesting....

So let's imagine that we want to add support for UAMI + object ID. You'd go to Abstractions and update `ManagedIdentityOptions` and add "UserAssignedObjectId" and release Abstractions. Then a few days later, release Id.Web with new dependency.

However, if a customer mismatches the version of Id.Web with the version of Abstractions, then they could get into a state where: 

- they set ManagedIdentityOptions's ObjectId  = "bla"
- older Id.Web doesn't know about "ObjectId" so it interpreset this as being System Assigned MSI (SAMI)
- if a SAMI or a single UAMI (even different the one specified) exists, a token for it will be returned. But not the expected one.

My recommendation would be to update ManagedIdentityOptions to: 


```cshapr
public class ManagedIdentityOptions
    
 [DefaultValue(null)]
 public string? UserAssignedClientId { get; set; }
  
 public bool IsSystemAssigned
 { 
      get{ return UserAssignedClientId != null; }
 }
```

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -674,6 +769,26 @@ private IConfidentialClientApplication BuildConfidentialClientApplication(Merged
...
+        /// Creates a managed identity client application.
+        /// </summary>
+        /// <param name="managedIdentityId">Indicates if system-assigned or user-assigned managed identity is used.</param>
+        /// <param name="enablePiiLogging">Indicates if logging that may contain personally identifiable information should be enabled.</param>
+        /// <returns>A managed identity application.</returns>
+        private IManagedIdentityApplication BuildManagedIdentityApplication(ManagedIdentityId managedIdentityId, bool enablePiiLogging)
+        {
+            return ManagedIdentityApplicationBuilder
+                .Create(managedIdentityId)
+                .WithHttpClientFactory(_httpClientFactory)
```
bgavrilMS, line 782: As discussed with @jmprieur , we'd like to bypass the HTTP pipeline customization for MSI. The reason for this that in MSIv2 (which @gladjohn is working on), there will be an MTLS connection happening, which doesn't play well with this extensibility point (we can go into details if needed)

The HTTP calls will be made to local service endpoints anyway, we don't think there are scenarios where apps would want to customize this. 

We can always add this on.

bgavrilMS (changes requested):
Please remove WithHttpClient modifier.
See my comment about `ManagedIdentityOptions` - this would be nice to have for future proofing, but not strictly required. 

Otherwise looks good! Great tests.

tests/Microsoft.Identity.Web.Test/Microsoft.Identity.Web.Test.csproj
JoshLozensky: thanks it is not needed any longer, I toyed with some classes in there but decided not to use it

tests/Microsoft.Identity.Web.Test/TokenAcquisitionAuthorityTests.cs
westin-m: I think avoiding magic numbers is good; this value could either be defined as a constant at the top of the file, or at least a comment on why 20 is used.

JoshLozensky: I agree I should mark it const but I don't want to put it at the top of the file vs. the top of the method because no other method uses it and then it is separated by 300+ lines from where it is used.

As far as 20 being a magic number, it is, but I don't know of any metric that could be used to determine an optimal number of threads for testing concurrency. Did you have any thoughts on it @westin-m?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
jennyf19: If we take this later, is the comment still blocking or what are you looking for as a resolution?

jennyf19: or just remove `WithHttpClient`...seems to be the suggestion in another comment?

tests/Microsoft.Identity.Web.Test/TokenAcquisitionAuthorityTests.cs
JoshLozensky: @jennyf19, right now my IDE is upset when I use var in this situation, do we have recommended settings so the IDE lines up with our preferred usage? Here are [the preference settings](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0007-ide0008).

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
JoshLozensky: I definitely agree and this is a neat way to handle the issue so I made [an ADO item for it](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2816904) since it requires a release in abstractions before being implemented, it is out of scope for this PR but should be done.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -585,10 +608,82 @@ private bool IsInvalidClientCertificateOrSignedAssertionError(MsalServiceExcepti
...
+                if (_managedIdentityApplicationsByClientId.TryGetValue(key, out application))
+                {
+                    return application;
+                }
+
+                // Set managedIdentityId to the correct value for either system or user assigned
+                ManagedIdentityId managedIdentityId;
+                if (key == SystemAssignedManagedIdentityKey)
+                {
+                    managedIdentityId = ManagedIdentityId.SystemAssigned;
```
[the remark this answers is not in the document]
JoshLozensky: [ManagedIdentityId](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/6f7017410e630dbe6851afd95b16e0520fbf77db/src/client/Microsoft.Identity.Client/AppConfig/ManagedIdentityId.cs) is from MSAL and is an argument in MSAL's [ManagedIdentityApplicationBuilder](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/6f7017410e630dbe6851afd95b16e0520fbf77db/src/client/Microsoft.Identity.Client/AppConfig/ManagedIdentityApplicationBuilder.cs)

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -18,12 +18,13 @@
 using Microsoft.Identity.Abstractions;
```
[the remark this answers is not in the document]
JoshLozensky: pulled it into its own file TokenAcquisition.ManagedIdentity.cs

tests/Microsoft.Identity.Web.Test/TokenAcquisitionAuthorityTests.cs
westin-m: > I don't want to put it at the top of the file vs. the top of the method

That's probably fine, I just say file because it's more conventional. Not too important.

> determine an optimal number of threads for testing concurrency

I can't say a ton about this, but I've seen others do three tests for concurrency: one thread (for correct behavior); _number of typical processors_ +  1  threads (depends on customers' production envs); and something large like 1000 threads to increase context switching and contention. You could swap 20 out for 33 (32 cores on a reasonable server + 1) to have a little more rhyme and reason, but feel free to find a second opinion.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
JoshLozensky: I mentioned in a different comment but [here are the docs for these options](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0007-ide0008), I agree we should align on one set as a team so it is unambiguous what to use and when.

tests/E2E Tests/TokenAcquirerTests/TokenAcquirer.cs
JoshLozensky: @jennyf19, I'm not sure what the ask is here... I thought customers would have these values from the appsettings.json config and not need to create their own options like I do here in the test and the method is essentially just to mock that config since I don't have an appsettings file to reference.

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderProvider.cs
jmprieur: I don't understand why you are introducing a new method here, @JoshLozensky?
As @jennyf19 asked, wasn't adding this line enough?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.ManagedIdentity.cs
```diff
@@ -0,0 +1,119 @@
...
+using Microsoft.Identity.Client.AppConfig;
+using Microsoft.IdentityModel.Tokens;
+
+namespace Microsoft.Identity.Web
+{
+    /// <summary>
+    /// Portion of the TokenAcquisition class that handles logic unique to managed identity.
+    /// </summary>
+#if NETSTANDARD2_0 || NET462 || NET472
+        internal partial class TokenAcquisition : ITokenAcquisitionInternal
```
jmprieur, line 18: I don't think that you need to re-write here that it implements this interface. Another partial already said it. I believe you can get rid of lines 17-19 included

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.ManagedIdentity.cs
```diff
@@ -0,0 +1,119 @@
...
+        /// Creates a managed identity client application.
+        /// </summary>
+        /// <param name="managedIdentityId">Indicates if system-assigned or user-assigned managed identity is used.</param>
+        /// <param name="enablePiiLogging">Indicates if logging that may contain personally identifiable information is enabled.</param>
+        /// <returns>A managed identity application.</returns>
+        private IManagedIdentityApplication BuildManagedIdentityApplication(ManagedIdentityId managedIdentityId, bool enablePiiLogging)
+        {
+            return ManagedIdentityApplicationBuilder
+                .Create(managedIdentityId)
+                .WithHttpClientFactory(_httpClientFactory)
```
jmprieur, line 93: I believe that for in the future it would be more beneficial to remove it. Bogdan can elaborate.

tests/E2E Tests/TokenAcquirerTests/TokenAcquirer.cs
jmprieur: Both are possible:
- customers can set it in the appsettings.json (in the DowntreamApis section)
-  but this is overridable by code (and therefore settable by code)

jmprieur (approved):
LGTM.
Nice job, @JoshLozensky  !

- I've the same question as Jenny for the TokenAcquisitionOptions. Let's do simple!
- Please add .zip to the .gitignore file: we really don't want to add zips of Mb to the repo, which we'll never look at!

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderProvider.cs
jennyf19: This still needs to be resolved before we can merge.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
jennyf19: I have no preference, @JoshLozensky can decide.

tests/IntegrationTests/PlaywrightTraces/B2CWebAppCallsWebApiLocally_TodoAppFunctionsCorrectly_net6.0.zip
jennyf19, line 1: I thought I had commented on these yesterday, but maybe it didn't save. These traces should not be included, you can add the PlaywriteTraces folder to the gitignore file.

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderProvider.cs
JoshLozensky: I was trying to uncouple the logic from the DefaultAuthorizationHeaderProvider and move it into the child class where it will be easier to find in the future to me it was really unintuitive that if I make changes to TokenAcquisitionOptions in abstractions I needed to also edit this ostensibly unrelated class in order for it to work. If it is in the child class it keeps the logic with the object it belongs to.

This isn't a hill to die on for me so if you'd prefer I'll just put it back @jennyf19 and @jmprieur.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisitionOptions.cs
JoshLozensky: See above comment, TLDR: it was a design decision to put the logic with the object it relates to rather than a random other class but if you'd prefer I put it back I can do that.

tests/Microsoft.Identity.Web.Test/TokenAcquisitionAuthorityTests.cs
JoshLozensky: Keegan recommended using ThreadPool.MaxThreads so I did that :)

keegan-caruso (changes requested):
zips are still in the change set.

jmprieur (approved):

.gitignore
```diff
@@ -350,3 +350,7 @@ MigrationBackup/
 # Ionide (cross platform F# VS Code tools) working folder
 .ionide/
 /tools/app-provisioning-tool/testwebapp
+
+# Playwright e2e testing trace files
+/tests/E2E Tests/PlaywrightTraces
+/tests/IntegrationTests/PlaywrightTraces
```
jennyf19, line 356: great! thank you

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderProvider.cs
```diff
@@ -58,6 +63,7 @@ private static TokenAcquisitionOptions CreateTokenAcquisitionOptionsFromApiOptio
                 ExtraQueryParameters = downstreamApiOptions?.AcquireTokenOptions.ExtraQueryParameters,
                 ForceRefresh = downstreamApiOptions?.AcquireTokenOptions.ForceRefresh ?? false,
                 LongRunningWebApiSessionKey = downstreamApiOptions?.AcquireTokenOptions.LongRunningWebApiSessionKey,
+                ManagedIdentity = downstreamApiOptions?.AcquireTokenOptions.ManagedIdentity,
```
jennyf19, line 66: this is great. we'll clean things up in the 3x.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.ManagedIdentity.cs
```diff
@@ -0,0 +1,114 @@
...
+            {
+                return SystemAssignedManagedIdentityKey;
+            }
+            else
+            {
+                return managedIdOptions.UserAssignedClientId!;
+            }
+        }
+    }
+}
```
jennyf19, line 114: awesome. love this.

jennyf19 (approved):
:shipit: Very nice work @JoshLozensky Looking forward to getting this out to customers very soon.

PR #2669, opened 2024-02-07
Stabilize UI and Unit Tests

JoshLozensky:
## Description

Addresses flakiness in UI tests as well as potential contention issues in the TokenAcquirer tests.

Ran the build pipeline 5 times with latest commit for testing, no failures encountered.

### DIFF
--- a/tests/E2E Tests/TokenAcquirerTests/TokenAcquirer.cs
+++ b/tests/E2E Tests/TokenAcquirerTests/TokenAcquirer.cs
@@ -21,7 +21,6 @@ using TaskStatus = System.Threading.Tasks.TaskStatus;
 namespace TokenAcquirerTests
 {
 #if !FROM_GITHUB_ACTION
-    [Collection("SerializeTestsUsingDefaultTokenAcquirerFactory")]
     public class TokenAcquirer
     {
         private static readonly string s_optionName = string.Empty;
@@ -40,7 +39,6 @@ namespace TokenAcquirerTests
         [Fact]
         public void TokenAcquirerFactoryDoesNotUseAspNetCoreHost()
         {
-            TokenAcquirerFactory.ResetDefaultInstance();
             TokenAcquirerFactory tokenAcquirerFactory = TokenAcquirerFactory.GetDefaultInstance();
             var serviceProvider = tokenAcquirerFactory.Build();
             var service = serviceProvider.GetService<ITokenAcquisitionHost>();
--- a/tests/E2E Tests/TokenAcquirerTests/TokenAcquirer.cs
+++ b/tests/E2E Tests/TokenAcquirerTests/TokenAcquirer.cs
@@ -313,7 +313,7 @@ namespace TokenAcquirerTests
             const string baseUrl = "https://vault.azure.net";
             const string clientId = "9c5896db-a74a-4b1a-a259-74c5080a3a6a";
             TokenAcquirerFactory tokenAcquirerFactory = TokenAcquirerFactory.GetDefaultInstance();
-            _ = tokenAcquirerFactory.Services;
+            //_ = tokenAcquirerFactory.Services;
             IServiceProvider serviceProvider = tokenAcquirerFactory.Build();
 
             // Act: Get the authorization header provider and add the options to tell it to use Managed Identity
--- a/tests/E2E Tests/TokenAcquirerTests/TokenAcquirer.cs
+++ b/tests/E2E Tests/TokenAcquirerTests/TokenAcquirer.cs
@@ -39,6 +39,7 @@ namespace TokenAcquirerTests
         [Fact]
         public void TokenAcquirerFactoryDoesNotUseAspNetCoreHost()
         {
+            TokenAcquirerFactory.ResetDefaultInstance();
             TokenAcquirerFactory tokenAcquirerFactory = TokenAcquirerFactory.GetDefaultInstance();
             var serviceProvider = tokenAcquirerFactory.Build();
             var service = serviceProvider.GetService<ITokenAcquisitionHost>();
@@ -312,8 +313,8 @@ namespace TokenAcquirerTests
             const string scope = "https://vault.azure.net/.default";
             const string baseUrl = "https://vault.azure.net";
             const string clientId = "9c5896db-a74a-4b1a-a259-74c5080a3a6a";
+            TokenAcquirerFactory.ResetDefaultInstance();
             TokenAcquirerFactory tokenAcquirerFactory = TokenAcquirerFactory.GetDefaultInstance();
-            //_ = tokenAcquirerFactory.Services;
             IServiceProvider serviceProvider = tokenAcquirerFactory.Build();
 
             // Act: Get the authorization header provider and add the options to tell it to use Managed Identity
--- a/tests/E2E Tests/TokenAcquirerTests/TokenAcquirer.cs
+++ b/tests/E2E Tests/TokenAcquirerTests/TokenAcquirer.cs
@@ -21,6 +21,7 @@ using TaskStatus = System.Threading.Tasks.TaskStatus;
 namespace TokenAcquirerTests
 {
 #if !FROM_GITHUB_ACTION
+    [Collection("SerializeTestsUsingDefaultTokenAcquirerFactory")]
     public class TokenAcquirer
     {
         private static readonly string s_optionName = string.Empty;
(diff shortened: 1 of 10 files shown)


tests/E2E Tests/WebAppUiTests/TestingWebAppLocally.cs
```diff
@@ -39,7 +39,7 @@ public TestingWebAppLocally(ITestOutputHelper output)
     public async Task ChallengeUser_MicrosoftIdFlow_LocalApp_ValidEmailPassword()
     {
         // Arrange
-        Process p = UiTestHelpers.StartProcessLocally(_uiTestAssemblyLocation, _devAppPath, _devAppExecutable);
+        Process? p = null;
```
jennyf19, line 42: nit: maybe just call this `process`?

jennyf19 (approved):
:shipit:

tests/E2E Tests/WebAppUiTests/TestingWebAppLocally.cs
JoshLozensky: done

kellyyangsong (approved):

tests/E2E Tests/TokenAcquirerTests/TokenAcquirer.cs
```diff
@@ -39,6 +39,7 @@ public TokenAcquirer()
         [Fact]
         public void TokenAcquirerFactoryDoesNotUseAspNetCoreHost()
         {
+            TokenAcquirerFactory.ResetDefaultInstance();
```
keegan-caruso, line 42: are these tests stateful?

bgavrilMS, line 42: Doesn't xunit provide a way to run some code before each test? Seems like you already have a ctor with this logic, but the ctor only runs once for all tests in this class.

https://stackoverflow.com/questions/46926852/xunit-constructor-runs-before-each-test

tests/E2E Tests/WebAppUiTests/UiTestNoParallelization.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Xunit;
+
+namespace WebAppUiTests
+{
+    [CollectionDefinition(nameof(UiTestNoParallelization), DisableParallelization = true)]
+    public class UiTestNoParallelization
```
bgavrilMS, line 9: Is it possible to use just a string ? Seems like all usages are `nameof(UiTestNoParallelization)` which is just "UiTestNoParallelization"

tests/E2E Tests/WebAppUiTests/TestingWebAppLocally.cs
```diff
@@ -18,8 +18,7 @@ namespace WebAppUiTests;
 #if !FROM_GITHUB_ACTION && !AZURE_DEVOPS_BUILD
 
 // since this test changes environment variables we'd prefer it not run at the same time as other tests
-[CollectionDefinition(nameof(TestingWebAppLocally), DisableParallelization = true)]
-[Collection("WebAppUiTests")]
+[CollectionDefinition(nameof(UiTestNoParallelization), DisableParallelization = true)]
```
bgavrilMS, line 21: How does this test change env variables? I am seeing a list of variables that are injected in the processes you start, but each process will maintain that. 

if the StartUrl is different (or can be made different), can you actually run some of these tests in parallel?

tests/E2E Tests/TokenAcquirerTests/TokenAcquirer.cs
JoshLozensky: I believe so, because they were failing on things like collections being altered so the enumeration fails or expecting things to be in cache but not finding them. It wasn't a problem before because this was the only test using it up until the

JoshLozensky: It does, but most of these tests aren't run in DevOps build so I didn't want to include them in the scope of this PR for the sake of the two that are.

tests/E2E Tests/WebAppUiTests/B2CWebAppCallsWebApiLocally.cs
```diff
@@ -77,8 +70,19 @@ public async Task Susi_B2C_LocalAccount_TodoAppFucntionsCorrectly()
...
+
             try
             {
+                // Start the web app and api processes.
+                // The delay before starting client prevents transient devbox issue where the client fails to load the first time after rebuilding.
+                // The delay after processes are started gives time to finish initial setup before attempted connection.
+                serviceProcess = UiTestHelpers.StartProcessLocally(_testAssemblyPath, _devAppPath + TC.s_todoListServicePath, TC.s_todoListServiceExe, serviceEnvVars);
+                Thread.Sleep(3000);
+                clientProcess = UiTestHelpers.StartProcessLocally(_testAssemblyPath, _devAppPath + TC.s_todoListClientPath, TC.s_todoListClientExe, clientEnvVars);
+                Thread.Sleep(5000);
```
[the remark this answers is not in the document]
JoshLozensky: The second wait was required as a result of the changes because Playwright was erroring out when the connection was refused as the apps were still starting up (afaik Playwright won't keep retrying in the case of refused connections, it retries for interactions with actual sites). However, I have toyed with the idea of putting it in a try catch loop to retry at a faster cadence to right-size the wait so I'll add that and see what you think.

tests/E2E Tests/WebAppUiTests/UiTestNoParallelization.cs
JoshLozensky: It can, I did it this way to make a clearer route for extensibility in the future as we could add collection level fixtures rather than class level. Would it be better to make it just a string?

tests/E2E Tests/WebAppUiTests/TestingWebAppLocally.cs
JoshLozensky: I can see how the comment lacks specificity, so I'll clean that up. Essentially, because we are launching multiple apps from the same environment that need to listen on different ports, it is the Kestrel:Endpoints:Http:Url that needs to change for each process that is started. This test only has one process so you don't see it getting set independently like in the other tests, however it does set this variable in the background which can cause a race condition when run in parallel with the other UI tests.

tests/E2E Tests/WebAppUiTests/B2CWebAppCallsWebApiLocally.cs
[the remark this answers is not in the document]
JoshLozensky: Doing the try-catch loop in local testing resulted in only 1 to 2 seconds spent waiting.

tests/E2E Tests/WebAppUiTests/TestingWebAppLocally.cs
bgavrilMS: You can still define a method that runs in a critical section (so only 1 thread) that starts the process to ensure there is no race condition on the env variables. 

That said, env variables seem complex as inter-process communication. Why not just start each processs with some command line arguments? 

note: i realize this may be outside the scope of this PR, so if needed pls open a new work item around test optimization.

tests/E2E Tests/WebAppUiTests/B2CWebAppCallsWebApiLocally.cs
```diff
@@ -77,16 +70,41 @@ public async Task Susi_B2C_LocalAccount_TodoAppFucntionsCorrectly()
...
+                while (InitialConnectionRetryCount > 0)
+                {
+                    try
+                    {
+                        await page.GotoAsync(TC.LocalhostUrl + TodoListClientPort);
+                        break;
+                    }
+                    catch (PlaywrightException ex)
+                    {
+                        Thread.Sleep(1000);
```
bgavrilMS, line 103: Don't use Thraed.Sleep for waiting. Use `await Task.Delay(1000)`. Thread.Sleep results in spin waiting. Task.Delay will cause .NET to pause your thread and give the CPU time to another thread. Might be quite helpful if you run tests in parallel.

tests/E2E Tests/WebAppUiTests/B2CWebAppCallsWebApiLocally.cs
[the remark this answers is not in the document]
bgavrilMS: I still see the second "wait 5 seconds" in the code.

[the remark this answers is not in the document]
JoshLozensky: Thanks, I missed taking out that line in one of the tests, removed now.

tests/E2E Tests/WebAppUiTests/B2CWebAppCallsWebApiLocally.cs
JoshLozensky: changed

tests/E2E Tests/WebAppUiTests/UiTestNoParallelization.cs
JoshLozensky: I appreciate the feedback but going to leave it as is.

PR #2685, opened 2024-02-27
Support for CIAM CUD

bgavrilMS:
Support for CIAM CUD

Example authority config: 

https://cats.ciamextensibility.com/123456-43bb-4ff9-89af-30ed8fe31c6d/v2.0/.well-known/openid-configuration?DC=test

Work: 

- [x] web app 
- [ ] web api
- [ ] a bunch of TODOs
- [ ] tests

@jmprieur, @jennyf19 - can you please let me know if this is going in the right direction?

Update:

- had to make a fix in MSAL to accept authority with query parameters (branch bogavril/ciamcud), fix already merged to main.
- with this new MSAL, I was able to get token in web app.

### DIFF
--- a/src/Microsoft.Identity.Web.TokenAcquisition/MicrosoftIdentityOptions.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/MicrosoftIdentityOptions.cs
@@ -49,7 +49,8 @@ namespace Microsoft.Identity.Web
 #endif
 
     /// <summary>
-    /// Options for configuring authentication using Azure Active Directory. It has both AAD and B2C configuration attributes.
+    /// Options for configuring authentication using Azure Active Directory. It supports several Identity Providers, 
+    /// including AAD, B2C and CIAM configuration attributes.
     /// </summary>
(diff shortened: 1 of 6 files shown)


src/Microsoft.Identity.Web.TokenAcquisition/MicrosoftIdentityOptions.cs
```diff
@@ -99,11 +100,20 @@ public class MicrosoftIdentityOptions : OpenIdConnectOptions
         /// <summary>
         /// Is considered B2C if the attribute SignUpSignInPolicyId is defined.
         /// </summary>
-        internal bool IsB2C
+        internal bool IsB2C // TODO: bogavril - how about an enum AuthorityType { AAD, B2C, CIAM, Oidc } instead?
```
jmprieur, line 103: This class is going to disappear in Id.Web 3.x
I would not change a line here.

jmprieur:
The first thing we need to do, is make sure we have tests for all the parts that will be touched.
Then, there are a few things that we'd like to change.
A design should come first, now that the spec is agreed.

Did you create git hub issue with the spec part for IdWeb?

PR #2696, opened 2024-03-08
Add support for custom user domains

westin-m:


src/Microsoft.Identity.Web/WebApiExtensions/MicrosoftIdentityWebApiAuthenticationBuilderExtensions.cs
```diff
@@ -176,7 +176,9 @@ public static class MicrosoftIdentityWebApiAuthenticationBuilderExtensions
                     // Process CIAM tenants
                     if (mergedOptions.Authority != null)
                     {
-                        mergedOptions.Authority = AuthorityHelpers.BuildCiamAuthorityIfNeeded(mergedOptions.Authority);
+                        bool preserveAuthority;
```
jennyf19, line 179: let's inline this variable declaration -> `out bool preserveAuthority` instead

src/Microsoft.Identity.Web/WebAppExtensions/MicrosoftIdentityWebAppAuthenticationBuilderExtensions.cs
```diff
@@ -312,7 +312,9 @@ public static class MicrosoftIdentityWebAppAuthenticationBuilderExtensions
 
                     if (mergedOptions.Authority != null)
                     {
-                        mergedOptions.Authority = AuthorityHelpers.BuildCiamAuthorityIfNeeded(mergedOptions.Authority);
+                        bool preserveAuthority;
```
jennyf19, line 315: let's inline this variable declaration -> `out bool preserveAuthority` instead

tests/Microsoft.Identity.Web.Test/AuthorityHelpersTests.cs
```diff
@@ -76,10 +76,12 @@ public void BuildAuthority_CiamAuthority_ReturnsValidAadAuthority()
             };
             string expectedResult = options.Authority + TestConstants.Domain;
 
-            string? result = AuthorityHelpers.BuildCiamAuthorityIfNeeded(options.Authority);
+            bool preserveAuthority;
+            string? result = AuthorityHelpers.BuildCiamAuthorityIfNeeded(options.Authority, out preserveAuthority);
```
jennyf19, line 80: same -> let's inline this variable declaration

tests/Microsoft.Identity.Web.Test/AuthorityHelpersTests.cs
```diff
@@ -91,10 +93,12 @@ public void BuildAuthority_CiamAuthorityWithTenant_ReturnsValidAadAuthority()
             };
             string expectedResult = options.Authority ;
 
-            string? result = AuthorityHelpers.BuildCiamAuthorityIfNeeded(options.Authority);
+            bool preserveAuthority;
+            string? result = AuthorityHelpers.BuildCiamAuthorityIfNeeded(options.Authority, out preserveAuthority);
```
jennyf19, line 97: same -> let's inline this variable declaration

jennyf19 (approved):
:shipit:

src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
```diff
@@ -49,6 +49,7 @@ public ConfidentialClientApplicationOptions ConfidentialClientApplicationOptions
         public string? RedirectUri { get; set; }
         public bool EnableCacheSynchronization { get; set; }
         internal bool MergedWithCca { get; set; }
+        internal bool PreserveAuthority { get; set; }
```
bgavrilMS, line 52: Consider adding some comments and a link to the work item here. We will forget why we ever needed this.

src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
```diff
@@ -426,7 +427,7 @@ internal static void ParseAuthorityIfNecessary(MergedOptions mergedOptions)
                 if (indexTenant >= 0)
                 {
                     mergedOptions.Instance = authority.Substring(0, indexTenant);
-                    mergedOptions.TenantId = authority.Substring(indexTenant + 1);
+                    mergedOptions.TenantId = mergedOptions.PreserveAuthority ? null : authority.Substring(indexTenant + 1);
```
bgavrilMS, line 430: I'm not sure about this. Let's take a CIAM CUD authority

`https://cats.com/12345/v2.0/`

then the `Instance` would be `https://cats.com/` and the `TenantId` would be null.

Where is "Instance" used elsewhere in the code?

src/Microsoft.Identity.Web/AuthorityHelpers.cs
```diff
@@ -34,7 +34,7 @@ internal static string EnsureAuthorityIsV2(string authority)
             return authority;
         }
 
-        internal static string? BuildCiamAuthorityIfNeeded(string? authority)
+        internal static string? BuildCiamAuthorityIfNeeded(string? authority, out bool preserveAuthority)
```
bgavrilMS, line 37: In the spirit of "Single responsability principle" (the S in SOLID), I would rename this new arg as "isCiamCudAuthority", i.e. the authority helper BuildCiamAuthorityIfNeeded should not need to know about concepts like "preseve authority".

src/Microsoft.Identity.Web/WebApiExtensions/MicrosoftIdentityWebApiAuthenticationBuilderExtensions.cs
```diff
@@ -176,7 +176,8 @@ public static class MicrosoftIdentityWebApiAuthenticationBuilderExtensions
                     // Process CIAM tenants
                     if (mergedOptions.Authority != null)
                     {
-                        mergedOptions.Authority = AuthorityHelpers.BuildCiamAuthorityIfNeeded(mergedOptions.Authority);
+                        mergedOptions.Authority = AuthorityHelpers.BuildCiamAuthorityIfNeeded(mergedOptions.Authority, out bool preserveAuthority);
+                        mergedOptions.PreserveAuthority = preserveAuthority;
```
bgavrilMS, line 180: This code, the "orchestrator" knows about `preserveAuthority`

src/Microsoft.Identity.Web/WebApiExtensions/MicrosoftIdentityWebApiAuthenticationBuilderExtensions.cs
```diff
@@ -176,7 +176,8 @@ public static class MicrosoftIdentityWebApiAuthenticationBuilderExtensions
                     // Process CIAM tenants
```
bgavrilMS, line 176: I think this does more than process CIAM, it can process any OIDC compliant authority.

tests/Microsoft.Identity.Web.Test/AuthorityHelpersTests.cs
```diff
@@ -76,10 +76,11 @@ public void BuildAuthority_CiamAuthority_ReturnsValidAadAuthority()
...
 
-            string? result = AuthorityHelpers.BuildCiamAuthorityIfNeeded(options.Authority);
+            string? result = AuthorityHelpers.BuildCiamAuthorityIfNeeded(options.Authority, out bool preserveAuthority);
 
             Assert.NotNull(result);
             Assert.Equal(expectedResult, result);
+            Assert.False(preserveAuthority);
         }
 
         [Fact]
```
bgavrilMS, line 86: Is there any way to get some coverage over the changes in  UpdateMergedOptionsFromMicrosoftIdentityOptions ?

src/Microsoft.Identity.Web/AuthorityHelpers.cs
westin-m: @jennyf19 do you have any thoughts on renaming this method?

src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
westin-m: In this code and after this method is called, `Instance` is stored in the same field of `confidentialClientApplicationOptions`.

src/Microsoft.Identity.Web/AuthorityHelpers.cs
jmprieur: But if we wanted to, BuildCiamOrOidcAuthority is ok.

tests/Microsoft.Identity.Web.Test/AuthorityHelpersTests.cs
westin-m: Yes, I would just need some support in understanding what the different cases look like for CIAM CUD

src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
jennyf19: For CIAM, just like for B2C, customers will use `"authority"` not `Instance` and `TenantId`. CIAM would not use `Instance`

src/Microsoft.Identity.Web/AuthorityHelpers.cs
trwalke: I am leaning more towards what @jennyf19 said. This method will only perform an action if the authority contains a Ciam suffix. but if it does not, then it does not build anything really. thus, the if needed is more like, "if we need to build then do it, otherwise, it does not."

src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
trwalke: I am trying to understand the logic of PreserveAuthority exactly, it seems to imply that the authority is not Ciam if set to true right? In that case, wouldnt we want to use the TenantId from the authority like we do in the original code?

Why wouldnt this line be 

`mergedOptions.PreserveAuthority ? authority.Substring(indexTenant + 1) : null;`

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -650,7 +650,12 @@ private async Task<IConfidentialClientApplication> BuildConfidentialClientApplic
 
                 string authority;
 
-                if (mergedOptions.IsB2C)
+                if (mergedOptions.PreserveAuthority && !string.IsNullOrEmpty(mergedOptions.Authority))
+                {
+                    authority = mergedOptions.Authority!;
+                    builder.WithGenericAuthority(authority);
```
trwalke, line 656: This will be switched to `WithOidcAuthority` right? Im assuming this is just for testing?

src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
bgavrilMS: Ok, then at least let's add a comment to "Instance". 

IMO, if a property isn't supposed to be used / trusted, there should be some protection against it. I'm worried we will forget about this convetion and we will use "Instance" later in some other part of the code that will blow up.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
bgavrilMS: `WithGenericAuthority` works today with MSAL 4.59. We soft deprecated this in MSAL 4.60, but this is yet to be released. In any case, it's just a rename. 

What doesn't work with MSAL 4.59 is an authority with extra query parameters, e.g. "https://cat.com/1235/v2.0?test=foo", but this is no longer needed (it used to be, the CIAM CUD authorities were behind a feature flag).

src/Microsoft.Identity.Web/WebApiExtensions/MicrosoftIdentityWebApiAuthenticationBuilderExtensions.cs
bgavrilMS: It was related to renaming `BuildCiamAuthority" which was discussed. Closing this.

tests/Microsoft.Identity.Web.Test/AuthorityHelpersTests.cs
bgavrilMS: CIAM (non-CUD) authority can be one of the following: 

- https://westin.ciamlogin.com
- https://westin.ciamlogin.com/{GUID} (the GUID is the tenant id)
- https://westin.ciamlogin.com/westin.onmicrosoft.com  (display name of the tenant)

CIAM CUD authority will be: 

- https://westin.com/{GUID}/v2.0

bgavrilMS: Not sure this helps, happy to chat directly about this if needed.

src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
bgavrilMS: I'm reopening this. `Instance` is used today in managing the dictionary of "CCA" objects, it's part of the key there. 

So there could be a potential conflict once we start supporting other generic authorities, for example: 

`https://demo.duendesoftware.com/A` vs `https://demo.duendesoftware.com/B` would both have the same instance and "Tenant" = null. 

IMO, if preseve authority = true then: 

- instance = authority
- tenant = null

FuPingFranco (approved):

bgavrilMS (approved):

PR #2708, opened 2024-03-16
Replace the user assigned managed identity in tests, by another one

jmprieur:
# Fix User assigned managed identity tests

The user assigned managed identity used in the release builds for the tests has changed.
@jennyf19 we might also want to take this in rel/v2?

running a release build to check the fix.


jennyf19 (approved):
You'll need to put in rel/v2 as you mention

Issue #2183, opened 2023-04-05
Razor Pages are entirely missing from DLLs since 2.0.2-preview


ascott18:
### Microsoft.Identity.Web Library

Microsoft.Identity.Web.UI

### Microsoft.Identity.Web version

2.7.0

### Web app

Sign-in users

### Web API

Not Applicable

### Token cache serialization

Not Applicable

### Description

Since version 2.0.2-preview of this library, the compiled Razor Pages are entirely missing from the nuget package for .net6.0+ targets.

This also obviously prevents these pages from being routed to, since the endpoints for the razor pages also doesn't exist, the URL generated [here](https://github.com/AzureAD/microsoft-identity-web/blob/13c77c31d46af7d6b19ac28279a2137d2c0e348f/src/Microsoft.Identity.Web.UI/Areas/MicrosoftIdentity/Controllers/AccountController.cs#L129) does not know of a page identified by `/Account/SignedOut`.

### Reproduction steps

1. Download Microsoft.Identity.Web.UI from nuget.org, any version from 2.0.2-preview onward
2. Open the nuget package with your tool of choice
3. Observe that Microsoft.Identity.Web.UI.Views.dll is not present in the nuget packages for net6.0 and net7.0 (but it is present for netcoreapp3.1)

![image](https://user-images.githubusercontent.com/5017521/230213200-fc864deb-7f79-4ba8-a5dd-b672f223a319.png)

### Relevant code snippets

```csharp
No code needed - issue is readily apparently from the nuget package contents.

Instantiate the ASP.NET Web App template from Visual Studio and select Microsoft Identity Platform as the authentication type. Then upgrade the Microsoft.Identity packages to the latest, remove the FallbackPolicy from Program.cs, and observe that https://localhost:7112/MicrosoftIdentity/Account/SignedOut returns a 404, while previous versions show a page with a message.
```

### Regression

2.0.0-beta, all 1.x versions

### Expected behavior

The razor pages exist

jennyf19:
We need to look into this:
[View components in ASP.NET Core | Microsoft Learn](https://learn.microsoft.com/en-us/aspnet/core/mvc/views/view-components?view=aspnetcore-7.0)
[Razor file compilation in ASP.NET Core | Microsoft Learn](https://learn.microsoft.com/en-us/aspnet/core/mvc/views/view-compilation?view=aspnetcore-7.0&tabs=visual-studio) 

[...]

PR #2712, opened 2024-03-18
rm netcoreapp3.1 and update UI targets

### LINKED ISSUE #2262: netcoreapp3.1 should probably be dropped
.net core 3.1 went out of support on Dec 13, 2022 as documented in https://learn.microsoft.com/en-us/lifecycle/products/microsoft-net-and-net-core 

Targeting of it should probably be dropped, or at least the planning of the next major version should consider dropping it.

kellyyangsong:
Fixes #2262


src/Microsoft.Identity.Web.UI/Microsoft.Identity.Web.UI.csproj
```diff
@@ -7,7 +7,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp3.1; net5.0</TargetFrameworks>
+    <TargetFrameworks>net6.0; net7.0; net8.0</TargetFrameworks>
```
keegan-caruso, line 10: can you explain a bit more?

src/Microsoft.Identity.Web.UI/Microsoft.Identity.Web.UI.csproj
```diff
@@ -23,4 +23,8 @@
     <ProjectReference Include="..\Microsoft.Identity.Web\Microsoft.Identity.Web.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
```
keegan-caruso, line 26: ?

kellyyangsong: I think this was a vestige of when I was trying to resolve the errors associated with IdWeb.UI target frameworks. Removing!

kellyyangsong: Hmm, sorry. Looks like this is necessary. Without: 
image

src/Microsoft.Identity.Web.UI/Microsoft.Identity.Web.UI.csproj
kellyyangsong: image
^ With only .net5

image
^ With .net5, .net6, .net7. ,net8

And with .net6, .net7, .net8 there are no errors :/

kellyyangsong: + this may be appropriate since we're moving to IdWeb 3x

src/Microsoft.Identity.Web.TokenAcquisition/CompatibilitySuppressions.xml
jennyf19, line 1: This should be removed but regenerated. We need this file to identify the breaking changes
`dotnet pack /p:GenerateCompatibilitySuppressionFile=true`

src/Microsoft.Identity.Web.UI/Microsoft.Identity.Web.UI.csproj
jennyf19: we need to check if this is still an issue:
https://github.com/AzureAD/microsoft-identity-web/issues/2183 
before making this change

src/Microsoft.Identity.Web.TokenAcquisition/CompatibilitySuppressions.xml
kellyyangsong: Done, thank you!!

kellyyangsong: Is it appropriate to regenerate these files for the entire project?

src/Microsoft.Identity.Web.UI/Microsoft.Identity.Web.UI.csproj
jmprieur: and this needs to go through a release build (as the issue was only seen on signed code)

src/Microsoft.Identity.Web.TokenAcquisition/CompatibilitySuppressions.xml
jmprieur: you mean all the IsPackable projects in the solution? yes.

PR #2722, opened 2024-03-26
Don't run b2c tests on ado

keegan-caruso:
# Don't run b2c tests on ado

## Description

Test is flaky and fails often


jennyf19 (approved):
:shipit:

MZOLN (approved):

Issue #2698, opened 2024-03-11
E2E testing of CIAM (non-CUD) in Id.Web


bgavrilMS:
Please have a look at existing E2E tests (e.g. the B2C ones). Ideally we will have tests for web app, web api and daemon.

PR #2728, opened 2024-03-27
Adding web app and web app call web api CIAM non CUD tests

trwalke:
# Adding web app and web app call web api CIAM non CUD tests

## Description
Work in progress.
https://github.com/AzureAD/microsoft-identity-web/issues/2698

Adding web app and web app call web api CIAM non CUD tests
Web app calls web api
web app calls graph
Deamon app calls graph
pending lab api integration

{Detail}

### DIFF
--- a/tests/DevApps/WebAppCallsMicrosoftGraph/appsettings.json
+++ b/tests/DevApps/WebAppCallsMicrosoftGraph/appsettings.json
@@ -17,15 +17,15 @@
             //    "SourceType": "SignedAssertionFromManagedIdentity",
             //    "ManagedIdentityClientId": ""
             //},
-            {
-                "SourceType": "KeyVault",
-                "KeyVaultUrl": "https://webappsapistests.vault.azure.net",
-                "KeyVaultCertificateName": "Self-Signed-5-5-22"
-            }
-            // {
-            //    "SourceType": "ClientSecret",
-            //    "ClientSecret": ""
-            // }
+            //{
+            //    "SourceType": "KeyVault",
+            //    "KeyVaultUrl": "https://webappsapistests.vault.azure.net",
+            //    "KeyVaultCertificateName": "Self-Signed-5-5-22"
+            //}
+             {
+                "SourceType": "ClientSecret",
+                "ClientSecret": ""
+             }
         ]
 
         // Id.Web v1.0 way of declaring client credentials
--- a/tests/DevApps/WebAppCallsMicrosoftGraph/appsettings.json
+++ b/tests/DevApps/WebAppCallsMicrosoftGraph/appsettings.json
@@ -17,15 +17,15 @@
             //    "SourceType": "SignedAssertionFromManagedIdentity",
             //    "ManagedIdentityClientId": ""
             //},
-            //{
-            //    "SourceType": "KeyVault",
-            //    "KeyVaultUrl": "https://webappsapistests.vault.azure.net",
-            //    "KeyVaultCertificateName": "Self-Signed-5-5-22"
-            //}
-             {
-                "SourceType": "ClientSecret",
-                "ClientSecret": ""
-             }
+            {
+                "SourceType": "KeyVault",
+                "KeyVaultUrl": "https://webappsapistests.vault.azure.net",
+                "KeyVaultCertificateName": "Self-Signed-5-5-22"
+            }
+             //{
+             //   "SourceType": "ClientSecret",
+             //   "ClientSecret": ""
+             //}
         ]
 
         // Id.Web v1.0 way of declaring client credentials
--- a/tests/DevApps/ciam/myWebApi/myWebApi.csproj
+++ b/tests/DevApps/ciam/myWebApi/myWebApi.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
-    <TargetFrameworks>net7.0</TargetFrameworks>
+    <TargetFrameworks>net8.0</TargetFrameworks>
     <Nullable>enable</Nullable>
     <ImplicitUsings>enable</ImplicitUsings>
     <UserSecretsId>aspnet-myWebApi-17be892d-3bb3-4282-92e2-301b02c86ed6</UserSecretsId>
--- a/tests/DevApps/ciam/myWebApp/myWebApp.csproj
+++ b/tests/DevApps/ciam/myWebApp/myWebApp.csproj
@@ -5,6 +5,7 @@
     <Nullable>enable</Nullable>
     <ImplicitUsings>enable</ImplicitUsings>
     <UserSecretsId>aspnet-myWebApp-23b8728b-775f-4796-8b63-e85cbaccc27c</UserSecretsId>
+    <UseWIP>true</UseWIP>
   </PropertyGroup>
 
   <ItemGroup Condition="'$(UseWIP)' == 'false' ">
--- a/tests/E2E Tests/TokenAcquirerTests/TokenAcquirer.cs
+++ b/tests/E2E Tests/TokenAcquirerTests/TokenAcquirer.cs
@@ -31,6 +31,13 @@ namespace TokenAcquirerTests
                 "Self-Signed-5-5-22")
         };
 
+        private static readonly CredentialDescription[] s_ciamClientCredentials = new[]
+{
+            CertificateDescription.FromKeyVault(
+                "https://buildautomation.vault.azure.net",
+                "AzureADIdentityDivisionTestAgentCert")
+        };
+
         public TokenAcquirer()
         {
             TokenAcquirerFactory.ResetDefaultInstance(); // Test only
@@ -87,6 +94,24 @@ namespace TokenAcquirerTests
                 option.TenantId = "msidentitysamplestesting.onmicrosoft.com";
                 option.ClientId = "6af093f3-b445-4b7a-beae-046864468ad6";
                 option.ClientCredentials = s_clientCredentials;
+            });
+
+            await CreateGraphClientAndAssert(tokenAcquirerFactory, services);
+        }
+
+        [IgnoreOnAzureDevopsFact]
+        //[Fact]
+        public async Task AcquireToken_WithMicrosoftIdentityApplicationOptions_ClientCredentialsCiamAsync()
+        {
+            TokenAcquirerFactory tokenAcquirerFactory = TokenAcquirerFactory.GetDefaultInstance();
+            IServiceCollection services = tokenAcquirerFactory.Services;
+
+            services.Configure<MicrosoftIdentityApplicationOptions>(s_optionName, option =>
+            {
+                option.Instance = "https://MSIDLABCIAM6.ciamlogin.com";
+                option.TenantId = "fe362aec-5d43-45d1-b730-9755e60dc3b9";
+                option.ClientId = "b244c86f-ed88-45bf-abda-6b37aa482c79";
+                option.ClientCredentials = s_ciamClientCredentials;
             });
 
             await CreateGraphClientAndAssert(tokenAcquirerFactory, services);
--- a/tests/E2E Tests/TokenAcquirerTests/TokenAcquirer.cs
+++ b/tests/E2E Tests/TokenAcquirerTests/TokenAcquirer.cs
@@ -99,7 +99,7 @@ namespace TokenAcquirerTests
             await CreateGraphClientAndAssert(tokenAcquirerFactory, services);
         }
 
-        [IgnoreOnAzureDevopsFact]
+        [IgnoreOnAzureDevopsFact(Skip = "https://github.com/AzureAD/microsoft-identity-web/issues/2732")]
         //[Fact]
         public async Task AcquireToken_WithMicrosoftIdentityApplicationOptions_ClientCredentialsCiamAsync()
         {
@@ -108,10 +108,9 @@ namespace TokenAcquirerTests
 
             services.Configure<MicrosoftIdentityApplicationOptions>(s_optionName, option =>
             {
-                option.Instance = "https://MSIDLABCIAM6.ciamlogin.com";
-                option.TenantId = "fe362aec-5d43-45d1-b730-9755e60dc3b9";
+                option.Authority = "https://MSIDLABCIAM6.ciamlogin.com";
                 option.ClientId = "b244c86f-ed88-45bf-abda-6b37aa482c79";
-                option.ClientCredentials = s_ciamClientCredentials;
+                option.ClientCredentials = s_clientCredentials;
             });
 
             await CreateGraphClientAndAssert(tokenAcquirerFactory, services);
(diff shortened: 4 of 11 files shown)


tests/DevApps/WebAppCallsMicrosoftGraph/appsettings.json
```diff
@@ -17,15 +17,15 @@
...
-            {
-                "SourceType": "KeyVault",
-                "KeyVaultUrl": "https://webappsapistests.vault.azure.net",
-                "KeyVaultCertificateName": "Self-Signed-5-5-22"
-            }
-            // {
-            //    "SourceType": "ClientSecret",
-            //    "ClientSecret": ""
-            // }
+            //{
```
trwalke, line 20: temporary change for testing. will create Cert in lab for CIAM app

tests/E2E Tests/WebAppUiTests/WebAppCallsApiCallsGraphLocally.cs
```diff
@@ -172,6 +176,120 @@ public async Task ChallengeUser_MicrosoftIdFlow_LocalApp_ValidEmailPasswordCreds
                 playwright.Dispose();
             }
         }
+
+        [Fact]
+        [SupportedOSPlatform("windows")]
+        public async Task ChallengeUser_MicrosoftIdFlow_LocalApp_ValidEmailPasswordCreds_TodoAppFunctionsCorrectlyWithCiam()
```
bgavrilMS, line 182: What is the difference between this test and the prev one? Does this call the web api and the prev one just "signs in"

jennyf19:
Please make sure the branch is updated and would like @JoshLozensky to review before this is merged. They will be OOF until Monday, just FYI. Thanks.

tests/DevApps/WebAppCallsMicrosoftGraph/appsettings.json
JoshLozensky: Has this run successfully with a hard-coded cert? I'm not able to run it myself without a secret.

tests/E2E Tests/WebAppUiTests/TestingWebAppLocally.cs
```diff
@@ -103,5 +103,86 @@ public async Task ChallengeUser_MicrosoftIdFlow_LocalApp_ValidEmailPassword()
...
+    [Fact]
+    [SupportedOSPlatform("windows")]
+    public async Task ChallengeUser_MicrosoftIdFlow_LocalApp_ValidEmailWithCiamPassword()
+    {
+        // Arrange
+        const string _clientSecret = "...";
+        Process? process = null;
+        const string TraceFileName = TraceFileClassName + "_ValidEmailPassword";
+        using IPlaywright playwright = await Playwright.CreateAsync();
+        IBrowser browser = await playwright.Chromium.LaunchAsync(new() { Headless = false });
```
JoshLozensky, line 116: Probably best to default headless to true so tests run faster and there aren't browsers popping up when people run Id Web's tests.

tests/E2E Tests/WebAppUiTests/TestingWebAppLocally.cs
```diff
@@ -103,5 +103,86 @@ public async Task ChallengeUser_MicrosoftIdFlow_LocalApp_ValidEmailPassword()
...
     }
+
+    [Fact]
+    [SupportedOSPlatform("windows")]
+    public async Task ChallengeUser_MicrosoftIdFlow_LocalApp_ValidEmailWithCiamPassword()
+    {
+        // Arrange
+        const string _clientSecret = "...";
+        Process? process = null;
+        const string TraceFileName = TraceFileClassName + "_ValidEmailPassword";
```
JoshLozensky, line 114: This needs to be changed or it will overwrite the other test's trace files. Each test should have a unique name.

JoshLozensky:
If we are mostly doing the same tasks with the same programs, should we instead make these theories and split out the unique fields into theory data?

JoshLozensky (changes requested):
Additionally, since these tests can be transiently unstable in the build pipeline while appearing fine locally, please test them multiple times if not adding a skip.

tests/E2E Tests/WebAppUiTests/WebAppCallsApiCallsGraphLocally.cs
trwalke: Both of the tests call the down stream api, the previous test simply adds additional clicks in the app for the todo list feature which is not present in this app.

tests/DevApps/WebAppCallsMicrosoftGraph/appsettings.json
trwalke: Not with cert but it does with a secret, will be updating with new app/credentials shortly

tests/E2E Tests/WebAppUiTests/WebAppCallsApiCallsGraphLocally.cs
bgavrilMS: Is it worth having both? Is one a superset of the other?

tests/E2E Tests/WebAppUiTests/TestingWebAppLocally.cs
bgavrilMS: For sure, this probably just a forgotten setting.

tests/E2E Tests/TokenAcquirerTests/TokenAcquirer.cs
```diff
@@ -92,6 +99,24 @@ public async Task AcquireToken_WithMicrosoftIdentityApplicationOptions_ClientCre
...
+        [IgnoreOnAzureDevopsFact]
+        //[Fact]
+        public async Task AcquireToken_WithMicrosoftIdentityApplicationOptions_ClientCredentialsCiamAsync()
+        {
+            TokenAcquirerFactory tokenAcquirerFactory = TokenAcquirerFactory.GetDefaultInstance();
+            IServiceCollection services = tokenAcquirerFactory.Services;
+
+            services.Configure<MicrosoftIdentityApplicationOptions>(s_optionName, option =>
+            {
+                option.Instance = "https://MSIDLABCIAM6.ciamlogin.com";
```
bgavrilMS, line 111: With CIAM we would not be supporting the `Instance` and `TenantId` configuration. We expect ppl to configure `Authority` instead.

tests/E2E Tests/WebAppUiTests/TestingWebAppLocally.cs
```diff
@@ -103,5 +103,86 @@ public async Task ChallengeUser_MicrosoftIdFlow_LocalApp_ValidEmailPassword()
...
+        IBrowser browser = await playwright.Chromium.LaunchAsync(new() { Headless = false });
+        IBrowserContext context = await browser.NewContextAsync(new BrowserNewContextOptions { IgnoreHTTPSErrors = true });
+        await context.Tracing.StartAsync(new() { Screenshots = true, Snapshots = true, Sources = true });
+
+        try
+        {
+            var clientEnvVars = new Dictionary<string, string>
+            {
+                {"AzureAd__ClientId", "16ee75bf-ebbe-42a8-9474-caa0fe856bc7"},//"Authority": "https://nativeauthasampleapp.ciamlogin.com"
+                {"AzureAd__Instance", "https://nativeauthasampleapp.ciamlogin.com"},
```
bgavrilMS, line 125: Same as above, expecting the Authority to be set (for both CUD and non-CUD). Instance + TenantID not expected to work.

tests/DevApps/WebAppCallsMicrosoftGraph/appsettings.json
```diff
@@ -17,15 +17,15 @@
             //    "SourceType": "SignedAssertionFromManagedIdentity",
```
bgavrilMS, line 17: @trwalke - Did you actually get the CIAM web app to get a token and call Graph? I am actually unsure if that scenario works today (it used to not work).

tests/DevApps/ciam/myWebApi/myWebApi.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk.Web">
 
   <PropertyGroup>
-    <TargetFrameworks>net7.0</TargetFrameworks>
+    <TargetFrameworks>net8.0</TargetFrameworks>
```
bgavrilMS, line 4: unrelated?

tests/E2E Tests/WebAppUiTests/WebAppCallsApiCallsGraphLocally.cs
```diff
@@ -25,20 +25,24 @@ public class WebAppCallsApiCallsGraphLocally : IClassFixture<InstallPlaywrightBr
...
         private readonly string _testAssemblyLocation = typeof(WebAppCallsApiCallsGraphLocally).Assembly.Location;
+        private readonly string _clientSecret = "...";
         private readonly ITestOutputHelper _output;
 
         public WebAppCallsApiCallsGraphLocally(ITestOutputHelper output)
         {
             _output = output;
         }
 
-        [Fact(Skip = "https://github.com/AzureAD/microsoft-identity-web/issues/2716")]
```
bgavrilMS, line 41: Was this fixed ?

tests/E2E Tests/WebAppUiTests/WebAppCallsApiCallsGraphLocally.cs
bgavrilMS: I am not sure this correct. I think the names are confusing. 

TestingWebAppLocally - tests just web app scenario (AcquireTokenByAuthcode / AcquireTokenSilent)
TestingWebAppCallsApiCallGraph - tests web app and web api scenario (so OBO too).

tests/E2E Tests/WebAppUiTests/TestingWebAppLocally.cs
trwalke: Yes, this is for testing. will be changed back

tests/E2E Tests/WebAppUiTests/WebAppCallsApiCallsGraphLocally.cs
trwalke: I agree, name will be updated. This test involves a web app calling a downstream web api (locally)

tests/DevApps/WebAppCallsMicrosoftGraph/appsettings.json
trwalke: Yes, it worked when testing locally
image

tests/E2E Tests/WebAppUiTests/WebAppCallsApiCallsGraphLocally.cs
trwalke: No, I will remove this change

tests/DevApps/WebAppCallsMicrosoftGraph/appsettings.json
trwalke: however, see https://github.com/AzureAD/microsoft-identity-web/issues/2732

tests/E2E Tests/WebAppUiTests/TestingWebAppLocally.cs
trwalke: Makes sense. Updated.

tests/DevApps/ciam/myWebApi/myWebApi.csproj
trwalke: When the file path for the test apps is calculated, it expects the file to be in the .net 8 target folder. The CIAM app was using net 7 however so I figured the path of least resistance would be to update the app to .NET 8. App appears to be functioning the same.

bgavrilMS: Kind of brittle to have the tfm in the output dir...

tests/E2E Tests/WebAppUiTests/WebAppCallsApiCallsGraphLocally.cs
```diff
@@ -172,6 +157,125 @@ public async Task ChallengeUser_MicrosoftIdFlow_LocalApp_ValidEmailPasswordCreds
...
+                _output.WriteLine("Web app sign-in flow successful using Sign in button after sign out.");
+            }
+            catch (Exception ex)
+            {
+                Assert.Fail($"the UI automation failed: {ex} output: {ex.Message}.");
+            }
+            finally
+            {
+                // Add the following to make sure all processes and their children are stopped.
+                Queue<Process> processes = new();
```
bgavrilMS, line 236: Can't this code be consolidated with other tests?

bgavrilMS (approved):

tests/E2E Tests/WebAppUiTests/WebAppCallsApiCallsGraphLocally.cs
trwalke: I was thinking about it but there are quite a few variables in this code block that would need to be passed into the consolidated method that it would end up being more complicated than its worth. I will take another look at it though.

tests/E2E Tests/WebAppUiTests/WebAppCallsApiCallsGraphLocally.cs
```diff
@@ -172,6 +157,125 @@ public async Task ChallengeUser_MicrosoftIdFlow_LocalApp_ValidEmailPasswordCreds
...
+                _output.WriteLine("Starting web app sign-in flow using sign in button after sign out.");
+                await page.GetByRole(AriaRole.Link, new() { Name = "Sign in" }).ClickAsync();
+                await UiTestHelpers.FirstLogin_MicrosoftIdFlow_ValidEmailPassword(page, email, LabUserHelper.FetchUserPassword("msidlabciam6"), _output);
+                await Assertions.Expect(page.GetByText("Welcome")).ToBeVisibleAsync(_assertVisibleOptions);
+                await Assertions.Expect(page.GetByText(email)).ToBeVisibleAsync(_assertVisibleOptions);
+                _output.WriteLine("Web app sign-in flow successful using Sign in button after sign out.");
+            }
+            catch (Exception ex)
+            {
+                Assert.Fail($"the UI automation failed: {ex} output: {ex.Message}.");
```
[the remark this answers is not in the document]
JoshLozensky: The point of the try, catch, finally is to make sure we shut down the processes we started in the finally. We still need to throw the exception from the catch or else we just swallow it right?

JoshLozensky (approved):

JoshLozensky (changes requested):
Can you please run this through our ADO build pipeline to ensure it doesn't increase UI test flakiness? They currently fail at about a 20 to 25% rate. They only fail in the build pipeline, not locally.

tests/DevApps/ciam/myWebApp/myWebApp.csproj
```diff
@@ -5,6 +5,7 @@
     <Nullable>enable</Nullable>
     <ImplicitUsings>enable</ImplicitUsings>
     <UserSecretsId>aspnet-myWebApp-23b8728b-775f-4796-8b63-e85cbaccc27c</UserSecretsId>
+    <UseWIP>true</UseWIP>
```
bgavrilMS, line 8: what does this do?

PR #2733, opened 2024-03-29
fix assertions being removed from dict before callback is executed

jennyf19:
#2734


jmprieur (approved):
Lgtm
Thanks @jennyf19

Issue #2747, opened 2024-04-05
Add New Member to MergedOptions post targeting NET9


FuPingFranco:
There is a new member missing in the MergedOptions which needs to be added.

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs -> run  PreventChangesInOpenIdConnectOptionsToBeOverlooked() test when targeting NET9 to repro. There're are now 61 vs 60 in NET8

Refer to this PR https://github.com/AzureAD/microsoft-identity-web/pull/2739/files/fd9112429db8cffac18da8dc94c6e20f0a125369#diff-e84d595785b0df423efb1d98a7e3035287e91aaf46a05f6b59cf431d32b1cef7 for more details.

PR #2758, opened 2024-04-14
Updates MergedOptions properties after targeting NET9

FuPingFranco:
# Updates MergedOptions properties after targeting NET9

## Description

Adds the new `AdditionalAuthorizationParameters` property from `OpenIdConnectOptions.cs` introduced in NET9 into  Id Web's `MergedOptions.cs`

Link to NET9 code: https://github.com/dotnet/aspnetcore/blob/release/9.0-preview2/src/Security/Authentication/OpenIdConnect/src/OpenIdConnectOptions.cs 

Fixes #[2747](https://github.com/AzureAD/microsoft-identity-web/issues/2747)

### DIFF
--- a/src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
@@ -214,6 +214,9 @@ namespace Microsoft.Identity.Web
 #if NET8_0_OR_GREATER
             mergedOptions.TimeProvider = microsoftIdentityOptions.TimeProvider;
 #endif
+#if NET9_0_OR_GREATER
+            mergedOptions.AdditionalAuthorizationParameters = microsoftIdentityOptions.AdditionalAuthorizationParameters;
+#endif
             mergedOptions.TokenValidationParameters = microsoftIdentityOptions.TokenValidationParameters.Clone();
             mergedOptions.UsePkce |= microsoftIdentityOptions.UsePkce;
 
--- a/src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
@@ -214,9 +214,6 @@ namespace Microsoft.Identity.Web
 #if NET8_0_OR_GREATER
             mergedOptions.TimeProvider = microsoftIdentityOptions.TimeProvider;
 #endif
-#if NET9_0_OR_GREATER
-            mergedOptions.AdditionalAuthorizationParameters = microsoftIdentityOptions.AdditionalAuthorizationParameters;
-#endif
             mergedOptions.TokenValidationParameters = microsoftIdentityOptions.TokenValidationParameters.Clone();
             mergedOptions.UsePkce |= microsoftIdentityOptions.UsePkce;
 
--- a/tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
+++ b/tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
@@ -877,19 +877,20 @@ namespace Microsoft.Identity.Web.Test
         {
             // If the number of public properties of OpenIdConnectOptions changes,
             // then, the PopulateOpenIdOptionsFromMergedOptions method
-            // needs to be updated. For this uncomment the System.IO.File.WriteAllLines() lines below, and run the test
-            // then diff the files to find what are the new properties
+            // needs to be updated. For this uncomment the variable "filePath" and line 893 below, then run the test
+            // and diff the files to find what are the new properties.
             int numberOfProperties = typeof(OpenIdConnectOptions).GetProperties(System.Reflection.BindingFlags.Public | System.Reflection.BindingFlags.Instance).Length;
 #if NET9_0_OR_GREATER
             int expectedNumberOfProperties = 61;
-            //System.IO.File.WriteAllLines(@"C:\temp\net9.txt", typeof(OpenIdConnectOptions).GetProperties().Select(p => p.Name));
+            //string filePath = @"C:\temp\net9.txt";
 #elif NET8_0
             int expectedNumberOfProperties = 60;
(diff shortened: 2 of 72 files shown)


src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
```diff
@@ -213,6 +213,9 @@ internal static void UpdateMergedOptionsFromMicrosoftIdentityOptions(MicrosoftId
             mergedOptions.StringDataFormat ??= microsoftIdentityOptions.StringDataFormat;
 #if NET8_0_OR_GREATER
             mergedOptions.TimeProvider = microsoftIdentityOptions.TimeProvider;
+#endif
+#if NET9_0_OR_GREATER
+            mergedOptions.AdditionalAuthorizationParameters = microsoftIdentityOptions.AdditionalAuthorizationParameters;
```
pmaytak, line 218: nit: I think the properties should be in alphabetical order (mostly)?

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
```diff
@@ -877,17 +877,19 @@ public void PreventChangesInOpenIdConnectOptionsToBeOverlooked()
         {
             // If the number of public properties of OpenIdConnectOptions changes,
             // then, the PopulateOpenIdOptionsFromMergedOptions method
-            // needs to be updated. For this uncomment the 2 lines below, and run the test
+            // needs to be updated. For this uncomment the System.IO.File.WriteAllLines() lines below, and run the test
             // then diff the files to find what are the new properties
             int numberOfProperties = typeof(OpenIdConnectOptions).GetProperties(System.Reflection.BindingFlags.Public | System.Reflection.BindingFlags.Instance).Length;
 #if NET9_0_OR_GREATER
             int expectedNumberOfProperties = 61;
+            //System.IO.File.WriteAllLines(@"C:\temp\net9.txt", typeof(OpenIdConnectOptions).GetProperties().Select(p => p.Name));
```
pmaytak, line 885: nit: save file path to a variable here and call `WriteAllLines` after `#endif` to reduce code duplication.

pmaytak (approved):

src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
jmprieur: no. This is based on the order of the properties in the class itself.

jmprieur (approved):
LGTM
Thanks @FuPingFranco

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
jennyf19: but all the file paths are different, so how sure how a variable name will help in this case?

pmaytak: A little cleaner, I think:
```csharp
            int numberOfProperties = typeof(OpenIdConnectOptions).GetProperties(System.Reflection.BindingFlags.Public | System.Reflection.BindingFlags.Instance).Length;
#if NET9_0_OR_GREATER
            int expectedNumberOfProperties = 61;
            string filePath = @"C:\temp\net9.txt";
#elif NET8_0
            int expectedNumberOfProperties = 60;
            string filePath = @"C:\temp\net8.txt";
#else
            int expectedNumberOfProperties = 57;
            string filePath = @"C:\temp\net7Below.txt";
#endif
            //System.IO.File.WriteAllLines(filePath, typeof(OpenIdConnectOptions).GetProperties().Select(p => p.Name));
            Assert.Equal(expectedNumberOfProperties, numberOfProperties);
        }
```

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
```diff
@@ -877,17 +877,19 @@ public void PreventChangesInOpenIdConnectOptionsToBeOverlooked()
         {
             // If the number of public properties of OpenIdConnectOptions changes,
             // then, the PopulateOpenIdOptionsFromMergedOptions method
-            // needs to be updated. For this uncomment the 2 lines below, and run the test
+            // needs to be updated. For this uncomment the System.IO.File.WriteAllLines() lines below, and run the test
             // then diff the files to find what are the new properties
             int numberOfProperties = typeof(OpenIdConnectOptions).GetProperties(System.Reflection.BindingFlags.Public | System.Reflection.BindingFlags.Instance).Length;
```
victorm-hernandez, line 882: >numberOfProperties [](http://example.com/codeflow?start=16&length=18)

While this works it can be improved, this test will not catch regressions where a property is added and another is removed. For a round test case you need a static definition of the properties expected on each .NET version and to compare against it.

victorm-hernandez: Since you have code here that drops the properties in a file, I would recommend using this file to add those arrays to the code.

src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
victorm-hernandez, line 218: >mergedOptions [](http://example.com/codeflow?start=12&length=13)

I am not familiar with this code, could you add to the PR description what is the impact of adding this? what is the use of merged options.

victorm-hernandez: This is for my education only :), thanks in advance

jennyf19: I would recommend asking CoPilot integration in GitHub for these kinds of questions, for example:
image

tests/Microsoft.Identity.Web.Test/WebAppExtensionsTests.cs
jennyf19: I see, thanks for clarifying.

FuPingFranco: Thanks @pmaytak  & @jennyf19 !

PR #2768, opened 2024-04-17
Reactivate tests using labs requiring admin consent (#2760)

jennyf19:
bringing over @jmprieur 's changes which were applied to rel/v2


jmprieur (approved):
LGTM
Thanks @jennyf19

PR #2798, opened 2024-04-24
Update OwinWebApp DevApp to configure the redirect_uri parameter.

victorm-hernandez:
# Update OwinWebApp DevApp to configure the redirect_uri parameter.

- [x ] You've read the [Contributor Guide](https://github.com/AzureAD/microsoft-identity-web/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://github.com/AzureAD/microsoft-identity-web/blob/master/CODE_OF_CONDUCT.md).
- [x ] You've included unit or integration tests for your change, where applicable.
- [x ] You've included inline docs for your change, where applicable.
- [x ] There's an open issue for the PR that you are making. If you'd like to propose a new feature or change, please open an issue to discuss the change or find an existing issue.

## Description

The redirect_uri parameter is appended to the authorize request while authenticating a user. It is used by the Identity Provider to redirect back to the relying party after the authentication is completed.

The current app implementation seems to suggest that the redirect_uri parameter is configured in the Startup.Auth.cs file already, as part of the ConfidentialClientApplicationOptions. However, this is not the case, this parameter is configured in OpenIdConnectAuthenticationOptions instead.

### DIFF
--- a/tests/DevApps/aspnet-mvc/OwinWebApp/App_Start/Startup.Auth.cs
+++ b/tests/DevApps/aspnet-mvc/OwinWebApp/App_Start/Startup.Auth.cs
@@ -1,13 +1,11 @@
-﻿using Microsoft.Owin.Security;
+﻿using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Identity.Client;
+using Microsoft.Identity.Web;
+using Microsoft.Identity.Web.OWIN;
+using Microsoft.Identity.Web.TokenCacheProviders.InMemory;
+using Microsoft.Owin.Security;
 using Microsoft.Owin.Security.Cookies;
 using Owin;
-using Microsoft.Identity.Web;
-using Microsoft.Identity.Web.TokenCacheProviders.InMemory;
-using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Identity.Client;
-using Microsoft.Identity.Abstractions;
-using Microsoft.Identity.Web.OWIN;
-using System.Web.Services.Description;
 
 namespace OwinWebApp
 {
@@ -21,7 +19,11 @@ namespace OwinWebApp
 
             OwinTokenAcquirerFactory factory = TokenAcquirerFactory.GetDefaultInstance<OwinTokenAcquirerFactory>();
 
-            app.AddMicrosoftIdentityWebApp(factory);
+            app.AddMicrosoftIdentityWebApp(factory, updateOptions: (options) =>
+            {
+                options.RedirectUri = "https://localhost:44386/";
+            });
+            
             factory.Services
                 .Configure<ConfidentialClientApplicationOptions>(options => { options.RedirectUri = "https://localhost:44386/"; })
                 .AddMicrosoftGraph()
--- a/tests/DevApps/aspnet-mvc/OwinWebApi/OwinWebApi.csproj
+++ b/tests/DevApps/aspnet-mvc/OwinWebApi/OwinWebApi.csproj
@@ -103,9 +103,6 @@
     <Content Include="Views\Shared\_Layout.cshtml" />
   </ItemGroup>
   <ItemGroup>
-    <Folder Include="App_Data\" />
-  </ItemGroup>
-  <ItemGroup>
     <PackageReference Include="Antlr">
       <Version>3.5.0.2</Version>
     </PackageReference>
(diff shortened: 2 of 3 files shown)


ciaozhang (approved):

tests/DevApps/aspnet-mvc/OwinWebApp/App_Start/Startup.Auth.cs
```diff
@@ -21,7 +19,11 @@ public void ConfigureAuth(IAppBuilder app)
 
             OwinTokenAcquirerFactory factory = TokenAcquirerFactory.GetDefaultInstance<OwinTokenAcquirerFactory>();
 
-            app.AddMicrosoftIdentityWebApp(factory);
+            app.AddMicrosoftIdentityWebApp(factory, updateOptions: (options) =>
+            {
+                options.RedirectUri = "https://localhost:44386/";
```
victorm-hernandez, line 24: >RedirectUri [](http://example.com/codeflow?start=24&length=11)

This is the option which affects the redirect_uri parameter.

JoshLozensky (approved):

jennyf19 (approved):
:shipit:

tests/DevApps/aspnet-mvc/OwinWebApp/App_Start/Startup.Auth.cs
```diff
@@ -1,13 +1,11 @@
-﻿using Microsoft.Owin.Security;
-using Microsoft.Owin.Security.Cookies;
-using Owin;
-using Microsoft.Identity.Web;
-using Microsoft.Identity.Web.TokenCacheProviders.InMemory;
-using Microsoft.Extensions.DependencyInjection;
+﻿using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Identity.Client;
-using Microsoft.Identity.Abstractions;
+using Microsoft.Identity.Web;
```
jennyf19, line 3: Is this needed?

jennyf19: In reading the PR description,  something doesn't seem right. Will need to look from laptop later

victorm-hernandez: This is just sorting using statements alphabetically

tests/DevApps/aspnet-mvc/OwinWebApp/App_Start/Startup.Auth.cs
jennyf19: I don't think this is needed, it looked to be an app registration issue. Added `https://localhost:44386/` as a redirectURI for this app in the lab. Not sure how it is even working, as this redirectURI was not registered to the app.

jennyf19 (changes requested):
I don't think this PR is needed. it was an app registration issue.

victorm-hernandez:
The intent of this PR is to show the reader how to populate the redirect_uri parameter since we have multiple mentions of options with a similar name but none of them affect this specific parameter. This is independent of the application registration. 

---
In reply to: [2026864027](https://github.com/AzureAD/microsoft-identity-web/pull/2798#issuecomment-2026864027) [](http://example.com/codeflow?ancestors=2026864027)

tests/DevApps/aspnet-mvc/OwinWebApp/App_Start/Startup.Auth.cs
victorm-hernandez: I found this issue with a dSTS application, this is unrelated with the configuration of the application. This shows how to affect a specific parameter called redirect_uri.

jennyf19:
> The intent of this PR is to show the reader how to populate the redirect_uri parameter since we have multiple mentions of options with a similar name but none of them affect this specific parameter. This is independent of the application registration.

I understand that it's independent of the app registration, but the app didn't work with this change. The app registration also needed to be updated. The recommended way is for customers to use the configuration to set the corresponding settings, which is set up [here](https://github.com/AzureAD/microsoft-identity-web/blob/master/src/Microsoft.Identity.Web.OWIN/OwinTokenAcquirerFactory.cs#L35).

victorm-hernandez:
Agreed, that is orthogonal to this change. This change shows how to define the desired return URL for this instance. As you know the configuration on the identity provider may hold multiple return URLs for a given application. The redirect_uri parameter defines which of those should be used for the current request. 

The current sample makes the reader think they are able to configure this by using some of the properties called RedirectUri from our options class, but that is not true, the only way to do this is by using the object I introduced on the sample. Hence the reason for the PR> 

---
In reply to: [2083123625](https://github.com/AzureAD/microsoft-identity-web/pull/2798#issuecomment-2083123625) [](http://example.com/codeflow?ancestors=2083123625)

victorm-hernandez:
Abandoning PR since sample owner doesnt see value on the change.

PR #2799, opened 2024-04-25
Disable the Ciam CUD tests as they don't work any longer since today.

jmprieur:
# Disable the Ciam CUD tests as they don't work any longer since today.


jennyf19 (approved):
:shipit:

PR #2813, opened 2024-04-30
Use ManagedIdentityCredential instead of DefaultAzureCredential for FIC scenario (PR for ID.Web 3.x)

christothes:
# Use `ManagedIdentityCredential` instead of `DefaultAzureCredential` for MI scenarios

## Description
For Azure.Identity credential scenarios where Managed Identity is needed, we should use `ManagedIdentityCredential` rather than `DefaultAzureCredential`.

`DefaultAzureCredential` now has more aggressive "fail fast" behavior which could lead to timeouts for token requests.

### DIFF
--- a/src/Microsoft.Identity.Web.Certificate/KeyVaultCertificateLoader.cs
+++ b/src/Microsoft.Identity.Web.Certificate/KeyVaultCertificateLoader.cs
@@ -45,8 +45,41 @@ namespace Microsoft.Identity.Web
             string? managedIdentityClientId,
             X509KeyStorageFlags x509KeyStorageFlags)
         {
-            Uri keyVaultUri = new Uri(keyVaultUrl);
-            ManagedIdentityCredential credential = new(managedIdentityClientId);
+            Uri keyVaultUri = new(keyVaultUrl);
+
+            bool disableInteractiveCreds = false;
+            var disableInteractiveCredsEnvVar = Environment.GetEnvironmentVariable("IDWEB_DISABLE_INTERACTIVE_AKV_CREDENTIALS");
+
+            if (disableInteractiveCredsEnvVar != null && (disableInteractiveCredsEnvVar == "1" || disableInteractiveCredsEnvVar.Equals("true", StringComparison.OrdinalIgnoreCase)))
+            {
+                disableInteractiveCreds = true;
+            }
+
+            DefaultAzureCredentialOptions options;
+
+            if (disableInteractiveCreds)
+            {
+                options = new DefaultAzureCredentialOptions
+                {
+                    ManagedIdentityClientId = managedIdentityClientId,
+                    ExcludeAzureCliCredential = true,
+                    ExcludeAzureDeveloperCliCredential = true,
+                    ExcludeAzurePowerShellCredential = true,
+                    ExcludeInteractiveBrowserCredential = true,
+                    ExcludeSharedTokenCacheCredential = true,
+                    ExcludeVisualStudioCodeCredential = true,
+                    ExcludeVisualStudioCredential = true
+                };
+            }
+            else
+            {
+                options = new DefaultAzureCredentialOptions
+                {
+                    ManagedIdentityClientId = managedIdentityClientId,
+                };
+            }
+
+            DefaultAzureCredential credential = new(options);
             CertificateClient certificateClient = new(keyVaultUri, credential);
             SecretClient secretClient = new(keyVaultUri, credential);
 
--- a/src/Microsoft.Identity.Web.Certificate/KeyVaultCertificateLoader.cs
+++ b/src/Microsoft.Identity.Web.Certificate/KeyVaultCertificateLoader.cs
@@ -45,41 +45,8 @@ namespace Microsoft.Identity.Web
             string? managedIdentityClientId,
             X509KeyStorageFlags x509KeyStorageFlags)
         {
-            Uri keyVaultUri = new(keyVaultUrl);
-
-            bool disableInteractiveCreds = false;
-            var disableInteractiveCredsEnvVar = Environment.GetEnvironmentVariable("IDWEB_DISABLE_INTERACTIVE_AKV_CREDENTIALS");
-
-            if (disableInteractiveCredsEnvVar != null && (disableInteractiveCredsEnvVar == "1" || disableInteractiveCredsEnvVar.Equals("true", StringComparison.OrdinalIgnoreCase)))
-            {
-                disableInteractiveCreds = true;
-            }
-
-            DefaultAzureCredentialOptions options;
-
-            if (disableInteractiveCreds)
-            {
-                options = new DefaultAzureCredentialOptions
-                {
-                    ManagedIdentityClientId = managedIdentityClientId,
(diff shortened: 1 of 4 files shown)


src/Microsoft.Identity.Web.Certificate/KeyVaultCertificateLoader.cs
```diff
@@ -46,11 +46,7 @@ public async Task LoadIfNeededAsync(CredentialDescription credentialDescription,
             X509KeyStorageFlags x509KeyStorageFlags)
         {
             Uri keyVaultUri = new Uri(keyVaultUrl);
-            DefaultAzureCredentialOptions options = new()
-            {
-                ManagedIdentityClientId = managedIdentityClientId,
-            };
-            DefaultAzureCredential credential = new(options);
+            ManagedIdentityCredential credential = new(managedIdentityClientId);
```
jmprieur, line 49: This one is the KeyVault certificate loader.
Aren't you interested to test things locally on your box using dev credentials?

src/Microsoft.Identity.Web.Certificateless/ManagedIdentityClientAssertion.cs
```diff
@@ -23,19 +23,7 @@ public class ManagedIdentityClientAssertion : ClientAssertionProviderBase
...
-                    WorkloadIdentityClientId = managedIdentityClientId,
-                    ExcludeAzureCliCredential = true,
-                    ExcludeAzureDeveloperCliCredential = true,
-                    ExcludeAzurePowerShellCredential = true,
-                    ExcludeInteractiveBrowserCredential = true,
-                    ExcludeSharedTokenCacheCredential = true,
-                    ExcludeVisualStudioCodeCredential = true,
-                    ExcludeVisualStudioCredential = true
-                });
+            _credential = new ManagedIdentityCredential(managedIdentityClientId);
```
jmprieur, line 26: We also need the workload identity credentials?

jmprieur (commented):
Thanks @christothes for this proposal
We are not going to take this PR as is, as this would be a breaking change for many customers (accessing KeyVault from your dev box when you are authorized to do so)

Also we are going to move these classes to MSAL.NET which already supports MSI and FIC+MSI directly, and will bring new scenarios in this area.
cc: @bgavrilMS

src/Microsoft.Identity.Web.Certificate/KeyVaultCertificateLoader.cs
christothes: I assumed that because a managedIdentityClientId is being used here unconditionally, it was only intended for MI scenarios.

src/Microsoft.Identity.Web.Certificateless/ManagedIdentityClientAssertion.cs
christothes: in that case, we should use `ChainedTokenCredential` with just workload and managed identity credentials.

bgavrilMS: We discuss about moving this implementation in MSAL but now I am not sure. Should MSAL have a dependnecy on workload identity logic @jmprieur ?

christothes: In Azure.Identity, we treat WIC as part of the managed identity flow because of k8s.

src/Microsoft.Identity.Web.Certificate/KeyVaultCertificateLoader.cs
jmprieur: no. It's for all scenarios. This is the only way to test locally.

src/Microsoft.Identity.Web.Certificateless/ManagedIdentityClientAssertion.cs
bgavrilMS: Not sure I understand your last sentence @christothes . Are you suggesting we add WIC logic to MSAL?

christothes: I think it is something to consider because it is a replacement for the Kubernetes pod-managed identity

bgavrilMS: And now we have a bug for it, reported by a partner - #2893

I am repurposing this PR to fix the bug, hope you agree @christothes

src/Microsoft.Identity.Web.Certificate/KeyVaultCertificateLoader.cs
bgavrilMS: I have resolved this.

bgavrilMS:
> Thanks @christothes for this proposal We are not going to take this PR as is, as this would be a breaking change for many customers (accessing KeyVault from your dev box when you are authorized to do so)
> 
> Also we are going to move these classes to MSAL.NET which already supports MSI and FIC+MSI directly, and will bring new scenarios in this area. cc: @bgavrilMS

I addressed all comments. it will also fix #2893 - please review again.

src/Microsoft.Identity.Web.Certificateless/ManagedIdentityClientAssertion.cs
```diff
@@ -21,30 +21,18 @@ public class ManagedIdentityClientAssertion : ClientAssertionProviderBase
...
             _tokenExchangeUrl = CertificatelessConstants.DefaultTokenExchangeUrl;
         }
 
         /// <summary>
         /// See https://aka.ms/ms-id-web/certificateless.
         /// </summary>
-        /// <param name="managedIdentityClientId">Optional ClientId of the Managed Identity or Workload Identity</param>
-        /// <param name="tokenExchangeUrl">Optional token exchange resource url. Default value is "api://AzureADTokenExchange/.default".</param>
+        /// <param name="managedIdentityClientId">Optional ClientId of the Managed Identity</param>
+        /// <param name="tokenExchangeUrl">Optional audience of the token to be requested from Managed Identity. Default value is "api://AzureADTokenExchange/.default". This value is different on other clouds.</param>
```
christothes, line 35:
```suggestion
        /// <param name="tokenExchangeUrl">Optional audience of the token to be requested from Managed Identity. Default value is "api://AzureADTokenExchange/.default". This value is different on clouds other than Azure Public.</param>
```

bgavrilMS:
Approach signed off by @jmprieur after sync meeting.

bgavrilMS:
Approach agreed on by @jmprieur during sync meeting.

neha-bhargava (approved):

jmprieur (approved):
LGTM
Thanks @bgavrilMS

bgavrilMS:
@jmprieur @jennyf19 - gentle ping on this.

jennyf19:
what testing was done for this? Also the branch looks to be outdated.

keegan-caruso:
Changes look right.

Seconding @jennyf19 comment around testing,

jennyf19 (approved):
:shipit:

PR #2816, opened 2024-05-02
add LogLevel.None predicate to logging decision

westin-m:
If `Logging:LogLevel:Microsoft.Identity.Web` is assigned to  `None` in appsettings.json, the default logger is not initialized and IdWeb does not record any logs.

Before, leaving the log level at `Information` and trying to use the no-op NullIdentityModelLogger (this decision would be ignored and the default logger would be used anyway) yielded about 500 lines of logging when launching an app and making a single API call. 
Changing the log level to `None` still resulted in logs being recorded by the default logger, about 10 lines for the above scenario of a single API call. 
Now, choosing `None` means the default logger is not instantiated and IdWeb logs nothing.


jennyf19:
@westin-m  can you share a screenshot of the diff now?

westin-m:
Here is an example of seeing fewer logs from the service and client when the LogLevel for IdWeb in both is set to None. The lines highlighted in red are either omitted because the default logger is not being used, or because of the time diffs when running the app. 
Service:
image

Client:
image

src/Microsoft.Identity.Web.TokenAcquisition/MicrosoftIdentityBaseAuthenticationBuilder.cs
```diff
@@ -40,18 +40,23 @@ public class MicrosoftIdentityBaseAuthenticationBuilder
...
                     return;
-
+                
                 // check if an ILogger was already created by user
                 ILogger? logger = serviceProvider.GetService<ILogger<IdentityLoggerAdapter>>();
                 if (logger == null)
                 {
                     var loggerFactory = serviceProvider.GetService<ILoggerFactory>();
-                    if (loggerFactory != null)
+                    if (loggerFactory != null)
```
jmprieur, line 59: nit: extra space?

jmprieur (approved):
LGTM
Thanks @westin-m

Please update the wiki on logging: https://github.com/AzureAD/microsoft-identity-web/wiki/Logging

src/Microsoft.Identity.Web.TokenAcquisition/MicrosoftIdentityBaseAuthenticationBuilder.cs
```diff
@@ -40,18 +40,23 @@ public class MicrosoftIdentityBaseAuthenticationBuilder
 
         internal static void SetIdentityModelLogger(IServiceProvider serviceProvider)
         {
+            var config = serviceProvider.GetRequiredService<IConfiguration>();
+            var loglevel = config.GetValue<string>("Logging:LogLevel:Microsoft.Identity.Web");
+            
             if (serviceProvider != null)
```
jennyf19, line 46: no need to check for serviceProvider here because it will throw on line 43 if it's not found.

westin-m: Good point. I'll move these lines inside that block so that it doesn't throw in that case

jennyf19 (approved):
:shipit: 

remember to add a section for the next changelog for this one...thx!

westin-m:
> Please update the wiki on logging: https://github.com/AzureAD/microsoft-identity-web/wiki/Logging

@jmprieur Done!

PR #2840, opened 2024-05-14
Process error AADSTS1000502 - expired certificate

westin-m:
Add a clause to process the error code for a certificate outside of the validity window.


JoshLozensky (approved):

jmprieur (approved):
LGTM
Thanks @westin-m

As discussed it's difficult to test end to end. But:
1. we have an E2E test for the mechanism (not for that particular case)
2.  you've done due diligence checking with ESTS to make sure that these are all the values we need.

🚢

PR #2826, opened 2024-05-08
add queryParams when building authority

kellyyangsong:
Fixes #2697

Issue #2697, opened 2024-03-08
Process the query part of the authority when present


jmprieur:
### Microsoft.Identity.Web Library

Microsoft.Identity.Web

### Microsoft.Identity.Web version

2.17.0

### Web app

Sign-in users

### Web API

Protected web APIs (validating tokens)

### Token cache serialization

In-memory caches

### Description

When the authority is provided with a query part, it would be great to keep it and propagate to the MetadataUrl and the MSAL authority

```json
"Authority" : "https://customdomain.com/tenant/v2.0?dc=dcxyw
```

### Reproduction steps

```json
"Authority" : "https://customdomain.com/tenant/v2.0?dc=dcxyw
```

### Relevant code snippets

```csharp
"Authority" : "https://customdomain.com/tenant/v2.0?dc=dcxyw
```

### Expected behavior

?dc=dcxyw is taken into account

pmaytak:
@jmprieur In what scenarios are the query params expected for an authority? When passed to MSAL, would these params be added to the /authorize or /token endpoints?

PR #2846, opened 2024-05-17
Revert "add queryParams when building authority (#2826)"

kellyyangsong:
The purpose of the issue #2697 was to respect user-provided query params. The goal is to actually parse the query params from options.Authority into options.ExtraQueryParams, rather than add options.ExtraQueryParams to the authority.


RojaEnnam:
@kellyyangsong can you please add the description for this PR. I would like to understand what we want to do here. We cannot stop sending query params because the telemetry we are relying on is based on these query params.

jennyf19 (approved):

PR #2869, opened 2024-05-28
Fix test issue (because TokenAcquirerFactory is a singleton :-())

jmprieur:
# Fix test issue (because TokenAcquirerFactory is a singleton :-())


jmprieur:
waiting for results of tests.

PR #2870, opened 2024-05-28
Use named http client for dSTS

ciaozhang:
Use named client for token acquistion propose to prefer IPv6 address.

Design doc:
https://microsoft.sharepoint.com/:w:/t/aad/devex/ERmovf1YilVOhNwNSKOWyeoBCumEwCaDiY-O0YIzv9i7oQ?e=6TfbRC&wdOrigin=TEAMS-MAGLEV.p2p_ns.rwc&wdExp=TEAMS-TREATMENT&wdhostclicktime=1715629585046&web=1

Spec: 
https://learn.microsoft.com/en-us/dotnet/api/system.net.http.ihttpclientfactory.createclient?view=net-8.0

IHttpClientFactory is guaranteed to return a new HttpClient if named http client is not registered.
The extension to add named client won't affect the current behaviour.

### DIFF
--- a/src/Microsoft.Identity.Web.TokenAcquisition/MsalAspNetCoreHttpClientFactory.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/MsalAspNetCoreHttpClientFactory.cs
@@ -17,7 +17,9 @@ namespace Microsoft.Identity.Web
 
         public HttpClient GetHttpClient()
         {
-            HttpClient httpClient = _httpClientFactory.CreateClient();
+            // The HttpClient will be created with the name of the client registered in the DI container
+            // If the named client(dSTS) is not found, the factory will fall back to the default client
+            HttpClient httpClient = _httpClientFactory.CreateClient("dSTS");
             httpClient.DefaultRequestHeaders.Add(Constants.TelemetryHeaderKey, IdHelper.CreateTelemetryInfo());
             return httpClient;
         }
--- a/src/Microsoft.Identity.Web.TokenAcquisition/MsalAspNetCoreHttpClientFactory.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/MsalAspNetCoreHttpClientFactory.cs
@@ -17,9 +17,9 @@ namespace Microsoft.Identity.Web
 
         public HttpClient GetHttpClient()
         {
-            // The HttpClient will be created with the name of the client registered in the DI container
-            // If the named client(dSTS) is not found, the factory will fall back to the default client
-            HttpClient httpClient = _httpClientFactory.CreateClient("dSTS");
+            // The HttpClient will be created with the named client registered in the DI container
+            // If the named client is not found, the factory will fall back to the default client
+            HttpClient httpClient = _httpClientFactory.CreateClient("token_acquisition_httpclient");
             httpClient.DefaultRequestHeaders.Add(Constants.TelemetryHeaderKey, IdHelper.CreateTelemetryInfo());
             return httpClient;
         }


jmprieur (changes requested):
We can't do that, of course

src/Microsoft.Identity.Web.TokenAcquisition/MsalAspNetCoreHttpClientFactory.cs
```diff
@@ -17,7 +17,9 @@ public MsalAspNetCoreHttpClientFactory(IHttpClientFactory httpClientFactory)
 
         public HttpClient GetHttpClient()
         {
-            HttpClient httpClient = _httpClientFactory.CreateClient();
+            // The HttpClient will be created with the name of the client registered in the DI container
+            // If the named client(dSTS) is not found, the factory will fall back to the default client
+            HttpClient httpClient = _httpClientFactory.CreateClient("dSTS");
```
jennyf19, line 22: Id Web is used by services outside of Microsoft, this will not work.

victorm-hernandez (approved):

src/Microsoft.Identity.Web.TokenAcquisition/MsalAspNetCoreHttpClientFactory.cs
```diff
@@ -17,7 +17,9 @@ public MsalAspNetCoreHttpClientFactory(IHttpClientFactory httpClientFactory)
 
         public HttpClient GetHttpClient()
         {
-            HttpClient httpClient = _httpClientFactory.CreateClient();
+            // The HttpClient will be created with the named client registered in the DI container
+            // If the named client is not found, the factory will fall back to the default client
+            HttpClient httpClient = _httpClientFactory.CreateClient("token_acquisition_httpclient");
```
jmprieur, line 22: I'm confused. 

I thought we had agreed that in the constructor of the `TokenAcquisition` class, you would no longer new-up and `MsalAspNetCoreHttpClientFactory(),` but you would use the locator pattern (that is attempt to get a service implementing `IMsalHttpClientFactory` from the service provider):

```CSharp
 _httpClientFactory = serviceProvider.GetService<IMsalHttpClientFactory>() ??  new MsalAspNetCoreHttpClientFactory(httpClientFactory);
```

did the design change @victorm-hernandez @ciaozhang ?

jmprieur (commented):
Please tell me if the design change.
Otherwise, let's do as agreed.

jmprieur (changes requested):
Please tell me if the design change.
Otherwise, let's do as agreed.

src/Microsoft.Identity.Web.TokenAcquisition/MsalAspNetCoreHttpClientFactory.cs
jmprieur: I think that the design we had agreed on is, in the TokenAcquisition constructor:

https://github.com/AzureAD/microsoft-identity-web/blob/77df2a63d759beb5bbeb9c1260c8efdffd9fff94/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs#L101

use the locator pattern to get a IMsalAspNetCoreHttpClientFactory if there is one in the service provider.

 _httpClientFactory = serviceProvider.GetService<IMsalAspNetCoreHttpClientFactory> ?? new MsalAspNetCoreHttpClientFactory(httpClientFactory);

src/Microsoft.Identity.Web.TokenAcquisition/MsalAspNetCoreHttpClientFactory.cs
victorm-hernandez: @jmprieur , as per our last conversation we agreed to use a named HTTP client. I will share the diagram we reviewed last time over email.

PR #2872, opened 2024-05-30
Add middleware to MS Graph HTTP client

### LINKED ISSUE #2684: Default Graph Middleware is not added when calling `AddMicrosoftGraph`

Microsoft.Identity.Web.Graph

2.17.0

### Web app

Not Applicable

### Web API

Not Applicable

### Token cache serialization

Not Applicable

### Description

Taking a look at the implementation of `AddMicrosoftGraph` at https://github.com/AzureAD/microsoft-identity-web/blob/847304540bbe6c6825861bf52155567c5bda575f/src/Microsoft.Identity.Web.GraphServiceClient/GraphServiceCollectionExtensions.cs#L113

The created httpClient does not setup the middleware for resilient applications for calling graph. Therefore, the client is missing functionalities such as the `RedirectHandler` and `RetryHandler`.

As the created client is probably a singleton, the httpClient should probably be created by calling and setting a static HttpClient instance to resolve this. 

```cs
var httpClient = GraphClientFactory.Create();
```

Related to https://github.com/microsoftgraph/msgraph-sdk-dotnet/issues/2357

### Reproduction steps

- Follow tutorial at https://learn.microsoft.com/en-us/entra/identity-platform/quickstart-console-app-netcore-acquire-token
- Modify code to change retry handler options

```cs
var retryHandlerOption = new RetryHandlerOption
{
    MaxRetry = 5,
    ShouldRetry = (delay,attempt,message) =>
    {
        Console.WriteLine($"Retrying attempt = {attempt}");
        return true;
    }
};

var user = await graphServiceClient
    .Users[userId]
    .GetAsync(requestConfiguration => requestConfiguration.Options.Add(retryHandlerOption));
```
- Run https://github.com/microsoft/dev-proxy to simulate errors

### Error message

Passing default options are not possible. 

### Id Web logs

_No response_

### Relevant code snippets

```csharp
var retryHandlerOption = new RetryHandlerOption
{
    MaxRetry = 5,
    ShouldRetry = (delay,attempt,message) =>
    {
        Console.WriteLine($"Retrying attempt = {attempt}");
        return true;
    }
};

var user = await graphServiceClient
    .Users[userId]
    .GetAsync(requestConfiguration => requestConfiguration.Options.Add(retryHandlerOption));
```

### Regression

_No response_

### Expected behavior

`AddMicrosoftGraph` should setup the default middleware by calling `GraphClientFactory.Create()` to create a new HttpClient. Or configure the client to have the list of middlware returned by calling `GraphClientFactory.CreateDefaultHandlers();` 

Alternatively, an overload that takes in a HttpClient or list of handlers should be possible to allow passing custom handlers to the HttpClient used by GraphServiceClient

iNinja:
# Add middleware to MS Graph HTTP client

Use GraphClientFactory to create the HTTP client used for MS Graph.

## Description
Removed use of IHTTPClientFactory in favour of GraphClientFactory, as the latter includes all the default middleware such as redirect and retry handlers, while also adding a telemetry handler specific for Graph and a compression handler required as a temporary solution. With this approach, once the temporary fix is removed from the GraphClientFactory we can consume it without further updates than the dependency version.

Note: I could not find any tests that validate retries, please advise on anything that can be added or updated to validate the updated client.
Note 2: Considered the alternative of keeping the IHTTPClientFactory and adding the middleware manually, but that carries the potential technical debt of updating this method when the Graph SDK revises their implementation.

Fixes #2684


src/Microsoft.Identity.Web.GraphServiceClient/GraphServiceCollectionExtensions.cs
```diff
@@ -103,14 +103,13 @@ private static GraphServiceClient CreateGraphServiceClient(IServiceProvider serv
...
             var options = serviceProvider.GetRequiredService<IOptions<GraphServiceClientOptions>>();
-            var httpClientFactory = serviceProvider.GetRequiredService<IHttpClientFactory>();
             var microsoftGraphOptions = options.Value;
             if (microsoftGraphOptions.Scopes == null)
             {
                 Throws.ArgumentNullException("scopes", IDWebErrorMessage.CalledApiScopesAreNull);
             }
 
-            var httpClient = httpClientFactory.CreateClient("GraphServiceClient");
+            var httpClient = GraphClientFactory.Create();
```
jmprieur, line 112: I think it's a breaking change for partners who would have provided their HttpClientFactory.
@jennyf19 this means we should not take this in rel/v2 and we should document it,

jmprieur: Maybe check if there is a factory for "GraphServiceClient" honor it, and otherwise?

jmprieur (commented):
LGTM
I wonder if we could keep the compatibility?

GetService<IHttpClientFactory<GraphServiceClient>> instead of GetRequiredService<>
and only use the GraphServiceFactory if it's not null?

would that twork?

src/Microsoft.Identity.Web.GraphServiceClient/GraphServiceCollectionExtensions.cs
jmprieur: In the TokenAcquisition constructor:

https://github.com/AzureAD/microsoft-identity-web/blob/77df2a63d759beb5bbeb9c1260c8efdffd9fff94/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs#L101

use the locator pattern to get a IMsalAspNetCoreHttpClientFactory if there is one in the service provider.

 _httpClientFactory = serviceProvider.GetService<IMsalAspNetCoreHttpClientFactory> ?? new MsalAspNetCoreHttpClientFactory(httpClientFactory);

iNinja: From what I can tell, this extension only adds the token acquisition service and GraphServiceClient if they are not already present, or if their lifetime does not match.
This code would not be executed in the case where the partners have configured their own.

The situation where an HTTPClient has been configured under the name of "GraphServiceClient" is one I am not sure if it can be checked without resulting in a default HTTPClient being created.

Any further thoughts are highly appreciated @jmprieur @jennyf19.

jennyf19 (approved):
:shipit:

PR #2874, opened 2024-06-03
PR Builds not failing on API breaking changes

keegan-caruso:
# PR Builds not failing on API breaking changes

## Description:

Pack needs to be done as part of the PR builds
to catch API breaking changes.


.github/workflows/dotnetcore.yml
```diff
@@ -74,3 +75,44 @@ jobs:
     - name: Test with .NET standard2.0
       run: dotnet test --no-restore --no-build Microsoft.Identity.Web.sln -f netstandard2.0 -p:FROM_GITHUB_ACTION=true --configuration Release --filter "(FullyQualifiedName!~Microsoft.Identity.Web.Test.Integration)&(FullyQualifiedName!~WebAppUiTests)&(FullyQualifiedName=IntegrationTests)"
 
+    - name: Pack Microsoft.Identity.Web
```
jmprieur, line 78: why not directly a dotnet pack --no-restore --no-build --configuration Microsoft.Identity.Web.sln? is it because of the Owin modules?

keegan-caruso: I'm going to split this up, one to fix the suppressions 

A separate one for getting these checks turned back on in PR

Issue #2482, opened 2023-09-27
[Documentation] - Clarify Azure Functions Support


dirthsj:
### Please check all that apply

- [ ] typo
- [ ] documentation doesn't exist
- [x] documentation needs clarification
- [ ] error(s) in the example
- [ ] needs an example

### Description of the issue

Currently, the azure functions documentation makes no mention of the issues surrounding Azure Functions support, e.g. #1548 which has been open for nearly two years.

Please update documentation to specify which version(s) of Azure Functions are supported and link to relevant issues users may encounter when attempting to use this library with Azure Functions.

jennyf19:
Azure functions is supported, but please use .NET 8+ Isolated model, as there were issues with the previous .NET versions, hence the open issues for an extended period of time. 
closing, but please re-open if needed.

PR #2876, opened 2024-06-04
Add empty DevApp template for Azure Functions

yangt49:
Part 1 of #2482


tools/app-provisioning-tool/tests/Tests.csproj
```diff
@@ -1,6 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
+    // Test says that app-provisioning-lib is not compatible with net7.0!
```
keegan-caruso, line 4: Can you clarify? What is the action item here?

tests/DevApps/AzureFunctions/AzureFunctions.csproj
```diff
@@ -0,0 +1,30 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
jennyf19, line 1: VS is a bit interesting, so via the file explorer, this project is in a folder called "AzureFunctions" but when you open it in VS, it looks like this:
image

Not in a folder. So you need to right click on the DevApps folder and select "add new folder" and name it the same as the one in the file explorer `AzureFunctions`

small nit, but it will keep things more organized 😸

tests/DevApps/AzureFunctions/AzureFunctions.csproj
```diff
@@ -0,0 +1,30 @@
...
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
```
jennyf19, line 3: needs to be `TargetFrameworks>net8.0</TargetFrameworks>` with an `s` 😅otherwise it will pull in .NET 6 & 7, which we don't want.

jennyf19: after you change that, there should be a warnign in VS to reload the project

tests/DevApps/AzureFunctions/Function1.cs
```diff
@@ -0,0 +1,24 @@
+using Microsoft.AspNetCore.Http;
```
jennyf19, line 1: when I pulled it locally, I had a SampleFunc folder showing up, which was messing things up:
image

so i deleted it and things seem to be building now. I don't see it in any of your files though... 🤔

jennyf19, line 1: add the license info to all the classes
// Copyright (c) Microsoft Corporation. All rights reserved.

tools/app-provisioning-tool/tests/Tests.csproj
yangt49: Ah thank you for catching that! We were running into an issue with building the project previously where I did not have the correct install -- that is resolved now.

jennyf19:
closing as this is not as straightforward as we thought

PR #2896, opened 2024-06-14
Implements GetAuthorizationHeaderAsync to support 1P specific scenarios

sruke:
What?
- Implements IAuthorizationHeaderProvider.GetAuthorizationHeaderAsync in DefaultAuthorizationHeaderProvider
- Updates DownstreamApi, MicrosoftGraph, GraphServiceClient and BaseAuthorizationHeaderProvider to use the new higher level API that can create a header for on-behalf-of an app or user instead of separate APIs (GetAuthorizationHeaderForAppAsync and GetAuthorizationHeaderForUserAsync). There is no change to the existing DevEx.
- Adds ConfigureAwait which was previously missing for some async calls

Why?
This is being done to accommodate 1P specific protocols.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -86,7 +86,7 @@
     <MicrosoftGraphVersion>4.36.0</MicrosoftGraphVersion>
     <MicrosoftGraphBetaVersion>4.57.0-preview</MicrosoftGraphBetaVersion>
     <MicrosoftExtensionsHttpVersion>3.1.3</MicrosoftExtensionsHttpVersion>
-    <MicrosoftIdentityAbstractions>5.3.0</MicrosoftIdentityAbstractions>    
+    <MicrosoftIdentityAbstractions>5.4.0-preview3-HeaderProvider</MicrosoftIdentityAbstractions>    
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(TargetFramework)' == 'net8.0'">
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <!--This should be passed from the VSTS build-->
     <!-- This needs to be greater than or equal to the validation baseline version -->
-    <ClientSemVer Condition="'$(ClientSemVer)' == ''">2.20.2-localbuild</ClientSemVer>
+    <ClientSemVer Condition="'$(ClientSemVer)' == ''">2.20.3-localbuild</ClientSemVer>
     <!--This will generate AssemblyVersion, AssemblyFileVersion and AssemblyInformationVersion-->
     <Version>$(ClientSemVer)</Version>
 
@@ -86,7 +86,7 @@
     <MicrosoftGraphVersion>4.36.0</MicrosoftGraphVersion>
     <MicrosoftGraphBetaVersion>4.57.0-preview</MicrosoftGraphBetaVersion>
     <MicrosoftExtensionsHttpVersion>3.1.3</MicrosoftExtensionsHttpVersion>
-    <MicrosoftIdentityAbstractions>6.0.0-preview1-headerprovider</MicrosoftIdentityAbstractions>    
+    <MicrosoftIdentityAbstractions>6.0.0</MicrosoftIdentityAbstractions>    
   </PropertyGroup>
 
   <PropertyGroup Condition="'$(TargetFramework)' == 'net8.0'">
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <!--This should be passed from the VSTS build-->
     <!-- This needs to be greater than or equal to the validation baseline version -->
-    <ClientSemVer Condition="'$(ClientSemVer)' == ''">2.16.0-localbuild</ClientSemVer>
+    <ClientSemVer Condition="'$(ClientSemVer)' == ''">2.20.2-localbuild</ClientSemVer>
     <!--This will generate AssemblyVersion, AssemblyFileVersion and AssemblyInformationVersion-->
     <Version>$(ClientSemVer)</Version>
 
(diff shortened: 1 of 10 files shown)


Directory.Build.props
```diff
@@ -86,7 +86,7 @@
     <MicrosoftGraphVersion>4.36.0</MicrosoftGraphVersion>
     <MicrosoftGraphBetaVersion>4.57.0-preview</MicrosoftGraphBetaVersion>
     <MicrosoftExtensionsHttpVersion>3.1.3</MicrosoftExtensionsHttpVersion>
-    <MicrosoftIdentityAbstractions>5.3.0</MicrosoftIdentityAbstractions>    
+    <MicrosoftIdentityAbstractions>5.4.0-preview3-HeaderProvider</MicrosoftIdentityAbstractions>
```
sruke, line 89: This PR will not be merged until we align on the interface definition in Microsoft.Identity.Abstractions and release it.

src/Microsoft.Identity.Web.GraphServiceClient/GraphAuthenticationProvider.cs
```diff
@@ -85,19 +101,31 @@ internal class GraphAuthenticationProvider : IAuthenticationProvider
...
-                    authorizationHeader = await _authorizationHeaderProvider.CreateAuthorizationHeaderForUserAsync(
-                         scopes!,
-                         authorizationHeaderProviderOptions,
-                         claimsPrincipal: user,
-                         cancellationToken);
+                    if (authorizationHeaderProviderOptions!.RequestAppToken)
+                    {
+                        authorizationHeader = await _authorizationHeaderProvider.CreateAuthorizationHeaderForAppAsync("https://graph.microsoft.com/.default",
+                            authorizationHeaderProviderOptions,
+                            cancellationToken).ConfigureAwait(false);
```
sruke, line 119: Was ConfigureAwait(false) intentionally not added previously?

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
```diff
@@ -336,7 +356,19 @@ internal partial class DownstreamApi : IDownstreamApi
             // which is done by not specifying any scopes
             if (effectiveOptions.Scopes != null && effectiveOptions.Scopes.Any())
             {
-                string authorizationHeader = appToken ?
+                string? authorizationHeader;
+                if (_authorizationHeaderProviderExtension != null)
```
jennyf19, line 360: due to dependency injection, this will not be null, so no need for the check.

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
```diff
@@ -336,7 +356,19 @@ internal partial class DownstreamApi : IDownstreamApi
...
+                {
+                    authorizationHeader = await _authorizationHeaderProviderExtension.CreateAuthorizationHeaderAsync(
+                       new RequestContext(),
```
jennyf19, line 363: How is RequestContext used?

sruke: RequestContext is not needed today. It was added for future use. For example, if serviceName needs to be passed down from DownstreamApi to AuthorizationHeaderProvider implementation.

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderProvider.cs
```diff
@@ -50,6 +51,46 @@ public DefaultAuthorizationHeaderProvider(ITokenAcquisition tokenAcquisition)
...
+            CancellationToken cancellationToken = default)
+        {
+            Client.AuthenticationResult result;
+
+            // Previously, with the API name we were able to distinguish between app and user token acquisition
+            // This context is missing in the new API, so can we enforce that downstreamApiOptions.RequestAppToken
+            // needs to be set to true to acquire a token for the app. We cannot rely on ClaimsPrincipal as it can be null for user token acquisition.
+            // DevEx Before:
+            // await authorizationHeaderProvider.CreateAuthorizationHeaderForAppAsync("https://graph.microsoft.com/.default").ConfigureAwait(false);
+            // DevEx with the new API:
```
jmprieur, line 68: The other DevEx is still possible?

jmprieur (commented):
This looks good.
did you run the test apps, in particular the Web App Calls Graph (there are links on the index page)

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderProvider.cs
sruke: Yes the other DevEx is still possible

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
sruke: RequestContext has been removed as agreed offline as there is no need for it today.

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
sruke: Good point! Moved away from the design of introducing a new interface so this change is so longer relevant.

Directory.Build.props
jmprieur: It would be a 6.0.0

sruke: Updated MicrosoftIdentityAbstractions to 6.0.0

sruke:
> This looks good. did you run the test apps, in particular the Web App Calls Graph (there are links on the index page)

Yes, all the ID.Web tests are passing including WebAppCallsApiCallsGraph test.

jennyf19 (approved):
:shipit: 
we want this in master as well, correct?

jmprieur (approved):
LGTM
Thanks @sruke

sruke:
> :shipit: we want this in master as well, correct?

Yes. Once this PR is merged, I'll cherry pick this change into master.

PR #2905, opened 2024-06-20
Stop non-OB build from running nightly

bgavrilMS:
I am also happy to delete this. The build will start failing soon because the signing step is deprecated

# {PR title}

## Description

{Detail}

### DIFF
--- a/build/pipeline-releasebuild.yaml
+++ b/build/pipeline-releasebuild.yaml
@@ -5,14 +5,7 @@
  
 trigger: none
 pr: none
- 
-# Create a daily midnight build for release builds on master to ensure our release builds function
-schedules:
-- cron: "0 0 * * *"
-  displayName: Daily midnight build
-  branches:
-    include:
-    - master
+
  
 pool:
   name: MwWilson1EsHostedPool


build/pipeline-releasebuild.yaml
```diff
@@ -5,14 +5,7 @@
...
 pr: none
- 
-# Create a daily midnight build for release builds on master to ensure our release builds function
-schedules:
-- cron: "0 0 * * *"
-  displayName: Daily midnight build
-  branches:
-    include:
-    - master
+
```
jmprieur, line 8: Thanks @bgavrilMS 
can we know about why we do that?

bgavrilMS: This is a non-compliant release build. It will stop working very soon. If you want to keep it around for reference, keep the file. It would be better to delete it fully. 

If this build is accidentally used to release binaries, it will cause an icm.

jmprieur: I like that it produces nightly bits, which the other does not do

bgavrilMS: https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=1317081&view=results - looks like OneBranch build does run nightly and produces nightly bits, which are also compliant.

PR #2906, opened 2024-06-21
Use registered IMsalHttpClientFactory if it is exist

ciaozhang:
The AddDstsHttpClientFactory() will create the dSTS prefer IPv6 http client and register the DstsHttpClientFactory.
This change will enforce the token acquisition flow to use dSTS http client to process http request.


jmprieur (approved):
LGTM
Thanks @ciaozhang

jennyf19 (approved):
Thanks @ciaozhang 

You might want to cherry-pick this into rel/v2 as well, Not sure which one we are releasing this week.

PR #2929, opened 2024-07-15
fix CVEs

jennyf19:


jmprieur (approved):
LGTM
Thanks @jennyf19 

.NET 9 is simpler, that's nice.

kellyyangsong (approved):

Directory.Build.props
```diff
@@ -88,21 +88,25 @@
     <MicrosoftGraphBetaVersion>4.57.0-preview</MicrosoftGraphBetaVersion>
     <MicrosoftExtensionsHttpVersion>3.1.3</MicrosoftExtensionsHttpVersion>
     <MicrosoftIdentityAbstractions>6.0.0</MicrosoftIdentityAbstractions>
-    <NetNineRuntimeVersion> 9.0.0-preview.4.24266.19</NetNineRuntimeVersion>
-    <AspNetCoreNineRuntimeVersion> 9.0.0-preview.4.24267.6</AspNetCoreNineRuntimeVersion>
+    <NetNineRuntimeVersion>9.0.0-preview.6.24327.7</NetNineRuntimeVersion>
+    <AspNetCoreNineRuntimeVersion>9.0.0-preview.6.24328.4</AspNetCoreNineRuntimeVersion>
+    <!--CVE-2024-30105-->
+    <SystemTextJsonVersion>8.0.4</SystemTextJsonVersion>
```
czb182, line 94: seems like this should have been the fixed version of the oldest currently support .net version which is .NET 6.  Then in .net 8 and higher specific fields set the higher versions.

With this going to pull in .NET 8 oriented libraries in .NET 6 targeted applicaitons

jmprieur: It's not going to pull .NET 8 libraries in .NET 6 applications as the libraries have different target framework

czb182:
This would be a nuget depenency for system.text.json 8.0.4. Wouldn't that mean it would load that at runtime instead of framework version?  Otherwise how would you be able to consume newer libraries for new features if for example I wanted to use some new system.text.json features in my .net 6 app.I don’t think this is big issue, just seems odd to cross .net libraries.  I don’t think the .net 6 version of this this library was vulnerable so it didn’t need to reference new versionSent from my iPhoneOn Jul 22, 2024, at 5:02 PM, Jean-Marc Prieur ***@***.***> wrote:﻿
@jmprieur commented on this pull request.



In Directory.Build.props:
> @@ -88,21 +88,25 @@
     <MicrosoftIdentityAbstractions>6.0.0</MicrosoftIdentityAbstractions>
-    <NetNineRuntimeVersion> 9.0.0-preview.4.24266.19</NetNineRuntimeVersion>
-    <AspNetCoreNineRuntimeVersion> 9.0.0-preview.4.24267.6</AspNetCoreNineRuntimeVersion>
+    <NetNineRuntimeVersion>9.0.0-preview.6.24327.7</NetNineRuntimeVersion>
+    <AspNetCoreNineRuntimeVersion>9.0.0-preview.6.24328.4</AspNetCoreNineRuntimeVersion>
+    
+    <SystemTextJsonVersion>8.0.4</SystemTextJsonVersion>

It's not going to pull .NET 8 libraries in .NET 6 applications as the libraries have different target framework

jmprieur:
The .NET team recommends Microsoft.Extensions 8.x

PR #2936, opened 2024-07-16
Lozensky/update deps for CVE

JoshLozensky:
updating sample dependency to follow component governance guidelines


jennyf19:
which CVE is this for?

JoshLozensky:
This PR addresses [this CVE](https://identitydivision.visualstudio.com/IDDP/_componentGovernance/Microsoft%20Identity%20Web%20-%20OneBranch/alert/11659442?typeId=24607031)

jennyf19 (approved):
not understanding why we need 9*, but it's just for the sample, so approving

Issue #2887, opened 2024-06-10
GetTokenAcquirer cannot set SendX5C


keegan-caruso:
### Microsoft.Identity.Web Library

Microsoft.Identity.Web

### Microsoft.Identity.Web version

2.19.0

### Web app

Not Applicable

### Web API

Not Applicable

### Token cache serialization

Not Applicable

### Description

As a user of 

https://github.com/AzureAD/microsoft-identity-web/blob/77df2a63d759beb5bbeb9c1260c8efdffd9fff94/src/Microsoft.Identity.Web.TokenAcquisition/DefaultTokenAcquirerFactoryImplementation.cs#L57

I have no way to set SendX5C to true.

### Reproduction steps

1. Try to use GetTokenAcquirer where sending the x5c header is required.

### Relevant code snippets

```csharp
GetTokenAcquirer(myOptionsWithX5CAsTrue)
```

### Expected behavior

GetTokenAcquirer should support sending x5c

bgavrilMS:
Is this a regression?

bgavrilMS:
I don't think this is a regression as this never worked. Still, an important bug since SN/I is pretty much mandatory internally. 

The problem is that [Id.Web does not add X5C to the assertion](https://github.com/AzureAD/microsoft-identity-web/blob/master/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs#L434), which is creates manually because this is how this protocol works. 

Here's an example of how [X5C should be added](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/tests/Microsoft.Identity.Test.Integration.netcore/HeadlessTests/LegacyPopTest.NetFwk.cs#L404). Well, there is probably easier way to inject X5C than using JSON, but this is to show where it is needed.

bgavrilMS:
I am taking this over as I have worked on this area before.

jennyf19:
https://github.com/AzureAD/microsoft-identity-web/releases/tag/3.1.0

PR #2950, opened 2024-07-23
Add X5C to MSAuth POP

bgavrilMS:
Fix for #2887

Testing: unit

### DIFF
--- a/src/Microsoft.Identity.Web.TokenAcquisition/MsAuth10AtPop.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/MsAuth10AtPop.cs
@@ -69,19 +69,17 @@ namespace Microsoft.Identity.Web
                 SigningCredentials = signingCredentials
             };
 
-            JwtSecurityTokenHandler tokenHandler = new JwtSecurityTokenHandler();
-            JwtSecurityToken token = (JwtSecurityToken)tokenHandler.CreateToken(securityTokenDescriptor);
-
             if (sendX5C)
             {
-                string exportedCertificate = Convert.ToBase64String(certificate.Export(X509ContentType.Cert));
-                token.Header.Add(
-                    IdentityModel.JsonWebTokens.JwtHeaderParameterNames.X5c, 
-                    new List<string> { exportedCertificate });
+                string x5cValue = Convert.ToBase64String(certificate.GetRawCertData());
+                securityTokenDescriptor.AdditionalHeaderClaims = 
+                    new Dictionary<string, object>() { { "x5c", new List<string> { x5cValue } } };
             }
+            
+            JsonWebTokenHandler tokenHandler = new JsonWebTokenHandler();
+            string token = tokenHandler.CreateToken(securityTokenDescriptor);
 
-            string stringToken =  tokenHandler.WriteToken(token);
-            return stringToken;
+            return token;
         }
     }
 }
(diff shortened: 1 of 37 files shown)


tests/Microsoft.Identity.Web.Test.Common/Mocks/MockHttpMessageHandler.cs
```diff
@@ -12,13 +13,13 @@ namespace Microsoft.Identity.Web.Test.Common.Mocks
 {
     public class MockHttpMessageHandler : HttpMessageHandler
     {
-        public Func<MockHttpMessageHandler, MockHttpMessageHandler> ReplaceMockHttpMessageHandler;
```
bgavrilMS, line 15: The httpClient mock will auto-ignore instance discovery, since this is an MSAL implementation detail.

src/Microsoft.Identity.Web.TokenAcquisition/MsAuth10AtPop.cs
```diff
@@ -57,14 +62,26 @@ internal static class MsAuth10AtPop
...
-            {
+            { 
                 Claims = claims,
-                SigningCredentials = new X509SigningCredentials(certificate)
+                SigningCredentials = signingCredentials
             };
 
-            var handler = new JsonWebTokenHandler();
-            return handler.CreateToken(securityTokenDescriptor);
+            JwtSecurityTokenHandler tokenHandler = new JwtSecurityTokenHandler();
```
jmprieur, line 72: Please use JsonWebTokenHandler.

src/Microsoft.Identity.Web.TokenAcquisition/MsAuth10AtPop.cs
```diff
@@ -57,14 +62,26 @@ internal static class MsAuth10AtPop
...
+            JwtSecurityToken token = (JwtSecurityToken)tokenHandler.CreateToken(securityTokenDescriptor);
+
+            if (sendX5C)
+            {
+                string exportedCertificate = Convert.ToBase64String(certificate.Export(X509ContentType.Cert));
```
jmprieur, line 77: Does this means that the certificate needs to be exportable?
It's only the public key?

tests/Microsoft.Identity.Web.Test.Common/Mocks/MockHttpClientFactory.cs
```diff
@@ -43,6 +66,8 @@ public MockHttpMessageHandler AddMockHandler(MockHttpMessageHandler handler)
 
         public HttpClient GetHttpClient()
         {
+
```
jmprieur, line 69: Do we need to have 2 spaces?

jmprieur (changes requested):
LGTM
Thanks @bgavrilMS 

I left a few questions and suggestions

jmprieur (approved):

src/Microsoft.Identity.Web.TokenAcquisition/MsAuth10AtPop.cs
bgavrilMS: The certificate is still protected by the password. See the unit test where I load a cert by the x5c value and I have to provide a password. 

Anyway, I switched to use `GetRawCertData` as this is what MSAL uses for x5c.

jennyf19 (approved):
:shipit:

bgavrilMS:
Thanks for the reviews and for merging this @jennyf19 and @jmprieur

Issue #2930, opened 2024-07-15
[Feature Request] Overloaded GetForUserAsync that takes a JsonTypeInfo<T> as a parameter to enable source generated Json deserialization (For NativeAOT compliance)


dualtagh:
**Is your feature request related to a problem? Please describe.**
microsoft-identity-web emits NativeAOT trim warnings when GetForUserAsync  is called from a NativeAOT compiled library. This is caused by using generics for JSON serialization/deserialization in DownstreamApi.cs.
```
ILC : Trim analysis error IL2026: Microsoft.Identity.Web.DownstreamApi.<DeserializeOutput>d__17`1.MoveNext(): Using member 'System.Text.Json.JsonSerializer.Deserialize<TOutput>(String,JsonSerializerOptions)' which has 'RequiresUnreferencedCodeAttribute' can break functionality when trimming application code. JSON serialization and deserialization might require types that cannot be statically analyzed. Use the overload that takes a JsonTypeInfo or JsonSerializerContext, or make sure all of the required types are preserved.
```

**Describe the solution you'd like**
Public interfaces (particularly GetForUserAsync) should accept a JsonTypeInfo<T> so that source generation can be used for Json serialization in a NativeAOT context.
Existing interfaces should be marked with the [RequiresUnreferencedCode] attribute. New interfaces that accept JsonTypeInfo should be made available only for net8+.

Example change for IDownstreamApi:
```
 [RequiresUnreferencedCode]
public Task<TOutput?> GetForUserAsync<TOutput>(
    string? serviceName,
    Action<DownstreamApiOptionsReadOnlyHttpMethod>? downstreamApiOptionsOverride = null,
    ClaimsPrincipal? user = null,
    CancellationToken cancellationToken = default) where TOutput : class;

public Task<TOutput?> GetForUserAsync<TOutput>(
    string? serviceName,
    JsonTypeInfo<TOutput> responseJsonTypeInfo,
    Action<DownstreamApiOptionsReadOnlyHttpMethod>? downstreamApiOptionsOverride = null,
    ClaimsPrincipal? user = null,
    CancellationToken cancellationToken = default) where TOutput : class;
```

**Describe alternatives you've considered**
Open to suggestions

**Additional context**
There is a short term need for GetForUserAsync to be NativeAOT compliant - but there are other interfaces with AOT warnings that could require similar changes. Will follow up with more details on this

keegan-caruso:
[...]

PR #2959, opened 2024-07-31
Implement IDownstreamApi overloads that take JsonTypeInfo<T> as a parameter to enable source generated Json deserialization for NativeAOT

dualtagh:
Addresses #2930

- Update Microsoft.Identity.Abstractions to 7.0.0 
- Implementation for new DownstreamApi interfaces in Microsoft.Identity.Abstractions for source gen Json serialization. Implementaitons are generated or copied as the existing with just the addition of jsonTypeInfo params


Directory.Build.props
```diff
@@ -111,6 +111,7 @@
     <SystemFormatsAsn1Version>$(NetNineRuntimeVersion)</SystemFormatsAsn1Version>
     <SystemTextJsonVersion>$(NetNineRuntimeVersion)</SystemTextJsonVersion>
     <MicrosoftExtensionsDependencyInjectionVersion>$(NetNineRuntimeVersion)</MicrosoftExtensionsDependencyInjectionVersion>
+    <MicrosoftIdentityAbstractionsVersion>7.0.0</MicrosoftIdentityAbstractionsVersion>
```
jmprieur, line 114: What happens if we reference IdentityAbstractions 7.0.0 in other frameworks? this should work too given the methods in Abstractions are defined only above 8.0.0?

I think we should systematically reference IdentityAbstractions 7.0.0

tests/Microsoft.Identity.Web.Test/Microsoft.Identity.Web.Test.csproj
```diff
@@ -20,7 +20,7 @@
 
   <ItemGroup>
     <PackageReference Include="System.Text.Json" Version="$(SystemTextJsonVersion)" />
-    <PackageReference Include="Microsoft.Identity.Abstractions" Version="$(MicrosoftIdentityAbstractionsVersion)" />
```
jmprieur, line 23: I would keep that

Directory.Build.props
jmprieur: Actually I checked. Just update line 91.
Let's just factorize in the common properties. No special case for net8.0 and net.9.0

jmprieur (approved):
I approved but requested that you don't special case the reference for net8.0 and net9.0 in the Directory.Build.props, but just use Abstractions 7.0.0 for all target framework (change line 91).

Otherwise this will break our unified / id4sbuild
cc: @jennyf19

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
```diff
@@ -219,7 +302,7 @@ public Task<HttpResponseMessage> CallApiForAppAsync(
             return clonedOptions;
         }
 
-        internal static HttpContent? SerializeInput<TInput>(TInput input, DownstreamApiOptions effectiveOptions)
+        internal static HttpContent? SerializeInput<TInput>(TInput input, DownstreamApiOptions effectiveOptions, JsonTypeInfo<TInput>? inputJsonTypeInfo = null)
```
pmaytak, line 305: Adding an optional parameter is a binary breaking change. I think this should be an additional overload and add it to NET8 only section.

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
```diff
@@ -234,17 +317,23 @@ public Task<HttpResponseMessage> CallApiForAppAsync(
...
+                        inputJsonTypeInfo == null ? JsonSerializer.Serialize(input) : JsonSerializer.Serialize(input, inputJsonTypeInfo),
+                        Encoding.UTF8,
+                        "application/json"),
                 };
             }
             return httpContent;
         }
 
-        internal static async Task<TOutput?> DeserializeOutput<TOutput>(HttpResponseMessage response, DownstreamApiOptions effectiveOptions)
+        internal static async Task<TOutput?> DeserializeOutput<TOutput>(HttpResponseMessage response, DownstreamApiOptions effectiveOptions, JsonTypeInfo<TOutput>? outputJsonTypeInfo = null)
```
pmaytak, line 336: Same comment as above.

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
pmaytak, line 1: I assume you won't be using `CallApiAsync`, so not adding overloads for them for now?

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
dualtagh: @pmaytak Thanks, I've updated it to have an overload and keep the original signature. I've put the implementation into a different method though - is this enough re binary breaking change for an internal method? Or does the original implementation need to stay the same?

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
```diff
@@ -166,8 +167,101 @@ public Task<HttpResponseMessage> CallApiForAppAsync(
...
+            where TOutput : class
+        {
+            DownstreamApiOptions effectiveOptions = MergeOptions(serviceName, downstreamApiOptionsOverride);
+            HttpResponseMessage response = await CallApiInternalAsync(serviceName, effectiveOptions, true,
+                                                                          null, null, cancellationToken).ConfigureAwait(false);
+
+            return await DeserializeOutput<TOutput>(response, effectiveOptions, outputJsonTypeInfo).ConfigureAwait(false);
+        }
+
+        internal static HttpContent? SerializeInput<TInput>(TInput input, DownstreamApiOptions effectiveOptions, JsonTypeInfo<TInput>? inputJsonTypeInfo)
```
pmaytak, line 253:
```suggestion
        internal static HttpContent? SerializeInput<TInput>(TInput input, DownstreamApiOptions effectiveOptions, JsonTypeInfo<TInput> inputJsonTypeInfo)
```

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
```diff
@@ -166,8 +167,101 @@ public Task<HttpResponseMessage> CallApiForAppAsync(
...
+        {
+            return SerializeInputImpl(input, effectiveOptions, inputJsonTypeInfo);
+        }
+
+        internal static async Task<TOutput?> DeserializeOutput<TOutput>(HttpResponseMessage response, DownstreamApiOptions effectiveOptions, JsonTypeInfo<TOutput>? outputJsonTypeInfo = null)
```
pmaytak, line 258: To match SerializeInput signature:
```suggestion
        internal static async Task<TOutput?> DeserializeOutput<TOutput>(HttpResponseMessage response, DownstreamApiOptions effectiveOptions, JsonTypeInfo<TOutput> outputJsonTypeInfo)
```

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
pmaytak: I think this is good.

pmaytak (approved):

jennyf19 (approved):
Great to see the additional test coverage @dualtagh 
:shipit:

PR #2966, opened 2024-08-06
Switch nuget source to dotnet public

keegan-caruso:
# Switch nuget source to dotnet public


keegan-caruso:
Microsoft.Identity.Lab.Api isn't available on this feed

PR #2971, opened 2024-08-09
Fix test instability

keegan-caruso:
Revert some of the test changes introduces in #2950

That change introduced test instability

fixes #2964

PR #2972, opened 2024-08-09
Fix AT POP tests

keegan-caruso:
Use existing pattern in cache extensions test in
MsAuth10AtPopTests

Continuation of #2971 

Fixes test instability introduced by #2950


jennyf19 (approved):
:shipit:

PR #2974, opened 2024-08-12
[test] updates for one build

jennyf19:
Ensures that the same build variable name (`MicrosoftIdentityWebVersion`) is used both for the version of the library itself, and in the libraries that take a dependency on Wilson. This name was already used in other libraries. This becomes the way to set the version in Release builds.

### DIFF
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -3,6 +3,5 @@
   <packageSources>
     <clear />
     <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
-    <add key="Package source 1" value="c:\one\packages" />
   </packageSources>
 </configuration>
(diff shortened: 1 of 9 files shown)


NuGet.Config
```diff
@@ -3,5 +3,6 @@
   <packageSources>
     <clear />
     <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
+    <add key="Package source 1" value="c:\one\packages" />
```
jennyf19, line 6:
```suggestion
```

bgavrilMS (approved):

jmprieur (approved):
LGTM
Thanks @jennyf19

jmprieur (approved):

PR #2977, opened 2024-08-13
Add BannedApiAnalyzers to prevent use of ClaimsIdentity constructors and AppContextSwitches for fallback

pmaytak:
Use CaseSensitiveClaimsIdentity by default.

Add AppContextSwitches to fallback to using ClaimsIdentity.

Add [BannedApiAnalyzers](https://github.com/dotnet/roslyn-analyzers/blob/main/src/Microsoft.CodeAnalysis.BannedApiAnalyzers/BannedApiAnalyzers.Help.md) to prevent use of ClaimsIdentity constructors and enforce use of CaseSensitiveClaimsIdentity, which makes claims retrieval consistent. 

The disallowed definitions are contained in BannedSymbols.txt. There is no wildcard to include all constructors for a type, so each one has to be added.

If a banned member is used, the following error is thrown:
`RS0030	The symbol has been marked as banned in this project, and an alternate should be used instead.`


jmprieur (approved):
LGTM
Thanks @pmaytak 

Assuming the RequireScopeXXX won't introduce a breaking change as we consume the claims identity ourselves?

tests/Microsoft.Identity.Web.Test.Common/TestConstants.cs
```diff
@@ -55,6 +55,8 @@ public static class TestConstants
         public const string GraphBaseUrlBeta = "https://graph.microsoft.com/beta";
         public const string GraphBaseUrl = "https://graph.microsoft.com/v1.0";
 
+        public const string IdToken = "ewogICJ0eXAiOiAiSldUIiwKICAiYWxnIjogIlJTMjU2IiwKICAia2lkIjogImsiCn0=.ewogICJuYW1lIjogIlVzZXIiCn0=.";
```
keegan-caruso, line 58: ?

jennyf19 (approved):
:shipit:

tests/Microsoft.Identity.Web.Test.Common/TestConstants.cs
pmaytak: Needed to have a well-constructed token for the test to run.

PR #2984, opened 2024-08-18
Ensure that SimulateOidc is built before IntegrationTests (that use it)

jmprieur:
# Ensure that SimulateOidc is built before IntegrationTests (that use it)


jennyf19 (approved):
:shipit:

Issue #2675, opened 2024-02-13
Microsoft.Identity.Web doesn't respect the use of NullIdentityModelLogger.


nlz242:
### Microsoft.Identity.Web Library

Microsoft.Identity.Web

### Microsoft.Identity.Web version

2.16.0

### Web app

Sign-in users and call web APIs

### Web API

Protected web APIs call downstream web APIs

### Token cache serialization

Not Applicable

### Description

A little bit of background before the issue itself...
Since 1.25.1 we'be been having some pretty serious performance issues. Our app has multiple authentication scheme defined and we switching to 1.25.1, we've noticed that under some load, we start loosing >2 seconds on some calls. After profiling the code, we've found that this time is spent failing IssuerValidation and logging failures, most of the wall-clock time is spent in logging failures. This wasn't an issue in 1.25.0. This has been reported in a couple of issues already. 

We've found two solution to our problem. The first one was suggested by @jmprieur, was basically to setup our authentication scheme and the defaut auth scheme logically. Doing that did bring the time loss down a bit, but we would still, under load, see >1 second degradation. 
So second solution was to create a logger that would do nothing and assign it to LogHelper.Logger. This worked great, performance was back where it should.

During a review of that code, we noticed we didn't have to create that logger, one already exists in Microsoft.IdentityModel.Abstraction, its called the NullIdentityModelLogger. So instead of re-inventing the wheel, we replaced our NullLogger with Microsoft's... and performance problems came back. I couldn't understand for the life of me WHY Microsoft's NullIdentityModelLogger would be slow, as it does NOTHING...

That is until I found this : 
https://github.com/AzureAD/microsoft-identity-web/blob/397fce764ad436a68934baa545f2b8884e70ae8c/src/Microsoft.Identity.Web.TokenAcquisition/MicrosoftIdentityBaseAuthenticationBuilder.cs#L41-L66

So if we explicitly set the logger to be the NullIdentityModelLogger, Microsoft.Identity.Web decides that it needs to it's own logger back in the game without any respect for our choice... Clearly a design shortcut to avoid having a configuration somewhere or something similar.

### Reproduction steps

1. Set LogHelper.Logger to NullIdentityModelLogger.Instance
2. Notice Wilson still logs
3. Create a duplicate of NullIdentityModelLogger, name it something else
4. Set LogHelper.Logger to this new instance
5. Notice Wilson doesn't log, like it should

### Relevant code snippets

```csharp
None
```

### Regression

[...]

PR #3003, opened 2024-09-03
Fix for 3000 - improve error messages when credentials fail to load

### LINKED ISSUE #3000: Improve the error message and logs if step1 of FIC fails

Microsoft.Identity.Web

3.1.0

### Web app

Sign-in users

### Web API

Not Applicable

### Token cache serialization

In-memory caches

### Description

Web Site calls Graph scenario. Instead of a secret, use a FIC. But configure a bad FIC, e.g. set a invalid id. 

**Actual**: ArgumentException: Delegate to an instance method cannot have null 'this'.
System.MulticastDelegate.ThrowNullThisInDelegateToInstance()
System.MulticastDelegate.CtorClosed(object target, IntPtr methodPtr)
Microsoft.Identity.Web.ConfidentialClientApplicationBuilderExtension.WithClientCredentialsAsync(ConfidentialClientApplicationBuilder builder, IEnumerable<CredentialDescription> clientCredentials, ILogger logger, ICredentialsLoader credentialsLoader, CredentialSourceLoaderParameters credentialSourceLoaderParameters)

**Expected**: error message and logs should inform the app developer that step1 of FIC failed

### Reproduction steps

Web Site calls Graph scenario. Instead of a secret, use a FIC. But configure a bad FIC, e.g. set a invalid id. 

### Error message

_No response_

### Id Web logs

_No response_

### Relevant code snippets

```csharp
"AzureAd": {
   "Domain": "bogavril.onmicrosoft.com",
   "TenantId": "839846c0-9cef-4455-9542-0c36d831d026",
   "ClientId": "089dfe99-af94-4a19-b803-1b01b456b271",
   "ClientCapabilities": [ "cp1" ],
   "CallbackPath": "/signin-oidc"
   
 }


```csharp
 options.ClientCredentials = new List<CredentialDescription>
 {
     new CredentialDescription() {
         SourceType= CredentialSource.SignedAssertionFromManagedIdentity,
         ManagedIdentityClientId = "089dfe99-af94-4a19-b803-1b01b456b271"  } 
   
 };
```

### Regression

no

### Expected behavior

Good error message.

bgavrilMS:
Fixes #3000

Function changes: 

- DefaultCredentialLoader is now responsible for setting the Skip flag, instead of individual requests
- Added more logging
- If all credentails fail, the exception messages now contains all the detail (similar to how AzureDefaultCredential has all the details of failing creds)

## Testing

- [x] Unit test that the correct assertion gets into MSAL from many combinations of credentials
- [x] Manual test (and unit test) for using a certificate chain of `MSI` and `cert`, where `MSI` fails
- [x] Manual test with cred chain - "AssertionFromFile" and then "CertificateFromKV".


src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
```diff
@@ -69,7 +69,25 @@ public async Task LoadCredentialsIfNeededAsync(CredentialDescription credentialD
...
+                            {
+                                _logger?.LogInformation(ex, $"Failed to load credential {credentialDescription.Id} from source {credentialDescription.SourceType}.");
+
+                                // Certificates are re-attempted, but assertions are not (behavior in Id.Web 3.1)
+                                // It makes sense to skip assertions (e.g. federated identity) because many developers use a "FIC (federated identity credential)" + "certificate fallback" setup
+                                // and we don't want to keep checking the FIC.
+                                if (credentialDescription.CredentialType == CredentialType.SignedAssertion)
+                                {
+                                    _logger.LogDebug("Setting Skip to true");                                    
+                                    credentialDescription.Skip = true;
```
bgavrilMS, line 87: I have kept the logic as is, i.e. only the "signed assertion" have this flag set to true. 

However, I think it makes sense to set it always. If a credential fails to load, is there any value in re-trying it? 

Note: this is the root cause of the issue #3000 - this flag wasn't set

src/Microsoft.Identity.Web.Certificate/SignedAssertionFromManagedIdentityCredentialLoader.cs
```diff
@@ -21,18 +21,14 @@ public async Task LoadIfNeededAsync(CredentialDescription credentialDescription,
...
                     managedIdentityClientAssertion = new ManagedIdentityClientAssertion(credentialDescription.ManagedIdentityClientId, credentialDescription.TokenExchangeUrl);
                 }
-                try
-                {
-                    // Given that managed identity can be not available locally, we need to try to get a
-                    // signed assertion, and if it fails, move to the next credentials
-                    _= await managedIdentityClientAssertion!.GetSignedAssertionAsync(null);
-                    credentialDescription.CachedValue = managedIdentityClientAssertion;
-                }
-                catch (AuthenticationFailedException)
```
bgavrilMS, line 31: This the root cause of the bug. When we moved from Azure SDK to MSAL, I forgot to update this exception type. Ultimately, it doesn't matter why the credential fails, so a catch all is better used (up the stack in DefaultCredentialLoader)

src/Microsoft.Identity.Web.TokenAcquisition/ConfidentialClientApplicationBuilderExtension.cs
```diff
@@ -58,51 +90,47 @@ public static async Task<ConfidentialClientApplicationBuilder> WithClientCredent
...
 
-            if (clientCredentials.Any(c => c.CredentialType == CredentialType.Certificate))
-            {
-                throw new ArgumentException(
-                    IDWebErrorMessage.ClientCertificatesHaveExpiredOrCannotBeLoaded,
-                    nameof(clientCredentials));
-            }
+            throw new ArgumentException(
+               IDWebErrorMessage.ClientCertificatesHaveExpiredOrCannotBeLoaded + errorMessage,
+               nameof(clientCredentials));
```
bgavrilMS, line 132: This also part of fix for #3000 - if none of the credentials can be loaded, we should fail with a meaningful message. Instead, the Exception here is conditioned on the existence of a certificate, so this code didn't execute for FIC.

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
```diff
@@ -69,7 +94,17 @@ public async Task LoadCredentialsIfNeededAsync(CredentialDescription credentialD
...
-                            await loader.LoadIfNeededAsync(credentialDescription, parameters);
+                        {
+                            try
+                            {
+                                await loader.LoadIfNeededAsync(credentialDescription, parameters);
+                            }
+                            catch (Exception ex)
+                            {
+                                Logger.CredentialLoadingFailure(_logger, credentialDescription, ex);
+                                throw;
```
bgavrilMS, line 105: Not changing the semantics of this method, just logging failures.

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
```diff
@@ -81,11 +116,18 @@ public async Task LoadCredentialsIfNeededAsync(CredentialDescription credentialD
...
+        /// Loads first valid credential which is not marked as Skipped. 
+        public async Task<CredentialDescription?> LoadFirstValidCredentialsAsync(
+            IEnumerable<CredentialDescription> credentialDescriptions, 
+            CredentialSourceLoaderParameters? parameters = null)
         {
             foreach (var credentialDescription in credentialDescriptions)
             {
                 await LoadCredentialsIfNeededAsync(credentialDescription, parameters);
+
+                // TODO:  Exceptions which occur in trying various credentials are logged and ignored.
```
bgavrilMS, line 128: I didn't change here but this method looks broken because it doesn't handle exceptions. There is no coverage of this.

src/Microsoft.Identity.Web.Certificate/SignedAssertionFilePathCredentialsLoader.cs
```diff
@@ -36,13 +36,14 @@ public async Task LoadIfNeededAsync(CredentialDescription credentialDescription,
...
                     // signed assertion, and if it fails, move to the next credentials
-                    _= await signedAssertion!.GetSignedAssertionAsync(null);
+                    _ = await signedAssertion!.GetSignedAssertionAsync(null);
                     credentialDescription.CachedValue = signedAssertion;
                 }
                 catch (Exception)
                 {
                     credentialDescription.Skip = true;
-                }
+                    throw;
```
bgavrilMS, line 45: This is part of the fix, because SignedAssertionFilePath loader would otherwise fail silently and lead to no credentail being used in MSAL. All loaders should throw.

src/Microsoft.Identity.Web.Certificateless/ManagedIdentityClientAssertion.cs
```diff
@@ -39,9 +46,26 @@ public ManagedIdentityClientAssertion(string? managedIdentityClientId)
...
+
+            var id = ManagedIdentityId.SystemAssigned;
+            if (!string.IsNullOrEmpty(managedIdentityClientId))
+            {
+                id = ManagedIdentityId.WithUserAssignedClientId(managedIdentityClientId);
+            }
+            var builder = ManagedIdentityApplicationBuilder.Create(id);
+            if (_logger != null)
+            {
+                builder.WithLogging(Log, ConvertMicrosoftExtensionsLogLevelToMsal(_logger));
```
bgavrilMS, line 63: Added logging into MSAL for this.

tests/Microsoft.Identity.Web.Test/Certificates/WithClientCredentialsTests.cs
```diff
@@ -0,0 +1,199 @@
...
+using NSubstitute;
+using NSubstitute.ExceptionExtensions;
+using Xunit;
+
+namespace Microsoft.Identity.Web.Test.Certificates
+{
+    public class WithClientCredentialsTests
+    {
+        [Fact]
+        public async Task FicFails_CertificateFallback()
```
bgavrilMS, line 21: This is the most important test.

src/Microsoft.Identity.Web.Certificate/SignedAssertionFromManagedIdentityCredentialLoader.cs
jmprieur: Makes sense. thanks for finding that out.

src/Microsoft.Identity.Web.TokenAcquisition/ConfidentialClientApplicationBuilderExtension.cs
```diff
@@ -30,23 +30,61 @@ public static async Task<ConfidentialClientApplicationBuilder> WithClientCredent
...
+                    return builder.WithCertificate(credential.Certificate);
+                case CredentialType.Secret:
+                    return builder.WithClientSecret(credential.ClientSecret);
+                default:
+                    throw new NotImplementedException();
+
+            }
+        }
+
+        internal /* for test */ async static Task<CredentialDescription?> LoadCredentialForMsalOrFailAsync(
```
bgavrilMS, line 61: Logic remains mostly the same, I just separated the existing method into 2, so that I can unit test.

src/Microsoft.Identity.Web.TokenAcquisition/ConfidentialClientApplicationBuilderExtension.cs
```diff
@@ -58,51 +96,54 @@ public static async Task<ConfidentialClientApplicationBuilder> WithClientCredent
...
                     if (credential.CredentialType == CredentialType.Secret)
                     {
-                        return builder.WithClientSecret(credential.ClientSecret);
+                        return credential;
                     }
                 }
             }
 
-            if (clientCredentials.Any(c => c.CredentialType == CredentialType.Certificate))
+            if (clientCredentials.Any(c => c.CredentialType == CredentialType.Certificate || c.CredentialType == CredentialType.SignedAssertion))
```
bgavrilMS, line 136: I changed this condition. I originally wanted to throw if no credential could be set, but this is not as simple because there are so MANY ways to set a secret (ASP options, CCA options etc.)

jmprieur (commented):
I would be great to create an end to end test

src/Microsoft.Identity.Web.Certificateless/ManagedIdentityClientAssertion.cs
jmprieur: Don't we have this conversion already in IdWeb?

bgavrilMS: We do, once for CCA, once for ManagedIdentityApplication. But I could not find a way to consolidate this code. 

I can move it all to the "Diagnostics" project, but in this case the Diagnostics project will depend on MSAL. Is this acceptable?

src/Microsoft.Identity.Web.Certificateless/ManagedIdentityClientAssertion.cs
```diff
@@ -39,9 +46,26 @@ public ManagedIdentityClientAssertion(string? managedIdentityClientId)
...
+                _logger?.LogInformation($"ManagedIdentityClientAssertion with tokenExchangeUrl={_tokenExchangeUrl}");
+            }
+
+            _managedIdentityApplication = builder.Build();
+            _tokenExchangeUrl = CertificatelessConstants.DefaultTokenExchangeUrl;
```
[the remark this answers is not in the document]
bgavrilMS: good catch

src/Microsoft.Identity.Web.Certificate/SignedAssertionFromManagedIdentityCredentialLoader.cs
jennyf19: That explains why this showed up recently then.

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
```diff
@@ -10,11 +11,34 @@
...
+
     /// <summary>
     /// Default credentials loader.
     /// </summary>
     public class DefaultCredentialsLoader : ICredentialsLoader
     {
+        /// <summary>
+        /// Logging infrastructure
+        /// </summary>
+        private static class Logger
```
jennyf19, line 23: Please use a similar set up to the high performance logging similar to [TokenAcquisition.Logger.cs](https://github.com/AzureAD/microsoft-identity-web/blob/master/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.Logger.cs)

src/Microsoft.Identity.Web.Certificateless/ManagedIdentityClientAssertion.cs
jennyf19: would like to not have it in two places if possible, but also not in Diagnostics

bgavrilMS: I'll just remove it for now.

jmprieur (approved):

jennyf19:
@bgavrilMS can you update w/master and respond to comments or mark them as resolved? thanks.

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
bgavrilMS: This has been adressed.

src/Microsoft.Identity.Web.Certificateless/ManagedIdentityClientAssertion.cs
```diff
@@ -52,11 +76,62 @@ public ManagedIdentityClientAssertion(string? managedIdentityClientId, string? t
...
             var result = await _managedIdentityApplication
-                .AcquireTokenForManagedIdentity(_tokenExchangeUrl)
+                .AcquireTokenForManagedIdentity(_tokenExchangeUrl)                
                 .ExecuteAsync(assertionRequestOptions?.CancellationToken ?? CancellationToken.None)
                 .ConfigureAwait(false);
 
             return new ClientAssertion(result.AccessToken, result.ExpiresOn);
         }
+
+        private void Log(
```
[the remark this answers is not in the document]
bgavrilMS: We need this.

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
```diff
@@ -81,11 +95,18 @@ public async Task LoadCredentialsIfNeededAsync(CredentialDescription credentialD
...
+        public async Task<CredentialDescription?> LoadFirstValidCredentialsAsync(
+            IEnumerable<CredentialDescription> credentialDescriptions, 
+            CredentialSourceLoaderParameters? parameters = null)
         {
             foreach (var credentialDescription in credentialDescriptions)
             {
                 await LoadCredentialsIfNeededAsync(credentialDescription, parameters);
+
+                // TODO:  Exceptions which occur in trying various credentials are logged and ignored.
+                // is this really intended behavior? This is different from WithClientCredentialsAsync
```
jmprieur, line 108: who should make the decision? me?

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
```diff
@@ -81,11 +95,18 @@ public async Task LoadCredentialsIfNeededAsync(CredentialDescription credentialD
...
+        /// Loads first valid credential which is not marked as Skipped. 
+        public async Task<CredentialDescription?> LoadFirstValidCredentialsAsync(
+            IEnumerable<CredentialDescription> credentialDescriptions, 
+            CredentialSourceLoaderParameters? parameters = null)
         {
             foreach (var credentialDescription in credentialDescriptions)
             {
                 await LoadCredentialsIfNeededAsync(credentialDescription, parameters);
+
+                // TODO:  Exceptions which occur in trying various credentials are logged and ignored.
```
jmprieur, line 107: who should do? me?

src/Microsoft.Identity.Web.Certificate/SignedAssertionFilePathCredentialsLoader.cs
jmprieur, line 45: If we throw at all stages, how will the chained credentials work?

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
bgavrilMS: Right, it was a question for the designer of this public API. There are no tests around it, so I cannot guess the intented behavior.  I can raise a bug if you like.

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
bgavrilMS: I will delete the comment and raise a bug.

src/Microsoft.Identity.Web.Certificate/SignedAssertionFilePathCredentialsLoader.cs
bgavrilMS: The orchestrator catches exceptions. You can see all the other loaders having this problem.

bgavrilMS: Silently failing here means you'll get a strange from MSAL, because you'll assume the credential worked and you'll use the `credentialDescription.CachedValue` which is null.

bgavrilMS: But I can undo this if you like.

jmprieur: that's not what I'm saying.
I just want to understand the new behavior and test it

jmprieur: I'll run end to end tests.

bgavrilMS: I can run a manual test @jmprieur . How about this: 

- cred chain of { Kubernetes FIC, certificate } 
- the Kubernetes FIC will fail 

Expected: the certificate will be attempted.

bgavrilMS: Done^^

src/Microsoft.Identity.Web.Certificateless/ManagedIdentityClientAssertion.cs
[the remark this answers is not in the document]
jennyf19: I know, but now we have the same code duplicated in two places.

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
```diff
@@ -69,7 +73,17 @@ public async Task LoadCredentialsIfNeededAsync(CredentialDescription credentialD
...
                         if (CredentialSourceLoaders.TryGetValue(credentialDescription.SourceType, out ICredentialSourceLoader? loader))
-                            await loader.LoadIfNeededAsync(credentialDescription, parameters);
+                        {
+                            try
+                            {
+                                await loader.LoadIfNeededAsync(credentialDescription, parameters);
+                            }
+                            catch (Exception ex)
+                            {
+                                Logger.CredentialLoadingFailure(_logger, credentialDescription, ex);
```
jennyf19, line 83: please address the warnings. thanks

jmprieur:
Started doing end to end testing, and there seem to be a regression on https://github.com/AzureAD/microsoft-identity-web/issues/2675

tests/DevApps/daemon-app/daemon-console-calling-downstreamApi/DaemonConsoleCallingDownstreamApi.csproj
```diff
@@ -10,6 +10,8 @@
 
   <ItemGroup>
     <Compile Include="..\minimal-web-api\WeatherForecast.cs" Link="WeatherForecast.cs" />
+    <PackageReference Include="Microsoft.Extensions.Logging" Version="8.0.0" />
```
jmprieur, line 13: probably not needed.

jmprieur: not indeed.

jmprieur:
> Started doing end to end testing, and there seem to be a regression on #2675

Actually no. It was a test error

jmprieur:
I ran regression tests, which passed

image

jmprieur (approved):

jennyf19 (approved):
:shipit:

PR #3017, opened 2024-09-17
Bump to MSAL 4.64.1

bgavrilMS:
Id.Web always needs to be on the latest MSAL. Otherwise we can't investigate issues.


jmprieur (approved):
LGTM
Thanks @bgavrilMS 

No reference to remove? (like wilson.abstractions)?

jmprieur:
> Id.Web always needs to be on the latest MSAL. Otherwise we can't investigate issues.

It should be in the MSAL.NET release process then!

bgavrilMS:
> LGTM Thanks @bgavrilMS
> 
> No reference to remove? (like wilson.abstractions)?

Thanks @jmprieur .. Id.Web.TokenCache does not depend directly on wilson.abstractions, but through MSAL. And MSAL still uses it (e.g. for logging). Did you mean anything else?

image

jmprieur:
no. that's all. Was checking

jmprieur (approved):

keegan-caruso (approved):

jennyf19:
> > Id.Web always needs to be on the latest MSAL. Otherwise we can't investigate issues.
> 
> It should be in the MSAL.NET release process then!

The MSAL team could provide the PRs for when they release and we don't generally take ones that are PCA Only with no CCA improvements.

PR #3024, opened 2024-09-19
Fix warnings

jmprieur:
# Fix Static analysis warnings

### DIFF
--- a/tests/Microsoft.Identity.Web.Test.Common/Mocks/QueryStringParser.cs
+++ b/tests/Microsoft.Identity.Web.Test.Common/Mocks/QueryStringParser.cs
@@ -57,7 +57,11 @@ namespace Microsoft.Identity.Web.Test.Common.Mocks
                 return message;
             }
 
+#if NET462 || NET472
             message = message.Replace("+", "%20");
+#else
+            message = message.Replace("+", "%20", StringComparison.OrdinalIgnoreCase);
+#endif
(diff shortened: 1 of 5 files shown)


jennyf19 (approved):
:shipit:

tests/Microsoft.Identity.Web.Test.Common/Mocks/QueryStringParser.cs
```diff
@@ -57,7 +57,11 @@ private static string UrlDecode(string message)
                 return message;
             }
 
+#if NET462 || NET472
```
bgavrilMS, line 60:
```suggestion
#if NETFRAMEWORK
```
I doubt they'll ever add it in...

bgavrilMS (approved):

tests/Microsoft.Identity.Web.Test.Common/Mocks/QueryStringParser.cs
henrik-me: should possibly not even be called this though. we are starting to remove core from the naming since v6. So perhaps LegacyNetframework?

bgavrilMS: `NETFRAMEWORK` is what the .NET team calls it. It's baked into all projects. No need to define a new pre-compilation flag.

PR #3039, opened 2024-09-26
Refactored UI Tests 

JoshLozensky:
## Description

Refactored UI tests to be more uniform in implementation so it is easier to maintain and understand.

Additionally, removed ConfigureAwait(false) from async test calls as XUnit guidance suggests this can negatively impact concurrence [as described here](https://xunit.net/xunit.analyzers/rules/xUnit1030). (I'll issue a separate PR suggesting removal from all Id Web tests).

### DIFF
--- a/tests/E2E Tests/WebAppUiTests/B2CWebAppCallsWebApiLocally.cs
+++ b/tests/E2E Tests/WebAppUiTests/B2CWebAppCallsWebApiLocally.cs
@@ -4,9 +4,10 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Globalization;
 using System.IO;
 using System.Runtime.Versioning;
-using System.Threading;
+using System.Text;
 using System.Threading.Tasks;
 using Azure.Identity;
 using Microsoft.Identity.Lab.Api;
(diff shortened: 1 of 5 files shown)


tests/E2E Tests/WebAppUiTests/B2CWebAppCallsWebApiLocally.cs
```diff
@@ -46,7 +49,7 @@ public async Task Susi_B2C_LocalAccount_TodoAppFucntionsCorrectly()
         {
             // Web app and api environmental variable setup.
             DefaultAzureCredential azureCred = new();
-            string clientSecret = await UiTestHelpers.GetValueFromKeyvaultWitDefaultCreds(_keyvaultUri, KeyvaultClientSecretName, azureCred);
+            string clientSecret = await UITH.GetValueFromKeyvaultWitDefaultCreds(_keyvaultUri, KeyvaultClientSecretName, azureCred);
```
pmaytak, line 52: I suggest keeping `UiTestHelpers` naming. `UITH` is not a common acronym, so it's not immediately obvious what it represents - makes code less readable. Plus with IntelliSense auto-complete, the name length is not a huge issue.

PR #3041, opened 2024-09-26
Onboard Id Web to Threading Analyzers

westin-m:
The following E2E scenarios were tested using the corresponding devapps:
Web app calls web api call graph
B2C web app calls web api
Owin
Blazer server calls web api
Blazer server B2C calls api
Ciam web app calls api
Web app calls graph
Deamon console calling downstream api, graph

All are working as expected.


src/Microsoft.Identity.Web.Certificate/DefaultCertificateLoader.cs
```diff
@@ -74,6 +75,29 @@ public static string? UserAssignedManagedIdentityClientId
...
+        
+        /// <summary>
+        /// Load the first certificate from the certificate description list.
+        /// </summary>
+        /// <param name="certificateDescriptions">Description of the certificates.</param>
+        /// <returns>First certificate in the certificate description list.</returns>
+        public static async Task<X509Certificate2?> LoadFirstCertificateAsync(IEnumerable<CertificateDescription> certificateDescriptions)
+        {
+            DefaultCertificateLoader defaultCertificateLoader = new(null);
+            CertificateDescription? certDescription = new();
```
jmprieur, line 87: not sure why you don't initialize it to null?

src/Microsoft.Identity.Web/AzureADB2COpenIDConnectEventHandlers.cs
```diff
@@ -31,7 +31,7 @@ public AzureADB2COpenIDConnectEventHandlers(
 
         public MicrosoftIdentityOptions Options { get; }
 
-        public Task OnRedirectToIdentityProvider(RedirectContext context)
```
jmprieur, line 34: Did you try pack the project?
Didn't the public analyzers didn't complain.

These events are defined by ASP.NET Core? do they have the Async in their name?

jmprieur (commented):
LGTM at first review, @westin-m 
Make sure you run end to end tests with credentials.
There was a race condition at some point, which is why the credential loaders used GetAwaiter.GetAsync().

I'm surprised there is no change in the public API?

src/Microsoft.Identity.Web.OWIN/OpenIdConnectCachingSecurityTokenProvider.cs
```diff
@@ -76,12 +76,12 @@ public IEnumerable<SecurityKey>? SecurityKeys
...
         }
 
-        private void RetrieveMetadata()
+        private async Task RetrieveMetadataAsync()
         {
             _synclock.EnterWriteLock();
             try
             {
-                OpenIdConnectConfiguration config = Task.Run(_configManager.GetConfigurationAsync).Result;
+                OpenIdConnectConfiguration config = await Task.Run(_configManager.GetConfigurationAsync);
```
msbw2, line 84:
```suggestion
                OpenIdConnectConfiguration config = await _configManager.GetConfigurationAsync();
```

src/Microsoft.Identity.Web/AzureADB2COpenIDConnectEventHandlers.cs
```diff
@@ -31,7 +31,7 @@ public AzureADB2COpenIDConnectEventHandlers(
...
+        public Task OnRedirectToIdentityProviderAsync(RedirectContext context)
```
kellyyangsong, line 34: Is this a breaking change?

src/Microsoft.Identity.Web/AzureADB2COpenIDConnectEventHandlers.cs
```diff
@@ -58,7 +58,7 @@ public Task OnRedirectToIdentityProvider(RedirectContext context)
             return Task.CompletedTask;
         }
 
-        public Task OnRemoteFailure(RemoteFailureContext context)
+        public Task OnRemoteFailureAsync(RemoteFailureContext context)
```
kellyyangsong, line 61: Is this also a breaking change?

src/Microsoft.Identity.Web/AzureADB2COpenIDConnectEventHandlers.cs
jennyf19: Yes, it is. Let's not change any public API and open an issue to track changes later.

jennyf19 (changes requested):
would like to have clarity on which E2E tests were run to verify things are working as expected, and which scenarios were covered.
would like to have no public API changes now.

src/Microsoft.Identity.Web/AzureADB2COpenIDConnectEventHandlers.cs
westin-m: I did pack it and there were no complaints. I must have missed this because it's an internal class. I'll keep the names and add a suppression.

src/Microsoft.Identity.Web/AzureADB2COpenIDConnectEventHandlers.cs
msbw2: Would the same apply to `internal` methods due to how they're shared between libraries?

westin-m: This was unintentional. The analyzers didn't catch it, and I think it got past me because the class is internal. I've added a suppression and kept the original name.

src/Microsoft.Identity.Web.Certificate/DefaultCertificateLoader.cs
westin-m: good call

westin-m:
> would like to have clarity on which E2E tests were run to verify things are working as expected, and which scenarios were covered. would like to have no public API changes now.

I've added this to the description

src/Microsoft.Identity.Web.Certificate/DefaultCertificateLoader.cs
```diff
@@ -74,6 +75,29 @@ public static string? UserAssignedManagedIdentityClientId
 
             return certDescription?.Certificate;
         }
+        
+        /// <summary>
+        /// Load the first certificate from the certificate description list.
+        /// </summary>
+        /// <param name="certificateDescriptions">Description of the certificates.</param>
+        /// <returns>First certificate in the certificate description list.</returns>
+        public static async Task<X509Certificate2?> LoadFirstCertificateAsync(IEnumerable<CertificateDescription> certificateDescriptions)
```
pmaytak, line 84: Not for this PR, but long-term we should add an overload that accepts a CancellationToken, which should be propagated down - LoadCredentialsIfNeededAsync uses a semaphore which accepts a cancellation token.

tests/E2E Tests/WebAppUiTests/TestingWebAppLocally.cs
```diff
@@ -37,20 +37,20 @@ public TestingWebAppLocally(ITestOutputHelper output)
...
     [SupportedOSPlatform("windows")]
-    public async Task ChallengeUser_MicrosoftIdFlow_LocalApp_ValidEmailPassword()
+    public async Task ChallengeUser_MicrosoftIdFlow_LocalApp_ValidEmailPasswordAsync()
     {
         LabResponse labResponse = await LabUserHelper.GetDefaultUserAsync();
 
         var clientEnvVars = new Dictionary<string, string>();
 
-        await ExecuteWebAppCallsGraphFlow(labResponse.User.Upn, labResponse.User.GetOrFetchPassword(), clientEnvVars, TraceFileClassName);
+        await ExecuteWebAppCallsGraphFlowAsync(labResponse.User.Upn, labResponse.User.GetOrFetchPassword(), clientEnvVars, TraceFileClassName).ConfigureAwait(false);
```
pmaytak, line 46: Remove ConfigureAwait in xUnit tests for better performance.

src/Microsoft.Identity.Web.OWIN/OpenIdConnectCachingSecurityTokenProvider.cs
```diff
@@ -76,12 +76,12 @@ public IEnumerable<SecurityKey>? SecurityKeys
...
         }
 
-        private void RetrieveMetadata()
+        private async Task RetrieveMetadataAsync()
         {
             _synclock.EnterWriteLock();
             try
             {
-                OpenIdConnectConfiguration config = Task.Run(_configManager.GetConfigurationAsync).Result;
+                OpenIdConnectConfiguration config = await _configManager.GetConfigurationAsync();
```
pmaytak, line 84: Need to revert and block it here.

https://stackoverflow.com/questions/59048102/c-sharp-fire-and-forget-task-and-discard

pmaytak (changes requested):

src/Microsoft.Identity.Web.OWIN/OpenIdConnectCachingSecurityTokenProvider.cs
westin-m: I'm a little confused, does `await` not block until the result of the `Task` is needed?

pmaytak: `await` doesn't block ("block" has a specific meaning here, `.Result` blocks), but it suspends the method and returns the control to the calling method, which since it doesn't await the result, just continues executing synchronously. So not awaiting ends up making this a fire-and-forget operation.

See:
- Step 6 in [What happens in an async method](https://learn.microsoft.com/en-us/dotnet/csharp/asynchronous-programming/task-asynchronous-programming-model#BKMK_WhatHappensUnderstandinganAsyncMethod) 
- [async and await](https://learn.microsoft.com/en-us/dotnet/csharp/asynchronous-programming/task-asynchronous-programming-model#BKMK_AsyncandAwait)
- More detail: [How Async/Await Really Works in C#](https://devblogs.microsoft.com/dotnet/how-async-await-really-works/)

westin-m: Understood. I will revert

pmaytak (approved):

tests/Microsoft.Identity.Web.Test.Integration/AcquireTokenForAppIntegrationTests.cs
```diff
@@ -117,17 +117,17 @@ public async Task GetAccessTokenForAppAndAuthResultForApp_WithMetaTenant_ShouldT
             Assert.Equal(0, _msalTestTokenCacheProvider.Count);
 
             // Act & Assert
-            async Task tokenResult() =>
-                await _tokenAcquisition.GetAccessTokenForAppAsync(TestConstants.s_scopeForApp, tenant: metaTenant);
+            async Task tokenResultAsync() =>
+                await _tokenAcquisition.GetAccessTokenForAppAsync(TestConstants.s_scopeForApp, tenant: metaTenant).ConfigureAwait(false);
```
JoshLozensky, line 121: Please remove ConfigureAwait(false) it is not compatible with XUnit

JoshLozensky (changes requested):
ConfigureAwait(false) is [not compatible with XUnit](https://xunit.net/xunit.analyzers/rules/xUnit1030) this is what had been causing our flaky tests previously

tests/Microsoft.Identity.Web.Test.Integration/AcquireTokenForAppIntegrationTests.cs
JoshLozensky: I did remove them all in the main branch fyi

tests/E2E Tests/SimulateOidc/Contollers/MetadataController.cs
```diff
@@ -34,7 +34,7 @@ public IActionResult  OpenIdConnectConfiguration()
         }
 
         [HttpGet("/v2.0/.well-known/keys.json")]
-        public async Task<IActionResult> Keys()
+        public async Task<IActionResult> KeysAsync()
         {
             HttpClient client = new HttpClient();
             string keysDocument = await client.GetStringAsync("https://login.microsoftonline.com/common/discovery/v2.0/keys").ConfigureAwait(false);
```
JoshLozensky, line 40:
```suggestion
            string keysDocument = await client.GetStringAsync("https://login.microsoftonline.com/common/discovery/v2.0/keys");
```

JoshLozensky (approved):

tests/E2E Tests/SimulateOidc/Contollers/MetadataController.cs
westin-m: The rule that you linked says it only affects methods marked with [Test] and [Theory]. This one isn't removed in main, so I'll leave it to avoid unnecessary changes

PR #3045, opened 2024-09-28
Removed ConfigureAwait(false) from flaky tests

JoshLozensky:
According to this guidance from [XUnit](https://xunit.net/xunit.analyzers/rules/xUnit1030), using `ConfigureAwait(false)` in tests can lead to issues with synchronization and parallelization. In local testing removing ConfigureAwait seemed to work as I could repro flakiness locally in about 1-3 attempts but post-fix there was no observed flakiness locally after 14 attempts.

That doesn't guarantee it fixes the build pipeline, but it looks promising so far. There is a risk of recurrence in the build pipeline even if this method is successful as I only edited tests that were observed as flaky in the last two weeks. Therefore, if a test still using ConfigureAwait(false) triggers flakiness but hasn't been failing itself, it would still be capable of breaking the build.

### DIFF
--- a/tests/E2E Tests/WebAppUiTests/B2CWebAppCallsWebApiLocally.cs
+++ b/tests/E2E Tests/WebAppUiTests/B2CWebAppCallsWebApiLocally.cs
@@ -103,7 +103,7 @@ namespace WebAppUiTests
                         if (InitialConnectionRetryCount == 0) { throw ex; }
                     }
                 }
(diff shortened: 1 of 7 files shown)


pmaytak (approved):

jennyf19 (approved):
Awesome. Thanks @JoshLozensky Please open an issue and attach to the PR for ease in tracking. 
:shipit:

PR #3083, opened 2024-10-11
Use nuget config file

SaurabhMSFT:
# Use nuget config file to use IDDP feed

## 
Nuget config file is updated during the build pipeline to use IDDP feed.

### DIFF
--- a/build/template-restore-build-MSIdentityWeb.yaml
+++ b/build/template-restore-build-MSIdentityWeb.yaml
@@ -17,6 +17,8 @@ steps:
     command: 'custom'
     custom: 'build'
     projects: '$(IdWebSourceDir)Microsoft.Identity.Web.sln'
+    feedsToUse: 'config'
+    nugetConfigPath: NuGet.config
     arguments: '-p:configuration=${{ parameters.BuildConfiguration }} -p:RunCodeAnalysis=true -p:MicrosoftIdentityWebVersion=${{ parameters.MicrosoftIdentityWebVersion }} -p:SourceLinkCreate=true'
 
 # This task is needed so that the 1CS Rolsyn analyzers task works.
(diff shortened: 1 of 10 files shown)


jmprieur (changes requested):
We need to do a slightly different fix. With a different nuget.config file for ado builds as we still want customers to be able to build the open source libraries

build/template-restore-build-MSIdentityWeb.yaml
```diff
@@ -11,6 +11,11 @@ steps:
     dotnet workload restore $(IdWebSourceDir)tests\DevApps\blazorwasm-b2c\blazorwasm2-b2c.csproj
   displayName: 'Install wasm-tools'
 
+- task: NuGetAuthenticate@1
+  displayName: NuGet Authenticate
+  inputs:
+      nuGetServiceConnections: 'Internal Analyzers'
```
jmprieur, line 17: Let's use 'IDDP Feed'

jmprieur: @SaurabhMSFT: what we want is to have another NugetConfig.
Or otherwise have a script task that removes a nuget source and adds another one (this is even better):

See for instance: [here](https://identitydivision.visualstudio.com/DevEx/_git/Id4sBuilds?path=/.pipelines/ID4S-Libaries-OneBranch-Official.yml&version=GBmain&line=196&lineEnd=199&lineStartColumn=1&lineEndColumn=1&lineStyle=plain&_a=contents)

SaurabhMSFT: I have done these changes.

jennyf19 (approved):

jmprieur (approved):
LGTM. Thanks @SaurabhMSFT 
Other builds will be affected, but we'll fix them too.

PR #3085, opened 2024-10-14
Fix reflection in MemoryCache tests to support both versions seen in .NET 9 RCs

msbw2:
Fix reflection in `MemoryCache` tests to support both versions seen in .NET 9 RCs. In RC1 there was one entry collection while in RC2 it was split back out as seen in previous versions of .NET: https://github.com/dotnet/runtime/commit/7ca789e12cbd9c1da8a9de68448d9835e5aa9784


jmprieur (approved):
Thanks @msbw2
There is a lot of chrun in the in memory cache

jmprieur (approved):
Thanks @msbw2
There is a lot of chrun in the. Net in memory cache!

jennyf19 (approved):
:shipit:

PR #3090, opened 2024-10-16
Split aot compatible and incompatible methods

SaurabhMSFT:
## Description

The Deserialize method would call an aot incompatible if the output info is null. We split the method into two such that the aot compatible code calls overload with JsonTypeInfo


jmprieur:
Thanks Saurabh. Is there duplication of code? I couldn't figure out without pulling the branch (which I have not done yet)

SaurabhMSFT:
I had to copy the method DeserializeOutputImplAsync<TOutput> and split the if block on left side(line 398-405) into two methods. Due to that if clause DeserializeOutputImplAsync method wasn't AOT compatible.

jennyf19 (approved):

jmprieur (approved):
LGTM
Thanks @SaurabhMSFT

PR #3093, opened 2024-10-18
Fix DevEx and IDDP builds

jmprieur:
# Fix DevEx and IDDP builds

**Goal:**
When building internally, use an internal Nuget feed, instead of nuget.org

**Did you know?** 
When you are accessing an artifact's feed in the same project, you don't need to specify a service connection in Azure DevOps, NuGetAuthenticate@1 without it a service connection works. If you specify a service connection, you'll need some credentials, like a PAT that will need to be rotated ...

**What** this PR does:
This PR modifies the build scripts to just run nuget authenticate (no service connection), and removes the Nuget.org source, and adds the IDDP source

### DIFF
--- a/build/template-restore-build-MSIdentityWeb.yaml
+++ b/build/template-restore-build-MSIdentityWeb.yaml
@@ -18,9 +18,11 @@ steps:
 
 - script: dotnet nuget remove source NuGet
   displayName: 'Remove external "NuGet" Source'
+  continueOnError: true # could have been run already in a different job on the same machine
 
 - script: dotnet nuget add source https://identitydivision.pkgs.visualstudio.com/_packaging/IDDP/nuget/v3/index.json -n IDDP
   displayName: 'Add IDDP artifacts as NuGet Source'
+  continueOnError: true # could have been run already in a different job on the same machine
 
 - task: DotNetCoreCLI@2
   displayName: 'Build solution Microsoft.Identity.Web.sln'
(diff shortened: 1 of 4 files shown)


build/template-restore-build-MSIdentityWeb.yaml
```diff
@@ -11,22 +11,25 @@ steps:
...
-- task: NuGetAuthenticate@1
-  displayName: NuGet Authenticate
-  inputs:
-      nuGetServiceConnections: 'IDDP Feed'
+# - task: NuGetAuthenticate@1
+#   displayName: NuGet Authenticate
+#   inputs:
+#       nuGetServiceConnections: 'IDDP Feed'
+
+- script: dotnet nuget remove source NuGet
```
kellyyangsong, line 19: For this task, ⁠I'm noticing this: 
image
Which is okay, since this task is configured to continue on error, right? Just double checking since I'm used to seeing a green check and not an orange icon:
image

jennyf19 (approved):
:shipit:

build/template-restore-build-MSIdentityWeb.yaml
jmprieur: thanks @kellyyangsong 
yes, I've noticed as well, and have changed this for a clear way (a small powershell script)

```Powershell
   $nugetSourceIsExternal = (dotnet nuget list source --format Short).Contains("https://api.nuget.org/v3/index.json")
    if ($nugetSourceIsExternal) {
        dotnet nuget remove source NuGet
        dotnet nuget add source https://identitydivision.pkgs.visualstudio.com/_packaging/IDDP/nuget/v3/index.json -n IDDP
    }
```

kellyyangsong (approved):

build/template-restore-build-MSIdentityWeb.yaml
kellyyangsong: Amazing.. and you came up with this script so quickly. Resolving this thread!

PR #3103, opened 2024-10-23
Add API and make ROPC call

### LINKED ISSUE #3091: [Feature Request] Add ROPC support for CCA
Add ROPC flow to Id web for CCA and potentially extend it to use user fic in MISE

```csharp
ClaimsPrincipal user= ClaimsPrincipalFactory.FromUsernamePassword("user@domain.com", "pwd");
downstreamApi.CallApiForUser("api", user: user);
```
 **Possible design**
- In `ClaimsPrincipalFactory` add a method to return a ClaimsPrincipal with 2 claims (xms-username and xms-password) based on the values provided.

- In [TokenAcquisition.GetAuthenticationResultForUserAsync](https://github.com/AzureAD/microsoft-identity-web/blob/e01306a40d3c8d6b7fcba5b44ce1171fdff7ceb8/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs#L255), if the user parameter is not null, and has the 2 claims xms-username and xms-password claims, then perform ROPC for CCA as a first step (in the try to still benefit from the client credentials rotation)

**Additional context**

neha-bhargava:
# Implement ROPC flow

Add ROPC flow to use username and password and call MSAL using the credentials

## Description

{Detail}

Fixes #3091


src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -251,6 +251,21 @@ public async Task<AuthenticationResult> GetAuthenticationResultForUserAsync(
             try
             {
                 AuthenticationResult? authenticationResult;
+
+                // If the user is not null and has claims xms-username and xms-password, perform ROPC for CCA
+                authenticationResult = await GetAuthenticationResultForConfidentialClientUsingRopcAsync(
```
jmprieur, line 256: It's more a TryGet?

src/Microsoft.Identity.Web.TokenAcquisition/ClaimConstants.cs
```diff
@@ -99,5 +99,15 @@ public static class ClaimConstants
...
+
+        /// <summary>
+        /// Username claims for ROPC flow.
+        /// </summary>
+        public const string Username = "xms-username";
+
+        /// <summary>
+        /// Password claims for ROPC flow.
+        /// </summary>
+        public const string Password = "xms-password";
```
jmprieur, line 111: You'll need to add a suppression line in credscan files in the builds

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -310,6 +325,47 @@ public async Task<AuthenticationResult> GetAuthenticationResultForUserAsync(
...
         }
 
+        private async Task<AuthenticationResult?> GetAuthenticationResultForConfidentialClientUsingRopcAsync(IConfidentialClientApplication application, IEnumerable<string> scopes, ClaimsPrincipal? user, MergedOptions mergedOptions, TokenAcquisitionOptions? tokenAcquisitionOptions)
+        {
+            if (user != null && user.HasClaim(c => c.Type == ClaimConstants.Username) && user.HasClaim(c => c.Type == ClaimConstants.Password))
+            {
+                string username = user.FindFirst(ClaimConstants.Username)?.Value ?? string.Empty;
+                string password = user.FindFirst(ClaimConstants.Password)?.Value ?? string.Empty;
+
+                var accounts = await application.GetAccountsAsync().ConfigureAwait(false);
```
jmprieur, line 335: I don't think that this exists in confidential client applications? does it?
or does it only contain the ROPC accounts?

We need to be very careful here.

src/Microsoft.Identity.Web/ClaimsPrincipalFactory.cs
```diff
@@ -107,5 +107,36 @@ public static ClaimsPrincipal FromTenantIdAndObjectId(string tenantId, string ob
...
+        /// <param name="username">UPN of the user for example username@domain.</param>
+        /// <param name="password">Password for the user.</param>
+        /// <returns>A <see cref="ClaimsPrincipal"/> containing these two claims.</returns>
+        public static ClaimsPrincipal FromUsernamePassword(string username, string password)
+        {
+            if (AppContextSwitches.UseClaimsIdentityType)
+            {
+#pragma warning disable RS0030 // Do not use banned APIs
+                return new ClaimsPrincipal(
+                    new ClaimsIdentity(new[]
```
jmprieur, line 124: You should instantiate a `CaseSensitiveClaimsIdentity`, not a `ClaimsIdentity`.
Please don't disable the warning. just do what it tells you to do.

src/Microsoft.Identity.Web/ClaimsPrincipalFactory.cs
```diff
@@ -107,5 +107,36 @@ public static ClaimsPrincipal FromTenantIdAndObjectId(string tenantId, string ob
...
+        /// <summary>
+        /// Instantiate a <see cref="ClaimsPrincipal"/> from a username and password.
+        /// This can be used for ROPC flow for testing purposes.
+        /// </summary>
+        /// <param name="username">UPN of the user for example username@domain.</param>
+        /// <param name="password">Password for the user.</param>
+        /// <returns>A <see cref="ClaimsPrincipal"/> containing these two claims.</returns>
+        public static ClaimsPrincipal FromUsernamePassword(string username, string password)
+        {
+            if (AppContextSwitches.UseClaimsIdentityType)
```
jmprieur, line 120: This if () is not needed.
this is a new API. Just use `CaseSensitiveClaimsIdentity`

jmprieur (commented):
I left a few suggestions

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
neha-bhargava: It exists in confidential client applications. It is currently obsolete but we still use in MSAL tests. I have created this [PR](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/4976) in MSAL.Net to remove the obsolete attribute from this method as suggested by @bgavrilMS

src/Microsoft.Identity.Web.TokenAcquisition/ClaimConstants.cs
neha-bhargava: Can you point me to the file?

neha-bhargava: Do you mean the PublicApi.Unshipped.txt files?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
jmprieur: Wow. I need explanations here.
- which account will they return be in a web app that signs-in users?

src/Microsoft.Identity.Web.TokenAcquisition/ClaimConstants.cs
neha-bhargava: Added

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
jennyf19: Agree that this seems to be going backwards.

src/Microsoft.Identity.Web.TokenAcquisition/ClaimConstants.cs
```diff
@@ -99,5 +99,15 @@ public static class ClaimConstants
         /// Name Identifier ID claim: "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/nameidentifier".
         /// </summary>
         public const string NameIdentifierId = "http://schemas.xmlsoap.org/ws/2005/05/identity/claims/nameidentifier";
+
+        /// <summary>
+        /// Username claims for ROPC flow.
+        /// </summary>
+        public const string Username = "xms-username";
```
jennyf19, line 106: Please resolve the warnings in the GitHub Actions. You can see them in the UI via GitHub, or in the IDE In VS.

bgavrilMS, line 106: Let's use `_` instead of `-`. This is how ESTS does it.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -313,6 +328,48 @@ public async Task<AuthenticationResult> GetAuthenticationResultForUserAsync(
...
 
+        private async Task<AuthenticationResult?> TryGetAuthenticationResultForConfidentialClientUsingRopcAsync(IConfidentialClientApplication application, IEnumerable<string> scopes, ClaimsPrincipal? user, MergedOptions mergedOptions, TokenAcquisitionOptions? tokenAcquisitionOptions)
+        {
+            if (user != null && user.HasClaim(c => c.Type == ClaimConstants.Username) && user.HasClaim(c => c.Type == ClaimConstants.Password))
+            {
+                string username = user.FindFirst(ClaimConstants.Username)?.Value ?? string.Empty;
+                string password = user.FindFirst(ClaimConstants.Password)?.Value ?? string.Empty;
+
+                var accounts = await application.GetAccountsAsync().ConfigureAwait(false);
+                var account = accounts.FirstOrDefault(account => account.Username == username);
```
bgavrilMS, line 339: YOu can't use `FirstOrDefault` because it will throw in case there are 2 accounts with the same username. Instead, yuou should throw a meaningful exception explaining that the scenario is not supported.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -313,6 +328,48 @@ public async Task<AuthenticationResult> GetAuthenticationResultForUserAsync(
...
+                        // Silent flow
+                        return await application.AcquireTokenSilent(
+                            scopes.Except(_scopesRequestedByMsal),
+                            account)
+                            .ExecuteAsync()
+                            .ConfigureAwait(false);
+                    }
+                    catch
+                    {
+                        // Ignore this exception as we will retry with ROPC
```
bgavrilMS, line 354: Can you at least log the exception?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
bgavrilMS: It exists and it is used @jmprieur, despite the API being deprecated. Some apps who choose to integrate directly with MSAL.NET for web app sceanrio will serialize their cache (for a single user) to the session. Then they apply the `GetAccounts` pattern. There is no need to remember the account_id in the session, it just extra info.

So in hindsight, the api makes sense if you know what you are doing. We kept it obsolete because we want ppl to use Id.Web and not MSAL directly and Id.Web holds on to the extra information in ClaimsPrincipal. 

`which account will they return be in a web app that signs-in users` - whatever account(s) you serialized. 

We can add API comments that it should not be used in web api scenario, but only with ROPC.

bgavrilMS: Alternative: Add support for `GetAccountById(null)` which will return the single cached account. And throw an exception if there are 2+ cached accounts.

Thoughts?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
neha-bhargava: Updated to accounts.Where

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
neha-bhargava: @jmprieur , @bgavrilMS Added an alternative to GetAccountsAsync by adding a claim account id to the user after first token acquisition. Please review.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -313,6 +328,58 @@ public async Task<AuthenticationResult> GetAuthenticationResultForUserAsync(
...
         }
 
+        private async Task<AuthenticationResult?> TryGetAuthenticationResultForConfidentialClientUsingRopcAsync(IConfidentialClientApplication application, IEnumerable<string> scopes, ClaimsPrincipal? user, MergedOptions mergedOptions, TokenAcquisitionOptions? tokenAcquisitionOptions)
+        {
+            if (user != null && user.HasClaim(c => c.Type == ClaimConstants.Username) && user.HasClaim(c => c.Type == ClaimConstants.Password))
+            {
+                string username = user.FindFirst(ClaimConstants.Username)?.Value ?? string.Empty;
+                string password = user.FindFirst(ClaimConstants.Password)?.Value ?? string.Empty;
+
+                if (user.GetMsalAccountId() != null)
```
jmprieur, line 338: It's never going to work as the user is created from the ClaimsPrincipalFactory.FromUsernamePassword(), and it does not have the home tenant id / home user id which are the components of GetMsalAccountId()

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -313,6 +328,58 @@ public async Task<AuthenticationResult> GetAuthenticationResultForUserAsync(
...
+                    catch (Exception ex)
+                    {
+                        // Log a message when the silent flow fails and try acquisition through ROPC.
+                        Logger.TokenAcquisitionError(_logger, ex.Message, ex);
+                    }
+
+                }
+
+                // ROPC flow
+                var authenticationResult = await ((IByUsernameAndPassword)application).AcquireTokenByUsernamePassword(
```
jmprieur, line 360: I think that this method should handle the cache itself. I see no other ways

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
neha-bhargava: I am adding these claims below [here](https://github.com/AzureAD/microsoft-identity-web/pull/3103/files#diff-3600735102cf2582ea4dc8277f799690b8a8135f4971e0be919fbecd053ae2caR370) after the first call executes

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -313,6 +328,59 @@ public async Task<AuthenticationResult> GetAuthenticationResultForUserAsync(
...
+            if (user != null && user.HasClaim(c => c.Type == ClaimConstants.Username) && user.HasClaim(c => c.Type == ClaimConstants.Password))
+            {
+                string username = user.FindFirst(ClaimConstants.Username)?.Value ?? string.Empty;
+                string password = user.FindFirst(ClaimConstants.Password)?.Value ?? string.Empty;
+
+                if (user.GetMsalAccountId() != null)
+                {
+                    try
+                    {
+                        var account = await application.GetAccountAsync(user.GetMsalAccountId()).ConfigureAwait(false);
```
jmprieur, line 343: what happens if the user does not contain hometid/homeoid?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
neha-bhargava: Discussed this, cannot handle the cache in this method.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
neha-bhargava: Then the call will be directly made to ROPC and hometid/homeoid will be added to the user. See [here](https://github.com/AzureAD/microsoft-identity-web/pull/3103/files/913fdf9051bd219aca402367958bd3f41d5601d8#diff-3600735102cf2582ea4dc8277f799690b8a8135f4971e0be919fbecd053ae2caR339) I am checking if we are able to get the msalAccountId, if not call ropc.

bgavrilMS (approved):

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -313,6 +328,59 @@ public async Task<AuthenticationResult> GetAuthenticationResultForUserAsync(
...
+
+                        // Silent flow
+                        return await application.AcquireTokenSilent(
+                            scopes.Except(_scopesRequestedByMsal),
+                            account)
+                            .ExecuteAsync()
+                            .ConfigureAwait(false);
+                    }
+                    catch (Exception ex)
```
bgavrilMS, line 352: Hmm, is this too broad? Would you want to catch only MsalException or even MsalServiceException and let the rest bubble up?

westin-m (approved):

jmprieur (approved):

PR #3105, opened 2024-10-23
Grouping Dependabot Updates

JoshLozensky:
Added grouping to dependabot config so PRs can contain multiple suggestions. The groupings are minor and patch updates with a security designation and minor and patch updates without a security designation. Major version updates will all still get their own PRs

Additionally, it will label grouped PRs with whether the updates are security-related or not based on the group names of security and notsecurity

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -17,3 +17,20 @@ updates:
     labels:
       - "dependabot"
       - "dependencies"
+    groups:
+      # Groups all recommended minor and patch updates with a security designation into a single PR
+      security:
+        applies-to: security-updates
+        patterns:
+        - "*"
+        update-types:
+        - "minor"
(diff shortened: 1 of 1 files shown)


westin-m:
**Note**: more info in the docs if anyone is curious like I was: [dependabot groups](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups)

westin-m (approved):

kellyyangsong (approved):

kellyyangsong:
My only concern is that it can get confusing to group updates, whereas separate PRs are very clear. We could see how this goes and always change back though. Thanks, @JoshLozensky!

jennyf19 (approved):
:shipit:

PR #3109, opened 2024-10-24
Remove nuget.config external feed to use

kellyyangsong:


jmprieur:
Sorry @kellyyangsong, this is not the fix I had in mind (assuming we are talking of the same thing)
But rather remove the parameters of the NugetAuthenticate@1 tasks.

kellyyangsong:
> Sorry @kellyyangsong, this is not the fix I had in mind (assuming we are talking of the same thing) But rather remove the parameters of the NugetAuthenticate@1 tasks.

Then we should be good, couldn't find any to remove! Will close this PR, thanks @jmprieur.

PR #3112, opened 2024-10-25
Use Nuget config file

SaurabhMSFT:


jmprieur:
What issue does this solve, @SaurabhMSFT?

SaurabhMSFT:
It solves two issues. The powershell script wasn't working properly earlier as it was not running in the correct folder. The powershell script should run in the folder containing Nuget.config. The second issue was that the Nuget.config file was not used while building the solution. The S360 item for api.nuget.org was showing that the build task was still using public nuget. Once this change is done we can lock our ADO pipeline to use a restrictive network policy to disallow connections to public internet.

SaurabhMSFT:
> /azp run

@jmprieur  is this correct command to start ADO job?

jmprieur (approved):

jennyf19 (approved):

PR #3118, opened 2024-10-31
Update locking and caching for OpenIdConnectCachingSecurityTokenProvider

### LINKED ISSUE #3078: Locking behavior too strict in `OpenIdConnectCachingSecurityTokenProvider`, struggles with enough concurrent requests

Microsoft.Identity.Web.OWIN

3.2.2

### Web app

Not Applicable

### Web API

Not Applicable

### Token cache serialization

Not Applicable

### Description

TL;DR: under enough concurrency, the locking behaviour of `OpenIdConnectCachingSecurityTokenProvider` creates a deadlock-like situation.

> [!NOTE]
> I've seen this class in enough Microsoft repositories to suspect that you don't really own it per-se, but this repo seems to be the go-to when it comes to OIDC auth across all stacks, and also the most recent commit for this file, so you're the "lucky" ones to get this report.

We've experienced cases where our app grinds to a halt when a certain concurrency threshold is reached.
Snapshot analysis pointed to `OpenIdConnectCachingSecurityTokenProvider`, more specifically the `Issuer` and `SecurityKeys` properties.

Looking at the implementation, and reading [the `ReaderWriterLockSlim` documentation](https://learn.microsoft.com/en-us/dotnet/fundamentals/runtime-libraries/system-threading-readerwriterlockslim), I think I can see how this happens.

The a-ha moment for me was when reading the following section from the documentation linked above (I left the upgradeable mode away as it's not using it)

> If a ReaderWriterLockSlim does not allow recursion, a thread that tries to enter the lock can block for several reasons:
>
> - A thread that tries to enter read mode blocks if there are threads waiting to enter write mode or if there is a single thread in write mode.
> - A thread that tries to enter write mode blocks if there is a thread in any of the three modes.

Looking at the implementation, we can see that accessing `Issuer` or `SecurityKeys` will:

1. Call `RetrieveMetadata` which enters "write" mode to get the configuration from the underlying `ConfigurationManager` instances, then
2. Enter "read" mode to return the value of the private field.

---

Here's a mental model of what happens for a single request:

1. The Issuer property is accessed:
   1. This calls `RetrieveMetadata`, which enters "write" mode and fetches the configuration.
   1. It then enters read mode to return the value.
1. The `SecurityKeys` properties is accessed:
   1. This calls `RetrieveMetadata`, which enters "write" mode and fetches the configuration.
   1. It then enters read mode to return the value.

Extrapolating this to many concurrent requests, here's what I think happens:

1. The first request comes in, and its thread _T1_ enters "write" mode.
2. Other requests come in, and their threads block trying to enter "write" mode because of _T1_.
3. _T1_ exits "write" mode, and cannot enter "read" mode because other threads are waiting to enter "write" mode.
4. Another thread _T2_ enters "write" mode.
5. More requests arrive, try to enter "write" mode, but can't, and get added to the "write" mode queue.

This means that concurrent requests are dependent on all other concurrent requests to have entered then exited write mode to be able to enter read mode and continue the processing of the request.
The bigger the "write" mode queue gets, the higher the chance of new requests arriving and making the problem worse.

The only way out of this situation is for no new requests to arrive, so the existing concurrent requests can enter and exit "write" mode one after the other, after which they will all be able to enter "read" mode.

However, the fact that we call `RetrieveMetadata` twice in a single request (`.Issuer` and `.SecurityKeys`, both accessed by `JwtFormat`) doesn't make it easy to get out of this situation.
Even more so when `ConfigurationManager` caches the OIDC configuration for 12 hours by default, so the vast majority of these calls will be no-ops when it comes to the actual config used to validate tokens.

Another aggravating factor is that we use `EnterReadMode` and `EnterWriteMode`, which will wait for as long as needed until they can enter the relevant mode. Note that `TryEnterReadMode` and `TryEnterWriteMode` exist, with timeout parameters, and their return value indicate whether entering the mode was successful.

---

Here are assumptions about Entra ID / B2C:

1. For a given tenant, the issuer never changes (I'm aware this is metadata endpoint-dependent, but the class can only target one endpoint).
2. Signing keys will be advertised in the metadata document long before they're used for signing so all consumers have the time to know about them.

Using these assumptions, I think we can loosen the locking behaviour:

1. Try to get a write lock "synchronously", otherwise use the information already available.
2. Do not use read locks, and use the information available.

Any thoughts?

### Reproduction steps

1. Set up an OWIN project with Microsoft.Identity.Web.
2. Fire enough concurrent requests to trigger the blocking behaviour (200 was the tipping point on my local machine)

### Error message

_No response_

### Id Web logs

_No response_

### Relevant code snippets

```csharp
N/A
```

### Regression

_No response_

### Expected behavior

No locking that brings the app to a halt.

msbw2:
Update locking and caching for `OpenIdConnectCachingSecurityTokenProvider`. Fixes #3078.

### DIFF
--- a/src/Microsoft.Identity.Web.OWIN/OpenIdConnectCachingSecurityTokenProvider.cs
+++ b/src/Microsoft.Identity.Web.OWIN/OpenIdConnectCachingSecurityTokenProvider.cs
@@ -1,13 +1,14 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
 using Microsoft.IdentityModel.Protocols;
 using Microsoft.IdentityModel.Protocols.OpenIdConnect;
 using Microsoft.IdentityModel.Tokens;
 using Microsoft.Owin.Security.Jwt;
-using System.Collections.Generic;
-using System.Threading;
-using System.Threading.Tasks;
 
 namespace Microsoft.Identity.Web
 {
@@ -15,19 +16,27 @@ namespace Microsoft.Identity.Web
     // the OpenID Connect metadata endpoint exposed by the STS by default.
     internal class OpenIdConnectCachingSecurityTokenProvider : IIssuerSecurityKeyProvider
     {
+        private readonly TimeProvider _timeProvider;
         public ConfigurationManager<OpenIdConnectConfiguration> _configManager;
         private string? _issuer;
-        private IEnumerable<SecurityKey>? _keys;
-        private readonly string _metadataEndpoint;
+        private ICollection<SecurityKey>? _keys;
+        private DateTimeOffset _syncAfter;
 
-        private readonly ReaderWriterLockSlim _synclock = new ReaderWriterLockSlim();
+        private readonly ReaderWriterLockSlim _syncLock = new();
+
+        private const int IdleState = 0;
+        private const int RunningState = 1;
+        private int _state = IdleState;
 
         public OpenIdConnectCachingSecurityTokenProvider(string metadataEndpoint)
-        {
-            _metadataEndpoint = metadataEndpoint;
-            _configManager = new ConfigurationManager<OpenIdConnectConfiguration>(metadataEndpoint, new OpenIdConnectConfigurationRetriever());
+            : this(new ConfigurationManager<OpenIdConnectConfiguration>(metadataEndpoint, new OpenIdConnectConfigurationRetriever()), TimeProvider.System) { }
 
-            RetrieveMetadata();
+        public OpenIdConnectCachingSecurityTokenProvider(ConfigurationManager<OpenIdConnectConfiguration> configManager, TimeProvider timeProvider)
+        {
+            _timeProvider = timeProvider;
+            _configManager = configManager;
+
+            RetrieveValues();
         }
 
         /// <summary>
@@ -36,22 +45,7 @@ namespace Microsoft.Identity.Web
         /// <value>
         /// The issuer the credentials are for.
         /// </value>
-        public string? Issuer
-        {
-            get
-            {
-                RetrieveMetadata();
-                _synclock.EnterReadLock();
-                try
-                {
-                    return _issuer;
-                }
-                finally
-                {
-                    _synclock.ExitReadLock();
-                }
-            }
-        }
+        public string? Issuer => RetrieveValues().Issuer;
 
         /// <summary>
         /// Gets all known security keys.
@@ -59,38 +53,43 @@ namespace Microsoft.Identity.Web
         /// <value>
         /// All known security keys.
         /// </value>
-        public IEnumerable<SecurityKey>? SecurityKeys
-        {
-            get
-            {
-                RetrieveMetadata();
-                _synclock.EnterReadLock();
-                try
-                {
-                    return _keys;
-                }
-                finally
-                {
-                    _synclock.ExitReadLock();
-                }
-            }
-        }
+        public IEnumerable<SecurityKey>? SecurityKeys => RetrieveValues().Keys;
 
-        private void RetrieveMetadata()
+        private (string? Issuer, ICollection<SecurityKey>? Keys) RetrieveValues()
         {
-            _synclock.EnterWriteLock();
-            try
+            _syncLock.EnterReadLock();
+            string? issuer = _issuer;
+            ICollection<SecurityKey>? keys = _keys;
+            DateTimeOffset syncAfter = _syncAfter;
+            _syncLock.ExitReadLock();
+
+            // Check if it's time to refresh the stored issuer and keys
+            if (syncAfter < _timeProvider.GetUtcNow())
             {
+                // Acquire lock to retrieve new metadata
+                if (Interlocked.CompareExchange(ref _state, RunningState, IdleState) == IdleState)
+                {
+                    try
+                    {
 #pragma warning disable VSTHRD002 // Avoid problematic synchronous waits
-                OpenIdConnectConfiguration config = Task.Run(_configManager.GetConfigurationAsync).Result;
+                        OpenIdConnectConfiguration config = Task.Run(_configManager.GetConfigurationAsync).Result;
 #pragma warning restore VSTHRD002 // Avoid problematic synchronous waits
-                _issuer = config.Issuer;
-                _keys = config.SigningKeys;
+                        // Acquire write lock to update stored values
+                        _syncLock.EnterWriteLock();
+                        issuer = _issuer = config.Issuer;
+                        keys = _keys = config.SigningKeys;
+                        // Metadata will refresh after AutomaticRefreshInterval + jitter, so refresh 1 hour after to account for jitter
+                        _syncAfter = _timeProvider.GetUtcNow() + _configManager.AutomaticRefreshInterval + TimeSpan.FromHours(1);
+                        _syncLock.ExitWriteLock();
+                    }
+                    finally
+                    {
+                        Interlocked.Exchange(ref _state, IdleState);
+                    }
+                }
             }
-            finally
-            {
-                _synclock.ExitWriteLock();
-            }
+
+            return (issuer, keys);
         }
     }
 }
--- a/src/Microsoft.Identity.Web.OWIN/InternalAPI.Unshipped.txt
+++ b/src/Microsoft.Identity.Web.OWIN/InternalAPI.Unshipped.txt
@@ -1,0 +1,1 @@
+Microsoft.Identity.Web.OpenIdConnectCachingSecurityTokenProvider.OpenIdConnectCachingSecurityTokenProvider(Microsoft.IdentityModel.Protocols.ConfigurationManager<Microsoft.IdentityModel.Protocols.OpenIdConnect.OpenIdConnectConfiguration!>! configManager, System.TimeProvider! timeProvider) -> void


jmprieur:
@alex-set and @mderriey.
Do you want to try this branch and check if your issues are fixed?

src/Microsoft.Identity.Web.OWIN/OpenIdConnectCachingSecurityTokenProvider.cs
```diff
@@ -36,61 +45,51 @@ public OpenIdConnectCachingSecurityTokenProvider(string metadataEndpoint)
...
-            }
-        }
+        public IEnumerable<SecurityKey>? SecurityKeys => RetrieveValues().Keys;
 
-        private void RetrieveMetadata()
+        private (string? Issuer, ICollection<SecurityKey>? Keys) RetrieveValues()
         {
-            _synclock.EnterWriteLock();
-            try
+            _syncLock.EnterReadLock();
```
jmprieur, line 60: Doesn't ConfigurationManagerBase already control the threadings?

msbw2: This is for reading/writing the fields in this class. This is to ensure they aren't changed in the middle of reading leading to unexpected behavior. The `ConfigurationManager` handles all the threading for making the actual metadata refresh call as needed internally, but that won't help here unless we change this to not cache the issuer and keys and instead just always call into the config manager (e.g., `public string? Issuer => _configManager.GetConfigurationAsync().Result.Issuer`), but it's my understanding that that is the issue being reported in the first place.

mderriey (commented):
_Disclaimer: I don't have much experience with multi-threading programming, locking, or synchronization._

First, thanks a lot for looking at this quickly after the issue was reported, it's much appreciated.

Overall, I think this version will improve the situation, since it heavily reduces the number of times we'll enter in a write lock.

I still have some concerns / questions.

### Read lock

> This is to ensure they aren't changed in the middle of reading leading to unexpected behavior.

I was thinking about this; is there a possibility that a thread reads the value of a variable while it's being written to by another thread, leading to a corrupt value being assigned?

I was under the impression that both variable assignment and read are atomic operations in .NET.

If they are, I'd question the need for a read lock altogether.

### Caching

I see this is managing a cache of its own through `syncAfter` to minimize the number of times we're calling `_configManager.GetConfigurationAsync`.

It feels to me that we're trying to recreate what `ConfigurationManager` already implements, since most calls to `GetConfigurationAsync` will return cached data.

Kind of related, @msbw2 also mentioned (emphasis mine):

> [...] won't help here unless we change this to not cache the issuer and keys and instead just always call into the config manager (e.g., public string? Issuer => _configManager.GetConfigurationAsync().Result.Issuer), but **it's my understanding that that is the issue being reported in the first place**.

I don't think that's what @alex-set and myself reported. The root issue is that under sufficient sustained concurrency, the original code leads to deadlock-style situations because of excessive locking and unbound synchronization, in the sense that threads will wait forever to acquire a read or write lock, without a timeout.

Calling `GetConfigurationAsync` on a threadpool thread every time is not ideal, I agree. However I'm not entirely sold on building our own cache on top of `ConfigurationManager` to avoid it.

### Proposal / Request for review

We had to push a hand-rolled fix for this, and landed on a solution that has the following characteristics:

1. Lightweight write lock through `Interlocked.Exchange`.
1. If a thread can't acquire a write lock, it will not wait, and will return the current issuer and signing keys.
3. No read lock.

The rationale behind this is that:

1. My understanding is that for a given tenant, the issuer never changes.
2. New signing keys are advertised way before they're being actually used to sign new tokens.

Here's the implementation (let me know if you'd like me to open a PR for a more practical discussion):

<details>
<summary>Click to expand</summary>

```csharp
public class OpenIdConnectCachingSecurityTokenProvider : IIssuerSecurityKeyProvider
{
    private int _lock = 0;

    public ConfigurationManager<OpenIdConnectConfiguration> _configManager;
    private string _issuer;
    private IEnumerable<SecurityKey> _keys;

    public OpenIdConnectCachingSecurityTokenProvider(string metadataEndpoint)
    {
        _configManager = new ConfigurationManager<OpenIdConnectConfiguration>(metadataEndpoint, new OpenIdConnectConfigurationRetriever());

        RetrieveMetadata();
    }

    /// <summary>
    /// Gets the issuer the credentials are for.
    /// </summary>
    /// <value>
    /// The issuer the credentials are for.
    /// </value>
    public string Issuer
    {
        get
        {
            RetrieveMetadata();
            return _issuer;
        }
    }

    /// <summary>
    /// Gets all known security keys.
    /// </summary>
    /// <value>
    /// All known security keys.
    /// </value>
    public IEnumerable<SecurityKey> SecurityKeys
    {
        get
        {
            RetrieveMetadata();
            return _keys;
        }
    }

    private void RetrieveMetadata()
    {
        // Try to acquire the lock
        //
        // Interlocked.Exchange returns the original value of _lock before it was swapped.
        // If it's 0, it means it went from 0 to 1, so we did acquire the lock.
        // If it's 1, then the lock was already acquired by another thread.
        //
        // See the example in the Exchange(Int32, Int32) overload: https://learn.microsoft.com/en-us/dotnet/api/system.threading.interlocked.exchange?view=netframework-4.7.2
        if (Interlocked.Exchange(ref _lock, 1) == 0)
        {
            OpenIdConnectConfiguration config = Task.Run(_configManager.GetConfigurationAsync).Result;
            _issuer = config.Issuer;
            _keys = config.SigningKeys;

            // Release the lock
            Interlocked.Exchange(ref _lock, 0);
        }
    }
}
```

</details>

It's still early days, but we've been running this in production for a few days now, it's been OK under concurrency levels that brought the application down before.

msbw2:
> _Disclaimer: I don't have much experience with multi-threading programming, locking, or synchronization._

Likewise.

> is there a possibility that a thread reads the value of a variable while it's being written to by another thread, leading to a corrupt value being assigned?
> I was under the impression that both variable assignment and read are atomic operations in .NET.
> If they are, I'd question the need for a read lock altogether.

Looking into this it looks like reading and writing for both `_issuer` and `_keys` is *not* atomic based on [this StackOverflow post](https://stackoverflow.com/a/9788). `string?` and `IEnumerable<SecurityKey>`/`ICollection<SecurityKey>` would not fit into the <=32-bit size category that's guaranteed to be atomic.

> It feels to me that we're trying to recreate what `ConfigurationManager` already implements, since most calls to `GetConfigurationAsync` will return cached data.

Yes basically.

> The root issue is that under sufficient sustained concurrency, the original code leads to deadlock-style situations because of excessive locking and unbound synchronization, in the sense that threads will wait forever to acquire a read or write lock, without a timeout.

I see what you mean. Tying this back to the question of atomicity, I think we need some lock.

> Calling `GetConfigurationAsync` on a threadpool thread every time is not ideal, I agree. However I'm not entirely sold on building our own cache on top of `ConfigurationManager` to avoid it.

Sure. My change was mainly in the spirit of the intent behind the original implementation, but I do agree with what you're stating. It makes sense that the `ConfigurationManager` should manage the configuration and not this class.

> ### Proposal / Request for review
> We had to push a hand-rolled fix for this, and landed on a solution that has the following characteristics:
> 
> 1. Lightweight write lock through `Interlocked.Exchange`.
> 2. If a thread can't acquire a write lock, it will not wait, and will return the current issuer and signing keys.
> 3. No read lock.
> 
> The rationale behind this is that:
> 
> 1. My understanding is that for a given tenant, the issuer never changes.
> 2. New signing keys are advertised way before they're being actually used to sign new tokens.

The rationales make sense. I'm just unsure on the atomicity of reads and writes since the StackOverflow post seems to suggest it's not guaranteed in this case hence a read and write lock is necessary.

msbw2:
> > is there a possibility that a thread reads the value of a variable while it's being written to by another thread, leading to a corrupt value being assigned?
> > I was under the impression that both variable assignment and read are atomic operations in .NET.
> > If they are, I'd question the need for a read lock altogether.
> 
> Looking into this it looks like reading and writing for both `_issuer` and `_keys` is _not_ atomic based on [this StackOverflow post](https://stackoverflow.com/a/9788). `string?` and `IEnumerable<SecurityKey>`/`ICollection<SecurityKey>` would not fit into the <=32-bit size category that's guaranteed to be atomic.

Never mind. All of these are reference types, and the references should fit into a native word, so the locks aren't necessary. Your implementation is what I would do to remove the locks, so you can create a PR with your changes. Thanks @keegan-caruso for pointing out the reference types and sizes.

alex-set:
To be clear, I think the biggest problem here is that ConfigurationManager is an async class and the IIssuerSecurityKeyProvider interface to which the OpenIdConnectCachingSecurityTokenProvider has to adhere is not async. If the interface had async get methods, this would be a non-issue. Under the current, larger design, no matter how you get this class to work, it will violate someone's rules for task concurrency. It's probably not realistic to address those issues in the short term, so I'll go back to talking about this specific class.

I agree with everything @mderriey stated. Though one thing that gives me hesitation about the implementation he suggests is the use of Task.Run in RetrieveMetadata. It would mean that every call to either of the properties will put work on a separate thread even though the called async method will very rarely ever do any async work since it usually just returns the cached config. However, the passive locking on calling the task ensures that it will never enter the thread starvation issue from which I'm still traumatized.

Here's what I used and it's been working fine for us:

```
    internal class OpenIdConnectCachingSecurityTokenProvider : IIssuerSecurityKeyProvider
    {
        private readonly ConfigurationManager<OpenIdConnectConfiguration> _configManager;
        private readonly Task<OpenIdConnectConfiguration> _initialRetrieval;

        public OpenIdConnectCachingSecurityTokenProvider(string metadataEndpoint)
        {
            _configManager = new ConfigurationManager<OpenIdConnectConfiguration>(
                metadataEndpoint, new OpenIdConnectConfigurationRetriever());

            _initialRetrieval = Task.Run(_configManager.GetConfigurationAsync);
        }

        public string? Issuer => GetConfig().Issuer;

        public IEnumerable<SecurityKey>? SecurityKeys => GetConfig().SigningKeys;

        private OpenIdConnectConfiguration GetConfig()
        {
            if (!_initialRetrieval.IsCompleted)
            {
#pragma warning disable VSTHRD002
                return _initialRetrieval.Result;
            }

            return _configManager.GetConfigurationAsync().Result;
#pragma warning restore VSTHRD002
        }
    }
```

One thing that's important to note is that the ConfigManager.GetConfigurationAsync method will only ever do async task work during the initial config retrieval. All other calls to the method will return the current config, though if it's time to refresh the config, it will start that on a separate thread, but it will still return the existing config rather than waiting for the new one. This means that GetConfigurationAsync will always returns a completed task after the first call to it completes. So when the class is constructed, we can just preserve the initial retrieval task. When the config is needed later, if the initial task is still running, we know that we have to wait for that to complete. If that task has already completed, we can just call GetConfigurationAsync again and know that it will return a completed task from which can just return the result.

The risk here is that it depends on the internal functionality of the ConfigurationManager and I don't think there's a guarantee that it wouldn't change in future versions. If it did change and the call to GetConfigurationAsync returned a task that is still running when we expect it to return a completed task, calling Result on it could theoretically cause a deadlock. However, I would argue that this class is likely only accessed from deep on async-heavy library call stacks, and it's highly unlikely that it could be called while attached to a synchronization context that could cause a deadlock.

If you want to stick with the Task.Run-always version instead of something like what I'm doing here, I have no problem with that. Like I said, it fixes the major bug that started the whole conversation. My remaining concerns are trivial by comparison.

msbw2:
> Though one thing that gives me hesitation about the implementation he suggests is the use of Task.Run in RetrieveMetadata. It would mean that every call to either of the properties will put work on a separate thread even though the called async method will very rarely ever do any async work since it usually just returns the cached config. However, the passive locking on calling the task ensures that it will never enter the thread starvation issue from which I'm still traumatized.

This makes sense to me.

> Here's what I used and it's been working fine for us:
> 
> ```
>     internal class OpenIdConnectCachingSecurityTokenProvider : IIssuerSecurityKeyProvider
>     {
>         private readonly ConfigurationManager<OpenIdConnectConfiguration> _configManager;
>         private readonly Task<OpenIdConnectConfiguration> _initialRetrieval;
> 
>         public OpenIdConnectCachingSecurityTokenProvider(string metadataEndpoint)
>         {
>             _configManager = new ConfigurationManager<OpenIdConnectConfiguration>(
>                 metadataEndpoint, new OpenIdConnectConfigurationRetriever());
> 
>             _initialRetrieval = Task.Run(_configManager.GetConfigurationAsync);
>         }
> 
>         public string? Issuer => GetConfig().Issuer;
> 
>         public IEnumerable<SecurityKey>? SecurityKeys => GetConfig().SigningKeys;
> 
>         private OpenIdConnectConfiguration GetConfig()
>         {
>             if (!_initialRetrieval.IsCompleted)
>             {
> #pragma warning disable VSTHRD002
>                 return _initialRetrieval.Result;
>             }
> 
>             return _configManager.GetConfigurationAsync().Result;
> #pragma warning restore VSTHRD002
>         }
>     }
> ```

The only thing I don't like here is the reference to the initial task that is kept for the entire lifetime of the class. Once `_initialRetrieval.IsCompleted` returns `true`, it can never change back to `false`, so we could theoretically just store that to avoid accessing the original task in the future, and, at some point, remove the reference. I don't know if it's worth the hassle to make this change, though.

alex-set:
One thing that finally occurred to me about the proposed implementation is that it calls RetrieveMetadata from the constructor which will block until the initial configuration loads. I think that's probably fine, and as a helpful side effect, it also makes the ConfigManager much easier to work with because the call to GetConfigurationAsync will always return a completed task without any async blocking. It handles keeping the configuration up to date separately, though it always returns the most recently cached version.

I think this could be an incredibly simple class. It shouldn't ever need to cache config requests, because that's exactly what the ConfigurationManager is already doing. It doesn't need to be concerned with concurrency for the reasons I already mentioned. With caching and concurrency no longer factors, there's no reason to worry about locking either. Obviously, it doesn’t need to maintain a reference to the initial retrieval task either, because that completes in the class constructor.

Here's what I have running in our moderately high-volume production environment:

```
internal class OpenIdConnectCachingSecurityTokenProvider : IIssuerSecurityKeyProvider
{
	private readonly ConfigurationManager<OpenIdConnectConfiguration> _configManager;

	public OpenIdConnectCachingSecurityTokenProvider(string metadataEndpoint)
	{
		_configManager = new ConfigurationManager<OpenIdConnectConfiguration>(
			metadataEndpoint, new OpenIdConnectConfigurationRetriever());

#pragma warning disable VSTHRD002
		Task.Run(_configManager.GetConfigurationAsync).Wait();
#pragma warning restore VSTHRD002
	}

	public string? Issuer => RetrieveMetadata().Issuer;

	public IEnumerable<SecurityKey>? SecurityKeys => RetrieveMetadata().SigningKeys;

#pragma warning disable VSTHRD002
	private OpenIdConnectConfiguration RetrieveMetadata() =>
		_configManager.GetConfigurationAsync().Result;
#pragma warning restore VSTHRD002
}
```

There still may be a valid question about whether this relies too much on the internal functionality of a separate library. Specifically, what if a future version of the ConfigurationManager’s GetConfigurationAsync method could occasionally block to do async work after the first retrieval? In that instance, calling Result on the task could potentially deadlock if a certain type of synchronization context were in effect. Here’s a potential way to safeguard against that without adding much overhead or complexity.

```
private OpenIdConnectConfiguration RetrieveMetadata()
{
    Task<OpenIdConnectConfiguration> task = _configManager.GetConfigurationAsync();
    if (task.IsCompleted)
    {
        return task.Result;
    }

    var sideTask = Task.Run(() => _configManager.GetConfigurationAsync());
    return sideTask.Result;
}
```

I know it looks a little hacky. Keep in mind that with the current implementation of ConfigurationManager, the returned task is always completed, so the second half of this method is effectively unreachable. Even if the GetConfigurationAsync implementation changed, I think it would be safe to assume it would still almost always return a completed task with the current configuration. As an added precaution, we can check if the returned task is still running in which case we would just want to restart the call from a separate thread using Task.Run, from which we can safely collect the configuration result. The important part is that we only call Task.Run when it’s truly necessary.

Sorry if I’m still laboring over some unimportant details. Hopefully this is helpful in some way, but I can drop it now that I’ve said my piece.

mderriey:
From my perspective, the more people chiming in the better, and I appreciate you thinking about potential ways to make this simpler.

I like your thinking, especially the perspective of having this class not use any locks or synchronization.

I can't speak about the dependency on the internals of `ConfigurationManager`, that'll be for the maintaining team to assess.

It's probably not important for the team here, but I'm in a position where we don't use Microsoft.Identity.Web.OWIN because reasons, and we vendored in the original implementation in our project. If your proposal gets in, I'm not certain we're in a position to use it as-is because the version of `ConfigurationManager` we use doesn't work this way, as we're still on v6.23.1.
I thought I'd mention it, but again, the responsibility of the team here is to get Microsoft.Identity.Web.OWIN working, not satisfy the expectation of people cherry-picking individual classes and have them work 😇

Thanks again.

alex-set:
I didn't even consider the functionality of an older version of ConfigManager. I just looked at the history and you're totally right; the functionality on which I was relying is a newer development. With that in mind, I think we have to use locking and caching to safely avoid thread pool starvation. In fact, the only thing I would really change about your original proposal is the use of Task.Run in RetrieveMetadata (which also means additional considerations for exception handling).

This is what I have running now:

```
    internal class OpenIdConnectCachingSecurityTokenProvider : IIssuerSecurityKeyProvider
    {
        private readonly ConfigurationManager<OpenIdConnectConfiguration> _configManager;

        private OpenIdConnectConfiguration _lastConfig;
        private int _lock = 0;
        private Exception? _lastException;

        public OpenIdConnectCachingSecurityTokenProvider(string metadataEndpoint)
        {
            _configManager = new ConfigurationManager<OpenIdConnectConfiguration>(
                metadataEndpoint, new OpenIdConnectConfigurationRetriever());

#pragma warning disable VSTHRD002
            _lastConfig = Task.Run(_configManager.GetConfigurationAsync).Result;
#pragma warning restore VSTHRD002
        }

        public string? Issuer => RetrieveMetadata().Issuer;

        public IEnumerable<SecurityKey>? SecurityKeys => RetrieveMetadata().SigningKeys;

        private OpenIdConnectConfiguration RetrieveMetadata()
        {
            if (Interlocked.Exchange(ref _lock, 1) == 0)
            {
                if (_lastException != null)
                {
                    Exception exc = _lastException;
                    _lastException = null;
                    _lock = 0;
                    throw exc;
                }

                Task<OpenIdConnectConfiguration> task = _configManager.GetConfigurationAsync();
                if (task.IsCompleted)
                {
                    UpdateConfig(task);
                }
                else
                {
                    _ = task.ContinueWith(UpdateConfig, TaskScheduler.Default);
                }
            }

            return _lastConfig;
        }

        private void UpdateConfig(Task<OpenIdConnectConfiguration> task)
        {
            if (task.IsFaulted)
            {
                _lastException = task.Exception;
            }
            else
            {
#pragma warning disable VSTHRD002 // This should always be a completed task
                _lastConfig = task.Result;
#pragma warning restore VSTHRD002
            }

            _lock = 0;
        }
    }
```

The call to GetConfigurationAsync still usually returns a completed task even with older versions (I think the default update interval is 12 hours). When it doesn't return a completed task, we can just schedule the continuation on a thread pool task and return the cached config without waiting for the result. That also means that if the async continuation has an exception, we have to preserve it and throw it on the next call to one of the getters.

I admit that this adds some non-trivial logic just to avoid Task.Run which may be overkill when passive locking is used to prevent overlapping calls. My only argument is that Task.Run is still really expensive to use in most calls to a property getter. I won't be offended if I don't persuade anyone.

msbw2:
@alex-set Which of the two PRs do you think is closer to your proposal? You could leave comments to try to move it in your direction. Or, you could create your own PR with your suggestion.

PR #3120, opened 2024-10-31
Remove use of Thread.Sleep

westin-m:
Remove unnecessary usage of `Thread.Sleep` in the `UiTestHelpers` class.

### DIFF
--- a/tests/E2E Tests/WebAppUiTests/UiTestHelpers.cs
+++ b/tests/E2E Tests/WebAppUiTests/UiTestHelpers.cs
@@ -324,7 +324,6 @@ namespace WebAppUiTests
                                                 processDataEntry.EnvironmentVariables);
 
                 processes.Add(processDataEntry.ExecutableName, process);
-                Thread.Sleep(5000);
             }
 
             //Verify that processes are running
@@ -357,7 +356,6 @@ namespace WebAppUiTests
                                                     processDataEntry.AppLocation,
                                                     processDataEntry.ExecutableName,
                                                     processDataEntry.EnvironmentVariables);
-                    Thread.Sleep(5000);
 
                     //Update process in collection
                     processes[processEntry.Key] = process;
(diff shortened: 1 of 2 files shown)


tests/E2E Tests/WebAppUiTests/UiTestHelpers.cs
```diff
@@ -324,7 +324,6 @@ internal static bool StartAndVerifyProcessesAreRunning(List<ProcessStartOptions>
                                                 processDataEntry.EnvironmentVariables);
 
                 processes.Add(processDataEntry.ExecutableName, process);
-                Thread.Sleep(5000);
```
msbw2, line 327: Is there a way to block the thread on waiting for all processes to start or error? I assume that's the intent behind this sleep.

tests/E2E Tests/WebAppUiTests/UiTestHelpers.cs
```diff
@@ -357,7 +356,6 @@ static void RestartProcesses(Dictionary<string, Process> processes, List<Process
                                                     processDataEntry.AppLocation,
                                                     processDataEntry.ExecutableName,
                                                     processDataEntry.EnvironmentVariables);
-                    Thread.Sleep(5000);
```
msbw2, line 360: Same here.

jmprieur (commented):
Thanks @westin-m 
Did you try this? I think we need to let some time for the web APIs to start?

tests/E2E Tests/WebAppUiTests/UiTestHelpers.cs
bgavrilMS: If you can check that all processes have started, you can poll on that condition. Like here https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/tests/Microsoft.Identity.Test.Common/TestCommon.cs#L258

westin-m: Interesting, it looks like something similar is being done with these tests, after the processes are checked to be running https://github.com/AzureAD/microsoft-identity-web/blob/011bd155ce3ddedeb43daa7348e09584cfa20552/tests/E2E%20Tests/WebAppUiTests/WebAppCallsApiCallsGraphLocally.cs#L335

PR #3124, opened 2024-11-02
Update locking and caching for `OpenIdConnectCachingSecurityTokenProvider`

### LINKED ISSUE #3078: Locking behavior too strict in `OpenIdConnectCachingSecurityTokenProvider`, struggles with enough concurrent requests

mderriey:
# Update locking and caching for `OpenIdConnectCachingSecurityTokenProvider`

- [X] You've read the [Contributor Guide](https://github.com/AzureAD/microsoft-identity-web/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://github.com/AzureAD/microsoft-identity-web/blob/master/CODE_OF_CONDUCT.md).
- [X] You've included inline docs for your change, where applicable.
- [X] There's an open issue for the PR that you are making. If you'd like to propose a new feature or change, please open an issue to discuss the change or find an existing issue.

Update locking and caching for `OpenIdConnectCachingSecurityTokenProvider`

## Description

This is a proposed alternative to #3118.

Changes:

1. Keep a lightweight write lock through `Interlocked.Exchange` to ensure a single thread calls `ConfigurationManager.GetConfigurationAsync` at a time.
2. Threads will not wait to acquire a write lock anymore, and will return existing data if they can't.
3. Remove read lock on the assumption that that variable assignments and reads are atomic operations for reference types (as per <https://github.com/AzureAD/microsoft-identity-web/pull/3118#issuecomment-2452708548>).

Thanks for offering me to open a PR for this.

Please get as many eyes as you can on this to validate my assumptions. I don't have much confidence in my multi-threading, locking, and synchronization abilities.

Also, feel free to take over that PR to adjust to your coding preferences; one thing I noticed is the code base seems to prefer using `Interlocked.CompareExchange` over `Interlocked.Exchange`.
Essentially, this is now yours, although I'll definitely keep an eye on it and answer questions where I can.

Fixes #3078

### DIFF
--- a/src/Microsoft.Identity.Web.OWIN/OpenIdConnectCachingSecurityTokenProvider.cs
+++ b/src/Microsoft.Identity.Web.OWIN/OpenIdConnectCachingSecurityTokenProvider.cs
@@ -1,13 +1,13 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
 using Microsoft.IdentityModel.Protocols;
 using Microsoft.IdentityModel.Protocols.OpenIdConnect;
 using Microsoft.IdentityModel.Tokens;
 using Microsoft.Owin.Security.Jwt;
-using System.Collections.Generic;
-using System.Threading;
-using System.Threading.Tasks;
 
 namespace Microsoft.Identity.Web
 {
@@ -15,16 +15,14 @@ namespace Microsoft.Identity.Web
     // the OpenID Connect metadata endpoint exposed by the STS by default.
     internal class OpenIdConnectCachingSecurityTokenProvider : IIssuerSecurityKeyProvider
     {
+        private int _lock = 0;
+
         public ConfigurationManager<OpenIdConnectConfiguration> _configManager;
         private string? _issuer;
         private IEnumerable<SecurityKey>? _keys;
-        private readonly string _metadataEndpoint;
-
-        private readonly ReaderWriterLockSlim _synclock = new ReaderWriterLockSlim();
 
         public OpenIdConnectCachingSecurityTokenProvider(string metadataEndpoint)
         {
-            _metadataEndpoint = metadataEndpoint;
             _configManager = new ConfigurationManager<OpenIdConnectConfiguration>(metadataEndpoint, new OpenIdConnectConfigurationRetriever());
 
             RetrieveMetadata();
@@ -41,15 +39,7 @@ namespace Microsoft.Identity.Web
             get
             {
                 RetrieveMetadata();
-                _synclock.EnterReadLock();
-                try
-                {
-                    return _issuer;
-                }
-                finally
-                {
-                    _synclock.ExitReadLock();
-                }
+                return _issuer;
             }
         }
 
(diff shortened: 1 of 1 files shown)


msbw2:
There's this comment from the other PR: https://github.com/AzureAD/microsoft-identity-web/pull/3118#issuecomment-2452916786

jennyf19:
> There's this comment from the other PR: [#3118 (comment)](https://github.com/AzureAD/microsoft-identity-web/pull/3118#issuecomment-2452916786)

@mderriey not sure if you saw this ^

mderriey:
> > There's this comment from the other PR: [#3118 (comment)](https://github.com/AzureAD/microsoft-identity-web/pull/3118#issuecomment-2452916786)
> 
> @mderriey not sure if you saw this ^

I did, yes, I replied further down in the thread.

Alex's comments make sense to me, but I can't comment on whether they're better than what's in this PR.

I opened this PR with the goal of potentially helping other people as much as getting some validation on the changes we've been running in production.

As mentioned in the PR description, I now consider this PR yours. I'm happy to contribute my thoughts, but I won't push in one direction or the other as I wouldn't be confident in my knowledge or experience when it comes to multithreading, thread synchronisation, or locking.

PR #3130, opened 2024-11-05
Add extensibility to update parameters for ROPC flow

### LINKED ISSUE #3129: [Feature Request] Add extensibility to support parameters change for ROPC
The current extensibility mechanism in Id web was only Enlighted for the client credential flow. 
To support user fic, similar extension in user flows (in particular ROPC) is needed to update parameters OnBeforeTokenRequest.

neha-bhargava:
# Add extension method to update the parameters for ROPC flow

## Description

Add extensibility to the ROPC flow to enable OnbeforeTokenRequest.

Fixes #3129

### DIFF
--- a/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
@@ -369,7 +369,7 @@ namespace Microsoft.Identity.Web
 
                 if (addInOptions != null)
                 {
-                    addInOptions.InvokeOnBeforeTokenAcquisitionForUsernamePassword(builder, tokenAcquisitionOptions);
+                    addInOptions.InvokeOnBeforeTokenAcquisitionForTestUser(builder, tokenAcquisitionOptions);
                 }
 
                 var authenticationResult = await builder.ExecuteAsync()
--- a/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
@@ -254,6 +254,21 @@ namespace Microsoft.Identity.Web
             try
             {
                 AuthenticationResult? authenticationResult;
+
+                // If the user is not null and has claims xms-username and xms-password, perform ROPC for CCA
+                authenticationResult = await TryGetAuthenticationResultForConfidentialClientUsingRopcAsync(
+                    application,
+                    scopes,
+                    user,
+                    mergedOptions,
+                    tokenAcquisitionOptions).ConfigureAwait(false);
+
+                if (authenticationResult != null)
+                {
+                    LogAuthResult(authenticationResult);
+                    return authenticationResult;
+                }
+
                 // Access token will return if call is from a web API
                 authenticationResult = await GetAuthenticationResultForWebApiToCallDownstreamApiAsync(
                     application,
@@ -311,6 +326,59 @@ namespace Microsoft.Identity.Web
             {
                 _retryClientCertificate = false;
             }
+        }
+
+        // This method mutate the user claims to include claims uid and utid to perform the silent flow for subsequent calls.
+        private async Task<AuthenticationResult?> TryGetAuthenticationResultForConfidentialClientUsingRopcAsync(IConfidentialClientApplication application, IEnumerable<string> scopes, ClaimsPrincipal? user, MergedOptions mergedOptions, TokenAcquisitionOptions? tokenAcquisitionOptions)
+        {
+            if (user != null && user.HasClaim(c => c.Type == ClaimConstants.Username) && user.HasClaim(c => c.Type == ClaimConstants.Password))
+            {
+                string username = user.FindFirst(ClaimConstants.Username)?.Value ?? string.Empty;
+                string password = user.FindFirst(ClaimConstants.Password)?.Value ?? string.Empty;
+
+                if (user.GetMsalAccountId() != null)
+                {
+                    try
+                    {
+                        var account = await application.GetAccountAsync(user.GetMsalAccountId()).ConfigureAwait(false);
+
+                        // Silent flow
+                        return await application.AcquireTokenSilent(
+                            scopes.Except(_scopesRequestedByMsal),
+                            account)
+                            .ExecuteAsync()
+                            .ConfigureAwait(false);
+                    }
+                    catch (MsalException ex)
+                    {
+                        // Log a message when the silent flow fails and try acquisition through ROPC.
+                        Logger.TokenAcquisitionError(_logger, ex.Message, ex);
+                    }
+
+                }
+
+                // ROPC flow
+                var authenticationResult = await ((IByUsernameAndPassword)application).AcquireTokenByUsernamePassword(
+                    scopes.Except(_scopesRequestedByMsal),
+                    username,
+                    password)
+                    .ExecuteAsync()
+                    .ConfigureAwait(false);
+
+                if (user.GetMsalAccountId() == null)
+                {
+                    // Add the account id to the user (in case of ROPC flow)
+                    user.AddIdentity(new CaseSensitiveClaimsIdentity(new[]
+                    {
+                        new Claim(ClaimConstants.UniqueObjectIdentifier, authenticationResult.Account.HomeAccountId.ObjectId),
+                        new Claim(ClaimConstants.UniqueTenantIdentifier, authenticationResult.Account.HomeAccountId.TenantId),
+                    }));
+                }
+
+                return authenticationResult;
+            }
+
+            return null;
         }
 
         private void LogAuthResult(AuthenticationResult? authenticationResult)
--- a/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
@@ -357,13 +357,23 @@ namespace Microsoft.Identity.Web
 
                 }
 
+                // Check for extension options for the ROPC flow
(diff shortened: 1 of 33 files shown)


src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisitionExtensionOptions.cs
```diff
@@ -28,5 +28,22 @@ internal void InvokeOnBeforeTokenAcquisitionForApp(AcquireTokenForClientParamete
                 OnBeforeTokenAcquisitionForApp(builder, acquireTokenOptions);
             }
         }
+
+        /// <summary>
+        /// Event fired when a ROPC flow request is being built.
+        /// </summary>        
+        public event BeforeTokenAcquisitionForUsernamePassword? OnBeforeTokenAcquisitionForUsernamePassword;
```
jmprieur, line 35: Would it be possible to have a more general event `OnBeforeTokenAcquisitionForUser`
with an additional parameter in the delegate of type enum saying {UsernamePassword or webApp }
?

tests/Microsoft.Identity.Web.Test/TokenAcquisitionAddInTests.cs
```diff
@@ -60,5 +61,51 @@ public async Task InvokeOnBeforeTokenAcquisitionForApp_InvokesEvent()
             Assert.NotNull(result);
             Assert.Equal(TokenSource.IdentityProvider, result.AuthenticationResultMetadata.TokenSource);
         }
+
+        [Fact]
+        public async Task InvokeOnBeforeTokenAcquisitionForUsernamePassword_InvokesEvent()
```
jmprieur, line 66: Not sure what this is testing?
Isn't it possible to test the TokenAcquirer feature?
Here this is testing MSAL?

tests/Microsoft.Identity.Web.Test/TokenAcquisitionAddInTests.cs
```diff
@@ -60,5 +61,51 @@ public async Task InvokeOnBeforeTokenAcquisitionForApp_InvokesEvent()
...
+                .AcquireTokenByUsernamePassword(new string[] { "scope" }, "username", "something");
+
+            bool eventInvoked = false;
+            options.OnBeforeTokenAcquisitionForUsernamePassword += (builder, options) =>
+            {
+                MsalAuthenticationExtension extension = new MsalAuthenticationExtension();
+                extension.OnBeforeTokenRequestHandler = async (request) =>
+                {
+                    eventInvoked = true;
+                };
```
jmprieur, line 95: you need to return a Task here

jmprieur (commented):
LGTM for the product changers
Not sure about the test

I think your branch is not up to date with master?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisitionExtensionOptions.cs
neha-bhargava: Then how will the builder be passed to the delegate? Currently I am passing builder for the ROPC flow. [See ](https://github.com/AzureAD/microsoft-identity-web/pull/3130/files#diff-1bcbce3e8aad1539a2c83fba7250ba30c20daecaba15b31844f5d2178c2eef18R22)

tests/Microsoft.Identity.Web.Test/TokenAcquisitionAddInTests.cs
neha-bhargava: Updated

tests/Microsoft.Identity.Web.Test/TokenAcquisitionAddInTests.cs
neha-bhargava: Updated the test to add body parameter through extensibility and validate it through the mock handler.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisitionExtensionOptions.cs
neha-bhargava: I don't think this can be made generic.

neha-bhargava:
> LGTM for the product changers Not sure about the test
> 
> I think your branch is not up to date with master?

Updated the test to validate the params sent in the request as changed through extensibility. I don't see updates from master

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -357,13 +357,23 @@ public async Task<AuthenticationResult> GetAuthenticationResultForUserAsync(
...
-                    scopes.Except(_scopesRequestedByMsal),
-                    username,
-                    password)
-                    .ExecuteAsync()
-                    .ConfigureAwait(false);
+                AcquireTokenByUsernameAndPasswordConfidentialParameterBuilder builder = ((IByUsernameAndPassword)application)
+                    .AcquireTokenByUsernamePassword(
+                        scopes.Except(_scopesRequestedByMsal),
+                        username,
+                        password);
```
bgavrilMS, line 368: I think you want to relax the condition at line 334 to allow this code to be hit if the end-user provides an empty password for user FIC

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisitionExtensionOptions.cs
bgavrilMS: Maybe Neha should add the `OnBefore` hooks for auth_code and for obo as well?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
neha-bhargava: I was thinking of sending a default value for password from MISE when a user assertion is passed instead since MSAL also validates that the password should not be null.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisitionExtensionOptions.cs
neha-bhargava: @jmprieur could you please suggest?

neha-bhargava: I can implement for the auth_code and obo as well if needed. I did not since those are not needed for now.

bgavrilMS: To me this level of granularity that exists right now is fine and avoids having to define an enum. Semantically the 2 options are the same ... the grant needs to be explicit.

bgavrilMS (approved):

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisitionExtensionOptions.cs
jennyf19: I don't think we want a specific event like `BeforeTokenAcquisitionForUsernamePassword` but one for user and one for app w/the needed parameters.

would appreciate some more time to think this through.

jennyf19 (changes requested):
Need more time to think this through. ETA is 11/8 eod.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisitionExtensionOptions.cs
jennyf19: So the issue really is the name, I would like to not surface "UsernamePassword" in the extensibility options.

bgavrilMS: I had a chat with @jmprieur and we'd like to propose a simple solution - instead of 

`public event BeforeTokenAcquisitionForUsernamePassword? OnBeforeTokenAcquisitionForUsernamePassword;` we will rename it all as 

`public event BeforeTokenAcquisitionForTestOnlyUser? OnBeforeTokenAcquisitionForTestOnlyUser;`

tests/Microsoft.Identity.Web.Test/TokenAcquisitionAddInTests.cs
```diff
@@ -60,5 +63,55 @@ public async Task InvokeOnBeforeTokenAcquisitionForApp_InvokesEvent()
...
+            // Act
+            options.InvokeOnBeforeTokenAcquisitionForUsernamePassword(builder, acquireTokenOptions);
+
+            var result = await builder.ExecuteAsync();
+
+            // Assert
+            Assert.True(eventInvoked);
+            Assert.NotNull(result);
+            Assert.Equal(TokenSource.IdentityProvider, result.AuthenticationResultMetadata.TokenSource);
+        }
```
bgavrilMS, line 115: @neha-bhargava - there are 2 GH warnings pending.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisitionExtensionOptions.cs
neha-bhargava: Updated the name. Thanks!

tests/Microsoft.Identity.Web.Test/TokenAcquisitionAddInTests.cs
neha-bhargava: I saw those too. I am not sure what those mean. Those are from the last PR.

neha-bhargava:
> Need more time to think this through. ETA is 11/8 eod.

@jennyf19 updates as suggested. Could you please re-review. Thanks!

jmprieur (approved):
LGTM
Thanks @neha-bhargava

jennyf19 (approved):

Issue #3121, opened 2024-11-01
[Feature Request] Allow multi-tenant applications to specify the AppHomeTenantId to be used for client credentials.


jmprieur:
**Is your feature request related to a problem? Please describe.**
When a multi-tenant application (TenantId: "common" or "organizations") wants to acquire a token on behalf of itself, Microsoft.Identity.Web can't use the "TenantId" as this is not accepted by ESTS.

Today we require all developers to specify the tenant ID to use by code, or by configuration in the case of IDownstreamApi:

**Describe the solution you'd like**
- [x] Add in the MicrosoftIdentityApplicationOptions (in Microsoft.Identity.Abstractions), a new string property `AppHomeTenantId`, which developers of multi-tenant apps could set in order to define the home tenant of the app (the tenant for which the app can get a token to call a downstream API on behalf of itself). This will usually be the app registration tenant. For instance, add it [here](https://github.com/AzureAD/microsoft-identity-abstractions-for-dotnet/blob/b92053d41909a73742e8241a29ffd5baab73f472/src/Microsoft.Identity.Abstractions/ApplicationOptions/MicrosoftIdentityApplicationOptions.cs#L76)
- [x] In Microsoft.Identity.Web,
   - [x] Add the same property to MergeOptions, and update the method that populates MergedOptions from MicrosoftIdentityApplicationOptions.
   - [x]  when the token acquisition happens "for App", and if the TenantId is "common" or "organization" and the developer has not overriden the "TenantId" in the AcquireTokenOptions, use the `AppHomeTenantId` value. For this we would need to change this expression [here](https://github.com/AzureAD/microsoft-identity-web/blob/011bd155ce3ddedeb43daa7348e09584cfa20552/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs#L437) to use `mergedOptions.AppHomeTenantId` if it's not null, and throw otherwise, like today
- [ ] For Auto-decrypt (SAL) use the `AppHomeTenantId` if the CredentialDescription.DownstreamApi.AcquireTokenOptions.Tenant is not specified. Might not be necessary as IdWeb does the token acquisition, but needs to be verified
- [ ] For MISE modules data providers, remove the configuration property that was using this tenant and let IdWeb do (or use the new property, instead of exposing it in each module configuration).
 
**Describe alternatives you've considered**
Multiply the same tenant value in all cases of auto-decrypt and MISE modules like today

msbw2:
[...]

PR #3132, opened 2024-11-07
Use AppHomeTenantId for acquiring app token when TenantId is not tenant

msbw2:
Use `AppHomeTenantId` for acquiring app token when `TenantId` is not tenant. https://github.com/AzureAD/microsoft-identity-web/issues/3121

### DIFF
--- a/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
@@ -72,7 +72,7 @@ namespace Microsoft.Identity.Web
         /// <summary>
         /// Meta-tenant identifiers which are not allowed in client credentials.
         /// </summary>
-        private readonly HashSet<string> _metaTenantIdentifiers = new HashSet<string>(
+        private static readonly HashSet<string> _metaTenantIdentifiers = new HashSet<string>(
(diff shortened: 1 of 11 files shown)


src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -430,6 +430,10 @@ public async Task<AuthenticationResult> GetAuthenticationResultForAppAsync(
             if (string.IsNullOrEmpty(tenant))
             {
                 tenant = mergedOptions.TenantId;
+                if (!string.IsNullOrEmpty(tenant) && _metaTenantIdentifiers.Contains(tenant!) && !string.IsNullOrEmpty(mergedOptions.AppHomeTenantId))
```
jennyf19, line 433: this class is tricky to test, what are your plans for testing?

jmprieur: Maybe we could just extract the logic to compute the effective tenant (including the lines above and below) in an internal method, and write a unit test for this method?

jmprieur (approved):
LGTM
Thanks @msbw2

jennyf19 (approved):
:shipit:

PR #3133, opened 2024-11-08
Fixing the ROPC test that broke the build

jmprieur:
# Fixing the ROPC test that broke the build.

The certificate that was used was not present on the build machine. Just using a certificate that is used by other tests.


jennyf19 (approved):
:shipit:

PR #3134, opened 2024-11-09
Update to .NET 9 GA. Update some test dependencies.

### LINKED ISSUE #3127: [Feature Request] Update to .NET 9 GA version
Update any .NET 9 dependencies to stable version, update .NET 9 version in build pipeline YAML, fix any new analyzer warnings.

pmaytak:
Closes #3127

### DIFF
--- a/.github/workflows/dotnetcore.yml
+++ b/.github/workflows/dotnetcore.yml
@@ -34,7 +34,6 @@ jobs:
     - name: 'Setup MSBuild'
       uses: microsoft/setup-msbuild@v2
       with:
-        vs-version: '17.12'
         vs-prerelease: true
         msbuild-architecture: x64
 
--- a/.github/workflows/dotnetcore.yml
+++ b/.github/workflows/dotnetcore.yml
@@ -26,21 +26,6 @@ jobs:
     - name: Checkout repository
       uses: actions/checkout@v4.2.1
 
-    - name: Setup .NET 6.0.x
-      uses: actions/setup-dotnet@v4.0.1
-      with:
-        dotnet-version: 6.0.x
-
-    - name: Setup .NET 7.0.x
-      uses: actions/setup-dotnet@v4.0.1
-      with:
-        dotnet-version: 7.0.x
-
-    - name: Setup .NET 8.0.x
-      uses: actions/setup-dotnet@v4.0.1
-      with:
-        dotnet-version: 8.0.x
-
     - name: Setup .NET 9.0.x
       uses: actions/setup-dotnet@v4.0.1
       with:
--- a/.github/workflows/dotnetcore.yml
+++ b/.github/workflows/dotnetcore.yml
@@ -20,14 +20,14 @@ on:
 jobs:
   build:
     runs-on: windows-latest
-    continue-on-error: true
+    continue-on-error: false
     name: "Build and run unit tests"
     steps:
     - name: Checkout repository
       uses: actions/checkout@v4.2.1
 
     - name: Setup .NET 9.0.x
-      uses: actions/setup-dotnet@v4.0.1
+      uses: actions/setup-dotnet@v4.1.0
       with:
         dotnet-version: 9.0.x
 
@@ -53,10 +53,10 @@ jobs:
 
     - name: Create code coverage report
       run: |
-        dotnet tool install -g dotnet-reportgenerator-globaltool
+        dotnet tool install -g dotnet-reportgenerator-globaltool --version 5.4.1
         reportgenerator -reports:./**/coverage.cobertura.xml -targetdir:CodeCoverage -reporttypes:'MarkdownSummaryGithub;Cobertura'
 
-    - name: Write Coverage to Job Summary
+    - name: Write coverage to job summary
       shell: bash
       run: |
         cat CodeCoverage/SummaryGithub.md >> $GITHUB_STEP_SUMMARY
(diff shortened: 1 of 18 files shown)


.github/workflows/dotnetcore.yml
```diff
@@ -44,7 +44,7 @@ jobs:
     - name: Setup .NET 9.0.x
```
keegan-caruso, line 44: Shouldn't need 6, 7, 8 with this. The 9 SDK can build 6,7,8.

pmaytak: Removed.

Directory.Build.props
```diff
@@ -26,8 +26,8 @@
   <PropertyGroup>
     <!-- For files to appear in the Visual Studio Solution explorer given we have conditional inclusion in some projects (IdWeb for instance)
          we need to have the higher framework, even if this produces a warning in the IDE -->
-    <TargetFrameworks>net6.0; net7.0; net8.0; net462; net472; netstandard2.0</TargetFrameworks>
-    <TargetFrameworks Condition="'$(TargetNet9)'== 'True'">$(TargetFrameworks); net9.0</TargetFrameworks>
+    <TargetFrameworks>net462; net472; netstandard2.0; net6.0; net7.0; net8.0; net9.0</TargetFrameworks>
```
jennyf19, line 29: there was a reason for the order, it was because it wasn't picking up all the frameworks. So not sure if it needs to move back to how it was previously.

pmaytak: Reverted.

pmaytak:
There are two things that are breaking:
- [AOT check](https://github.com/AzureAD/microsoft-identity-web/actions/runs/11813374464/job/32910296600?pr=3134#step:5:72) `Actual warning count: 61 is not as expected. Expected warning count is: 63`
- [Unit tests on .NET 9](https://github.com/AzureAD/microsoft-identity-web/actions/runs/11813126227/job/32909564564?pr=3134#step:11:632). 3 tests are strangely failing in Action but succeed locally. `PreventChangesInOpenIdConnectOptionsToBeOverlooked` is one of them - `Expected: 57 Actual:   62`. 62 is the correct number of properties on .NET 9, however it expects 57 for some reason.

src/Microsoft.Identity.Web.GraphServiceClient/Microsoft.Identity.Web.GraphServiceClient.csproj
```diff
@@ -20,7 +20,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Graph" Version="5.40.0" />
+    <PackageReference Include="Microsoft.Graph" Version="5.51.0" />
```
pmaytak, line 23: Updated to this version because the transitive dependency ([Microsoft.Kiota.Http.HttpClientLibrary](https://www.nuget.org/packages/Microsoft.Kiota.Http.HttpClientLibrary/1.3.4#dependencies-body-tab)) for the previous version doesn't support System.Text.Json 9.0.0. This is the minimum Graph version that does. 

Alternative is to include 5.51.0 on .NET 9 and 5.40.0 on other targets, but I think this is more consistent.

build/test-aot.ps1
```diff
@@ -16,7 +16,7 @@ foreach ($line in $($publishOutput -split "`r`n"))
 }
 
 Write-Host "Actual warning count is: ", $actualWarningCount
-$expectedWarningCount = 63
+$expectedWarningCount = 61
```
pmaytak, line 19: I had to upgrade Microsoft.Graph and the transitive dependencies were updated. These 2 were removed:
> ILC : Trim analysis warning IL2075: Microsoft.Kiota.Abstractions.RequestInformation.GetEnumName<T>(!!0): 'this' argument does not satisfy 'DynamicallyAccessedMemberTypes.PublicConstructors', 'DynamicallyAccessedMemberTypes.PublicMethods', 'DynamicallyAccessedMemberTypes.PublicFields', 'DynamicallyAccessedMemberTypes.PublicNestedTypes', 'DynamicallyAccessedMemberTypes.PublicProperties', 'DynamicallyAccessedMemberTypes.PublicEvents' in call to 'System.Type.GetMember(String)'. The return value of method 'System.Object.GetType()' does not have matching annotations. The source value must declare at least the same requirements as those declared on the target location it is assigned to. [D:\a\microsoft-identity-web\microsoft-identity-web\tests\Microsoft.Identity.Web.AotCompatibility.TestApp\Microsoft.Identity.Web.AotCompatibility.TestApp.csproj::TargetFramework=net8.0]
ILC : AOT analysis warning IL3050: Microsoft.Kiota.Abstractions.RequestInformation.ReplaceEnumValueByStringRepresentation(Object): Using member 'System.Enum.GetValues(Type)' which has 'RequiresDynamicCodeAttribute' can break functionality when AOT compiling. It might not be possible to create an array of the enum type at runtime. Use the GetValues<TEnum> overload or the GetValuesAsUnderlyingType method instead. [D:\a\microsoft-identity-web\microsoft-identity-web\tests\Microsoft.Identity.Web.AotCompatibility.TestApp\Microsoft.Identity.Web.AotCompatibility.TestApp.csproj::TargetFramework=net8.0]

jennyf19 (approved):
:shipit:

westin-m (approved):

PR #3137, opened 2024-11-12
Dont modify the merged options when building the confidential client

keegan-caruso:
# Don't mutate options

## Description

PrepareAuthorityInstanceForMsal modifies a shared options object, we shouldn't do that.

### DIFF
--- a/src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
@@ -6,10 +6,9 @@ using System.Collections.Generic;
 using System.Linq;
 using Microsoft.Identity.Abstractions;
 
-#if !NETSTANDARD2_0 && !NETFRAMEWORK
+#if !NETSTANDARD2_0 && !NET462 && !NET472
 using Microsoft.AspNetCore.Authentication.JwtBearer;
 using Microsoft.IdentityModel.Protocols.OpenIdConnect;
-using Microsoft.AspNetCore.Authentication.OpenIdConnect;
 #endif
 using Microsoft.Identity.Client;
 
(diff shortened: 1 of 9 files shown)


keegan-caruso:
Don't merge yet, need to do more validation

jennyf19 (approved):
:shipit:

I like it.

jennyf19:
@keegan-caruso is this one ready?

src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
```diff
@@ -466,14 +474,14 @@ public void PrepareAuthorityInstanceForMsal()
             if (IsB2C && Instance.EndsWith("/tfp/", StringComparison.OrdinalIgnoreCase))
```
jmprieur, line 474: Maybe a better fix could be to do it:
- here: https://github.com/AzureAD/microsoft-identity-web/blob/455c68776f44ec574738a0fc1b7b1f6039789e44/src/Microsoft.Identity.Web/WebApiExtensions/MicrosoftIdentityWebApiAuthenticationBuilderExtensions.cs#L183
- and here: https://github.com/AzureAD/microsoft-identity-web/blob/455c68776f44ec574738a0fc1b7b1f6039789e44/src/Microsoft.Identity.Web/WebAppExtensions/MicrosoftIdentityWebAppAuthenticationBuilderExtensions.cs#L325

Moving the code in AuthorityHelpers?

jmprieur (approved):
LGTM but we'd want to:
- test the B2C test apps manually
- run a unified build with this branch

I also proposed an alternative fix

src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
keegan-caruso: That should also work.

PR #3140, opened 2024-11-12
Update the extensibility to add user

### LINKED ISSUE #3139: [Feature Request] Update the extensibility for test user to include user claims
The extensibility needs to be updated to pass the user claims to make use of the pattern for user FIC scenario.

Pass the user claims along with the other parameters. Modify the delegate to pass user claims as `public delegate void BeforeTokenAcquisitionForTestUser(AcquireTokenByUsernameAndPasswordConfidentialParameterBuilder builder, AcquireTokenOptions? acquireTokenOptions, ClaimsPrincipal user);`

**Additional context**

neha-bhargava:
# Update the extensibility `BeforeTokenAcquisitionForTestUser` to pass user claims 

## Description

Need to pass the user to the extensibility to ensure the parameters can be updated before token request on basis of the user.

Fixes #3139

### DIFF
--- a/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisitionExtensionOptions.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisitionExtensionOptions.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License.
 
+using System.Security.Claims;
 using Microsoft.Identity.Abstractions;
 using Microsoft.Identity.Client;
 
@@ -38,11 +39,11 @@ namespace Microsoft.Identity.Web
         /// Invoke the BeforeTokenAcquisitionForTestUser event.
         /// </summary>
         internal void InvokeOnBeforeTokenAcquisitionForTestUser(AcquireTokenByUsernameAndPasswordConfidentialParameterBuilder builder,
-                                                           AcquireTokenOptions? acquireTokenOptions)
+                                                           AcquireTokenOptions? acquireTokenOptions, ClaimsPrincipal user)
         {
             if (OnBeforeTokenAcquisitionForTestUser != null)
             {
-                OnBeforeTokenAcquisitionForTestUser(builder, acquireTokenOptions);
+                OnBeforeTokenAcquisitionForTestUser(builder, acquireTokenOptions, user);
             }
         }
     }
--- a/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisitionExtensionOptions.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisitionExtensionOptions.cs
@@ -41,10 +41,7 @@ namespace Microsoft.Identity.Web
         internal void InvokeOnBeforeTokenAcquisitionForTestUser(AcquireTokenByUsernameAndPasswordConfidentialParameterBuilder builder,
                                                            AcquireTokenOptions? acquireTokenOptions, ClaimsPrincipal user)
         {
-            if (OnBeforeTokenAcquisitionForTestUser != null)
-            {
-                OnBeforeTokenAcquisitionForTestUser(builder, acquireTokenOptions, user);
-            }
+            OnBeforeTokenAcquisitionForTestUser?.Invoke(builder, acquireTokenOptions, user);
         }
     }
 }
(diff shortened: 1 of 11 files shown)


src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisitionExtensionOptions.cs
```diff
@@ -38,12 +39,9 @@ internal void InvokeOnBeforeTokenAcquisitionForApp(AcquireTokenForClientParamete
         /// Invoke the BeforeTokenAcquisitionForTestUser event.
         /// </summary>
         internal void InvokeOnBeforeTokenAcquisitionForTestUser(AcquireTokenByUsernameAndPasswordConfidentialParameterBuilder builder,
-                                                           AcquireTokenOptions? acquireTokenOptions)
+                                                           AcquireTokenOptions? acquireTokenOptions, ClaimsPrincipal user)
```
jennyf19, line 42: FYI, that this is a breaking change. Had we already shipped, we would have had to take a major version.

jennyf19: Also, has this been tested? Surprised this wasn't caught previously.

neha-bhargava: Agree that this is a breaking change. I noticed this needs to be added when implementing the user FIC scenario in MISE for which this extensibility is being added. I did add a unit test but the complete test will be in MISE as we cannot have an integration test for user fic in this repo. MISE and Id web are new code base for me. Sorry I missed this before.

bgavrilMS: We can introduce as non-breaking if needed, but it's more work + soft deprecation etc.
It would be a second callback that fire.

bgavrilMS (approved):

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisitionExtensionOptions.cs
jennyf19: Breaking change is consistent across libraries, so it's not library specific. Any change in the public API (not an addition) is considered a breaking change. I'm more concerned that E2E testing has not been done to ensure there are not more changes needed to the public API. Ex. how was this not caught previously before merging?

neha-bhargava: Discussed offline.

pmaytak: @neha-bhargava What was the result of the conversation? Is this delegate only used by us?

pmaytak: Ok, I see that the delegate was added recently and has not been released yet; so not a breaking change. Similar comment to Jenny's then is just making sure this was tested E2E and is all that is needed so we don't need to update it after the release.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisitionExtensions.cs
```diff
@@ -19,5 +20,6 @@ namespace Microsoft.Identity.Web
     /// </summary>
     /// <param name="builder">Builder</param>
     /// <param name="acquireTokenOptions">Token acquisition options for the request. Can be null.</param>
-    public delegate void BeforeTokenAcquisitionForTestUser(AcquireTokenByUsernameAndPasswordConfidentialParameterBuilder builder, AcquireTokenOptions? acquireTokenOptions);
+    /// <param name="user">User claims.</param>
+    public delegate void BeforeTokenAcquisitionForTestUser(AcquireTokenByUsernameAndPasswordConfidentialParameterBuilder builder, AcquireTokenOptions? acquireTokenOptions, ClaimsPrincipal user);
```
pmaytak, line 24: @neha-bhargava  Is this naming correct? Why does it have `Test` in it?

jennyf19: already discussed last week, will follow up offline
JM signed off on the API as well.

pmaytak (approved):

PR #3142, opened 2024-11-13
updates for .NET 9

jennyf19:


.github/workflows/dotnetcore.yml
```diff
@@ -22,14 +22,18 @@ jobs:
     runs-on: windows-latest
     continue-on-error: true
     name: "Build and run unit tests"
+
+    strategy: 
+      matrix:
+        dotnet-version:[ '6.0.x', '7.0.x', '8.0.x', '9.0.x' ]
```
jennyf19, line 28: this seems to help against any issues w/the frameworks not running properly, there seemed to be an issue w/the wrong framework running on the OIDC tests

build/test-aot.ps1
```diff
@@ -16,7 +16,7 @@ foreach ($line in $($publishOutput -split "`r`n"))
 }
 
 Write-Host "Actual warning count is: ", $actualWarningCount
-$expectedWarningCount = 63
+$expectedWarningCount = 61
```
jennyf19, line 19: if the count is lower, that is okay. we don't want it going up. I assume some dependencies were updated.

.github/workflows/dotnetcore.yml
pmaytak: I tried installing all frameworks yesterday, it didn't help. But if this works, then it works...

jennyf19: it worked when I did a test PR

pmaytak (approved):

.github/workflows/dotnetcore.yml
pmaytak: The run failed?
https://github.com/AzureAD/microsoft-identity-web/actions/runs/11822912264

pmaytak: The run failed?
https://github.com/AzureAD/microsoft-identity-web/actions/runs/11822912264

jennyf19: That's not what it showed. and then I changed the base branch so I don't have proof.

PR #3152, opened 2024-11-18
Install all .NET versions in pipeline to fix run tests task

msbw2:
Install all .NET versions in pipeline to fix run tests task. Partially reverts 1e955bfec45d779763c954fd1b8adf015f23c053.

### DIFF
--- a/build/template-install-dependencies.yaml
+++ b/build/template-install-dependencies.yaml
@@ -4,6 +4,26 @@
 
 steps:
 - task: UseDotNet@2
+  displayName: 'Use .NET SDK 3.1'
+  inputs:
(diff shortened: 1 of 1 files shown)


build/template-install-dependencies.yaml
```diff
@@ -3,6 +3,26 @@
 #install dotnet core
 
 steps:
```
keegan-caruso, line 5: If this is actually what we need to do, lets do it as a strategy matrix and test per target.

jennyf19 (approved):
:shipit:

westin-m (approved):

PR #3175, opened 2025-01-03
Update dotnet actions

sebastienros:
Required to prevent build failures due to the upcoming CDN changes

see https://github.com/dotnet/core/issues/9671

### DIFF
--- a/.github/workflows/aot-check.yml
+++ b/.github/workflows/aot-check.yml
@@ -23,7 +23,7 @@ jobs:
         fetch-depth: 1
 
     - name: Setup .NET 9.0.x
-      uses: actions/setup-dotnet@v4.1.0
+      uses: actions/setup-dotnet@v4
       with:
         dotnet-version: 9.0.x
 
--- a/.github/workflows/dotnetcore.yml
+++ b/.github/workflows/dotnetcore.yml
@@ -24,10 +24,10 @@ jobs:
     name: "Build and run unit tests"
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v4.2.1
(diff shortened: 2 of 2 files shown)


.github/workflows/aot-check.yml
```diff
@@ -23,7 +23,7 @@ jobs:
         fetch-depth: 1
 
     - name: Setup .NET 9.0.x
-      uses: actions/setup-dotnet@v4.1.0
+      uses: actions/setup-dotnet@v4
```
pmaytak, line 26: @sebastienros Does targeting only a major version here always pulls the latest version of the action? (While targeting major.minor.patch will pin to that specific one?) Or is it that just the major action versions where updated with the fix?

pmaytak (approved):

.github/workflows/aot-check.yml
sebastienros: Correct, you could have done it on purpose because you want your builds to be super deterministic, but in this case the expected behavior won't change: install dotnet.

sebastienros: https://github.com/actions/setup-dotnet/issues/570

sebastienros: And this one raising your exact concern https://github.com/dotnet/core/issues/9671#issuecomment-2569473755

keegan-caruso (approved):

.github/workflows/aot-check.yml
jennyf19: Thanks @sebastienros

bgavrilMS:
Thanks @sebastienros

PR #3184, opened 2025-01-07
Add null handling for process output/error data in UiTestHelpers

kllysng:


tests/E2E Tests/WebAppUiTests/UiTestHelpers.cs
```diff
@@ -182,8 +182,8 @@ public static Process StartProcessLocally(
             else
             {
                 // Log the output and error streams
-                process.OutputDataReceived += (sender, e) => output.WriteLine(e.Data);
-                process.ErrorDataReceived += (sender, e) => output.WriteLine(e.Data);
+                process.OutputDataReceived += (sender, e) => output.WriteLine(e?.Data ?? "null output data received.");
```
jmprieur, line 185: You might also want to write (before the data): the Process name of Process ID, as otherwise, there will be a mixed in any order of the output from all the processes in the test output.

Something like:
```csharp
process.OutputDataReceived += (sender, e) => {output.Write($"{process.Id} "); output.WriteLine(e?.Data ?? "null output data received.");}
```

and same for the following line?

jmprieur (approved):
LGTM
I provided a suggestion to improve the readability of the logs.

tests/E2E Tests/WebAppUiTests/UiTestHelpers.cs
kllysng: I like that and I've added it. Thank you!

keegan-caruso (approved):

PR #3185, opened 2025-01-07
package updates from dependabot

jennyf19:


keegan-caruso:
We should update the groupings for dependabot so it can handle more this itself.

We should separate out asp.net and microsoft.extensions into groups.

keegan-caruso (approved):

pmaytak (approved):

PR #3189, opened 2025-01-09
Revert WaitForProcess in UI Tests

jmprieur:
# Revert WaitForProcess in UI Tests

Removes the call to "await currentProcess.WaitForExitAsync();" as the web APIs never terminate by themselves (they are a loop
that serves requests until the process is killed).


jennyf19 (approved):
:shipit:

Issue #3186, opened 2025-01-07
 update the groupings for dependabot


jennyf19:
We should update the groupings for dependabot so it can handle more this itself.

We should separate out asp.net and microsoft.extensions into groups.

_Originally posted by @keegan-caruso in https://github.com/AzureAD/microsoft-identity-web/issues/3185#issuecomment-2576026615_

PR #3196, opened 2025-01-13
update dependabot

jennyf19:
#3186

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -37,8 +37,14 @@ updates:
       # Separate group for ASP.NET dependencies
       aspnet:
         patterns:
-          - "Microsoft.AspNetCore.*"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -34,3 +34,11 @@ updates:
         update-types:
         - "minor"
         - "patch"
+      # Separate group for ASP.NET dependencies
+      aspnet:
```
keegan-caruso, line 38: I think we want to limit these to minor and patch?

jennyf19: done, thanks!

keegan-caruso (approved):
thanks

kllysng (approved):

pmaytak (approved):

Issue #3021, opened 2024-09-19
Public api DefaultCredentialsLoader.LoadFirstValidCredentialsAsync fails on first failure


bgavrilMS:
### Microsoft.Identity.Web Library

Microsoft.Identity.Web

### Microsoft.Identity.Web version

3.10

### Web app

Sign-in users

### Web API

Protected web APIs (validating tokens)

### Token cache serialization

In-memory caches

### Description

1. Create a list of CredentialDescription with 2 credentials, e.g. a certificate from a non-existent file and a certificate from string. Ensure the first credential fails. 
2. Call the public API `LoadFirstValidCredentialsAsync(list)`

Actual: exception caused by failure to load the first credential
Expected: load second credential

### Reproduction steps

^^

### Relevant code snippets

```csharp
LoadFirstValidCredentialsAsync
```

### Expected behavior

no exception

jmprieur:
Is it a regression?

bgavrilMS:
I don't think so.

jennyf19:
Are you expecting my team to pick this up?

bgavrilMS:
Looks like a good first issue for someone from MSAL team to pick this up.

jennyf19:
Exit criteria: E2E testing is required to verify the logging works as expected.

bgavrilMS:
Yes, Id.Web uses a different API internally - [LoadCredentialForMsalOrFailAsync](https://github.com/AzureAD/microsoft-identity-web/blob/master//src/Microsoft.Identity.Web.TokenAcquisition/ConfidentialClientApplicationBuilderExtension.cs#L61)

This API is not used, but it is a public API:

![image](https://github.com/user-attachments/assets/b05f9749-bfe7-4ce3-a4b1-387af0c27c12)

bgavrilMS:
Consider deprecating this API. Is it useful on its own? 

CC @jmprieur

PR #3197, opened 2025-01-13
fix for 3021 but needs another test

jennyf19:
#3021


jennyf19:
got confused w/another issue.

PR #3210, opened 2025-01-22
Suppress TFM Build Warnings

kllysng:
Passing on https://identitydivision.visualstudio.com/IDDP/_build/results?buildId=1417266&view=results

I had to add suppressions to additional dir.build.props files as the setting is not propagating from the top level .props.

### DIFF
--- a/tests/Directory.Build.props
+++ b/tests/Directory.Build.props
@@ -8,6 +8,10 @@
     <IsPackable>false</IsPackable>
     <EnablePackageValidation>false</EnablePackageValidation>
     <AssemblyOriginatorKeyFile>../../build/MSAL.snk</AssemblyOriginatorKeyFile>
+  </PropertyGroup>
+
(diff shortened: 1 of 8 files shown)


tests/Directory.Build.props
```diff
@@ -8,6 +8,7 @@
     <IsPackable>false</IsPackable>
     <EnablePackageValidation>false</EnablePackageValidation>
     <AssemblyOriginatorKeyFile>../../build/MSAL.snk</AssemblyOriginatorKeyFile>
+    <SuppressTfmSupportBuildWarnings>true</SuppressTfmSupportBuildWarnings>
```
keegan-caruso, line 11: We should scope this down to the frameworks that need it. e.g. net8.0/9.0 don't need this.

jennyf19: like this: https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet/pull/3081/files

westin-m (approved):

pmaytak (approved):

PR #3226, opened 2025-01-30
Adding Extensibility for Custom Signed Assertion Providers

### LINKED ISSUE #3220: Expose credential loader extensibility to support custom signed assertion providers
## Summary

This design proposal aims at adding extensibility to DefaultCredentialsLoader so that partner teams, or SDK on top of Microsoft.Identity.Web can bring their own credential providers.

See also https://github.com/AzureAD/microsoft-identity-abstractions-for-dotnet/issues/146 and https://github.com/AzureAD/microsoft-identity-abstractions-for-dotnet/issues/153 which are pre-requisites

## Motivation and goals

See [Add extensibility to CredentialDescription to support custom signed assertion providers](https://github.com/AzureAD/microsoft-identity-abstractions-for-dotnet/issues/146) for the devex for users of the extensibility. 

In that article, we mentioned that the developer using this extensibility provides `CredentialDescription` with a specific `CustomSignedAssertionProviderName` and calls an extension method on the service collection (for instance `services.AddFmiProvider();`). This design proposal explains how the extender provides
- the extension method on the service collection.
- the signed assertion credential source loader, and the signed assertion provider
- the value that developers need to use in `CustomSignedAssertionProviderName` (possibly the data), 

## In scope

- This design proposal is for the DevEx for the provider of the extension (that is the extender)

## Out of scope

- See [Add extensibility to CredentialDescription to support custom signed assertion providers](https://github.com/AzureAD/microsoft-identity-abstractions-for-dotnet/issues/146) for the devex for users of the extensions.

## Proposed developer experience for the extender providing a custom assertion provider

Remember, the developer experience for the developer using the extension is described in [Add extensibility to CredentialDescription to support custom signed assertion providers](https://github.com/AzureAD/microsoft-identity-abstractions-for-dotnet/issues/146)

### Configuration file

Nothing to do. This is not about the developer using the extensibility

### Startup.cs

Nothing to do. This is not about the developer using the extensibility

### Classes provided as the extension

The developer providing the extension needs to provide three classes:

1. A class deriving from [ClientAssertionProviderBase](https://github.com/AzureAD/microsoft-identity-web/blob/master/src/Microsoft.Identity.Web.Certificateless/ClientAssertionProviderBase.cs) and that will compute or retrieve the signed assertion.

   **Exemple:**

   ```csharp
      internal class MyCustomSignedAssertionProvider : ClientAssertionProviderBase
      {
          public MyCustomSignedAssertionProvider(Dictionary<string, object>? properties)
          {
              // Here you would implement the logic to extract what you need from the properties passed in
              // the configuration
          }
          protected override Task<ClientAssertion> GetClientAssertionAsync(AssertionRequestOptions? assertionRequestOptions)
          {
              // Here you would implement the logic to get the signed assertion, which is probably going
              // to be a call to a service. This call can be parameterized by the parameters in the properties
              // of the constructor.
  
              // In this sample code we just create an empty signed assertion and return it.
              var clientAssertion = new ClientAssertion("FakeAssertion", DateTimeOffset.Now);
              return Task.FromResult(clientAssertion);
          }
      }
   ```

1. Provide a class implementing [ICustomSignedAssertionProvider](https://github.com/AzureAD/microsoft-identity-abstractions-for-dotnet/issues/153) which needs to:
    - get from the cached value or instantiate the class in 1) (MyCustomSignedAssertionProvider ) in 
       the LoadIfNeededAsync overload
    - Call GetSignedAssertionAsync
    - process the exception and set the Skip flag. Set the CachedValue to the new MyCustomSignedAssertionProvider in case there is no exception.

    ```csharp
    namespace ExtensibilityTests
    {
        internal class MyCustomSignedAssertionLoader : ICustomSignedAssertionProvider
        {
            public MyCustomSignedAssertionLoader(ILogger<DefaultCredentialsLoader> logger)
            {
                _logger = logger;
            }
            public CredentialSource CredentialSource => CredentialSource.CustomSignedAssertion;
    
            public string Name => "MyCustomExtension";
    
            private ILogger<DefaultCredentialsLoader> _logger;
    
            public async Task LoadIfNeededAsync(CredentialDescription credentialDescription, CredentialSourceLoaderParameters? parameters = null)
            {
                MyCustomSignedAssertionProvider? signedAssertion = credentialDescription.CachedValue as MyCustomSignedAssertionProvider;
                if (credentialDescription.CachedValue == null)
                {
                    signedAssertion = new MyCustomSignedAssertionProvider(credentialDescription.CustomSignedAssertionProviderData);
                }
                try
                {
                    // Given that managed identity can be not available locally, we need to try to get a
                    // signed assertion, and if it fails, move to the next credentials
                    _ = await signedAssertion!.GetSignedAssertionAsync(null);
                    credentialDescription.CachedValue = signedAssertion;
                }
                catch (Exception)
                {
                    credentialDescription.Skip = true;
                    throw;
                }
            }
        }
    }
    ```

1. Write an extension method to register the extension with the service collection. This is the only class the developer
   using the extension will need to use.

   ```csharp
      public static IServiceCollection AddCustomSignedAssertionProvider(
         this IServiceCollection services)
     {
         services.AddSingleton<ICustomSignedAssertionProvider, MyCustomSignedAssertionLoader>();
         return services;
     }
   ```

# Design/Work to be done in Microsoft.Identity.Web
1. In Microsoft.Identity.Web.DefaultCredentialsLoader (in the Microsoft.Identity.Web.Certificate)

    - [x] Add a public property `CustomSignedAssertionCredentialSourceLoaders` of type 
       ` IDictionary<string, ICustomSignedAssertionProvider>`
      
      ```csharp
        /// <summary>
        /// Dictionary of custom signed assertion credential source loaders, by name (fully qualified type name).
        /// </summary>
        public IDictionary<string, ICustomSignedAssertionProvider> CustomSignedAssertionCredentialSourceLoaders { get; }
      ```

   - [x] Add an extra parameter to the constructor of DefaultCredentialLoader: signedAssertionProviders 
 of type `IEnumerable<ICustomSignedAssertionProvider>?` with a default value of null, to avoid a breaking change.
   - [x] Initialize CustomSignedAssertionCredentialSourceLoaders it in the constructor of DefaultCredentialsLoader based
          on the values in signedAssertionProviders. For each provider, the key would be `provider.Name ?? provider.GetType().FullName!`

   - [x] In the LoadCredentialsIfNeededAsync method, when the CachedValue is null, if the sourceType of the credential 
          description is CredentialSource.CustomSignedAssertion and credentialDescription.CustomSignedAssertionProviderName
          is defined, instead of using the CredentialsSourceLoaders, use the CustomSignedAssertionSourceLoaders. The code will
          be very similar in both cases (custom providers or in the box providers).
          If the CustomSignedAssertionCredentialSourceLoaders of name 
          credentialDescription.CustomSignedAssertionProviderName is not found, we need to log a new error 
          (CustomSignedAssertionProviderNotFound):

          "Failed to find custom signed assertion provider {name} from source {sourceType}. Will it be skipped in the future ? {skip}.");

   - [x] Modify the DefaultCredentialLoader.Logger.cs file to add the new log message.

1. In Microsoft.Identity.Web.DefautCertificateLoader 
   - [x] Add the same kind of constructor in DefaultCertificateLoader (calling the constructor of the base class). This will
          be used by the Dependency injection to inject the ICustomSignedAssertionProvider (custom providers)
  
1. In ConfidentialClientApplicationBuilderExtension, in LoadCredentialForMsalOrFailAsync:
   - [x] process the case where credential.SourceType == CredentialSource.CustomSignedAssertion (if not skiped, return the credentials). and add
  logging

1. Unit tests

1. Integration tests.

   Create an integration test that demonstrate the extension (MyCustomSignedAssertionProvider)
   and the way to use it. (for instance with a daemon configuration like the following, but that's just an example)

   ```csharp
            TokenAcquirerFactory tokenAcquirerFactory = TokenAcquirerFactory.GetDefaultInstance();
            tokenAcquirerFactory.Services.Configure<MicrosoftIdentityApplicationOptions>(options =>
            {
                options.Instance = "https://login.microsoftonline.com/";
                options.TenantId = "msidlab4.onmicrosoft.com";
                options.ClientId = "f6b698c0-140c-448f-8155-4aa9bf77ceba";
                options.ClientCredentials = [ new CredentialDescription() { SourceType = CredentialSource.CustomSignedAssertion,
                CustomSignedAssertionProviderName = "MyCustomExtension"}];
            });
            tokenAcquirerFactory.Services.AddCustomSignedAssertionProvider();

            // Get the authorization request creator service
            IAuthorizationHeaderProvider authorizationHeaderProvider = serviceProvider.GetRequiredService<IAuthorizationHeaderProvider>();
            
            string authorizationHeader = await authorizationHeaderProvider.CreateAuthorizationHeaderForAppAsync("https://graph.microsoft.com/.default");
            Console.WriteLine(authorizationHeader.Substring(0, authorizationHeader.IndexOf(" ", StringComparison.OrdinalIgnoreCase) + 4) + "...");
            }
   ```

1. Add documentation
   
   Add a wiki article to document the extensibility (probably based on [the ](https://github.com/AzureAD/microsoft-identity-abstractions-for-dotnet/issues/146) and this design proposal)

JoshLozensky:
This PR adds extensibility for customers to be able to define their own signed assertion providers. The integration test in \tests\E2E Tests\CustomSignedAssertionProviderTests\CustomSignedAssertionProviderExtensibilityTests.cs has an example of the sort of work a customer would need to do in order to use this functionality.

Fixes #3220


src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoaderCustomSignedAssertion.cs
```diff
@@ -0,0 +1,111 @@
...
+
+            CustomSignedAssertionCredentialSourceLoaders = sourceLoaderDict;
+        }
+
+        /// <summary>
+        /// Dictionary of custom signed assertion credential source loaders, by name (fully qualified type name).
+        /// </summary>
+        public IDictionary<string, ICredentialSourceLoader>? CustomSignedAssertionCredentialSourceLoaders { get; }
+
+
```
jennyf19, line 38:
```suggestion
```

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoaderCustomSignedAssertion.cs
```diff
@@ -0,0 +1,111 @@
...
+            CustomSignedAssertionProviderNotFoundException providerNotFoundException;
+
+            // No source loader(s)
+            if (CustomSignedAssertionCredentialSourceLoaders == null || !CustomSignedAssertionCredentialSourceLoaders.Any())
+            {
+                providerNotFoundException = CustomSignedAssertionProviderNotFoundException.SourceLoadersNullOrEmpty();
+            }
+
+            // No provider name
+            else if (string.IsNullOrEmpty(credentialDescription.CustomSignedAssertionProviderName))
```
jennyf19, line 50: do we need to check if `credentialDescription` is null first?

jmprieur (commented):
Good start
I left some questions/comments. I'm puzzled by the exceptions.

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoaderCustomSignedAssertion.cs
```diff
@@ -0,0 +1,111 @@
...
+                await sourceLoader.LoadIfNeededAsync(credentialDescription, parameters);
+                return;
+            }
+
+            Logger.CustomSignedAssertionProviderLoadingFailure(_logger, credentialDescription, providerNotFoundException);
+            throw providerNotFoundException;
+        }
+    }
+
+    internal class CustomSignedAssertionProviderNotFoundException : Exception
```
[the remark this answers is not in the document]
jennyf19: even if not, we cannot create new exception types, it creates a lot of confusion for customers

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
```diff
@@ -72,7 +72,11 @@ public async Task LoadCredentialsIfNeededAsync(CredentialDescription credentialD
                 {
                     if (credentialDescription.CachedValue == null)
                     {
-                        if (CredentialSourceLoaders.TryGetValue(credentialDescription.SourceType, out ICredentialSourceLoader? loader))
+                        if (credentialDescription.SourceType == CredentialSource.CustomSignedAssertion)
+                        {
+                            await ProcessCustomSignedAssertionAsync(credentialDescription, parameters);
```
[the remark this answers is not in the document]
JoshLozensky: @jmprieur 
I'm a bit confused as to why we would need to catch it here since it is logged inside the ProcessCustomSignedAssertionAsync method. In the similar example on lines 87 and 88 for other source types when we do catch the exception, we do so to log it before throwing.

Should the exception be made public? Wasn't sure we wanted to add to our public API.

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoaderCustomSignedAssertion.cs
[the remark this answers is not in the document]
jmprieur: We shouldn't need to have exceptions here. CredentialDescriptions are chained, so even if one fails, the next can succeed. We have 2 cases here:
- Mis-configuration. Logging is enough (Error is fine). We don't want to throw exceptions at least in this code. If we feel strongly about warning the developer, we should not do it here, but in the validation of the configuration (have another work item for that)
-

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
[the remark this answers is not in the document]
jmprieur: We don't need to throw at all. We are in control of the code.

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoaderCustomSignedAssertion.cs
```diff
@@ -0,0 +1,110 @@
...
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Microsoft.Identity.Abstractions;
+
+namespace Microsoft.Identity.Web
+{
+    public partial class DefaultCredentialsLoader : ICredentialsLoader
```
jmprieur, line 14: do you need to repeat the base interface here?

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoaderCustomSignedAssertion.cs
```diff
@@ -0,0 +1,110 @@
...
+            // No provider name
+            else if (string.IsNullOrEmpty(credentialDescription.CustomSignedAssertionProviderName))
+            {
+                providerNotFoundException = CustomSignedAssertionProviderNotFoundException.ProviderNameNullOrEmpty();
+            }
+
+            // No source loader for provider name
+            else if (!CustomSignedAssertionCredentialSourceLoaders!.TryGetValue(credentialDescription.CustomSignedAssertionProviderName!, out ICredentialSourceLoader? sourceLoader))
+            {
+                providerNotFoundException = CustomSignedAssertionProviderNotFoundException.ProviderNameNotFound(credentialDescription.CustomSignedAssertionProviderName!);
```
jmprieur, line 57: Just log errors. Please don't throw

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoaderCustomSignedAssertion.cs
JoshLozensky: I don't think so, it is already dereferenced in multiple if statements required to reach this code path in the public LoadCredentialsIfNeededAsync method in the DefaultCredentialsLoader.cs file. The only way for this to end up needing a null check is if someone else uses this private method irresponsibly in the future.

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoaderCustomSignedAssertion.cs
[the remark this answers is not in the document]
JoshLozensky: I will try to find a relevant existing exception unless I hear otherwise @jmprieur

JoshLozensky:
> Good start I left some questions/comments. I'm puzzled by the exceptions.

I mistook "log a new error" from the spec for throw a new exception

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoaderCustomSignedAssertion.cs
```diff
@@ -0,0 +1,111 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
```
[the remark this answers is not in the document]
JoshLozensky: changed

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoaderCustomSignedAssertion.cs
JoshLozensky: I do not, removed

src/Microsoft.Identity.Web.Certificate/CertificateErrorMessage.cs
```diff
@@ -16,6 +16,9 @@ internal static class CertificateErrorMessage
         public const string BothClientSecretAndCertificateProvided = "IDW10105: Both client secret and client certificate, " +
                    "cannot be included in the configuration of the web app when calling a web API. ";
         public const string ClientCertificatesHaveExpiredOrCannotBeLoaded = "IDW10109: All client certificates passed to the configuration have expired or can't be loaded. ";
+        public const string CustomProviderNameNullOrEmpty = "IDW10111 The name of the custom signed assertion provider is null or empty.";
+        public const string CustomProviderNotFound = "IDW10112: The custom signed assertion provider with name '{0}' was not found.";
```
jmprieur, line 20:
```suggestion
        public const string CustomProviderNotFound = "IDW10112: The custom signed assertion provider with name '{0}' was not found. Was it registered in the service collection?";
```

jmprieur:
Let's have an aka.ms link for the error messages for this feature.
https://aka.ms/msidweb/ms-id-web/custom-signed-assertion-providers, that will link to the wiki page

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.CustomSignedAssertion.cs
```diff
@@ -0,0 +1,75 @@
...
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Microsoft.Identity.Abstractions;
+
+
```
jennyf19, line 11:
```suggestion
```

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.CustomSignedAssertion.cs
```diff
@@ -0,0 +1,75 @@
...
+namespace Microsoft.Identity.Web
+{
+    public partial class DefaultCredentialsLoader
+    {
+        /// <summary>
+        /// Constructor for DefaultCredentialsLoader when using custom signed assertion provider source loaders.
+        /// </summary>
+        /// <param name="logger"></param>
+        /// <param name="customSignedAssertionProviders">Set of custom signed assertion providers.</param>
+        public DefaultCredentialsLoader(ILogger<DefaultCredentialsLoader>? logger, IEnumerable<ICustomSignedAssertionProvider> customSignedAssertionProviders) : this(logger)
```
jennyf19, line 21: nit: the logger param should come last, as required parameters should come first, followed by option or nullable params.

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.CustomSignedAssertion.cs
```diff
@@ -0,0 +1,75 @@
...
+        {
+            var sourceLoaderDict = new Dictionary<string, ICredentialSourceLoader>();
+
+            foreach (var provider in customSignedAssertionProviders)
```
jennyf19, line 25: do we need a null check on `customSignedAssertionProviders`?

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.CustomSignedAssertion.cs
```diff
@@ -0,0 +1,75 @@
...
+            {
+                sourceLoaderDict.Add(provider.Name ?? provider.GetType().FullName!, provider);
```
jennyf19, line 27: do we care about handling duplicate keys?
example:
```csharp:
 var key = provider.Name ?? provider.GetType().FullName!;
 if (!sourceLoaderDict.TryAdd(key, provider))
 {
     _logger.LogWarning("Duplicate provider key detected: {Key}. The provider was not added.", key);
 }
 ```

File: src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.CustomSignedAssertion.cs [hunk 13]
```diff
@@ -0,0 +1,75 @@
...
+            }
+
+            CustomSignedAssertionCredentialSourceLoaders = sourceLoaderDict;
+        }
+
+        /// <summary>
+        /// Dictionary of custom signed assertion credential source loaders, by name (fully qualified type name).
+        /// </summary>
+        public IDictionary<string, ICredentialSourceLoader>? CustomSignedAssertionCredentialSourceLoaders { get; }
```
jennyf19, line 36: could this be a `IReadOnlyDictionary` instead? this would ensure external code can read but not modify the dictionary.

tests/Microsoft.Identity.Web.Test/CustomSignedAssertionProviderTests.cs
```diff
@@ -0,0 +1,178 @@
...
+using System.Threading.Tasks;
+using Moq;
+using Microsoft.Extensions.Logging;
+using Microsoft.Identity.Abstractions;
+using Xunit;
+
+namespace Microsoft.Identity.Web.Test
+{
+    public class CustomSignedAssertionProviderTests
+    {
```
jennyf19, line 16: make sure that null parameters are handled correctly.

jennyf19: seems like there is too much focus on verifying the logs and not on the core functionalities.

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.CustomSignedAssertion.cs
jennyf19, line 36: this should be `ICustomSignedAssertionProvider` according to the spec?

tests/Microsoft.Identity.Web.Test/CustomSignedAssertionProviderTests.cs
jennyf19: would be good to have a concurrency test if possible.

tests/Directory.Build.props
```diff
@@ -34,6 +34,7 @@
     <MicrosoftApplicationInsightsEventCounterCollectionVersion>2.22.0</MicrosoftApplicationInsightsEventCounterCollectionVersion>
     <MicrosoftExtensionsCachingStackExchangeRedisVersion>6.0.12</MicrosoftExtensionsCachingStackExchangeRedisVersion>
     <MicrosoftPlaywrightVersion>1.48.0</MicrosoftPlaywrightVersion>
+    <MoqVersion>4.20.72</MoqVersion>
```
jennyf19, line 37: would rather not add this if we can avoid it.

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.CustomSignedAssertion.cs
JoshLozensky: Since the customSignedAssertionProviders arg in this constructor is not defined as nullable, we shouldn't need to check as any attempt to provide a null value to the constructor would result in a compile time error. It could be empty but that wouldn't hurt the foreach loop.

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.CustomSignedAssertion.cs
JoshLozensky: fixed

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.CustomSignedAssertion.cs
JoshLozensky: Agreed, the spec stated this property twice, one as `ICustomSignedAssertionProvider` the other as `ICredentialSourceLoader` (which is inherited by `ICustomSignedAssertionProvider`) so I edited it to be consistent as `ICustomSignedAssertionProvider` in both places and fixed the code.

tests/Directory.Build.props
JoshLozensky: It would be hard to mock the logger on my own unless there is a different mocking framework you'd prefer I use? 

For my learning, what are the concerns with adding package to our test project @jennyf19?

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.CustomSignedAssertion.cs
JoshLozensky: My understanding is the intent is for users to be able to modify it so they can add credential sources similar to `DefaultCredentialsLoader.CredentialSourceLoaders`. Added language to the summary to indicate as much.

@jmprieur are we concerned with users modifying or deleting entries from `CustomSignedAssertionCredentialSourceLoaders`?

tests/Microsoft.Identity.Web.Test/CustomSignedAssertionProviderTests.cs
JoshLozensky: > make sure that null parameters are handled correctly.
Are you saying we should handle potential null values for non-nullable members? There are very few nullable members in this work and if someone forces through a null value into a non-nullable arg/member wouldn't we want them to get the relevant null value exception?

> seems like there is too much focus on verifying the logs and not on the core functionalities.
I'm a bit confused as most "core functionality" is written in the extension of `ICustomSignedAssertionProvider` by the customer, I have a couple of simple mocks of extension implementation for the unit tests but the integration test is where any further functionality in that respect will be handled.

Also, why isn't logging considered core functionality when what we are building is code for other developers to use? Doesn't that make logging a feature for customers?

> would be good to have a concurrency test if possible.
What sort of concurrency scenario are you imagining i.e. the same `DefaultCredentialsLoader` being used repeatedly, many 'DefaultCredentialsLoader' objects being created/used, or something else?

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.CustomSignedAssertion.cs
JoshLozensky: added

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoaderCustomSignedAssertion.cs
jennyf19: this is the part that I was missing `dereferenced in multiple if statements required to reach this code path`

tests/Directory.Build.props
jennyf19: For example: https://medium.com/@michalsitek/critical-security-vulnerability-in-moq-4-20-0-ffd24739cc49

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.CustomSignedAssertion.cs
JoshLozensky: Make protected from offline convo

tests/Microsoft.Identity.Web.Test/CustomSignedAssertionProviderTests.cs
JoshLozensky: Offline answers:
- Yes check the non-nullable values
- Core functionality meaning similar to the checking null values and other ways someone might call public methods
- Concurrency shouldn't be an issue in this case since we control the internals of the object post-instantiation

tests/Directory.Build.props
JoshLozensky: Offline conclusion, try doing our own mocking

tests/Microsoft.Identity.Web.Test/CustomSignedAssertionProviderTests.cs
JoshLozensky: Addressed non-nullable arg in public method and accompanying test

tests/Directory.Build.props
JoshLozensky: Was able to do a custom mock logger myself (with help from copilot)

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.CustomSignedAssertion.cs
```diff
@@ -0,0 +1,84 @@
...
+
+namespace Microsoft.Identity.Web
+{
+    public partial class DefaultCredentialsLoader
+    {
+        /// <summary>
+        /// Constructor for DefaultCredentialsLoader when using custom signed assertion provider source loaders.
+        /// </summary>
+        /// <param name="customSignedAssertionProviders">Set of custom signed assertion providers.</param>
+        /// <param name="logger"></param>
```
keegan-caruso, line 19: Empty param

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.CustomSignedAssertion.cs
```diff
@@ -0,0 +1,84 @@
...
+
+            CustomSignedAssertionCredentialSourceLoaders = sourceLoaderDict;
+        }
+
+        /// <summary>
+        /// Dictionary of custom signed assertion credential source loaders, by name (fully qualified type name).
+        /// The application can add more to process additional credential sources.
+        /// </summary>
+        protected IDictionary<string, ICustomSignedAssertionProvider>? CustomSignedAssertionCredentialSourceLoaders { get; }
+
```
keegan-caruso, line 46:
```suggestion
```

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.CustomSignedAssertion.cs
```diff
@@ -0,0 +1,84 @@
...
+
+        private async Task ProcessCustomSignedAssertionAsync(CredentialDescription credentialDescription, CredentialSourceLoaderParameters? parameters)
+        {
+            // No source loader(s)
+            if (CustomSignedAssertionCredentialSourceLoaders == null || CustomSignedAssertionCredentialSourceLoaders.Count == 0)
+            {
+                _logger.LogError(CertificateErrorMessage.CustomProviderSourceLoaderNullOrEmpty);
+            }
+
```
keegan-caruso, line 55: This is not our typical style. There shouldn't be new lines between if / else if /else blocks. The comments can move inside the block if it helps with organization.

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.Logger.cs
```diff
@@ -18,13 +24,25 @@ private static class Logger
...
+                    "Failed to load credential {id} from source {sourceType}. Will it be skipped in the future ? {skip}."
+                );
 
             public static void CredentialLoadingFailure(ILogger logger, CredentialDescription cd, Exception? ex)
                 => s_credentialLoadingFailure(logger, cd.Id, cd.SourceType.ToString(), cd.Skip, ex);
+
+            private static readonly Action<ILogger, string, string, bool, Exception?> s_customSignedAssertionProviderLoadingFailure =
+                LoggerMessage.Define<string, string, bool>(
+                    LogLevel.Information,
+                    new EventId(7, nameof(CustomSignedAssertionProviderLoadingFailure)),
```
keegan-caruso, line 37: Event Ids should be unique

tests/E2E Tests/CustomSignedAssertionProviderTests/UnitTest1.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace CustomSignedAssertionProviderTests
+{
+    public class UnitTest1
```
keegan-caruso, line 6: ?

tests/E2E Tests/CustomSignedAssertionProviderTests/MyCustomSignedAssertionLoader.cs
```diff
@@ -0,0 +1,45 @@
...
+            try
+            {
+                // Try to get a signed assertion, and if it fails, move to the next credentials
+                _ = await signedAssertion!.GetSignedAssertionAsync(null);
+                credentialDescription.CachedValue = signedAssertion;
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(22, ex.Message);
+                _logger.LogError(42, "Failed to get signed assertion.");
```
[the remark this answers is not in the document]
JoshLozensky: This is for a test, not an actual error

tests/Microsoft.Identity.Web.Test/CustomSignedAssertionProviderTests.cs
```diff
@@ -0,0 +1,223 @@
...
+            else
+            {
+                Assert.DoesNotContain(loggerMock.LoggedMessages, log => log.LogLevel == expectedLogLevel);
+            }
+        }
+
+        public static IEnumerable<object[]> CustomSignedAssertionLoggingTestData()
+        {
+            // No source loaders
+            yield return new object[]
```
keegan-caruso, line 70: No real reason to use yield here, we can just return a list. This talks a little bit about discovery vs execution which applies here: https://xunit.net/faq/theory-data-stability-in-vs

tests/Microsoft.Identity.Web.Test/CustomSignedAssertionProviderTests.cs
```diff
@@ -0,0 +1,223 @@
...
+            {
+                Assert.Contains(loggerMock.LoggedMessages, log => log.LogLevel == expectedLogLevel && log.Message.Contains(expectedLogMessage, StringComparison.InvariantCulture));
+            }
+            else
+            {
+                Assert.DoesNotContain(loggerMock.LoggedMessages, log => log.LogLevel == expectedLogLevel);
+            }
+        }
+
+        public static IEnumerable<object[]> CustomSignedAssertionLoggingTestData()
```
keegan-caruso, line 67: Can we use a theory data object instead of a list of objects? This is hard to understand.

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.CustomSignedAssertion.cs
JoshLozensky: filled

tests/E2E Tests/CustomSignedAssertionProviderTests/MyCustomSignedAssertionLoader.cs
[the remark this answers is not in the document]
JoshLozensky: Cleaned this up, this class is a dummy version of what customers will need to implement, effectively a sample as opposed to code anyone will directly use

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.CustomSignedAssertion.cs
JoshLozensky: moved comments inside blocks

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.Logger.cs
JoshLozensky: They sure should, and now are

tests/E2E Tests/CustomSignedAssertionProviderTests/UnitTest1.cs
JoshLozensky: Hadn't gotten there yet, I wasn't the person to take it out of draft... This has a test now and is ironically an integration test

tests/Microsoft.Identity.Web.Test/CustomSignedAssertionProviderTests.cs
JoshLozensky: fixed

tests/Microsoft.Identity.Web.Test/CustomSignedAssertionProviderTests.cs
JoshLozensky: yep done

jennyf19 (approved):

src/Microsoft.Identity.Web.TokenAcquisition/ConfidentialClientApplicationBuilderExtension.Logger.cs
```diff
@@ -131,6 +137,14 @@ public static void UsingSignedAssertionFromVault(
                 ILogger logger,
                 string signedAssertionUri) => s_usingSignedAssertionFromVault(logger, signedAssertionUri, default!);
 
+            /// <summary>
+            /// Logger for handling information specific to ConfidentialClientApplicationBuilderExtension.
+            /// </summary>
+            /// <param name="logger">ILogger.</param>
```
keegan-caruso, line 143: This is an internal class? Either fill these out or remove them.

tests/E2E Tests/CustomSignedAssertionProviderTests/CustomSignedAssertionProviderExtensibilityTests.cs
```diff
@@ -0,0 +1,56 @@
...
+            {
+                // Act
+                _ = await authorizationHeaderProvider.CreateAuthorizationHeaderForAppAsync("https://graph.microsoft.com/.default");
+            }
+            catch (MsalServiceException MsalEx)
+            {
+                // Assert
+                Assert.Contains(expectedExceptionCode, MsalEx.Message, StringComparison.InvariantCulture);
+            }
+            catch (Exception ex)
```
keegan-caruso, line 50:
```suggestion
            catch (Exception ex) when (e is not XunitException)
```

Otherwise you will catch the exception from the failed assert and have a more confusing error

tests/E2E Tests/CustomSignedAssertionProviderTests/CustomSignedAssertionProviderTests.csproj
```diff
@@ -0,0 +1,41 @@
...
+    <TargetFrameworks>net9.0</TargetFrameworks>
+    <AssemblyOriginatorKeyFile>../../../build/MSAL.snk</AssemblyOriginatorKeyFile>
+    <Nullable>enable</Nullable>
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <None Remove="appsettings.json" />
+  </ItemGroup>
+
```
keegan-caruso, line 13:
```suggestion
```

tests/E2E Tests/CustomSignedAssertionProviderTests/appsettings.json
```diff
@@ -0,0 +1,13 @@
+{
+    "AzureAd": {
+        "Instance": "https://login.microsoftonline.com/",
+        "TenantId": "msidlab4.onmicrosoft.com",
+        "ClientId": "f6b698c0-140c-448f-8155-4aa9bf77ceba",
+        "ClientCredentials": [
+            {
+                "SourceType": "CustomSignedAssertion",
+                "CustomSignedAssertionProviderName": "MyCustomExtension"
```
keegan-caruso, line 9: I'm confused on this, the docs say this should be the FullName of the type, that should include the namespace, shouldn't it?

keegan-caruso: This lines up with the work item but the code comments contradict the behavior.

Can we fix this comment here: 

https://github.com/AzureAD/microsoft-identity-web/blob/ebc7128489ccdb52b70fcd60c413f4a502ea8e60/src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.CustomSignedAssertion.cs#L41

To match the implementation that it can be provider name or fully qualified type name?

src/Microsoft.Identity.Web.TokenAcquisition/ConfidentialClientApplicationBuilderExtension.Logger.cs
JoshLozensky: "ILogger." as the param explanation is consistent with how the other ones are done

keegan-caruso (approved):
Thanks for the quick iterations!

PR #3229, opened 2025-02-04
update benchmark

jennyf19:
benchmarks were not working #3230


jmprieur (approved):

benchmark/GraphServiceClientBenchmark.cs
```diff
@@ -0,0 +1,57 @@
...
+// Licensed under the MIT License.
+
+using BenchmarkDotNet.Attributes;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Identity.Web;
+using Microsoft.Graph;
+using Microsoft.Extensions.Configuration;
+using Microsoft.AspNetCore.Authentication.OpenIdConnect;
+
+#pragma warning disable CA1050 // Declare types in namespaces
```
keegan-caruso, line 11: Why not declare in the same ns as Program.cs ?daemon_console

benchmark/Program.cs
```diff
@@ -21,6 +21,7 @@ static void Main(string[] args)
             var benchmarkConfig = ManualConfig.Union(DefaultConfig.Instance, new BenchmarkConfig());
 #endif
             BenchmarkSwitcher.FromAssembly(typeof(Program).Assembly).Run(args, benchmarkConfig);
+            BenchmarkRunner.Run<GraphServiceClientBenchmark>();
```
keegan-caruso, line 24: This doesn't look right. BenchmarkSwitcher above should select the benchmark to run, we should run it directly like this.

jennyf19: yes...was just trying somethings out. I'll move this back to draft, for when I have time to come back to it. the benchmarks were not working for the other though

jennyf19: reverted some things

benchmark/GraphServiceClientBenchmark.cs
jennyf19: reverted

keegan-caruso (approved):

Issue #3240, opened 2025-02-11
[Feature Request] When AcquireTokenOptions.FmiPath is not null or empty, Token acquisition should call WithFmiPath()


jmprieur:
**Is your feature request related to a problem? Please describe.**
MSAL now exposes .WithFmiPath(string) on AcquireTokenForClientBuilder. We need to also expose it in the higher level APIs to support Federation Managed Identity.

**Describe the solution you'd like**
When AcquireTokenOptions.FmiPath is not null or empty and developers call (directly or through downstream API) the GetAuthenticationResultAsync method, IdWeb should call .WithFmiPath() on the ConfidentialClientApplicationBuilder, for instance [here](https://github.com/AzureAD/microsoft-identity-web/blob/56a434987c0b087f5a7ac54a0e4803ab617efde2/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs#L556)

**Describe alternatives you've considered**
None

**Additional context**
- See also https://github.com/AzureAD/microsoft-identity-abstractions-for-dotnet/issues/159 for the AcquireTokenOptions.FmiPath property
- For end to end testing work with @bgavrilMS (this is not requested yet part of this work item)

PR #3247, opened 2025-02-19
Lozensky/add fmi path

JoshLozensky:
For testing, will be working with Bogdan/Travis on E2E testing as this isn't really approachable via unit tests.

Since MSAL made `AcquireTokenForClient.WithFmiPath()` available, in the one place we use the AcquireTokenForClient method, adding a check for and insertion of the FMI Path so Id Web customers can make use of this new functionality. Addresses issue #3240

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -86,10 +86,8 @@
   </PropertyGroup>
 
   <PropertyGroup Label="Common dependency versions">
-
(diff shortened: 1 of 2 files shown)


jmprieur (commented):
This is a great start

jmprieur (approved):
LGTM
Thanks, @JoshLozensky 

cc: @bgavrilMS

kllysng (approved):

PR #3261, opened 2025-02-25
Fix authority parsing logic

JoshLozensky:
This pull request changes logic in the `ParseAuthorityIfNecessary` method in the `MergedOptions` class to fix a bug where the check on the `indexTenant` always evaluates to true.

Originally, there had been a bug here because the '/' chars in the scheme prefix of the URI were not taken into account leading to false positives when checking for the presence of the instance/tenantId

[This PR](https://github.com/AzureAD/microsoft-identity-web/commit/0fdfc96216cb6828f462b85bca21842fc9664915) went into Id Web in 3.1.0 and attempted to address the above bug but accidentally created another where the `indexTenant` >=0 condition always evaluated to true as its min possible value was 7.

The current PR fixes this by making the starting index of the authority URI scheme-agnostic and uses that dynamic value to compare against the `indexTenant` variable.

### DIFF
--- a/src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
@@ -442,9 +442,10 @@ namespace Microsoft.Identity.Web
         {
             if (string.IsNullOrEmpty(mergedOptions.TenantId) && string.IsNullOrEmpty(mergedOptions.Instance) && !string.IsNullOrEmpty(mergedOptions.Authority))
             {
-                ReadOnlySpan<char> httpsPrefix = "https://".AsSpan();
+                ReadOnlySpan<char> doubleSlash = "//".AsSpan();
                 ReadOnlySpan<char> authoritySpan = mergedOptions.Authority.AsSpan().TrimEnd('/');
-                int startingIndex = authoritySpan.StartsWith(httpsPrefix) ? httpsPrefix.Length : 0;
+                int doubleSlashIndex = authoritySpan.IndexOf(doubleSlash);
+                int startingIndex = doubleSlashIndex == -1 ? 0 : doubleSlashIndex + doubleSlash.Length;
 
                 // Gets position of first '/' after the "https://" prefix, will return -1 if not found, this checks for presence of instance/tenantId
                 int indexTenant = authoritySpan.Slice(startingIndex).IndexOf('/') + startingIndex;
--- a/src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
@@ -442,11 +442,13 @@ namespace Microsoft.Identity.Web
         {
             if (string.IsNullOrEmpty(mergedOptions.TenantId) && string.IsNullOrEmpty(mergedOptions.Instance) && !string.IsNullOrEmpty(mergedOptions.Authority))
             {
-                const int StartingIndex = 8; // length of "https://"
+                ReadOnlySpan<char> httpsPrefix = "https://".AsSpan();
                 ReadOnlySpan<char> authoritySpan = mergedOptions.Authority.AsSpan().TrimEnd('/');
+                int startingIndex = authoritySpan.StartsWith(httpsPrefix) ? httpsPrefix.Length : 0;
 
-                int indexTenant = authoritySpan.Slice(StartingIndex).IndexOf('/') + StartingIndex;
-                if (indexTenant >= 0)
+                // Gets position of first '/' after the "https://" prefix, will return -1 if not found, this checks for presence of instance/tenantId
+                int indexTenant = authoritySpan.Slice(startingIndex).IndexOf('/') + startingIndex;
+                if (indexTenant >= startingIndex)
                 {
                     int indexVersion = authoritySpan.Slice(indexTenant + 1).IndexOf('/');
                     int indexEndOfTenant = indexVersion == -1 ? authoritySpan.Length : indexVersion + indexTenant + 1;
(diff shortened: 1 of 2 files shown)


jmprieur (commented):
Thanks @JoshLozensky 
We probably need to add a unit test that would fail before the bug was fixed?

src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
```diff
@@ -442,11 +442,13 @@ internal static void ParseAuthorityIfNecessary(MergedOptions mergedOptions)
         {
             if (string.IsNullOrEmpty(mergedOptions.TenantId) && string.IsNullOrEmpty(mergedOptions.Instance) && !string.IsNullOrEmpty(mergedOptions.Authority))
             {
-                const int StartingIndex = 8; // length of "https://"
+                ReadOnlySpan<char> doubleSlash = "//".AsSpan();
                 ReadOnlySpan<char> authoritySpan = mergedOptions.Authority.AsSpan().TrimEnd('/');
+                int startingIndex = authoritySpan.IndexOf(doubleSlash) == -1 ? 0 : authoritySpan.IndexOf(doubleSlash) + doubleSlash.Length;
```
ksaaf, line 447: Call IndexOf() once.

```suggestion
                int doubleSlashIndex = authoritySpan.IndexOf(doubleSlash)
                int startingIndex = doubleSlashIndex == -1 ? 0 : doubleSlashIndex + doubleSlash.Length;
```

src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
```diff
@@ -442,11 +442,13 @@ internal static void ParseAuthorityIfNecessary(MergedOptions mergedOptions)
         {
             if (string.IsNullOrEmpty(mergedOptions.TenantId) && string.IsNullOrEmpty(mergedOptions.Instance) && !string.IsNullOrEmpty(mergedOptions.Authority))
             {
-                const int StartingIndex = 8; // length of "https://"
+                ReadOnlySpan<char> doubleSlash = "//".AsSpan();
```
ksaaf, line 445: Would we be better off declaring double slash as a private const and use it?

bgavrilMS, line 445: Authority is an URI. I wonder if just using the URI class (which will perform some parsing on its own) will suffice here?

src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
```diff
@@ -442,11 +442,13 @@ internal static void ParseAuthorityIfNecessary(MergedOptions mergedOptions)
         {
```
bgavrilMS, line 442: What is the bug that you are trying to fix @JoshLozensky ? Higher level SDKs should not be parsing the authority at all, that is MSAL's responsibility.

src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
JoshLozensky: A ReadOnlySpan cannot be declared as a const and the IndexOf method will only accept a ReadOnlySpan

src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
JoshLozensky: good point thanks

src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
JoshLozensky: @bgavrilMS was something unclear in the description of the PR explaining the bug that I should alter? 

In this case the parsing is only happening when there is no Instance or Tenant Id defined by the user but there is an authority the parsing is to pull those values out and add them to the merged options. Does MSAL have a way to spit out Instance and Tenant Id when given an authority?

src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
JoshLozensky: The parsing is to get the tenant Id and instance values out of the authority, I don't see a way to do that using the URI class while maintaining the performance of the ReadOnlySpan, open to suggestions though :)

JoshLozensky:
> Thanks @JoshLozensky We probably need to add a unit test that would fail before the bug was fixed?

Added :)

src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
bgavrilMS: How is it going to work for a dSTS authority like ? Is it a relevant scenario?

`https://some.url.dsts.core.azure-test.net/dstsv2/72f988bf-86f1-41af-91ab-2d7cd011db47`

src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
bgavrilMS: I understood the problem better after you added tests.

src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
bgavrilMS: In any case, not blocking this PR.

jennyf19: I think this was discussed offline, what's the resolution?

bgavrilMS: The conclusion is that there should only be 1 place where we parse authorities, probably in MSAL. It's tricky business. But I am not blocking this PR on this very precise fix.

JoshLozensky: Agreed when discussed offline with Jenny, this will only be hit when users don't define their tenant id so if a case isn't covered here all they have to do is define it in their config and it won't be a problem.

westin-m (approved):

PR #3273, opened 2025-02-28
Add Support for Custom Saml Bearer in HttpRequest Headers

### LINKED ISSUE #3269: [Feature Request] Support for using SAML bearer tokens to call Downstream API using Headers.TryAddWithoutValidation()
Currently, the library uses `httpRequestMessage.Headers.Add(Authorization, authorizationHeader)` in `DownstreamApi.cs` for adding the authorization header. However, this does not work when adding a SAML bearer token, where it would be required to use `httpRequestMessage.Headers.TryAddWithoutValidation(Authorization, authorizationHeader);`

In `UpdateRequestAsync()` method of `DownstreamApi.cs`, a check can be added to identify if the authorizationHeader contains saml-bearer. If so httpRequestMessage.Headers.TryAddWithoutValidation(Authorization, authorizationHeader); can be used instead of httpRequestMessage.Headers.Add(Authorization, authorizationHeader) for adding the token in the header

```csharp
if(authorizationHeader.StartsWith("Identifier for saml bearer token"))
{
httpRequestMessage.Headers.TryAddWithoutValidation(Authorization, authorizationHeader);
}
else
{
httpRequestMessage.Headers.Add(Authorization, authorizationHeader)
}
```

**Additional context**
This feature is needed for scenarios where SAML tokens is used to call the Downstream API in OBO flows

sthanu98:
# Add Support for Saml Authorization Scheme in HttpRequest Headers

## Description

This PR introduces a change to the way authorization headers are added to HTTP requests, specifically to support SAML tokens. The change ensures that custom SAML authorization schemes are correctly added using TryAddWithoutValidation. This approach allows for more flexible handling of SAML tokens, even when they might not pass standard validation.

Changes:

Replaced the below line to improve compatibility with custom SAML authorization schemes.
```csharp 
httpRequestMessage.Headers.Add(Authorization, authorizationHeader);
``` 
with 
```csharp
httpRequestMessage.Headers.TryAddWithoutValidation(Authorization, authorizationHeader);
```
 
Added unit tests to validate the new behavior and ensure headers are correctly added in various scenarios.

Fixes #3269

### DIFF
--- a/src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
+++ b/src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
@@ -522,7 +522,7 @@ namespace Microsoft.Identity.Web
                        user,
                        cancellationToken).ConfigureAwait(false);
 
-                httpRequestMessage.Headers.Add(Authorization, authorizationHeader);
+                httpRequestMessage.Headers.TryAddWithoutValidation(Authorization, authorizationHeader);
             }
             else
             {
--- a/src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
+++ b/src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
@@ -522,6 +522,7 @@ namespace Microsoft.Identity.Web
                        user,
                        cancellationToken).ConfigureAwait(false);
 
+                // The TryAddWithoutValidation method is used to bypass the strict checks on format of Authorization header
                 httpRequestMessage.Headers.TryAddWithoutValidation(Authorization, authorizationHeader);
             }
             else
--- a/src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
+++ b/src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
@@ -28,6 +28,7 @@ namespace Microsoft.Identity.Web
         private readonly IOptionsMonitor<DownstreamApiOptions> _namedDownstreamApiOptions;
         private const string Authorization = "Authorization";
         protected readonly ILogger<DownstreamApi> _logger;
+        private const string AuthSchemeDstsSamlBearer = "http://schemas.microsoft.com/dsts/saml2-bearer";
 
         /// <summary>
         /// Constructor.
@@ -522,8 +523,15 @@ namespace Microsoft.Identity.Web
                        user,
                        cancellationToken).ConfigureAwait(false);
 
-                // The TryAddWithoutValidation method is used to bypass the strict checks on format of Authorization header
-                httpRequestMessage.Headers.TryAddWithoutValidation(Authorization, authorizationHeader);
+                if (authorizationHeader.StartsWith(AuthSchemeDstsSamlBearer, StringComparison.OrdinalIgnoreCase))
+                {
+                    // TryAddWithoutValidation method bypasses strict validation, allowing non-standard headers to be added for custom Header schemes that cannot be parsed.
+                    httpRequestMessage.Headers.TryAddWithoutValidation(Authorization, authorizationHeader);
+                }
+                else
+                {
+                    httpRequestMessage.Headers.Add(Authorization, authorizationHeader);
+                }
             }
             else
             {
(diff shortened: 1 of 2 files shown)


sthanu98:
> @sthanu98 please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

@microsoft-github-policy-service agree company="Microsoft"

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
```diff
@@ -522,7 +522,7 @@ public Task<HttpResponseMessage> CallApiForAppAsync(
                        user,
                        cancellationToken).ConfigureAwait(false);
 
-                httpRequestMessage.Headers.Add(Authorization, authorizationHeader);
+                httpRequestMessage.Headers.TryAddWithoutValidation(Authorization, authorizationHeader);
```
sruke, line 525: TryAddWithoutValidation returns false if the header cannot be added. Should an exception be thrown in this case?

pmaytak:
Couple of questions:
- What exactly is the error now when using the SAML auth header?
- What validation is skipped when using the new method?
- Looks like there's property NonValidated for such headers, does HTTP client treat those headers differently, will it cause issues E2E?
- Is there a way to add the SAML header value as valid to the HTTP client?
  - I am a bit skeptical on turning off validation for all header values. This means we should make sure we have proper test coverage that makes sure the header creator doesn't create invalid headers somehow (also, for ex. if we add a new header provider implementation).

sthanu98:
@pmaytak 

1. System.FormatException: The format of value 'http://schemas.microsoft.com/dsts/saml2-bearer token-redacted' is invalid.
         at System.Net.Http.Headers.HttpHeaderParser.ParseValue(String value, Object storeValue, Int32& index)
         at System.Net.Http.Headers.HttpHeaders.ParseAndAddValue(HeaderDescriptor descriptor, HeaderStoreItemInfo info, String value)
         at System.Net.Http.Headers.HttpHeaders.Add(HeaderDescriptor descriptor, String value)
         at ActorServer_WebAPI_MISE.Controllers.HelloWorldController.GetAsync() in C:\repos\Microsoft.Identity.Client.Internal\samples\SamlOBO\ActorServer_WebAPI_MISE\Controllers\HelloWorldController.cs:line 61
 
This is the error when using Add() instead of TryAddWithoutValidation()

2. HttpHeaders.ParseAndAddValue method is skipped in TryAddWithoutValidation()
3. the NonValidated headers are treated the same as validated headers when sending requests in HttpClient.
4. The SAML header value we are passing fails the ParseAndAddValue validation, so we cannot make it valid for httpclient. If you are skeptical of this change affecting existing JWT scenarios, we can change to either of the below implementations

```csharp
if(authorizationHeader.StartsWith("https://schemas.microsoft.com/saml-bearer"))
{
httpRequestMessage.Headers.TryAddWithoutValidation(Authorization, authorizationHeader);
}
else
{
httpRequestMessage.Headers.Add(Authorization, authorizationHeader)
}
```

or 

```csharp
try
{
httpRequestMessage.Headers.Add(Authorization, authorizationHeader)
}
catch(FormatException)
{
httpRequestMessage.Headers.TryAddWithoutValidation(Authorization, authorizationHeader);
}
```

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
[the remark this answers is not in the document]
sthanu98: done

sthanu98: @sruke TryAddWithoutValidation removes some validations done by the Add method. So essentially all the headers that were previously allowed should still be allowed. Given that there were no exceptions thrown before, should these be added now?

pmaytak (approved):

jmprieur (approved):
LGTM
Thanks @sthanu98

PR #3274, opened 2025-02-28
Locking Down the TokenAcquirerFactory

JoshLozensky:
As demonstrated by the unit tests in this PR failing when the locks (also added here) are removed, it is possible in the currently released version to get multiple threads of execution through the `TokenAcquirerFactory.GetDefaultInstance` and `TokenAcquirerFactory.Build` methods when called concurrently. 

These methods perform setup of services critical to the library's functionality so protecting them from concurrent access is essential to avoiding unexpected behavior.

Additionally, due to the static, singleton nature of GetDefaultInstance this PR adds all test classes using the method to a collection to prevent them from being run in parallel.

### DIFF
--- a/.github/workflows/dotnetcore.yml
+++ b/.github/workflows/dotnetcore.yml
@@ -42,14 +42,14 @@ jobs:
     - name: Build solution
       run: msbuild Microsoft.Identity.Web.sln -r -t:build -verbosity:m -property:Configuration=Release
 
-    - name: Test with .NET 6.0.x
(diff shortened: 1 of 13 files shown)


jennyf19:
@JoshLozensky can you please clean up the commits? If you need help, lmk.

jmprieur (approved):
LGTM
Thanks @JoshLozensky

jennyf19 (approved):
:shipit:

PR #3278, opened 2025-03-03
Catch exception precisely, and change an api url

rayluo:
# {PR title}

Summary of the changes (Less than 80 chars):

Before this change, the `ViewData["json"]` was not populated.


tests/DevApps/WebAppCallsMicrosoftGraph/Pages/Index.cshtml.cs
```diff
@@ -29,30 +29,37 @@ public IndexModel(ILogger<IndexModel> logger, GraphServiceClient graphServiceCli
...
-         
+            try
+            {
+                var user = await _graphServiceClient.Me.GetAsync(r =>
+                r.Options.WithScopes("user.read")
+                //.WithUser(User)
+                );
+                ViewData["name"] = user.DisplayName;
+            }
+            catch (Exception)
```
jmprieur, line 40: actually no, you should not do that. as there is an attribute to catch the exception.
Did you see the `[AuthorizeForScopes(Scopes = new[] { "user.read" })]` attribute, @rayluo ?

See [Managing incremental consent and conditional access](https://github.com/AzureAD/microsoft-identity-web/wiki/Managing-incremental-consent-and-conditional-access) for details

jennyf19 (changes requested):
blocking due to JM's comment

tests/DevApps/WebAppCallsMicrosoftGraph/Pages/Index.cshtml.cs
rayluo: It is true that the sample still works as-is without this part of change.

The reason I made this change is because I ran into an exception in `await _graphServiceClient.Me.GetAsync(...)` *after* I tentatively changed the "BaseUrl" in appsettings.json from `https://graph.microsoft.com/beta` to `https://graph.microsoft.com/v1.0/me`. Presumably the `_graphServiceClient.Me.GetAsync(...)` is only expected to work with that MS Graph beta endpoint. So, this change will make the sample more robust.

rayluo: @jmprieur , if we take the context above into consideration, would you suggest wrapping this code path with exception handling, or not?

jmprieur:
@rayluo : let's start to raise a GitHub issue with the customer problem (and repro steps)?
Then we can decide. For the moment I don't understand what the problem is

jmprieur:
What happens if you close the app to empty the msal cache and restart it?
Is the user asked to re signing?

PR #3283, opened 2025-03-05
Add singleton protection for test class using TokenAcquirerFactory

JoshLozensky:
Since there was a second class in this test file it was missed when adding the singleton protection collection definition. Added the definition and also the call to reset the TokenAcquirerFactory's default instance in the test itself.

### DIFF
--- a/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquirerFactory.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquirerFactory.cs
@@ -215,11 +215,14 @@ namespace Microsoft.Identity.Web
         /// </summary>
         internal /* for unit tests */ static void ResetDefaultInstance()
         {
-            if (defaultInstance?.ServiceProvider != null)
+            lock (s_defaultInstanceLock)
             {
-                (defaultInstance.ServiceProvider as IDisposable)?.Dispose();
+                if (defaultInstance?.ServiceProvider != null)
+                {
+                    (defaultInstance.ServiceProvider as IDisposable)?.Dispose();
+                }
+                defaultInstance = null;
             }
-            defaultInstance = null;
         }
 
         // Move to a derived class?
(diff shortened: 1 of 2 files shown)


src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquirerFactory.cs
```diff
@@ -215,11 +215,14 @@ protected virtual void PreBuild()
         /// </summary>
         internal /* for unit tests */ static void ResetDefaultInstance()
         {
-            if (defaultInstance?.ServiceProvider != null)
+            lock (s_defaultInstanceLock)
             {
-                (defaultInstance.ServiceProvider as IDisposable)?.Dispose();
+/*                if (defaultInstance?.ServiceProvider != null)
```
jmprieur, line 220: Why is it commented out?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquirerFactory.cs
```diff
@@ -215,11 +215,14 @@ protected virtual void PreBuild()
         /// </summary>
         internal /* for unit tests */ static void ResetDefaultInstance()
         {
-            if (defaultInstance?.ServiceProvider != null)
+            lock (s_defaultInstanceLock)
```
jmprieur, line 218: Is it needed?

jmprieur (commented):
I don't understand

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquirerFactory.cs
JoshLozensky: It isn't required but I was thinking it might mitigate some race conditions in parallel tests by making sure a defaultInstance isn't made null while still doing work in the constructor

JoshLozensky:
> @JoshLozensky : I don't understand what problem this is solving?

@jmprieur trying to fix the Id Web build failures

jmprieur (commented):
I think it's fixed already. Look in master the last PRs

PR #3289, opened 2025-03-07
Disable the CIAM UI test as the kmsi interupt takes too long and Plaright times out

jmprieur:
# Disable the CIAM UI test as the kmsi interupt takes too long and Plaright times out


jennyf19 (approved):
LGTM
:shipit:

PR #3291, opened 2025-03-07
Post release cleanup

jmprieur:
# Post release 3.8.0 cleanup
- Reset the public API Unshipped.txt files, and update the Shipped ones
- Update the IDWeb next version to 3.8.1 for the moment
- Enable Package Validation for the new Microsoft.Identity.Web.FicOidc package and sets its baselineVersion to 3,8,0 as its it's first version.
-

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <!--This should be passed from the VSTS build-->
     <!-- This needs to be greater than or equal to the validation baseline version -->
-    <MicrosoftIdentityWebVersion Condition="'$(MicrosoftIdentityWebVersion)' == ''">3.8.0</MicrosoftIdentityWebVersion>
+    <MicrosoftIdentityWebVersion Condition="'$(MicrosoftIdentityWebVersion)' == ''">3.8.1</MicrosoftIdentityWebVersion>
     <!--This will generate AssemblyVersion, AssemblyFileVersion and AssemblyInformationVersion-->
     <Version>$(MicrosoftIdentityWebVersion)</Version>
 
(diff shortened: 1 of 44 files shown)


Directory.Build.props
```diff
@@ -2,7 +2,7 @@
   <PropertyGroup>
     <!--This should be passed from the VSTS build-->
     <!-- This needs to be greater than or equal to the validation baseline version -->
-    <MicrosoftIdentityWebVersion Condition="'$(MicrosoftIdentityWebVersion)' == ''">3.8.0</MicrosoftIdentityWebVersion>
+    <MicrosoftIdentityWebVersion Condition="'$(MicrosoftIdentityWebVersion)' == ''">3.8.1</MicrosoftIdentityWebVersion>
```
mdchennu, line 5: Since version 3.8.0 is being released, why would this be 3.8.1? Not a blocking comment but I was just wondering!

mdchennu (approved):

Directory.Build.props
sruke: The patch part of major.minor.patch version is incremented to one higher than the released version so that any future previews built are of a higher version.

sruke (approved):

sruke (commented):
Consider updating IdentityModel version to 8.6.1. See [here](https://github.com/AzureAD/microsoft-identity-web/blob/06b2954cdd2887bf752b65bf43daa7f43525f5cd/Directory.Build.props#L89C5-L89C126)

Directory.Build.props
mdchennu: answered by sruthi, so resolving comment

PR #3299, opened 2025-03-14
Support FIC with AT_POP

sruke:


src/Microsoft.Identity.Web.TokenAcquisition/MsAuth10AtPop.cs
```diff
@@ -28,19 +29,43 @@ internal static AcquireTokenForClientParameterBuilder WithAtPop(
...
+        {
+            _ = Throws.IfNull(popPublicKey);
+            _ = Throws.IfNull(jwkClaim);
 
-                 data.BodyParameters.Remove("client_assertion");
-                 data.BodyParameters.Add("request", signedAssertion);
+            builder.WithProofOfPosessionKeyId(popPublicKey);
+            builder.OnBeforeTokenRequest((data) =>
+            {
+                string? signedAssertion = credentialDescription.CachedValue as string;
```
jmprieur, line 62: I don't understand that. Why not let MSAL use the client assertion that was provided with the delegate at the construction of the confidential client application instance. I thought that the whole point of the new MS Auth 1.0 at pop was to be able to allow for FIC etc ?

Are lines 62 and 63 really needed? and also do we need to pass-in the `credentialDescription` ?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -629,6 +629,39 @@ public async Task<AuthenticationResult> GetAuthenticationResultForAppAsync(
...
 
+        private static void ConfigureAtPopWithCredentials(
+            AcquireTokenForClientParameterBuilder builder,
+            IAppConfig appConfig,
+            TokenAcquisitionOptions tokenAcquisitionOptions,
+            IEnumerable<CredentialDescription>? clientCredentials,
+            bool sendX5C)
+        {
+            // Try to configure AtPop with custom signed assertion first
+            if (clientCredentials != null)
```
jmprieur, line 640: It sounds like it's redoing what IdWeb does.

jmprieur (changes requested):
I don't think that these are the right changes
(or I miss something)

src/Microsoft.Identity.Web.TokenAcquisition/MsAuth10AtPop.cs
```diff
@@ -1,85 +1,27 @@
...
-            };
-
-            if (sendX5C)
+            AtPopOperation op = new AtPopOperation(popPublicKey, jwkClaim);
+            builder.WithAuthenticationExtension(new MsalAuthenticationExtension()
             {
-                string x5cValue = Convert.ToBase64String(certificate.GetRawCertData());
-                securityTokenDescriptor.AdditionalHeaderClaims = 
-                    new Dictionary<string, object>() { { "x5c", new List<string> { x5cValue } } };
-            }
```
sruke, line 77: Will removing this section cause any problems?

src/Microsoft.Identity.Web.TokenAcquisition/MsAuth10AtPop.cs
sruke: Removed custom signed assertion configuration.
It will now be set up via delegate during CCA initialization.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
sruke: Removed custom signed assertion configuration.
It will now be set up via delegate during CCA initialization.

src/Microsoft.Identity.Web.TokenAcquisition/MsAuth10AtPop.cs
jmprieur: I should not. If the client credentials are a certificate and the customer set SendX5C, will will be sent.

jmprieur (approved):
LGTM
Thanks @sruke

jennyf19 (approved):
:shipit: 
approve w/small suggestions

src/Microsoft.Identity.Web.TokenAcquisition/AtPopOperation.cs
```diff
@@ -0,0 +1,42 @@
...
+
+namespace Microsoft.Identity.Web
+{
+    internal class AtPopOperation : IAuthenticationOperation
+    {
+        private readonly string _reqCnf;
+
+        public AtPopOperation(string keyId, string reqCnf)
+        {
+            KeyId = keyId;
```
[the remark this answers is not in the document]
sruke: Good point, added tests.

PR #3326, opened 2025-04-08
Added .Net10 preview

### LINKED ISSUE #3300: [Feature Request] Update to .NET 10 Preview 3
.NET 10 can be downloaded from [here](https://dotnet.microsoft.com/en-us/download/dotnet/10.0). You will need the latest Visual Studio Preview to run the project with this version.

To-do:
- Add .NET 10 Preview 3 to targets for this project.
- Resolve any new analyzer warnings and errors.
- Confirm all tests pass.
- Update the build templates to use this .NET version.
- In some projects and build pipelines, we have `TargetNet9` conditional build variable. Rename this to `TargetNetNext` so it can be reused for subsequent updates more easily.
- Check [.NET 10 breaking changes](https://learn.microsoft.com/en-us/dotnet/core/compatibility/10.0) for anything that may affect our code.

mdchennu:
#Added .Net10 preview to dotnetcore.yml 

- [X] You've read the [Contributor Guide](https://github.com/AzureAD/microsoft-identity-web/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://github.com/AzureAD/microsoft-identity-web/blob/master/CODE_OF_CONDUCT.md).
- [X] You've included unit or integration tests for your change, where applicable.
- [X] You've included inline docs for your change, where applicable.
- [X] There's an open issue for the PR that you are making. If you'd like to propose a new feature or change, please open an issue to discuss the change or find an existing issue.

## Description

Updated the dotnetcore.yaml file to include a preview version of .Net10

Fixes #3300  (in this specific format)


.github/workflows/dotnetcore.yml
```diff
@@ -26,6 +26,11 @@ jobs:
     - name: Checkout repository
       uses: actions/checkout@v4
 
+    - name: Setup .NET 10.0.x
+      uses: actions/setup-dotnet@v2.1.0
+      with:
+        dotnet-version: 10.0.100-preview.2.24157.14
```
mdchennu, line 32: note for me: need to update to the correct version of dotnet

.github/workflows/dotnetcore.yml
```diff
@@ -26,6 +26,11 @@ jobs:
     - name: Checkout repository
       uses: actions/checkout@v4
 
+    - name: Setup .NET 10.0.x
+      uses: actions/setup-dotnet@v4
+      with:
+        dotnet-version: 10.0.100-preview.2
```
mdchennu, line 32: I thought we were supposed to use 10.0.100-preview.3, but I only see 10.0.100-preview.2 here: https://dotnet.microsoft.com/en-us/download/dotnet/10.0

Which one should I use?

jmprieur: you can't use what is not available :-)

jennyf19: You might have to specify this: 10.0.100-preview.2.25164.34 

https://dotnet.microsoft.com/en-us/download/dotnet/10.0

Directory.Build.props
```diff
@@ -104,6 +104,23 @@
     <PublicApiAnalyzersVersion>3.3.4</PublicApiAnalyzersVersion>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net10.0'">
+    <NetNineRuntimeVersion>10.0.0</NetNineRuntimeVersion>
```
jennyf19, line 108: these need to be on the respective preview versions from the link to the .NET 10 download page

Directory.Build.props
```diff
@@ -104,6 +104,23 @@
     <PublicApiAnalyzersVersion>3.3.4</PublicApiAnalyzersVersion>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(TargetFramework)' == '10.0.0-preview.3'">
```
keegan-caruso, line 107: I would think this would be "net10.0-preview3"?

jennyf19: right

mdchennu: updated!

build/template-postbuild-code-analysis.yaml
```diff
@@ -4,13 +4,15 @@
 steps:
 - task: securedevelopmentteam.vss-secure-development-tools.build-task-roslynanalyzers.RoslynAnalyzers@3
   displayName: 'Run Roslyn Analyzers'
+  condition: eq(variables['TargetNet10'], 'False')
```
mdchennu, line 7: does this condition need to be added? is `TargetNet10` the right value to use?

keegan-caruso: Not quite, "TargetNetNext"

keegan-caruso: Hopefully we can use copilot for this part, the way this should be:

- controlled by "TargetNetNext"
- If true, then use one variable for frameworks that includes 10
- If false, then use another variable that doesn't include 10
- then when we upgrade versions, all we do is update the two variables

mdchennu:
I've been trying to build the projects as I update the `.csproj` files. 

When I try to build and run the tests, I see the following error that I need Visual Studio version 17.16 or higher:
image

I currently have Visual Studio Version 17.14.0 Preview 2.0 installed. 

So I tried:

1. Googling online for 17.16 -> lol couldn't find it
2. Checking my Visual Studio Installer to see is 17.16 preview is available -> it wasn't
3. Searching in the developer community (link here: https://developercommunity.visualstudio.com/VisualStudio?q=+Visual+Studio+version+17.16+&fTime=6m)

I couldn't find it anywhere :/. Is there something I'm missing?

build/template-postbuild-code-analysis.yaml
mdchennu: okay, I'm starting to understand! These are the updates I've made: 

- Directory.Build.Props -> added in two variables for frameworks -> one for when "TargetNetNext" is true (which includes net10.0) and one for when "TargetNetNext" is false 
- in the `build/template-postbuild-code-analysis.yaml` -> I removed the reference to TargetNetNext, because I think we want the analyzers to run whether we are on a preview version or not. 
- - in `'build\template-install-dependecies.yaml` -> im checking whther `TargetNetNext` is set to true before using the 10.0 preview version

is this more on track?

mdchennu:
Parking this PR until a newer version of .NET10 has been released.

pmaytak:
VS 2022 Preview 6 is released. Maybe it has better support for this .NET version.

PR #3339, opened 2025-04-19
Linux friendly devcontainer

keegan-caruso:
# Linux friendly devcontainer

## Description

Include a slnf to build SDK style projects that are on netcore.

- Can only build SDK style projects


.devcontainer/scripts/container-creation.sh
```diff
@@ -2,6 +2,8 @@
 
 set -e
 
+dotnet restore
```
keegan-caruso, line 5: We need the solution to be restore first, so move that to setup.

.vscode/settings.json
```diff
@@ -5,5 +5,6 @@
       "*.props": "xml",
       "*.targets": "xml",
       "*.tasks": "xml"
-    }
+    },
+    "dotnet.preferCSharpExtension": true
```
keegan-caruso, line 9: cSharp extension only supports SDK style projects, this is fine since we can only build SDK style on linux

Directory.Build.props
```diff
@@ -213,4 +213,4 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
     </PackageReference>
   </ItemGroup>
-</Project>
+</Project>
```
keegan-caruso, line 216: auto fixed, but this is correct, file should end with a newline

jmprieur (approved):
Lgtm
Thanks @keegan-caruso

westin-m (approved):

PR #3361, opened 2025-05-14
Add SigningCredential loading to DefaultCredentialsLoader

RojaEnnam:
## Description

Add SigningCredential loading to DefaultCredentialsLoader

Fixes #{3359}

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -95,7 +95,7 @@
     <AzureSecurityKeyVaultCertificatesVersion>4.6.0</AzureSecurityKeyVaultCertificatesVersion>
     <MicrosoftGraphVersion>4.36.0</MicrosoftGraphVersion>
     <MicrosoftGraphBetaVersion>4.57.0-preview</MicrosoftGraphBetaVersion>
-    <MicrosoftIdentityAbstractionsVersion>9.0.0</MicrosoftIdentityAbstractionsVersion>
+    <MicrosoftIdentityAbstractionsVersion>9.1.0</MicrosoftIdentityAbstractionsVersion>
     <!--CVE-2024-43485-->
     <SystemTextJsonVersion>8.0.5</SystemTextJsonVersion>
     <!--CVE-2023-29331-->
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -3,5 +3,6 @@
   <packageSources>
     <clear />
     <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
+    <add key="Local" value="C:\repos\Builds\abstractions" />
   </packageSources>
 </configuration>
--- a/docs/blog-posts/symmetric-key-support.md
+++ b/docs/blog-posts/symmetric-key-support.md
@@ -11,6 +11,38 @@ This proposal outlines the addition of symmetric key support for signing credent
 3. Follow existing patterns in the codebase
 4. Maintain backward compatibility
 
+## Developer Experience
+The implementation provides a straightforward and type-safe approach to working with symmetric keys while maintaining clean separation of concerns:
+
+### Key Management
+When working with symmetric keys, developers can utilize two primary sources:
+
+1. **Azure Key Vault Integration**
+   ```csharp
+   var credentials = new CredentialDescription
+   {
+       SourceType = CredentialSource.SymmetricKeyFromKeyVault,
+       KeyVaultUrl = "https://your-vault.vault.azure.net",
+       KeyVaultSecretName = "your-secret-name"
+   };
+   ```
+
+2. **Direct Base64 Encoded Keys**
+   ```csharp
+   var credentials = new CredentialDescription
+   {
+       SourceType = CredentialSource.SymmetricKeyBase64Encoded,
+       Base64EncodedValue = "your-base64-encoded-key"
+   };
+   ```
+
+### Implementation Details
+- The DefaultCredentialLoader automatically selects the appropriate loader based on the SourceType
+- Key material is loaded and converted to a SymmetricSecurityKey
+- The security key is stored in the CachedValue property of CredentialDescription
+- This design maintains independence from Microsoft.IdentityModel types in the abstractions layer
+- The implementation follows the same pattern as certificate handling for consistency
+
 ## Design
 
 ### 1. New CredentialSource Values(Abstractions Layer)
--- a/docs/blog-posts/symmetric-key-support.md
+++ b/docs/blog-posts/symmetric-key-support.md
@@ -1,0 +1,188 @@
+# Adding Support for Symmetric Keys in Microsoft.Identity.Web
+
+## Overview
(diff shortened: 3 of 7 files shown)


jmprieur:
Thanks @RojaEnnam
It's a good start. You might want to create a new interface

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
```diff
@@ -116,6 +117,31 @@ public async Task LoadCredentialsIfNeededAsync(CredentialDescription credentialD
...
+            try
+            {
+                await LoadCredentialsIfNeededAsync(credentialDescription, parameters);
+
+                if (credentialDescription.Certificate != null)
+                {
+                    return new X509SigningCredentials(credentialDescription.Certificate, credentialDescription.Algorithm);
+                }
+
+                return null;
```
RojaEnnam, line 136: Returning null if Certificate is not loaded because a null certificate is a valid state. Open to suggestions.

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
```diff
@@ -116,6 +117,31 @@ public async Task LoadCredentialsIfNeededAsync(CredentialDescription credentialD
...
+        CredentialDescription credentialDescription,
+        CredentialSourceLoaderParameters? parameters = null)
+        {
+            _ = Throws.IfNull(credentialDescription);
+
+            try
+            {
+                await LoadCredentialsIfNeededAsync(credentialDescription, parameters);
+
+                if (credentialDescription.Certificate != null)
```
RojaEnnam, line 131: This shows creating the SigningCredentials only from the certificate. I've included the proposal to add support for loading SymmetricSecurityKey in this PR.

Directory.Build.props
```diff
@@ -95,7 +95,7 @@
     <AzureSecurityKeyVaultCertificatesVersion>4.6.0</AzureSecurityKeyVaultCertificatesVersion>
     <MicrosoftGraphVersion>4.36.0</MicrosoftGraphVersion>
     <MicrosoftGraphBetaVersion>4.57.0-preview</MicrosoftGraphBetaVersion>
-    <MicrosoftIdentityAbstractionsVersion>9.0.0</MicrosoftIdentityAbstractionsVersion>
+    <MicrosoftIdentityAbstractionsVersion>9.1.0</MicrosoftIdentityAbstractionsVersion>
```
jmprieur, line 98: It's not published, @RojaEnnam FYI

NuGet.Config
```diff
@@ -3,5 +3,6 @@
   <packageSources>
     <clear />
     <add key="NuGet" value="https://api.nuget.org/v3/index.json" />
+    <add key="Local" value="C:\repos\Builds\abstractions" />
```
jmprieur, line 6: No longer needed

docs/blog-posts/symmetric-key-support.md
```diff
@@ -0,0 +1,220 @@
...
+    SymmetricKeyFromKeyVault = 6,
+    SymmetricKeyBase64Encoded = 7
+}
+```
+
+### 2. SymmetricKeyDescription Class(IdWeb Layer)
+Following the same pattern as CertificateDescription:
+
+```csharp
+public class SymmetricKeyDescription : CredentialDescription
```
bgavrilMS, line 70: The `CredentialDescription` class already has the responsibility of managing different types of credentials - see the `CredentialType` property - it can handle Certificates, Signed Assertions (i.e. JWT tokens with a special audience to denote they are actually credentials), client secrets and decrypt keys. 

Have you considered just extending `CredentialDescription` instead of creating a new class? What are the pros / cons?

docs/blog-posts/symmetric-key-support.md
```diff
@@ -0,0 +1,220 @@
...
+        // Load secret from Key Vault
+        var secret = await _secretClient.GetSecretAsync(description.KeyVaultSecretName).ConfigureAwait(false);
+        if (secret?.Value == null)
+            throw new InvalidOperationException($"Secret {description.KeyVaultSecretName} not found in Key Vault");
+
+        try
+        {
+            // Convert secret value to bytes and create SymmetricSecurityKey
+            var keyBytes = Convert.FromBase64String(secret.Value.Value);
+            description.CachedValue = new SymmetricSecurityKey(keyBytes);
```
bgavrilMS, line 128: The fact that `CachedValue` is an object is ... unpleasant. It requires users to cast to the right object (string, X509Certificate2 etc.). 

At least for certificates, a `Certificate` property also exists to avoid a cast. I suggest you do the same for symmetric keys?

docs/blog-posts/symmetric-key-support.md
```diff
@@ -0,0 +1,220 @@
...
+            { CredentialSource.StoreWithDistinguishedName, new StoreWithDistinguishedNameCertificateLoader() },
+            { CredentialSource.Base64Encoded, new Base64EncodedCertificateLoader() },
+
+            // New symmetric key loaders
+            { CredentialSource.SymmetricKeyFromKeyVault, new KeyVaultSymmetricKeyLoader(_secretClient) },
+            { CredentialSource.SymmetricKeyBase64Encoded, new Base64EncodedSymmetricKeyLoader() }
+        };
+    }
+
+    public async Task<SigningCredentials?> LoadSigningCredentialsAsync(
```
bgavrilMS, line 190: I wonder if a simpler design would be to rely on the existing object model, i.e. `ICredentialsLoader` and to treat the `SymmetricSecurityKey` as another type of credential - see `CredentialType` enum (alongside `DecryptKeys`, which is also not used by MSAL).

Apart from simpelr object model, this would avoid mixing the concept of `SigningCredentials` in Id.Web. I believe `SigningCredentials` + X509 has some perf implications (i.e. loading it often is expensive, caching needs to be done).

docs/blog-posts/symmetric-key-support.md
bgavrilMS: (read my comments below about extending CredentialDescription first)

src/Microsoft.Identity.Web.Certificate/ISigningCredentialsLoader.cs
```diff
@@ -0,0 +1,25 @@
...
+    /// </summary>
+    public interface ISigningCredentialsLoader
+    {
+        /// <summary>
+        /// Loads SigningCredentials from the credential description.
+        /// </summary>
+        /// <param name="credentialDescription">Credential description.</param>
+        /// <param name="parameters">Optional parameters for loading credentials.</param>
+        /// <returns>SigningCredentials if successful, null otherwise.</returns>
+        Task<SigningCredentials?> LoadSigningCredentialsAsync(
```
bgavrilMS, line 21: Not sure this is needed if `CredentialDescription` can alraedy return an X509Certificate2 and potentially a SymmetricKey . It becomes just a convenience method.

bgavrilMS:
Pls reopen if needed.

PR #3381, opened 2025-06-04
Adding fix for OBO cache key error in long running OBO with DefaultAuthorizationHeaderProvider

trwalke:

## Description

Updates the `DefaultAuthorizationHeaderProvider` to update the `AcquireTokenOptions.LongRunningWebApiSessionKey` after the token is acquired so that the key can be used in the next OBO call.
 
Fixes https://github.com/AzureAD/microsoft-identity-web/issues/3382


src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderProvider.cs
```diff
@@ -113,5 +121,14 @@ private static TokenAcquisitionOptions CreateTokenAcquisitionOptionsFromApiOptio
...
             };
         }
+
+        /// <summary>
+        /// Since AcquireTokenOptions is recreated, we need to update the original TokenAcquisitionOptions wth the parameters that were
+        /// updated in the new TokenAcquisitionOptions.
+        /// </summary>
+        private void UpdateOriginalTokenAcquisitionOptions(AcquireTokenOptions? acquireTokenOptions, TokenAcquisitionOptions newTokenAcquisitionOptions)
+        {
+            acquireTokenOptions!.LongRunningWebApiSessionKey = newTokenAcquisitionOptions.LongRunningWebApiSessionKey;
```
keegan-caruso: this is unsafe, we can't say not null here (!) and maybe null where this is called (?)

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderProvider.cs
```diff
@@ -27,13 +27,16 @@ public async Task<string> CreateAuthorizationHeaderForUserAsync(
             ClaimsPrincipal? claimsPrincipal = null,
             CancellationToken cancellationToken = default)
         {
+            var newTokenAcquisitionOptions = CreateTokenAcquisitionOptionsFromApiOptions(downstreamApiOptions, cancellationToken);
```
keegan-caruso, line 30: can you help me understand this a bit more?

Does this option get captured in the result object as part of GetAuthenticationResultForUserAsync/GetAuthenticationResultForAppAsync?

Can you point to where the modification is to AcquireTokenOptions?

Would it be a better fix to make it so that the code that is changing LongRunningWebApiSessionKey doesn't mutate input?

trwalke: Yeah, it is an interesting design, I would also prefer that it did not mutate the `AcquireTokenOptions` after the authentication but this is how it works at the moment. This PR is to fix this bug in particular, but a follow up PR can be used to refactor how the Long running OBO works all together. It would take some discussion on the API and user experience. 

When the options are provided with a `LongRunningWebApiSessionKey` that is equal to `LongRunningWebApiSessionKeyAuto` ("AllocateForMe") , the code will set this value to the cache key that MSAL calculates after initiating the long running OBO session. It is then expected for the developer to save the updated key from the modified `AcquireTokenOptions` and provide it in subsequent authentication calls.

See https://github.com/AzureAD/microsoft-identity-web/wiki/get-token-in-event-handler#principle
and https://github.com/AzureAD/microsoft-identity-web/blob/177e83df9749618826e1e0e54b1f68dccf62df74/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs#L951

trwalke: Also, it does not happen for `GetAuthenticationResultForAppAsync` since this code path does not use msal's OBO flow

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderProvider.cs
trwalke: Yes it is, updated

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderProvider.cs
trwalke: FYI @jmprieur @bgavrilMS ^

tests/Microsoft.Identity.Web.Test/AuthorizationHeaderProviderTests.cs
```diff
@@ -0,0 +1,149 @@
...
+
+                // Step 3: First call with ClaimsPrincipal to initiate LR session
+                var scopes = new[] { "User.Read" };
+                var result = await authorizationHeaderProvider.CreateAuthorizationHeaderForUserAsync(
+                    scopes,
+                    options,
+                    claimsPrincipal);
+
+                Assert.NotNull(result);
+                Assert.NotEqual(options.AcquireTokenOptions.LongRunningWebApiSessionKey, TokenAcquisitionOptions.LongRunningWebApiSessionKeyAuto);
```
bgavrilMS, line 67: What is it equal to then?

trwalke: It is set to the hash of the AT, which is acquired from MSAL

tests/Microsoft.Identity.Web.Test/AuthorizationHeaderProviderTests.cs
```diff
@@ -0,0 +1,149 @@
...
+            var tokenAcquirerFactory = InitTokenAcquirerFactoryForTest();
+            IServiceProvider serviceProvider = tokenAcquirerFactory.Build();
+
+            IAuthorizationHeaderProvider authorizationHeaderProvider =
+                serviceProvider.GetRequiredService<IAuthorizationHeaderProvider>();
+            var mockHttpClient = serviceProvider.GetRequiredService<IMsalHttpClientFactory>() as MockHttpClientFactory;
+
+            using (mockHttpClient)
+            {
+                mockHttpClient!.AddMockHandler(MockHttpCreator.CreateClientCredentialTokenHandler());
```
bgavrilMS: You should add a mock response that has AT, RT, IdT

bgavrilMS: You should add a mock response that has AT, RT, IdT, client_info

bgavrilMS: You should add a mock response that has AT, RT, IdT, client_info

tests/Microsoft.Identity.Web.Test/AuthorizationHeaderProviderTests.cs
bgavrilMS: Please add a comment in the test around this. You can even assert.

tests/Microsoft.Identity.Web.Test/AuthorizationHeaderProviderTests.cs
```diff
@@ -0,0 +1,148 @@
...
+                var identity = new CaseSensitiveClaimsIdentity(claims, "TestAuth");
+                identity.BootstrapContext = CreateTestJwt();
+                var claimsPrincipal = new ClaimsPrincipal(identity);
+
+                // Create options with LongRunningWebApiSessionKey
+                var options = new AuthorizationHeaderProviderOptions
+                {
+                    AcquireTokenOptions = new AcquireTokenOptions
+                    {
+                        LongRunningWebApiSessionKey = TokenAcquisitionOptions.LongRunningWebApiSessionKeyAuto
```
bgavrilMS, line 52: Can you have a test where the user doesn't set this to "Auto" but uses 1-2 separate keys instead?

tests/Microsoft.Identity.Web.Test/AuthorizationHeaderProviderTests.cs
```diff
@@ -0,0 +1,148 @@
...
+using Microsoft.Identity.Web.Test.Common.Mocks;
+using Microsoft.Identity.Web.TestOnly;
+using Microsoft.IdentityModel.Tokens;
+using Xunit;
+
+namespace Microsoft.Identity.Web.Test
+{
+    public class AuthorizationHeaderProviderTests
+    {
+        [Fact]
```
bgavrilMS, line 22: Great test!

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderProvider.cs
trwalke: This will be in a follow up PR FYI. Lets merge this one in to unblock the customer

jennyf19 (approved):

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderProvider.cs
keegan-caruso: Thanks

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderProvider.cs
```diff
@@ -27,13 +27,16 @@ public async Task<string> CreateAuthorizationHeaderForUserAsync(
...
             var result = await _tokenAcquisition.GetAuthenticationResultForUserAsync(
                 scopes,
                 downstreamApiOptions?.AcquireTokenOptions.AuthenticationOptionsName,
                 downstreamApiOptions?.AcquireTokenOptions.Tenant,
                 downstreamApiOptions?.AcquireTokenOptions.UserFlow,
                 claimsPrincipal,
-                CreateTokenAcquisitionOptionsFromApiOptions(downstreamApiOptions, cancellationToken)).ConfigureAwait(false);
+                newTokenAcquisitionOptions).ConfigureAwait(false);
+
+            UpdateOriginalTokenAcquisitionOptions(downstreamApiOptions?.AcquireTokenOptions, newTokenAcquisitionOptions);
```
keegan-caruso, line 39: Can you help me understand why we need to update the original instance if we are passing in a new instance of this option?

jmprieur (commented):
Wouldn't it be better to use the new interface that Keegan created:

https://github.com/AzureAD/microsoft-identity-abstractions-for-dotnet/blob/main/src/Microsoft.Identity.Abstractions/DownstreamApi/IAuthorizationHeaderProvider_T.cs

this way we could return something?

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderProvider.cs
trwalke: Because the original instance is what the developer expects to use to get the new cache key. I am not sure why this code is creating a new instance of the options. I am also not sure if everything in the options are properly copied over to the new instance so it is best to update the customer's original instance updated with the relevant data.

trwalke:
> Wouldn't it be better to use the new interface that Keegan created:
> 
> https://github.com/AzureAD/microsoft-identity-abstractions-for-dotnet/blob/main/src/Microsoft.Identity.Abstractions/DownstreamApi/IAuthorizationHeaderProvider_T.cs
> 
> this way we could return something?

I'm am not sure it is the best way to do it. The returning of the cache key also happens when using the `ITokenAcquisition` so I wouldn't want to limit the solution to only `IAuthorizationHeaderProvider_T`.

Lets fix this for the customer since it is the currently released behavior and change the design in a follow up PR. I can create a work item.

keegan-caruso (approved):

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderProvider.cs
trwalke: Discussed with Keegan. We are now on the same page

PR #3401, opened 2025-06-27
Update OidcFicSignedAssertionProviderExtensions.cs to use TryAddEnumerable

### LINKED ISSUE #3400: When expecting DI with an IEnumerable<T>, one should use services.TryAddEnumerable([ ServiceDescriptor.Singleton<T, Impl>(), not just AddSingleton.

Microsoft.Identity.Web

3.10.0
### Web app

Not Applicable

### Web API

Not Applicable

### Token cache serialization

Not Applicable

### Description

When expecting DI with an IEnumerable<T>, one should use

```csharp
 services.TryAddEnumerable([ ServiceDescriptor.Singleton<T, Impl>()
```
 not just `AddSingleton`

### Reproduction steps

`TryAddSingleton()` looks for whether any implementation type has been registered for the given service type. If yes then it does not register the implementation type (given in the call) for the service type (given in the call). This is not what we want for `ICustomSignedAssertionLoader` as it's meant to have several

On the other hand, `TryAddEnumerable(ServiceDescriptor)` on the other hand looks for the SAME implementation type as the implementation type given in the call has already been registered for the given service type. If yes, then it does not register the implementation type (given in the call) for the service type. If no, then it does. 
the Enumerable suffix indicates that it can register more than one implementation types for the same service type which is what we want for the ICustomSignedAssertionLoader

### Expected behavior

 services.TryAddEnumerable([ ServiceDescriptor.Singleton<ICustomSignedAssertionLoader Impl>()

jmprieur:
# Update OidcFicSignedAssertionProviderExtensions.cs to use TryAddEnumerable

## Description

{Detail}

Fixes #3400


src/Microsoft.Identity.Web.OidcFIC/OidcFicSignedAssertionProviderExtensions.cs
```diff
@@ -19,7 +20,11 @@ public static class OidcFicSignedAssertionProviderExtensions
         /// <returns>the service collection for chaining.</returns>
         public static IServiceCollection AddOidcFic(this IServiceCollection services)
         {
-            services.AddSingleton<ICustomSignedAssertionProvider, OidcIdpSignedAssertionLoader>();
+            services.TryAddEnumerable(
+                new[]
+                {
+                ServiceDescriptor.Singleton<ICustomSignedAssertionProvider, OidcIdpSignedAssertionLoader>()
+                });
```
keegan-caruso:
```suggestion
            services.TryAddEnumerable(ServiceDescriptor.Singleton<ICustomSignedAssertionProvider, OidcIdpSignedAssertionLoader>());
```

This overload of TryAddEnumerable that takes an IEnumerable of ServiceDescriptor only calls the other overload on the individual items. So we can just use the method that takes a single item. Also there is a more simple syntax for creating ServiceDescriptor that I found recently.

src/Microsoft.Identity.Web.OidcFIC/OidcFicSignedAssertionProviderExtensions.cs
```diff
@@ -19,7 +20,11 @@ public static class OidcFicSignedAssertionProviderExtensions
         /// <returns>the service collection for chaining.</returns>
         public static IServiceCollection AddOidcFic(this IServiceCollection services)
```
keegan-caruso, line 21: Should be easy to add a test for this. 

```cs
var services = new ServiceCollection();
// add a test ICustomSignedAssertionProvider
services.AddOidcFic()
// assert that both ICustomSignedAssertionProvider are registered

```

keegan-caruso (approved):
Small suggestion to clean it up, we should add a test, but otherwise LGTM.

pmaytak (approved):

PR #3430, opened 2025-07-22
Reload certificates for all client credential based issues

### LINKED ISSUE #3429: [Feature Request] Reload certificates for all client credential based issues

The current behavior of the library is to reload and retry certificates in the event that the certificate fails for one of the following two issues:

* The certificate has been revoked.
* The certificate has expired.

There is a [great many](https://learn.microsoft.com/en-us/entra/identity-platform/reference-error-codes#aadsts-error-codes) other reasons that authentication can fail that fall outside of these two scenarios. If any of those additional scenarios occur the certificate will never be reloaded.

As a result, if a bad certificate is installed on the machine and picked up, and subsequently rotated, a service restart is needed for the new certificate to be used.

https://github.com/AzureAD/microsoft-identity-web/blob/deedd9dd817935a8f6c9049c250c66269152115b/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs#L812

I would like any issue with the certificate to trigger a reload, so that a new certificate can be selected. This can be done by just checking if the error code is `invalid_client`.

Adding more reasons to the acceptable list of reasons. While this would work for known reasons, it is far safer to just retry for _everything_ in case additional things are added or cases are missed.

tlupes:
# Reload certificates for all client credential based issues

Change retry logic to reload certificates on all credential issues, not just revocation/expiration.

## Description

The current behavior of the library is to reload and retry certificates in the event that the certificate fails for one of the following two issues:

* The certificate has been revoked.
* The certificate has expired.

There is a [great many](https://learn.microsoft.com/en-us/entra/identity-platform/reference-error-codes#aadsts-error-codes) other reasons that authentication can fail that fall outside of these two scenarios. If any of those additional scenarios occur the certificate will never be reloaded.

As a result, if a bad certificate is installed on the machine and picked up, and subsequently rotated, a service restart is needed for the new certificate to be used.

Fixes #3429

### DIFF
--- a/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
@@ -812,18 +812,7 @@ namespace Microsoft.Identity.Web
         private bool IsInvalidClientCertificateOrSignedAssertionError(MsalServiceException exMsal)
         {
             return !_retryClientCertificate &&
-                string.Equals(exMsal.ErrorCode, Constants.InvalidClient, StringComparison.OrdinalIgnoreCase) &&
-#if !NETSTANDARD2_0 && !NET462 && !NET472
-                (exMsal.Message.Contains(Constants.InvalidKeyError, StringComparison.OrdinalIgnoreCase)
-                || exMsal.Message.Contains(Constants.SignedAssertionInvalidTimeRange, StringComparison.OrdinalIgnoreCase)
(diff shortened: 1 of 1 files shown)


tlupes:
Integration tests all happy locally!

bgavrilMS (approved):

jmprieur (approved):

jmprieur:
discussed risks with Bogdan for other invalid_client (bad GUID). This is only one retry so taking it

PR #3433, opened 2025-07-25
Support for Agent user identities

jmprieur:
## Support Agent user identities

#3432


src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisitionExtensionOptions.cs
```diff
@@ -35,13 +36,26 @@ internal void InvokeOnBeforeTokenAcquisitionForApp(AcquireTokenForClientParamete
...
         /// </summary>
-        internal void InvokeOnBeforeTokenAcquisitionForTestUser(AcquireTokenByUsernameAndPasswordConfidentialParameterBuilder builder,
-                                                           AcquireTokenOptions? acquireTokenOptions, ClaimsPrincipal user)
+        internal async Task InvokeOnBeforeTokenAcquisitionForTestUserAsync(AcquireTokenByUsernameAndPasswordConfidentialParameterBuilder builder,
+                                                                           AcquireTokenOptions? acquireTokenOptions, ClaimsPrincipal user)
         {
+            // Run the async event if it is not null.
+            if (OnBeforeTokenAcquisitionForTestUserAsync != null)
+            {
+                await OnBeforeTokenAcquisitionForTestUserAsync(builder, acquireTokenOptions, user);
```
keegan-caruso: async event handlers require a little more to ensure that all complete before continuing execution.

One option would be to get the invocation list

src/Microsoft.Identity.Web.AgentIdentities/MsalAgentUserIdentityAddIn.cs
```diff
@@ -0,0 +1,80 @@
...
+            options.ExtraParameters.TryGetValue(Constants.UsernameKey, out object? usernameObject);
+            if (agentIdentityObject is string agentIdentity && usernameObject is string username)
+            {
+                ITokenAcquirerFactory tokenAcquirerFactory = serviceProvider.GetRequiredService<ITokenAcquirerFactory>();
+                IAuthenticationSchemeInformationProvider authenticationSchemeInformationProvider =
+                    serviceProvider.GetRequiredService<IAuthenticationSchemeInformationProvider>();
+
+                ITokenAcquirer tokenAcquirer = tokenAcquirerFactory.GetTokenAcquirer(authenticationSchemeInformationProvider.GetEffectiveAuthenticationScheme(options.AuthenticationOptionsName));
+
+                AcquireTokenOptions options1 = options.Clone();
```
keegan-caruso: can we use a better name here?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -334,88 +340,107 @@ public async Task<AuthenticationResult> GetAuthenticationResultForUserAsync(
...
+                var dict = MergeExtraQueryParameters(mergedOptions, tokenAcquisitionOptions);
+                if (dict != null)
                 {
-                    addInOptions.InvokeOnBeforeTokenAcquisitionForTestUser(builder, tokenAcquisitionOptions, user);
+                    builder.WithExtraQueryParameters(dict);
                 }
-
-                // Pass the token acquisition options to the builder
-                if (tokenAcquisitionOptions != null)
+                if (tokenAcquisitionOptions.ExtraHeadersParameters != null)
```
keegan-caruso, line 416:
```suggestion

                if (tokenAcquisitionOptions.ExtraHeadersParameters != null)
```

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -334,88 +340,107 @@ public async Task<AuthenticationResult> GetAuthenticationResultForUserAsync(
...
-                        builder.WithSignedHttpRequestProofOfPossession(tokenAcquisitionOptions.PoPConfiguration);
-                    }
+                    builder.WithExtraHttpHeaders(tokenAcquisitionOptions.ExtraHeadersParameters);
                 }
+                if (tokenAcquisitionOptions.CorrelationId != null)
+                {
+                    builder.WithCorrelationId(tokenAcquisitionOptions.CorrelationId.Value);
+                }
+                builder.WithClaims(tokenAcquisitionOptions.Claims);
+                if (tokenAcquisitionOptions.PoPConfiguration != null)
```
keegan-caruso, line 425:
```suggestion

                if (tokenAcquisitionOptions.CorrelationId != null)
                {
                    builder.WithCorrelationId(tokenAcquisitionOptions.CorrelationId.Value);
                }
                
                builder.WithClaims(tokenAcquisitionOptions.Claims);
                
                if (tokenAcquisitionOptions.PoPConfiguration != null)
```

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -334,88 +340,107 @@ public async Task<AuthenticationResult> GetAuthenticationResultForUserAsync(
...
+                {
+                    builder.WithSignedHttpRequestProofOfPossession(tokenAcquisitionOptions.PoPConfiguration);
+                }
+            }
 
-                var authenticationResult = await builder.ExecuteAsync()
-                .ConfigureAwait(false);
+            var authenticationResult = await builder.ExecuteAsync()
```
keegan-caruso:
```suggestion
            var authenticationResult = await builder.ExecuteAsync().ConfigureAwait(false);
```

tests/E2E Tests/AgentApplications/TokenAcquirerFactorySingletonProtection.cs
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace AgentApplicationsTests
+{
+    [CollectionDefinition(nameof(TokenAcquirerFactorySingletonProtection))]
```
keegan-caruso:
```suggestion
     [CollectionDefinition(nameof(TokenAcquirerFactorySingletonProtection))]
     public class TokenAcquirerFactorySingletonProtection
     {
         // This class has no code, and is never created. Its purpose is to prevent test classes using the
         // static singleton DefaultTokenAcquirerFactory from running in parallel as some tests modify this singleton.
    }
```

src/Microsoft.Identity.Web.AgentIdentities/MsalAgentUserIdentityAddIn.cs
jmprieur: Thanks!

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisitionExtensionOptions.cs
jmprieur: Thanks for the heads up!

src/Microsoft.Identity.Web.AgentIdentities/MsalAgentUserIdentityAddIn.cs
```diff
@@ -0,0 +1,80 @@
...
+                        request.BodyParameters.Remove("password");
+
+                        if (request.BodyParameters.TryGetValue("client_secret", out var secret)
+                                && secret.Equals("default", StringComparison.OrdinalIgnoreCase))
+                        {
+                            request.BodyParameters.Remove("client_secret");
+                        }
+
+                        // For the moment
+                        request.RequestUri = new Uri(request.RequestUri + "?slice=first");
```
bgavrilMS, line 72: Please don't hardcode test slices. Use `Extra Query Parameters` in tests

src/Microsoft.Identity.Web.AgentIdentities/MsalAgentUserIdentityAddIn.cs
```diff
@@ -0,0 +1,80 @@
...
+                {
+                    throw new InvalidOperationException("Failed to acquire the signed assertions.");
+                }
+
+                string userFicAssertion = agentIdentityFic.AccessToken!;
+
+                // Register the MSAL extension that will modify the token request just in time.
+                MsalAuthenticationExtension extension = new()
+                {
+                    OnBeforeTokenRequestHandler = (request) =>
```
bgavrilMS, line 53: This does not tie the user token to any of the parameters in here (username, user_fic). THis means that the user token is cached by the typical (instance, tenantID, home_account_id, scopes). 

Should we associate the user token to the `username` and to the `oid` ?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisitionExtensionOptions.cs
```diff
@@ -35,13 +37,33 @@ internal void InvokeOnBeforeTokenAcquisitionForApp(AcquireTokenForClientParamete
...
         /// </summary>
-        internal void InvokeOnBeforeTokenAcquisitionForTestUser(AcquireTokenByUsernameAndPasswordConfidentialParameterBuilder builder,
-                                                           AcquireTokenOptions? acquireTokenOptions, ClaimsPrincipal user)
+        internal async Task InvokeOnBeforeTokenAcquisitionForTestUserAsync(AcquireTokenByUsernameAndPasswordConfidentialParameterBuilder builder,
+                                                                           AcquireTokenOptions? acquireTokenOptions, ClaimsPrincipal user)
         {
+            // Run the async event if it is not null
+            if (OnBeforeTokenAcquisitionForTestUserAsync != null)
+            {
+                // (cannot directly await an async event because events are not tasks
```
bgavrilMS, line 54: Is it to difficult to move away from events to a list of delegates?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -254,6 +254,12 @@ public async Task<AuthenticationResult> GetAuthenticationResultForUserAsync(
...
 
+            if (tokenAcquisitionOptions is not null)
+            {
+                tokenAcquisitionOptions.ExtraParameters ??= new Dictionary<string, object>();
+                tokenAcquisitionOptions.ExtraParameters[Constants.ExtensionOptionsServiceProviderKey] = _serviceProvider;
+            }
+
             try
             {
                 AuthenticationResult? authenticationResult;
```
bgavrilMS, line 265: Can you please expose `bool FromCache` property to be able to assert in tests if the token comes from cache or not?

src/Microsoft.Identity.Web.AgentIdentities/MsalAgentUserIdentityAddIn.cs
```diff
@@ -0,0 +1,80 @@
...
+                    {
+                        // Already in the request:
+                        // - client_id = agentIdentity;
+                        // - client_assertion is the AA FIC
+
+                        // User FIC parameters
+                        request.BodyParameters["username"] = username;
```
bgavrilMS, line 60: Is this needed? ROPC already adds the username.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -334,88 +340,107 @@ public async Task<AuthenticationResult> GetAuthenticationResultForUserAsync(
...
-                        username,
-                        password);
+            // Check for extension options for the ROPC flow
+            TokenAcquisitionExtensionOptions? addInOptions = tokenAcquisitionExtensionOptionsMonitor?.CurrentValue;
 
-                if (addInOptions != null)
+            // ROPC flow
+            AcquireTokenByUsernameAndPasswordConfidentialParameterBuilder builder = ((IByUsernameAndPassword)application)
+                .AcquireTokenByUsernamePassword(
+                    scopes.Except(_scopesRequestedByMsal),
```
bgavrilMS, line 398: nit: there is no need to do this "Except" logic. MSAL already takes care of it.

bgavrilMS (changes requested):
Tests need to work with static caches.

src/Microsoft.Identity.Web.AgentIdentities/MsalAgentUserIdentityAddIn.cs
```diff
@@ -0,0 +1,80 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Identity.Abstractions;
+using Microsoft.Identity.Client;
+using Microsoft.Identity.Client.Extensibility;
+
+namespace Microsoft.Identity.Web.AgentIdentities
+{
+    internal static class AgentUserIdentityMsalAddIn
+    {
+        private const string azureAdTokenExchangeScope = "api://AzureADTokenExchange/.default";
```
bgavrilMS, line 18: This should not be hardcoded, as it is different on different clouds. It needs to be an input.

src/Microsoft.Identity.Web.AgentIdentities/MsalAgentUserIdentityAddIn.cs
```diff
@@ -0,0 +1,80 @@
...
+            if (agentIdentityObject is string agentIdentity && usernameObject is string username)
+            {
+                ITokenAcquirerFactory tokenAcquirerFactory = serviceProvider.GetRequiredService<ITokenAcquirerFactory>();
+                IAuthenticationSchemeInformationProvider authenticationSchemeInformationProvider =
+                    serviceProvider.GetRequiredService<IAuthenticationSchemeInformationProvider>();
+
+                ITokenAcquirer tokenAcquirer = tokenAcquirerFactory.GetTokenAcquirer(authenticationSchemeInformationProvider.GetEffectiveAuthenticationScheme(options.AuthenticationOptionsName));
+
+                AcquireTokenOptions optionsForAgentIdFic = options.Clone();
+                optionsForAgentIdFic.ExtraParameters = new Dictionary<string, object>();
```
bgavrilMS, line 40: optionsForAgentIdFic  is not used.

src/Microsoft.Identity.Web.AgentIdentities/MsalAgentUserIdentityAddIn.cs
```diff
@@ -0,0 +1,80 @@
...
+            IServiceProvider serviceProvider = (IServiceProvider)options.ExtraParameters![Constants.ExtensionOptionsServiceProviderKey];
+            options.ExtraParameters.TryGetValue(Constants.AgentIdentityKey, out object? agentIdentityObject);
+            options.ExtraParameters.TryGetValue(Constants.UsernameKey, out object? usernameObject);
+            if (agentIdentityObject is string agentIdentity && usernameObject is string username)
+            {
+                ITokenAcquirerFactory tokenAcquirerFactory = serviceProvider.GetRequiredService<ITokenAcquirerFactory>();
+                IAuthenticationSchemeInformationProvider authenticationSchemeInformationProvider =
+                    serviceProvider.GetRequiredService<IAuthenticationSchemeInformationProvider>();
+
+                ITokenAcquirer tokenAcquirer = tokenAcquirerFactory.GetTokenAcquirer(authenticationSchemeInformationProvider.GetEffectiveAuthenticationScheme(options.AuthenticationOptionsName));
```
bgavrilMS, line 37: This tokenAcquirer should be inside the `OnBeforeTokenRequest`, because it does not need to be evaluated if `OnBeforeTokenRequest` is not fired (e.g. when tokens are cached).

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -334,88 +340,107 @@ public async Task<AuthenticationResult> GetAuthenticationResultForUserAsync(
...
-                            account)
-                            .ExecuteAsync()
-                            .ConfigureAwait(false);
-                    }
-                    catch (MsalException ex)
-                    {
-                        // Log a message when the silent flow fails and try acquisition through ROPC.
-                        Logger.TokenAcquisitionError(_logger, ex.Message, ex);
-                    }
+                    var account = await application.GetAccountAsync(user.GetMsalAccountId()).ConfigureAwait(false);
```
bgavrilMS, line 375: nit: avoid calling `user.GetMsalAccountId()` twice.

src/Microsoft.Identity.Web.AgentIdentities/MsalAgentUserIdentityAddIn.cs
jmprieur: yes, I think we should. username at least?

src/Microsoft.Identity.Web.AgentIdentities/MsalAgentUserIdentityAddIn.cs
jmprieur: Lol. This is what Neha wrote in the MISE add-in

src/Microsoft.Identity.Web.AgentIdentities/MsalAgentUserIdentityAddIn.cs
jmprieur: Do you have the mapping instance => exchange string?

jmprieur:
replaced by https://github.com/AzureAD/microsoft-identity-web/pull/3435

PR #3435, opened 2025-07-29
Support for Agent User identities

jmprieur:
#Support for Agent User identities


src/Microsoft.Identity.Web.AgentIdentities/Microsoft.Identity.Web.AgentIdentities.csproj
```diff
@@ -8,8 +8,7 @@
 
     <!-- The package is new in 3.10.0.-->
     <PackageValidationBaselineVersion>3.10.0</PackageValidationBaselineVersion>
-    <EnablePackageValidation>false</EnablePackageValidation>
-    
+    <EnablePackageValidation>true</EnablePackageValidation>
```
jmprieur, line 11: now that we released a first version of Microsoft.Identity.Web.AgentIdentities, we can enable package (version) validation

src/Microsoft.Identity.Web.AgentIdentities/MsalAgentUserIdentityAddIn.cs
```diff
@@ -0,0 +1,89 @@
...
+using Microsoft.Extensions.Options;
+using Microsoft.Identity.Abstractions;
+using Microsoft.Identity.Client;
+using Microsoft.Identity.Client.Extensibility;
+
+namespace Microsoft.Identity.Web.AgentIdentities
+{
+    internal static class AgentUserIdentityMsalAddIn
+    {
+        private const string azureAdTokenExchangeScope = "api://AzureADTokenExchange/.default";
```
jmprieur: not needed

src/Microsoft.Identity.Web.AgentIdentities/MsalAgentUserIdentityAddIn.cs
```diff
@@ -0,0 +1,88 @@
...
+using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Options;
+using Microsoft.Identity.Abstractions;
+using Microsoft.Identity.Client;
+using Microsoft.Identity.Client.Extensibility;
+
+namespace Microsoft.Identity.Web.AgentIdentities
+{
+    internal static class AgentUserIdentityMsalAddIn
```
jennyf19: The file name and class name are different, do you need `msal` in the name?

src/Microsoft.Identity.Web.AgentIdentities/AgentIdentitiesExtension.cs
```diff
@@ -51,24 +53,50 @@ public static AuthorizationHeaderProviderOptions WithAgentIdentity(this Authoriz
...
         }
 
-        // TODO:make public?
-        private static AcquireTokenOptions ForAgentIdentity(this AcquireTokenOptions options, string agentApplicationId)
+        /// <summary>
+        /// Updates the options to acquire a token for the agent user identity.
+        /// </summary>
+        /// <param name="options">Authorization header provider options.</param>
+        /// <param name="agentApplicationId">The agent identity GUID.</param>
+        /// <param name="username">upn of the user.</param>
```
jennyf19:
```suggestion
        /// <param name="username">UPN of the user.</param>
```

src/Microsoft.Identity.Web.AgentIdentities/AgentIdentitiesExtension.cs
```diff
@@ -51,24 +53,50 @@ public static AuthorizationHeaderProviderOptions WithAgentIdentity(this Authoriz
...
+            };
+            
+            // Set the username and agent identity parameters
+            options.AcquireTokenOptions.ExtraParameters[Constants.UsernameKey] = username;
+            options.AcquireTokenOptions.ExtraParameters[Constants.AgentIdentityKey] = agentApplicationId;
+            
+            return options;
+        }
+
+        // TODO:would it make sense to have it public?
```
jennyf19, line 82: Let's take offline

src/Microsoft.Identity.Web.AgentIdentities/MsalAgentUserIdentityAddIn.cs
```diff
@@ -0,0 +1,88 @@
...
+                        request.BodyParameters.Remove("password");
+
+                        if (request.BodyParameters.TryGetValue("client_secret", out var secret)
+                                && secret.Equals("default", StringComparison.OrdinalIgnoreCase))
+                        {
+                            request.BodyParameters.Remove("client_secret");
+                        }
+
+                        // For the moment
+                        request.RequestUri = new Uri(request.RequestUri + "?slice=first");
```
jennyf19: let's discuss this too

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquirerExtensions.cs
```diff
@@ -0,0 +1,84 @@
...
+
+
+        /// <summary>
+        /// Consider the client assertion for the token request.
+        /// </summary>
+        /// <param name="options">Options (can be null)</param>
+        /// <param name="clientAssertion">client assertion (shouldn't be null)</param>
+        /// <returns>The modified options. New options with the signed assertion if <paramref name="options"/> was null.</returns>
+        public static AcquireTokenOptions WithClientAssertion(this AcquireTokenOptions options, string clientAssertion)
+        {
```
jennyf19, line 68: Is the options object shared elsewhere?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -334,88 +340,107 @@ public async Task<AuthenticationResult> GetAuthenticationResultForUserAsync(
...
-                    {
-                        builder.WithSignedHttpRequestProofOfPossession(tokenAcquisitionOptions.PoPConfiguration);
-                    }
+                    builder.WithExtraHttpHeaders(tokenAcquisitionOptions.ExtraHeadersParameters);
                 }
+                if (tokenAcquisitionOptions.CorrelationId != null)
+                {
+                    builder.WithCorrelationId(tokenAcquisitionOptions.CorrelationId.Value);
+                }
+                builder.WithClaims(tokenAcquisitionOptions.Claims);
```
jennyf19, line 424: Now there are two ways to provide user identity? Claims on the ClaimsPrincipal and ExtraParameters dict?

tests/E2E Tests/AgentApplications/AgentUserIdentityTestscs.cs
```diff
@@ -0,0 +1,122 @@
...
+using System.Security.Claims;
+using System.Security.Cryptography.X509Certificates;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Graph;
+using Microsoft.Identity.Abstractions;
+using Microsoft.Identity.Web;
+
+namespace AgentApplicationsTests
+{
+    public class AgentUserIdentityTests
```
jennyf19, line 15: I think this needs to be above the name of the class to take effect
#if !FROM_GITHUB_ACTION

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
jmprieur: Yes. There is the legacy one (for MISE user FIC)
and the experience for agents, which start from:

```csharp
Options.WithAgentUserIdentity(agentIdentityGuid, username)
```

the options don't have access to the user (which could exist btw in an authenticated web API, but would have nothing to do with the agent user identity.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquirerExtensions.cs
jmprieur: only in the call.
would you suggest to clone it?

src/Microsoft.Identity.Web.AgentIdentities/MsalAgentUserIdentityAddIn.cs
jmprieur: Fixed

src/Microsoft.Identity.Web.AgentIdentities/AgentUserIdentityMsalAddIn.cs
```diff
@@ -0,0 +1,88 @@
...
+                        request.BodyParameters.Remove("password");
+
+                        if (request.BodyParameters.TryGetValue("client_secret", out var secret)
+                                && secret.Equals("default", StringComparison.OrdinalIgnoreCase))
+                        {
+                            request.BodyParameters.Remove("client_secret");
+                        }
+
+                        // For the moment
+                        request.RequestUri = new Uri(request.RequestUri + "?slice=first");
```
keegan-caruso: Can this not be in the query params?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquirerExtensions.cs
```diff
@@ -0,0 +1,84 @@
...
+        /// </param>
+        /// <param name="cancellationToken">Cancellation token.</param>
+        /// <returns></returns>
+        public static async Task<AcquireTokenResult> GetFicTokenAsync(this ITokenAcquirer tokenAcquirer,
+            AcquireTokenOptions? options = null,
+            string? clientAssertion = null,
+            string scope = "api://AzureAdTokenExchange/.default",
+            CancellationToken cancellationToken = default)
+        {
+            AcquireTokenOptions? tokenAcquisitionOptions;
```
keegan-caruso, line 46:
```suggestion
            AcquireTokenOptions tokenAcquisitionOptions;
```

Double check this, but I don't think this should be marked as nullable. It will always be assigned after the if else below.

tests/E2E Tests/AgentApplications/AutonomousAgentTests.cs
```diff
@@ -1,53 +1,68 @@
...
+            services.AddInMemoryTokenCaches();
+            services.Configure<MicrosoftIdentityApplicationOptions>(configuration.GetSection("AzureAd"));
             services.AddAgentIdentities();
             services.AddMicrosoftGraph(); // If you want to call Microsoft Graph
-            var serviceProvider = tokenAcquirerFactory.Build();
+            var serviceProvider = services.BuildServiceProvider();
+
+            string agentIdentity = "d84da24a-2ea2-42b8-b5ab-8637ec208024"; // Replace with the actual agent identity
+
+            //// Get an authorization header and handle the call to the downstream API yoursel
```
keegan-caruso, line 42:
```suggestion
            // Get an authorization header and handle the call to the downstream API yoursel
```

tests/E2E Tests/AgentApplications/AutonomousAgentTests.cs
```diff
@@ -1,53 +1,68 @@
...
+            IAuthorizationHeaderProvider authorizationHeaderProvider = serviceProvider.GetService<IAuthorizationHeaderProvider>()!;
+            AuthorizationHeaderProviderOptions options = new AuthorizationHeaderProviderOptions().WithAgentIdentity(agentIdentity);
 
+            //// Request user tokens in autonomous agents.
```
keegan-caruso, line 46:
```suggestion
            // Request user tokens in autonomous agents.
```

keegan-caruso (approved):

src/Microsoft.Identity.Web.AgentIdentities/AgentUserIdentityMsalAddIn.cs
jmprieur: This doesn't work without for the moment.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquirerExtensions.cs
```diff
@@ -0,0 +1,84 @@
...
+        /// <term>api://AzureAdTokenExchange/.default</term><description>for public cloud</description>
+        /// <term>api://AzureADTokenExchangeChina/.default</term><description>for China cloud</description>
+        /// <term>api://AzureADTokenExchangeFrance/.default</term><description>for Bleu</description>
+        /// <term>api://AzureADTokenExchangeGermany/.default</term><description>for the German cloud</description>
+        /// </item>
+        /// </list>
+        /// </param>
+        /// <param name="cancellationToken">Cancellation token.</param>
+        /// <returns></returns>
+        public static async Task<AcquireTokenResult> GetFicTokenAsync(this ITokenAcquirer tokenAcquirer,
```
bgavrilMS, line 40: Do we need to make this public now? Maybe we can solve the `api://AzureAdTokenExchange` problem (e.g. by using a GUID) or have we given that up?

src/Microsoft.Identity.Web.AgentIdentities/AgentUserIdentityMsalAddIn.cs
bgavrilMS: I tried to set ExtraQueryParaterms in the test and they do not show up as `AcquireTokenOptions` here. Not sure why. 

Maybe you can use a special options.ExtraParameters or smth? We should not put this slice param in an official release, this has caused problems in the past.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -334,88 +340,107 @@ public async Task<AuthenticationResult> GetAuthenticationResultForUserAsync(
...
-                    }
-                    catch (MsalException ex)
-                    {
-                        // Log a message when the silent flow fails and try acquisition through ROPC.
-                        Logger.TokenAcquisitionError(_logger, ex.Message, ex);
-                    }
+                    var account = await application.GetAccountAsync(user.GetMsalAccountId()).ConfigureAwait(false);
 
+                    // Silent flow
+                    return await application.AcquireTokenSilent(
```
bgavrilMS, line 378: When I ran the test - AgentUserIdentityGetsTokenForGraphAsync, it did not hit the AcquireTokenSilent path (I changed it to request 2 tokens). The user is always null here.

bgavrilMS: (which makes sense, because the ClaimsPrincipal is always null)

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquirerExtensions.cs
jmprieur: Actually no. The options can still be null!

tests/E2E Tests/AgentApplications/AutonomousAgentTests.cs
jmprieur: Fixed

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquirerExtensions.cs
jmprieur: I tried the GUID and this didn't work.
Do you propose to keep GetFicTokenAsync internal?

bgavrilMS: Yes.

bgavrilMS: Not a strong push on this, but maybe we can have a better API :)

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
jmprieur: You would have to do like in this test:
https://github.com/AzureAD/microsoft-identity-web/blob/450e7b70629a2c109960ebf845fff109bd694619/tests/E2E%20Tests/AgentApplications/AgentUserIdentityTestscs.cs#L99-L113

bgavrilMS (approved):

PR #3460, opened 2025-08-26
Enabling IdWeb to use ExtraBodyParameters Prototype

trwalke:
# {PR title}

## Description

{Detail}

### DIFF
--- a/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
@@ -643,6 +643,9 @@ namespace Microsoft.Identity.Web
                            tokenAcquisitionOptions.PopClaim!);
                     }
                 }
+
+                //Add extra body parameters configured by extensions
+                AddExtraBodyParametersIfNeeded(tokenAcquisitionOptions, builder);
             }
 
             try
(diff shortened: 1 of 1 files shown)


src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -655,6 +655,9 @@ public async Task<AuthenticationResult> GetAuthenticationResultForAppAsync(
                            tokenAcquisitionOptions.PopClaim!);
                     }
                 }
+
+                //Add extra body parameters configured by extensions
+                AddExtraBodyParametersIfNeeded(tokenAcquisitionOptions, builder);
```
jmprieur: Why not have this call in the block starting line 591? You would be sure that tokenAcquisitionOptions is not null.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -688,6 +691,18 @@ public async Task<AuthenticationResult> GetAuthenticationResultForAppAsync(
             }
         }
 
+        private void AddExtraBodyParametersIfNeeded(TokenAcquisitionOptions tokenAcquisitionOptions, AcquireTokenForClientParameterBuilder builder)
+        {
+            if (tokenAcquisitionOptions.ExtraParameters != null
```
jmprieur, line 695: Do we want customers to be able to use ExtraBodyParameters? or not at all
If yes, we'd want to add it directly to the `AcquireTokenOptions`

jmprieur (commented):
LGTM
I left a couple of suggestions. Will approve when we have aligned.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
jmprieur: Discussed with Bodgan offline. Let's not expose it

jmprieur (approved):

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
bgavrilMS: Done

PR #3472, opened 2025-09-10
Combining or Fixing Test Collections

JoshLozensky:
Combining two different collections intended for tests using token acquirer factory. Also, fixing UI test collection definition


jmprieur (approved):
LGTM
Thanks @JoshLozensky

PR #3480, opened 2025-09-16
Update tests to new behavior + 401 on AuthorizationHeader when we are sure it's an auth issue

jmprieur:
# Update tests to new behavior + 401 on AuthorizationHeader when we are sure it's an auth issue


src/Microsoft.Identity.Web.Sidecar/Endpoints/AuthorizationHeaderEndpoint.cs
```diff
@@ -22,6 +23,7 @@ public static void AddAuthorizationHeaderRequestEndpoints(this WebApplication ap
             ProducesProblem(StatusCodes.Status401Unauthorized);
     }
 
+    [AllowAnonymous]
```
keegan-caruso:
```suggestion
```

keegan-caruso: Instead, add             

AllowAnonymous().

To the builder call above chained off of MapPost

jennyf19 (approved):

PR #3483, opened 2025-09-18
add downstream api

keegan-caruso:
Add downstream api endpoint

### DIFF
--- a/src/Microsoft.Identity.Web.Sidecar/Microsoft.Identity.Web.Sidecar.http
+++ b/src/Microsoft.Identity.Web.Sidecar/Microsoft.Identity.Web.Sidecar.http
@@ -1,5 +1,7 @@
 @Microsoft.Identity.Web.Sidecar_HostAddress = http://localhost:5178
 
+@AccessToken = token_here
+
 GET {{Microsoft.Identity.Web.Sidecar_HostAddress}}/openapi/v1.json
 Accept: application/json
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.Identity.Web.Sidecar/Microsoft.Identity.Web.Sidecar.http
```diff
@@ -1,5 +1,7 @@
 @Microsoft.Identity.Web.Sidecar_HostAddress = http://localhost:5178
 
+@AccessToken = token_here
```
keegan-caruso, line 3: using the env file was unreliable, it wouldn't always use an updated value

src/Microsoft.Identity.Web.Sidecar/Endpoints/AuthorizationHeaderEndpoint.cs
```diff
@@ -71,7 +72,7 @@ private static async Task<Results<Ok<AuthorizationHeaderResult>, ProblemHttpResu
                 options.Scopes,
                 options,
                 httpContext.User,
-                httpContext.RequestAborted);
+                cancellationToken);
```
jmprieur, line 75: why this change?

jmprieur (approved):
LGTM
Thanks @keegan-caruso

src/Microsoft.Identity.Web.Sidecar/Endpoints/AuthorizationHeaderEndpoint.cs
keegan-caruso: I'm letting the framework abstract this for me. I don't care that it is cancelled from the request aborted specifically, I just want it to respond to cancellations that the framework provides

PR #3500, opened 2025-09-26
Add tests and fixes for downstream api sidecar endpoint

keegan-caruso:
# Add tests and fixes for downstream api sidecar endpoint

## Description

Changes were required for the downstream api endpoint.
Options override cannot be done through body as we need to
be able to forward the body of a request sent to the service

### DIFF
--- a/src/Microsoft.Identity.Web.Sidecar/Models/DownstreamApiRequest.cs
+++ b/src/Microsoft.Identity.Web.Sidecar/Models/DownstreamApiRequest.cs
@@ -1,0 +1,22 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Identity.Abstractions;
+
+namespace Microsoft.Identity.Web.Sidecar.Models;
+
+/// <summary>
+/// Represents the inputs to the downstream API endpoint.
+/// </summary>
+public readonly struct DownstreamApiRequest
+{
+    [FromQuery]
+    public string? AgentIdentity { get; init; }
+
+    [FromQuery]
+    public string? AgentUsername { get; init; }
+
+    [FromQuery]
+    public string? Tenant { get; init; }
+}
--- a/src/Microsoft.Identity.Web.Sidecar/OpenAPI/Microsoft.Identity.Web.Sidecar.json
+++ b/src/Microsoft.Identity.Web.Sidecar/OpenAPI/Microsoft.Identity.Web.Sidecar.json
@@ -146,13 +146,20 @@
             "schema": {
               "type": "string"
             }
+          },
+          {
+            "name": "Tenant",
+            "in": "query",
+            "schema": {
+              "type": "string"
+            }
           }
         ],
         "requestBody": {
           "content": {
             "application/json": {
               "schema": {
-                "$ref": "#/components/schemas/DownstreamApiOptions"
+                "$ref": "#/components/schemas/DownstreamApiOptions2"
               }
             }
           }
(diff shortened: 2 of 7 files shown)


tests/Directory.Build.props
```diff
@@ -38,6 +38,7 @@
     <MicrosoftIdentityLabApiVersion>1.0.2</MicrosoftIdentityLabApiVersion>
     <SystemNetHttpVersion>4.3.4</SystemNetHttpVersion>
     <SystemTextRegularExpressions>4.3.1</SystemTextRegularExpressions>
+    <MoqVersion>4.18.4</MoqVersion>
```
keegan-caruso, line 41: added as this is called out as part of the testing strategy in agents.md

src/Microsoft.Identity.Web.Sidecar/Models/DownstreamApiRequest.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Microsoft.Identity.Web.Sidecar.Models;
+
+/// <summary>
+/// Represents the inputs to the downstream API endpoint.
+/// </summary>
+public readonly struct DownstreamApiRequest
+{
+    [FromQuery]
+    public string? AgentIdentity { get; init; }
```
jennyf19, line 15: is this the right name? someone was floating around "blueprint"

jmprieur: It is
The blueprint is the app. The agent identity is a principal

keegan-caruso:
> Is it possible to override scopes? And AcquireTokenOptions?

@jmprieur Scopes yes, we could do this through query parameters. the binding should supported repeated keys or indexed keys e.g.

Repeated keys:
?scopes=read&scopes=write
Indexed keys:
?scopes[0]=read&scopes[1]=write

I'm not sure about AcquireTokenOptions as an object. It should be possible to pass a JSON object as a header would you like me to make that change?

src/Microsoft.Identity.Web.Sidecar/Models/DownstreamApiRequest.cs
```diff
@@ -0,0 +1,22 @@
...
+
+    [FromQuery]
+    public string? AgentUsername { get; init; }
+
+    [FromQuery]
+    public string? Tenant { get; init; }
+}
```
jmprieur, line 22: Could we add:

```csharp
[FromQuery]
DownstreamApiOptions DownstreamApiOptions {get;init}
```

jmprieur: Or otherwise:
- Scopes
- AuthorizationHeaderProviderOptions?

src/Microsoft.Identity.Web.Sidecar/OpenAPI/Microsoft.Identity.Web.Sidecar.json
```diff
@@ -146,13 +146,20 @@
...
+              "type": "string"
+            }
           }
         ],
         "requestBody": {
           "content": {
             "application/json": {
               "schema": {
-                "$ref": "#/components/schemas/DownstreamApiOptions"
+                "$ref": "#/components/schemas/DownstreamApiOptions2"
```
jmprieur, line 162: Why is there a 2 here, @keegan-caruso ?

keegan-caruso: Yeah, good question, this is how the generator created the spec. Two different input objects reference the same C# object (DownstreamApiOptions). That reference is created twice. Not sure why.

PR #3506, opened 2025-09-30
Allow multiple certificate observers

tlupes:
# Allow multiple certificate observers

Adds the ability to register multiple ICertificatesOberserver

## Description

Currently only the most recently registered certificates observer will be used. This can be brittle as the observer can be used for multiple purposes and consumers could accidentally overwrite an existing observer with a custom one or vice versa. There is no functional reason to only support one, so this change will send notifications to all registered observers.


jmprieur (approved):
LGTM
Maybe a small unit test?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -58,9 +58,12 @@ class OAuthConstants
         protected readonly IServiceProvider _serviceProvider;
         protected readonly ITokenAcquisitionHost _tokenAcquisitionHost;
         protected readonly ICredentialsLoader _credentialsLoader;
-        protected readonly ICertificatesObserver? _certificatesObserver;
+        protected readonly ICertificatesObserver[] _certificatesObservers;
```
keegan-caruso: Can this be IReadOnlyList<ICertificatesObserver> or another immutable collection?

Do derived classes need to be able to modify the contents of the collection?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -106,7 +109,10 @@ public TokenAcquisition(
             _serviceProvider = serviceProvider;
             _tokenAcquisitionHost = tokenAcquisitionHost;
             _credentialsLoader = credentialsLoader;
+            _certificatesObservers = [.. serviceProvider.GetServices<ICertificatesObserver>()];
+#pragma warning disable CS0618 // Type or member is obsolete. Setup for backward compatibility.
             _certificatesObserver = serviceProvider.GetService<ICertificatesObserver>();
```
keegan-caruso, line 114: checking my understanding, this isn't used in this class because any additions here would be handled by the collection?

tests/Microsoft.Identity.Web.Test/CertificatesObserverTests.cs
```diff
@@ -0,0 +1,389 @@
...
+using Microsoft.Identity.Abstractions;
+using Microsoft.Identity.Web.Experimental;
+using Xunit;
+
+namespace Microsoft.Identity.Web.Test
+{
+    public class CertificatesObserverTests
+    {
+        [Fact]
+        public async Task ObesrverSendsCorrectEvents()
```
keegan-caruso:
```suggestion
        public async Task ObserverSendsCorrectEvents()
```

keegan-caruso (approved):
Few comments, otherwise, looks correct

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
tlupes: Correct. I don't know if I can remove it as I don't fully understand why you are tracking the internal API.

keegan-caruso: tracking the internal API is because a few places use friend assemblies. That effectively means those assemblies need to treat internals as publics

keegan-caruso (approved):

PR #3508, opened 2025-10-01
Sidecar python adapter devapp

keegan-caruso:
# Sidecar python adapter devapp

## Description

Generated client from co-pilot over the open api spec.


tests/DevApps/SidecarAdapter/python/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+Display the available commands:
+
+```sh
+uv run --with requests main.py --help
+```
+
+Example: validate an authorization header returned by `get_token.py`:
+
+```sh
```
saurabhsathe-ms, line 34: Are these commands shell specific? If not, is it possible to make it shell independent like maybe a python script that demonstrates the process?

tests/DevApps/SidecarAdapter/python/README.md
```diff
@@ -0,0 +1,40 @@
+# Sidecar Adapter Python
+
+This folder contains helper scripts for interacting with the Microsoft Identity Web Sidecar from Python.
```
saurabhsathe-ms, line 3: We should have a requirements section here I guess like need to install uv and python package requests.

tests/DevApps/SidecarAdapter/python/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Contents
+
+- `MicrosoftIdentityWebSidecarClient.py` – Typed client covering the Sidecar's `/Validate`, `/AuthorizationHeader`, and `/DownstreamApi` endpoints.
+- `main.py` – Command-line harness that exercises the client and prints JSON responses.
+- `get_token.py` – Helper for obtaining a user token via MSAL.
+```
+
+## Usage
+
```
saurabhsathe-ms, line 17: Since the below commands work with windows powershell should we add a command saying please run them in powershell?

keegan-caruso: these should work in linux shells, no? msal python has an issue with interactive flows on WSL, yes.

PR #3559, opened 2025-10-16
[POC] Use secure HTTP client factory and force certificate credential for token binding

cpp11nullptr:


src/Microsoft.Identity.Web.TokenAcquisition/MsalMtlsHttpClientFactory.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System.Net.Http;
+using System.Security.Cryptography.X509Certificates;
+using Microsoft.Identity.Client;
+
+namespace Microsoft.Identity.Web
+{
+    internal sealed class MsalMtlsHttpClientFactory : IMsalMtlsHttpClientFactory
```
bgavrilMS: I don't understand the purpose of this class. It derives from IMsalMtlsHttpClientFactory and encapsulates an IMsalMtlsHttpClientFactory . 

I think what you mean is for a class that dervices from IMsalMtlsHttpClientFactory but it should encapsulate [System.Net.Http](https://learn.microsoft.com/en-us/dotnet/api/system.net.http?view=net-9.0-pp)IHttpClientFactory

cpp11nullptr: I consider it more like an adaptor because IMsalMtlsHttpClientFactory doesn't derive from IHttpClientFactory, which means that it this class needed need to implement what SecureHttpClientFactory does but we'd have a code duplication between MSAL.NET and Identity.Web in this case. I don't have strong opinion there though, so it's better to have encapsulated IHttpClientFactory rather than IMsalMtlsHttpClientFactory, is it correct?

src/Microsoft.Identity.Web.TokenAcquisition/ConfidentialClientApplicationBuilderExtension.cs
```diff
@@ -44,6 +45,17 @@ public static async Task<ConfidentialClientApplicationBuilder> WithClientCredent
                 return builder;
             }
 
+            // regardless of credential type being set, bound token requires a certificate
+            if (isTokenBinding)
+            {
+                if (credential.Certificate == null)
```
bgavrilMS, line 51: I think this works for a POC, but we probably need more abstraction here. Each credential may, or may not, be compatible with mTLS POP. In particular, I am thinking of the 2 FIC credentials. 

CC @jmprieur @MZOLN

src/Microsoft.Identity.Web.TokenAcquisition/MsalMtlsHttpClientFactory.cs
bgavrilMS: I think the simplest for now is to intervene here: https://github.com/AzureAD/microsoft-identity-web/blob/a8b927344ae71d41f8518b19778f86c21d095001/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs#L933 - if mtls POP is used, don't add `WithHttpClientFactory`. 

If you do want to inject HTTP transport, then you must provide an adapter between IHttpClientFactory and IMsalMtlsHttpClientFactory i.e.

```csharp
public class Impl : IMsalMtlsHttpClientFactory 
{
  Impl(IHttpClientFactory aspFactory) { }
}

```

However, I am not sure how / if this ^^ adapter can be implemnted.

cpp11nullptr: I changed it to use IHttpClientFactory for non-mTLS HTTP clients with managing a pool of mTLS clients, which are created in place (due to a fact that its handler needs to be configured before an actual instance is created).

PR #3562, opened 2025-10-17
Correctly compute Application Key when credential usage fails

tlupes:
# Correctly compute Application Key when credential usage fails

Modify `GetApplicationKey` to use the same computation in both `GetOrBuildConfidentialClientApplicationAsync` as well as various catch statements.

## Description

If a certificate does not work, then the library is supposed to invalidate the cache for the certificate so that a new certificate can be used on the next call. However, different keys are used for these purposes, so the cache is never correctly invalidated. This means that the retry behavior will never work as intended and certificates need to be explicitly rotated.

This PR adjusts cache key generation to be consistent throughout.

In addition, it changes where the keys are generated to ensure that the correct value is used. This ensures that the correct key is used. For example, when a certificate is loaded or reset, the cache key changes which causes the code to not perform as expected.

### DIFF
--- a/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
@@ -919,6 +919,9 @@ namespace Microsoft.Identity.Web
 
                 // Build and store the application
                 var newApp = await BuildConfidentialClientApplicationAsync(mergedOptions);
+
+                // Recompute the key as BuildConfidentialClientApplicationAsync can cause it to change.
+                key = GetApplicationKey(mergedOptions);
                 _applicationsByAuthorityClientId[key] = newApp;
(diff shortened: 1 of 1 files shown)


tlupes:
I am going to assume that the CertificateRotationTest.cs would fail, however as they rely on external resources which aren't currently functional, I assume they haven't been run in a while. The observer tests (which I modified here) were just looking for the events to occur, but I did not have it set up to check that the correct certs were being rotated.

Once changed to check the certs, it became apparent that the certificate was always null on retry as it was cleared but never correctly retried. The observer tests now function well as end-to-end rotation tests; I will probably come back at a later time and move them into CertificateRotationTest.cs for clarity.

As for why they failed on the MISE tests, I'm not entirely sure. I'll take a look to see what configuration is different there, however the fault there was simply because the events came in a slightly different order than before which caused the tests to fail.

PR #3583, opened 2025-11-03
Bump MSAL to version 4.79.2 and handle changes to deprecated WithExtraQueryParameters APIs

Avery-Dunn:
In MSAL.NET the `WithExtraQueryParameters` APIs are receiving two changes: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5536
- Adding the extra query parameters to the cache keys, so that the correct token can be uniquely found
- Deprecating the existing APIs in favor of one that can allow a customer to control the cache key behavior

This PR updates to the latest MSAL.NET version and handles the new API and cache behavior. It does so by making changes to the `TokenAcquisition` class, mainly by adjusting the existing `MergeExtraQueryParameters` helper method to return the `Dictionary` style that the new `WithExtraQueryParameters` API expects.

The changes were tested with a locally built version of MSAL.NET, and all existing tests that dealt with extra query parameters worked with minimal or no changes (in `TokenAcquisitionAuthorityTests`, `ExtraParametersTests`, `DownstreamApiOptionsMergeTests`, and `DownstreamApiEndpointTests`)

### DIFF
--- a/.github/ISSUE_TEMPLATE/sidecar_issue.md
+++ b/.github/ISSUE_TEMPLATE/sidecar_issue.md
@@ -1,0 +1,23 @@
+---
+name: Entra ID sidecar issue
+about: Submit a bug or feature request for the Entra ID sidecar
+title: "[Entra ID sidecar] "
+labels: sidecar, needs attention, untriaged
+assignees: ''
+
+---
+
+**Is your feature request or bug related to a problem? Please describe.**
+A clear and concise description of what the problem is. Ex. I'm always frustrated when [...].
+
+**Describe the solution you'd like**
+A clear and concise description of what you want to happen.
+
+**Describe alternatives you've considered**
+A clear and concise description of any alternative solutions or features you've considered.
+
+**Additional context**
+Add any other context or screenshots about the feature request here.
+
+**Are you using a language adapter to call the sidecar?**
+Can you provide a repro of the issue?


bgavrilMS (approved):

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -1362,25 +1361,40 @@ private Task<AuthenticationResult> GetAuthenticationResultForWebAppWithAccountFr
...
-            if (tokenAcquisitionOptions.ExtraQueryParameters != null)
+            // Return null if both sources are empty
+            if (tokenAcquisitionOptions?.ExtraQueryParameters == null && mergedOptions.ExtraQueryParameters == null)
             {
-                var mergedDict = new Dictionary<string, string>(tokenAcquisitionOptions.ExtraQueryParameters);
-                if (mergedOptions.ExtraQueryParameters != null)
+                return null;
+            }
+
+            var mergedDict = new Dictionary<string, (string value, bool includeInCacheKey)>(StringComparer.OrdinalIgnoreCase);
```
jmprieur, line 1374: Do you think we would have more in the future?
Would it deserve a little structure instead of the tuple?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -1362,25 +1361,40 @@ private Task<AuthenticationResult> GetAuthenticationResultForWebAppWithAccountFr
...
+
+            // Add from tokenAcquisitionOptions first (these take precedence)
+            if (tokenAcquisitionOptions?.ExtraQueryParameters != null)
+            {
+                foreach (var pair in tokenAcquisitionOptions.ExtraQueryParameters)
+                {
+                    mergedDict[pair.Key] = (pair.Value, true);
```
jmprieur, line 1381: So if partners provide EQP, we always consider them part of the cache key?

jmprieur (commented):
LGTM
I left a couple of comments

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
Avery-Dunn: The public API has already been released in MSAL: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5536

We discussed other structures a bit, but couldn't think of a use case that would need such extendibility. We can always overload it in the future if we need to (just like the old API), and will likely make it default to "true" and remove the boolean in the next major MSAL version anyway.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
Avery-Dunn: Yes, we decided to have ID web always add them to the cache key. The alternatives would've been either:
1) Add a duplicate public API in ID web to mimic MSAL's, and do all the same deprecation here as well
2) Default to false, and let the caching issue still be a problem if using ID web

This does potentially change the existing behavior of cache lookups, but it's more like an intentional bugfix than an unintentional breaking change.

neha-bhargava (approved):

Directory.Build.props
```diff
@@ -95,7 +95,7 @@
     <!--CVE-2024-43485-->
     <SystemTextJsonVersion>8.0.5</SystemTextJsonVersion>
     <!--CVE-2023-29331-->
-    <SystemFormatsAsn1Version>8.0.1</SystemFormatsAsn1Version>
+    <SystemFormatsAsn1Version>9.0.8</SystemFormatsAsn1Version>
```
jmprieur: @Avery-Dunn: was this necessary? In general we try to keep the dependencies minimal?

bgavrilMS: Ah, this one is a bit tricky. We took a dependency on version 9.0.8 on MSAL for this. But I don't think it is needed, the code does compile ok with 8.0.1 ... let's check.

bgavrilMS: @Avery-Dunn @Robbie-Microsoft - can you please check ? We may need yet another MSAL release for this.

Avery-Dunn: Yeah, as @bgavrilMS said that dependency was added to MSAL.NET a while back, but it appears like we can use 8.0.1 over than and keep the version here in ID Web as-is. I've just made this PR in MSAL.NET: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5583

neha-bhargava (approved):

Directory.Build.props
Avery-Dunn: Released an MSAL hotfix so we no longer need to update that version here.

jmprieur (approved):

PR #3586, opened 2025-11-10
Add token binding scenario

cpp11nullptr:
## Purpose

Introduce a “token binding scenario” for confidential client (app-only) token acquisition and downstream API calls:

- Bind access tokens to a client certificate (expected `cnf` claim – thumbprint / key reference).
- Use mutual TLS (mTLS) when calling downstream APIs.
- Support Bearer / PoP / mTLS protocol selection via `AuthorizationHeaderProviderOptions.ProtocolScheme`.
- Provide a service collection extension (one-line enablement) wiring token acquisition + binding + downstream calls.

## High-level scenario

1. Developer enables token binding scenario via new service collection extension (PR states: “add service collection extension to enable token binding”).
2. A binding certificate is sourced from ClientCredentials (direct cert or via Managed Identity).
3. Developer specifies a protocol scheme (e.g. `ProtocolScheme = "Pop"` or future `ProtocolScheme = "mtls-pop"` / `"mtls"`) in `AuthorizationHeaderProviderOptions`.
4. Token acquisition (client credentials flow) is configured accordingly; Entra ID returns a token containing a binding claim (`cnf`) when certificate binding is active.
5. The authorization header provider produces the header (currently `Bearer <token>`; protocol-specific transformations for schemes like Pop are applied internally).
6. An mTLS-capable HTTP client presents the same certificate during TLS handshake.
7. Downstream API validates that presented client certificate matches token’s `cnf` claim (x5t thumbprint).
8. Fallback path: if no certificate is available (or protocol does not require it), a regular unbound token is retrieved.

### DIFF
--- a/tests/DevApps/MtlsPop/MtlsPopClient/Program.cs
+++ b/tests/DevApps/MtlsPop/MtlsPopClient/Program.cs
@@ -1,0 +1,55 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using Microsoft.Graph;
+using Microsoft.Identity.Abstractions;
+using Microsoft.Identity.Web;
+
+namespace MtlsPopSample
+{
+    public class Program
+    {
+        public static async Task Main(string[] args)
+        {
+            var tokenAcquirerFactory = TokenAcquirerFactory.GetDefaultInstance();
+
+            tokenAcquirerFactory.Services.AddLogging(builder =>
+                builder.AddConsole().SetMinimumLevel(LogLevel.Information));
+
+            tokenAcquirerFactory.Services.AddDownstreamApi("WebApi",
+                tokenAcquirerFactory.Configuration.GetSection("WebApi"));
+
+            tokenAcquirerFactory.Services.AddMicrosoftGraph();
+
+            var sp = tokenAcquirerFactory.Build();
+
+            Console.WriteLine("Scenario 1: calling web API with mTLS PoP token...");
+            var webApi = sp.GetRequiredService<IDownstreamApi>();
+            var result = await webApi.GetForAppAsync<IEnumerable<WeatherForecast>>("WebApi",
+                options =>
+                {
+                    options.AcquireTokenOptions.ExtraParameters ??= new Dictionary<string, object>();
+                    options.AcquireTokenOptions.ExtraParameters["RequestBoundToken"] = true; // mTLS PoP
+                }).ConfigureAwait(false);
+
+            Console.WriteLine("Web API result:");
+            foreach (var forecast in result!)
+            {
+                Console.WriteLine($"{forecast.Date}: {forecast.Summary} - {forecast.TemperatureC}C/{forecast.TemperatureF}F");
+            }
+
+            Console.WriteLine();
+            Console.WriteLine("Scenario 2: Calling Microsoft Graph with Bearer (non mTLS PoP) token...");
+            var graphServiceClient = sp.GetRequiredService<GraphServiceClient>();
+            var users = await graphServiceClient.Users
+                .Request()
+                .WithAppOnly()
+                .GetAsync();
+
+            Console.WriteLine("Microsoft Graph result:");
+            Console.WriteLine($"{users.Count} users");
+        }
+    }
+}
--- a/tests/DevApps/MtlsPop/MtlsPopClient/MtlsPopClient.csproj
+++ b/tests/DevApps/MtlsPop/MtlsPopClient/MtlsPopClient.csproj
@@ -1,0 +1,31 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>net8.0</TargetFramework>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <Nullable>enable</Nullable>
+    <UseWip>true</UseWip>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Extensions.Logging.Console" Version="8.0.0" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(UseWip)' != 'true'">
+    <PackageReference Include="Microsoft.Identity.Web.DownstreamApi" Version="3.6.2" />
+    <PackageReference Include="Microsoft.Identity.Web.MicrosoftGraph" Version="3.6.2" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(UseWip)' == 'true'">
+    <ProjectReference Include="..\..\..\..\src\Microsoft.Identity.Web.DownstreamApi\Microsoft.Identity.Web.DownstreamApi.csproj" />
+    <ProjectReference Include="..\..\..\..\src\Microsoft.Identity.Web.MicrosoftGraph\Microsoft.Identity.Web.MicrosoftGraph.csproj" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Content Include="appsettings.json">
+      <CopyToOutputDirectory>Always</CopyToOutputDirectory>
+    </Content>
+  </ItemGroup>
+
+</Project>
--- a/tests/DevApps/MtlsPop/MtlsPopClient/WeatherForecast.cs
+++ b/tests/DevApps/MtlsPop/MtlsPopClient/WeatherForecast.cs
@@ -1,0 +1,16 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+namespace MtlsPopSample
+{
+    public class WeatherForecast
+    {
+        public DateOnly Date { get; set; }
+
+        public int TemperatureC { get; set; }
+
+        public int TemperatureF => 32 + (int)(TemperatureC / 0.5556);
+
+        public string? Summary { get; set; }
+    }
+}
--- a/tests/DevApps/MtlsPop/MtlsPopClient/appsettings.json
+++ b/tests/DevApps/MtlsPop/MtlsPopClient/appsettings.json
@@ -1,0 +1,22 @@
+{
+    "AzureAd": {
+        "Instance": "https://login.microsoftonline.com/",
+        "TenantId": "bea21ebe-8b64-4d06-9f6d-6a889b120a7c",
+        "ClientId": "163ffef9-a313-45b4-ab2f-c7e2f5e0e23e",
+        "AzureRegion": "westus3",
+        "ClientCredentials": [
+            {
+                "SourceType": "StoreWithDistinguishedName",
+                "CertificateStorePath": "CurrentUser/My",
+                "CertificateDistinguishedName": "CN=LabAuth.MSIDLab.com"
+            }
+        ],
+        "SendX5c": true
+    },
+    "WebApi": {
+        "BaseUrl": "https://localhost:7060/",
+        "RelativePath": "WeatherForecast",
+        "RequestAppToken": true,
+        "Scopes": [ "https://graph.microsoft.com/.default" ]
+    }
+}
--- a/tests/DevApps/MtlsPop/MtlsPopWebApi/Controllers/WeatherForecastController.cs
+++ b/tests/DevApps/MtlsPop/MtlsPopWebApi/Controllers/WeatherForecastController.cs
@@ -1,0 +1,39 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.Identity.Web.Resource;
+
+namespace MtlsPopSample.Controllers
+{
+    [ApiController]
+    [Route("[controller]")]
+    [Authorize(Policy = "MtlsPop")]
+    public class WeatherForecastController : ControllerBase
+    {
+        private static readonly string[] Summaries = new[]
+        {
+            "Freezing", "Bracing", "Chilly", "Cool", "Mild", "Warm", "Balmy", "Hot", "Sweltering", "Scorching"
+        };
+
+        private readonly ILogger<WeatherForecastController> _logger;
+
+        public WeatherForecastController(ILogger<WeatherForecastController> logger)
+        {
+            _logger = logger;
+        }
+
+        [HttpGet(Name = "GetWeatherForecast")]
+        public IEnumerable<WeatherForecast> Get()
+        {
+            return Enumerable.Range(1, 5).Select(index => new WeatherForecast
+            {
+                Date = DateOnly.FromDateTime(DateTime.Now.AddDays(index)),
+                TemperatureC = Random.Shared.Next(-20, 55),
+                Summary = Summaries[Random.Shared.Next(Summaries.Length)]
+            })
+            .ToArray();
+        }
+    }
+}
--- a/tests/DevApps/MtlsPop/MtlsPopWebApi/MtlsPopAuthorizationHandler.cs
+++ b/tests/DevApps/MtlsPop/MtlsPopWebApi/MtlsPopAuthorizationHandler.cs
@@ -1,0 +1,120 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.IdentityModel.Tokens;
+using System.IdentityModel.Tokens.Jwt;
+using System.Security.Cryptography;
+using System.Security.Cryptography.X509Certificates;
+using System.Text.Json;
+
+namespace MtlsPopSample
+{
+    public class MtlsPopRequirement : IAuthorizationRequirement
+    {
+    }
+
+    public class MtlsPopAuthorizationHandler : AuthorizationHandler<MtlsPopRequirement>
+    {
+        private const string ProtocolName = "MTLS_POP";
+
+        private readonly ILogger<MtlsPopAuthorizationHandler> _logger;
+
+        public MtlsPopAuthorizationHandler(ILogger<MtlsPopAuthorizationHandler> logger)
+        {
+            _logger = logger;
+        }
+
+        protected override Task HandleRequirementAsync(AuthorizationHandlerContext context, MtlsPopRequirement requirement)
+        {
+            _logger.LogInformation("MtlsPopAuthorizationHandler invoked");
+
+            if (context.Resource is not HttpContext httpContext)
+            {
+                _logger.LogWarning("Resource is not HttpContext");
+                return Task.CompletedTask;
+            }
+
+            var authHeader = httpContext.Request.Headers.Authorization.FirstOrDefault();
+            var authToken = !string.IsNullOrEmpty(authHeader) && authHeader.StartsWith($"{ProtocolName} ", StringComparison.OrdinalIgnoreCase)
+                ? authHeader.Substring($"{ProtocolName} ".Length).Trim()
+                : null;
+
+            if (string.IsNullOrEmpty(authToken))
+            {
+                _logger.LogWarning("No auth token found");
+                return Task.CompletedTask;
+            }
+
+            try
+            {
+                var handler = new JwtSecurityTokenHandler();
+                var token = handler.ReadJwtToken(authToken);
+
+                var cnfClaim = token.Claims.FirstOrDefault(c => c.Type == "cnf");
+                if (cnfClaim == null)
+                {
+                    _logger.LogWarning("mTLS PoP token does not contain 'cnf' claim");
+                    context.Fail(new AuthorizationFailureReason(this, "Missing 'cnf' claim in  token"));
+                    return Task.CompletedTask;
+                }
+
+                _logger.LogInformation($"The 'cnf' claim value: {cnfClaim.Value}");
+
+                var cnfJson = JsonDocument.Parse(cnfClaim.Value);
+                if (!cnfJson.RootElement.TryGetProperty("x5t#S256", out var x5tS256Element))
+                {
+                    _logger.LogWarning("The 'cnf' claim does not contain 'x5t#S256' property");
+                    context.Fail(new AuthorizationFailureReason(this, "Missing 'x5t#S256' property in mTLS PoP 'cnf' claim"));
+                    return Task.CompletedTask;
+                }
+
+                var x5tS256 = x5tS256Element.GetString();
+                if (string.IsNullOrEmpty(x5tS256))
+                {
+                    _logger.LogWarning("The 'cnf' claim contains an empty 'x5t#S256' property");
+                    context.Fail(new AuthorizationFailureReason(this, "Empty 'x5t#S256' property in mTLS PoP 'cnf' claim"));
+                    return Task.CompletedTask;
+                }
+
+                _logger.LogInformation($"Token bound to certificate with x5t#S256: {x5tS256}");
+
+                var clientCert = httpContext.Connection.ClientCertificate;
+                if (clientCert != null)
+                {
+                    var certThumbprint = GetCertificateThumbprint(clientCert);
+                    _logger.LogInformation($"Client cert thumprint: {certThumbprint}");
+
+                    if (!string.Equals(certThumbprint, x5tS256, StringComparison.OrdinalIgnoreCase))
+                    {
+                        _logger.LogWarning($"Mismatch between cert thumbprint and 'x5t#S256' from mTLS PoP 'cnf' claim property: cert thumbprint - {certThumbprint}, x5t#S256 = {x5tS256}");
+                        context.Fail(new AuthorizationFailureReason(this, "Cert thumbprint and mTLS PoP 'cnf' claim 'x5t#S256' property mismatch"));
+                        return Task.CompletedTask;
+                    }
+
+                    _logger.LogInformation("mTLS PoP token validation successful");
+                }
+                else
+                {
+                    _logger.LogInformation("No client certificate in request");
+                }
+
+                context.Succeed(requirement);
+            }
+            catch (Exception ex)
+            {
+                _logger.LogError(ex, "Error validating mTLS PoP token");
+                context.Fail(new AuthorizationFailureReason(this, $"mTLS PoP validation error: {ex.Message}"));
+            }
+
+            return Task.CompletedTask;
+        }
+
+        private static string GetCertificateThumbprint(X509Certificate2 certificate)
+        {
+            using var sha256 = SHA256.Create();
+            var hash = sha256.ComputeHash(certificate.RawData);
+            return Base64UrlEncoder.Encode(hash);
+        }
+    }
+}
--- a/tests/DevApps/MtlsPop/MtlsPopWebApi/MtlsPopWebApi.csproj
+++ b/tests/DevApps/MtlsPop/MtlsPopWebApi/MtlsPopWebApi.csproj
@@ -1,0 +1,22 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+
+  <PropertyGroup>
+    <TargetFramework>net8.0</TargetFramework>
+    <Nullable>enable</Nullable>
+    <ImplicitUsings>enable</ImplicitUsings>
+    <UseWip>true</UseWip>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.AspNetCore.OpenApi" Version="8.0.10" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(UseWip)' != 'true'">
+    <PackageReference Include="Microsoft.Identity.Web" Version="3.6.2" />
+  </ItemGroup>
+
+  <ItemGroup Condition="'$(UseWip)' == 'true'">
+    <ProjectReference Include="..\..\..\..\src\Microsoft.Identity.Web\Microsoft.Identity.Web.csproj" />
+  </ItemGroup>
+
+</Project>
--- a/tests/DevApps/MtlsPop/MtlsPopWebApi/Program.cs
+++ b/tests/DevApps/MtlsPop/MtlsPopWebApi/Program.cs
@@ -1,0 +1,172 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Microsoft.AspNetCore.Authentication.JwtBearer;
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.Identity.Web;
+
+namespace MtlsPopSample
+{
+    public class Program
+    {
+        public static void Main(string[] args)
+        {
+            var builder = WebApplication.CreateBuilder(args);
+
+            builder.Services.AddControllers();
+
+            // Learn more about configuring OpenAPI at https://learn.microsoft.com/aspnet/core/fundamentals/openapi/aspnetcore-openapi
+            builder.Services.AddEndpointsApiExplorer();
+
+            builder.Services.AddMicrosoftIdentityWebApiAuthentication(builder.Configuration);
+
+            builder.Services.AddAuthorization(options =>
+            {
+                options.AddPolicy("MtlsPop", policy =>
+                    policy.Requirements.Add(new MtlsPopRequirement()));
+            });
+
+            builder.Services.AddSingleton<IAuthorizationHandler, MtlsPopAuthorizationHandler>();
+
+            var app = builder.Build();
+
+            app.UseHttpsRedirection();
+
+            app.UseAuthentication();
+            app.UseAuthorization();
+
+            app.MapControllers();
+
+            app.Run();
+        }
+    }
+}
+
+/*
+
+using Microsoft.AspNetCore.Authentication.JwtBearer;
+using Microsoft.Identity.Web;
+using Microsoft.IdentityModel.Tokens;
+using System.IdentityModel.Tokens.Jwt;
+using System.Security.Cryptography;
+using System.Security.Cryptography.X509Certificates;
+
+var builder = WebApplication.CreateBuilder(args);
+
+// Add services to the container.
+builder.Services.AddControllers();
+builder.Services.AddEndpointsApiExplorer();
+
+// Configure Microsoft Identity Web API authentication
+builder.Services.AddMicrosoftIdentityWebApiAuthentication(builder.Configuration, "AzureAd", "Bearer", true);
+
+// Configure custom mTLS PoP token validation
+builder.Services.Configure<JwtBearerOptions>(JwtBearerDefaults.AuthenticationScheme, options =>
+{
+    var existingOnTokenValidated = options.Events?.OnTokenValidated;
+
+    options.Events ??= new JwtBearerEvents();
+
+    options.Events.OnTokenValidated = async context =>
+    {
+        // Call the existing handler first (if any)
+        if (existingOnTokenValidated != null)
+        {
+            await existingOnTokenValidated(context);
+        }
+
+        var logger = context.HttpContext.RequestServices.GetRequiredService<ILogger<Program>>();
+
+        try
+        {
+            // Get the JWT token
+            var token = context.SecurityToken as JwtSecurityToken;
+            if (token == null)
+            {
+                logger.LogWarning("SecurityToken is not a JwtSecurityToken");
+                context.Fail("Invalid token type");
+                return;
+            }
+
+            // Check for cnf claim (confirmation claim for mTLS PoP)
+            var cnfClaim = token.Claims.FirstOrDefault(c => c.Type == "cnf");
+            if (cnfClaim == null)
+            {
+                logger.LogWarning("Token does not contain cnf claim - not an mTLS PoP token");
+                context.Fail("Missing cnf claim - mTLS PoP token required");
+                return;
+            }
+
+            logger.LogInformation("Found cnf claim in token: {CnfValue}", cnfClaim.Value);
+
+            // Parse the cnf claim to get x5t#S256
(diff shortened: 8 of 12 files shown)


Directory.Build.props
```diff
@@ -213,4 +214,8 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers</IncludeAssets>
     </PackageReference>
   </ItemGroup>
+
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net8.0' or '$(TargetFramework)' == 'net9.0' or '$(TargetFramework)' == 'net472' or '$(TargetFramework)' == 'netstandard2.0'">
+    <DefineConstants>$(DefineConstants);SUPPORTS_MTLS;</DefineConstants>
```
jmprieur, line 205: Net4.6.2 won't support it? Maybe you'd want a negative condition instead (if not net462)

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
```diff
@@ -26,6 +27,12 @@ internal partial class DownstreamApi : IDownstreamApi
     {
         private readonly IAuthorizationHeaderProvider _authorizationHeaderProvider;
         private readonly IHttpClientFactory _httpClientFactory;
+
+        // This MSAL HTTP client factory is used to create HTTP clients with mTLS bindign certificate.
```
jmprieur:
```suggestion
        // This MSAL HTTP client factory is used to create HTTP clients with mTLS binding certificate.

```

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
```diff
@@ -38,15 +45,18 @@ internal partial class DownstreamApi : IDownstreamApi
...
         /// <param name="httpClientFactory">HTTP client factory.</param>
         /// <param name="logger">Logger.</param>
+        /// <param name="serviceProvider">The service provider.</param>
         public DownstreamApi(
             IAuthorizationHeaderProvider authorizationHeaderProvider,
             IOptionsMonitor<DownstreamApiOptions> namedDownstreamApiOptions,
             IHttpClientFactory httpClientFactory,
-            ILogger<DownstreamApi> logger)
+            ILogger<DownstreamApi> logger,
+            IServiceProvider serviceProvider)
```
jmprieur: You'd want to avoid the locator pattern. Rather add a constructor with the IMsalHttpClientFactory? and keep this one?

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
```diff
@@ -558,15 +573,37 @@ public Task<HttpResponseMessage> CallApiForAppAsync(
...
             if (effectiveOptions.Scopes != null && effectiveOptions.Scopes.Any())
             {
-                string authorizationHeader = await _authorizationHeaderProvider.CreateAuthorizationHeaderAsync(
-                       effectiveOptions.Scopes,
-                       effectiveOptions,
-                       user,
-                       cancellationToken).ConfigureAwait(false);
+                string authorizationHeader = string.Empty;
+
+                // Firtsly check if it's token binding scenario so authorization header provider will return
```
jmprieur:
```suggestion
                // Firtsly check if it's token binding scenario so authorization header provider returns

```

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderBoundProvider.cs
```diff
@@ -0,0 +1,89 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Security.Claims;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Identity.Abstractions;
+
+namespace Microsoft.Identity.Web
+{
+    internal sealed class DefaultAuthorizationHeaderBoundProvider : IAuthorizationHeaderProvider, IAuthorizationHeaderBoundProvider
```
jmprieur: We should release Abstractions to make your life easier

jmprieur: Why not change the DefaultAuthorizationHeaderProvider?

src/Microsoft.Identity.Web.TokenAcquisition/ServiceCollectionExtensions.cs
```diff
@@ -160,10 +163,78 @@ public static IServiceCollection AddTokenAcquisition(
...
 
+        /// <summary>
+        /// Enables token binding by registering required for it services in the service collection.
+        /// </summary>
+        /// <remarks>
+        /// This extension should be called when token acquisition is already configured.
+        /// </remarks>
+        /// <param name="services">The service collection to which the token binding support will be added. Must not be null.</param>
+        /// <returns>The same IServiceCollection instance, allowing for method chaining.</returns>
+        public static IServiceCollection EnableTokenBinding(this IServiceCollection services)
```
jmprieur: why not just provide this? there are 2 ways of doing the same thing? (calling this method, or using the boolean)

src/Microsoft.Identity.Web.TokenAcquisition/ServiceCollectionExtensions.cs
```diff
@@ -160,10 +163,78 @@ public static IServiceCollection AddTokenAcquisition(
...
+        {
+            ServiceDescriptor? msalHttpClientFactory = services.FirstOrDefault(s => s.ServiceType == typeof(IMsalHttpClientFactory));
+            if (msalHttpClientFactory != null)
+            {
+                services.Remove(msalHttpClientFactory);
```
jmprieur: Might require logging if we didn't add it (But the customer did)

src/Microsoft.Identity.Web.TokenAcquisition/PublicAPI/net462/PublicAPI.Shipped.txt
```diff
@@ -156,7 +156,6 @@ Microsoft.Identity.Web.TokenAcquisitionOptions.PoPConfiguration.set -> void
 Microsoft.Identity.Web.TokenAcquisitionOptions.TokenAcquisitionOptions() -> void
 static Microsoft.Identity.Web.Internal.WebApiBuilders.EnableTokenAcquisition(System.Action<Microsoft.Identity.Client.ConfidentialClientApplicationOptions!>! configureConfidentialClientApplicationOptions, string! authenticationScheme, Microsoft.Extensions.DependencyInjection.IServiceCollection! services, Microsoft.Extensions.Configuration.IConfigurationSection? configuration) -> Microsoft.Identity.Web.MicrosoftIdentityAppCallsWebApiAuthenticationBuilder!
 static Microsoft.Identity.Web.PrincipalExtensionsForSecurityTokens.GetBootstrapToken(this System.Security.Principal.IPrincipal! claimsPrincipal) -> Microsoft.IdentityModel.Tokens.SecurityToken?
-static Microsoft.Identity.Web.ServiceCollectionExtensions.AddTokenAcquisition(this Microsoft.Extensions.DependencyInjection.IServiceCollection! services, bool isTokenAcquisitionSingleton = false) -> Microsoft.Extensions.DependencyInjection.IServiceCollection!
```
bgavrilMS, line 159: What happened here?

src/Microsoft.Identity.Web.TokenAcquisition/ConfidentialClientApplicationBuilderExtension.cs
```diff
@@ -58,6 +58,29 @@ public static async Task<ConfidentialClientApplicationBuilder> WithClientCredent
...
+        {
+            var credential = await LoadCredentialForMsalOrFailAsync(
+                clientCredentials,
+                logger,
+                credentialsLoader,
+                credentialSourceLoaderParameters).ConfigureAwait(false);
+
+            if (credential?.Certificate == null)
+            {
+                logger.LogError("Loaded credentials for token binding doesn't contain a certificate");
```
bgavrilMS: I think in this case you should throw exception. I believe that `WithClientCredentialsAsync` returns the builder simply because the ASP.NET middleware has a different way to configure a secret. 

However, in the case of mtls pop, the credential needs to result in a certificate.

src/Microsoft.Identity.Web.TokenAcquisition/ConfidentialClientApplicationBuilderExtension.cs
bgavrilMS, line 28: Maybe add a boolean flag here like "forTokenBinding" or smth. 

And if the credential doesn't result in a certificate, then throw exception. Including on the null path at line 42 (which was intended for some obscure scnario where the secret is set somewhere else)

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderBoundProvider.cs
bgavrilMS: Pls add a link to your PR in abstractions with `IAuthorizationHeaderBoundProvider`

bgavrilMS: Have you considered using the existing `DefaultAuthorizationHeaderProvider` and having it implement the newer `IAuthorizationHeaderBoundProvider` ?

bgavrilMS: It seems to me that `IAuthorizationHeaderBoundProvider` is more general purpose that being tied to mtls pop. Maybe we need better name.

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderBoundProvider.cs
```diff
@@ -0,0 +1,89 @@
...
+            var newTokenAcquisitionOptions = TokenAcquisitionOptionsHelper.CreateTokenAcquisitionOptionsFromApiOptions(downstreamApiOptions, cancellationToken);
+
+            // Token binding flow currently supports only app tokens.
+            var tokenAcquisitionResult = await _tokenAcquisition.GetAuthenticationResultForAppAsync(
+                downstreamApiOptions.Scopes?.FirstOrDefault() ?? string.Empty,
+                downstreamApiOptions?.AcquireTokenOptions.AuthenticationOptionsName,
+                downstreamApiOptions?.AcquireTokenOptions.Tenant,
+                newTokenAcquisitionOptions).ConfigureAwait(false);
+
+            TokenAcquisitionOptionsHelper.UpdateOriginalTokenAcquisitionOptions(downstreamApiOptions?.AcquireTokenOptions, newTokenAcquisitionOptions);
```
bgavrilMS: I don't understand why this is needed. It seems to copy a LongRunningOBO property and this is a User flow.

bgavrilMS: Ah never mind, it's not your code.

src/Microsoft.Identity.Web.TokenAcquisition/ServiceCollectionExtensions.cs
```diff
@@ -37,7 +39,8 @@ public static class ServiceCollectionExtensions
         /// </example>
         public static IServiceCollection AddTokenAcquisition(
             this IServiceCollection services,
-            bool isTokenAcquisitionSingleton = false)
+            bool isTokenAcquisitionSingleton = false,
+            bool enableTokenBinding = false)
```
bgavrilMS: I don't think the end-user should have to add `AddTokenAcquisition(enableTokenBinding)` and then to also request POP tokens via the config / options.

src/Microsoft.Identity.Web.TokenAcquisition/ServiceCollectionExtensions.cs
```diff
@@ -160,10 +163,78 @@ public static IServiceCollection AddTokenAcquisition(
...
+        public static IServiceCollection EnableTokenBinding(this IServiceCollection services)
+        {
+            ServiceDescriptor? msalHttpClientFactory = services.FirstOrDefault(s => s.ServiceType == typeof(IMsalHttpClientFactory));
+            if (msalHttpClientFactory != null)
+            {
+                services.Remove(msalHttpClientFactory);
+            }
+
+            // Registering HTTP client factory which allows to create HTTP client with client certificate
+            services.AddSingleton<IMsalHttpClientFactory>(sp =>
```
bgavrilMS: This is fine, but if we can't get it to work at all, it's acceptable to not allow IHttpClient injection in the TokenAcquisition flow and even to disablow usage of DownstreamAPI + POP.

tests/Microsoft.Identity.Web.Test.Common/Mocks/MockMtlsHttpClientFactory.cs
```diff
@@ -0,0 +1,97 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Http;
+using System.Net.Http.Headers;
+using System.Runtime.InteropServices;
+using System.Security.Cryptography.X509Certificates;
+using Microsoft.Identity.Client;
+using Xunit;
+
+namespace Microsoft.Identity.Web.Test.Common.Mocks
```
bgavrilMS, line 14: It would be good to modify an existing dev app to showcase the developer experience. Ideally we'd have a dev app that can get bearer tokens for API1 and pop tokens for API2 (e.g. Graph)

Directory.Build.props
MZOLN, line 205: Is this the idea of a feature flag?

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderBoundProvider.cs
MZOLN: I'm not sure about the name, did we consult it with Jean-Marc?

MZOLN: Token acquisition requiring a new interface seems like an optimal design decision in my opinion? Why would we need a new interface?

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderProvider.cs
```diff
@@ -27,7 +27,7 @@ public async Task<string> CreateAuthorizationHeaderForUserAsync(
             ClaimsPrincipal? claimsPrincipal = null,
             CancellationToken cancellationToken = default)
         {
-            var newTokenAcquisitionOptions = CreateTokenAcquisitionOptionsFromApiOptions(downstreamApiOptions, cancellationToken);
+            var newTokenAcquisitionOptions = TokenAcquisitionOptionsHelper.CreateTokenAcquisitionOptionsFromApiOptions(downstreamApiOptions, cancellationToken);
```
MZOLN: why static helper? Couldnt we use some instance or private methods?

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderProvider.cs
```diff
@@ -93,43 +93,8 @@ public async Task<string> CreateAuthorizationHeaderAsync(
                     newTokenAcquisitionOptions).ConfigureAwait(false);
             }
 
-            UpdateOriginalTokenAcquisitionOptions(downstreamApiOptions?.AcquireTokenOptions, newTokenAcquisitionOptions);
+            TokenAcquisitionOptionsHelper.UpdateOriginalTokenAcquisitionOptions(downstreamApiOptions?.AcquireTokenOptions, newTokenAcquisitionOptions);
             return result.CreateAuthorizationHeader();
         }
-
-        private static TokenAcquisitionOptions CreateTokenAcquisitionOptionsFromApiOptions(
```
MZOLN, line 100: If we are extracting those out maybe they warrant a proper abstraction vs a static helper class. Maybe a TokenAcquisitonOptions factory or converter?

src/Microsoft.Identity.Web.TokenAcquisition/MsalMtlsHttpClientFactory.cs
```diff
@@ -0,0 +1,120 @@
...
+namespace Microsoft.Identity.Web
+{
+    /// <summary>
+    /// Provides a factory for creating HTTP clients configured for mTLS authentication with using binding certificate.
+    /// It uses a hybrid approach with leveraging IHttpClientFactory for non-mTLS HTTL clients and maintaining
+    /// a pool of mTLS clients with using certificate as a key.
+    /// </summary>
+    public sealed class MsalMtlsHttpClientFactory : IMsalMtlsHttpClientFactory
+    {
+        private const long MaxMtlsHttpClientCountInPool = 1000;
```
MZOLN, line 20: Those values should be configurable.

src/Microsoft.Identity.Web.TokenAcquisition/MsalMtlsHttpClientFactory.cs
```diff
@@ -0,0 +1,120 @@
...
+        private const long MaxResponseContentBufferSizeInBytes = 1024 * 1024;
+
+        // Please see (https://aka.ms/msal-httpclient-info) for important information regarding the HttpClient.
+        private static readonly ConcurrentDictionary<string, HttpClient> s_mtlsHttpClientPool = new ConcurrentDictionary<string, HttpClient>();
```
MZOLN, line 24: why s_?

src/Microsoft.Identity.Web.TokenAcquisition/MsalMtlsHttpClientFactory.cs
```diff
@@ -0,0 +1,120 @@
...
+
+            string key = x509Certificate2.Thumbprint;
+            HttpClient httpClient = CreateMtlsHttpClient(x509Certificate2);
+            httpClient = s_mtlsHttpClientPool.GetOrAdd(key, httpClient);
+            return httpClient;
+        }
+
+        private HttpClient CreateMtlsHttpClient(X509Certificate2 bindingCertificate)
+        {
+#if SUPPORTS_MTLS
```
MZOLN, line 80: wouldnt #if !WINDOWS suffice?

src/Microsoft.Identity.Web.TokenAcquisition/MsalMtlsHttpClientFactory.cs
```diff
@@ -0,0 +1,120 @@
...
+#else
+            throw new NotSupportedException("mTLS is not supported on this platform.");
+#endif
+        }
+
+        private static void CheckAndManageCache()
+        {
+            lock (s_cacheLock)
+            {
+                if (s_mtlsHttpClientPool.Count >= 1000)
```
MZOLN: def needs to be configurable

src/Microsoft.Identity.Web.TokenAcquisition/MsalMtlsHttpClientFactory.cs
```diff
@@ -0,0 +1,120 @@
...
+            string key = x509Certificate2.Thumbprint;
+            HttpClient httpClient = CreateMtlsHttpClient(x509Certificate2);
+            httpClient = s_mtlsHttpClientPool.GetOrAdd(key, httpClient);
+            return httpClient;
+        }
+
+        private HttpClient CreateMtlsHttpClient(X509Certificate2 bindingCertificate)
+        {
+#if SUPPORTS_MTLS
+            CheckAndManageCache();
```
MZOLN, line 81: Every thousand calls we will just  serialize all the threads and abruptly clear the cache? Wouldnt something like LRU be a better approach here?

This could create massive unexpected spikes in latency for users every now and then that we might get ICMs on

src/Microsoft.Identity.Web.TokenAcquisition/MsalMtlsHttpClientFactory.cs
```diff
@@ -0,0 +1,120 @@
...
+                    s_mtlsHttpClientPool.Clear();
+                }
+            }
+        }
+
+        private static void ConfigureRequestHeadersAndSize(HttpClient httpClient)
+        {
+            httpClient.MaxResponseContentBufferSize = MaxResponseContentBufferSizeInBytes;
+            httpClient.DefaultRequestHeaders.Accept.Clear();
+            httpClient.DefaultRequestHeaders.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
```
MZOLN, line 116: Is that always true?

src/Microsoft.Identity.Web.TokenAcquisition/MsalMtlsHttpClientFactory.cs
```diff
@@ -0,0 +1,120 @@
...
+using System.Net.Http.Headers;
+using System.Security.Cryptography.X509Certificates;
+using Microsoft.Identity.Client;
+
+namespace Microsoft.Identity.Web
+{
+    /// <summary>
+    /// Provides a factory for creating HTTP clients configured for mTLS authentication with using binding certificate.
+    /// It uses a hybrid approach with leveraging IHttpClientFactory for non-mTLS HTTL clients and maintaining
+    /// a pool of mTLS clients with using certificate as a key.
```
MZOLN: .... with using certificate thumbprint as a key...

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisitionOptionsHelper.cs
```diff
@@ -0,0 +1,60 @@
...
+    /// </summary>
+    internal static class TokenAcquisitionOptionsHelper
+    {
+        /// <summary>
+        /// Creates a new instance of the TokenAcquisitionOptions class based on the specified downstream API options.
+        /// </summary>
+        /// <param name="downstreamApiOptions">The options used to configure the authorization header provider for the downstream API. Can be null.</param>
+        /// <param name="cancellationToken">A cancellation token that can be used to cancel the token acquisition operation.</param>
+        /// <returns>A TokenAcquisitionOptions object initialized with values from the provided downstream API options.</returns>
+        public static TokenAcquisitionOptions CreateTokenAcquisitionOptionsFromApiOptions(
```
MZOLN: This should go to a separate class, not named a helper.

Directory.Build.props
jmprieur: a feature flag would be dynamic. Here this is compile time.

src/Microsoft.Identity.Web.TokenAcquisition/ConfidentialClientApplicationBuilderExtension.cs
jmprieur: Where does it come from?
- Client credentials always? (Client certificate or result of a call to the hosting infrastructure?)

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderBoundProvider.cs
jmprieur: In an Ideal world we would just do a breaking change and have `IAuthorizationHeaderProvider` return `AuthorizationProviderOptions` but here is another interface that already exists which is IAuthorizationHeaderProvider<T> which we can use with T being an `OperationResult<AuthorizationHeaderResult, Error>`.

Thiking more, I would prefer we say IAuthorizationHeaderProvider<OperationResult<AuthorizationHeaderInformation, AuthorizationHeaderError>>` instead of the IAuthorizationHeaderBoundProvider because the certificate is one of the result in  AuthorizationProviderOptions. There could be more in the future, given Ivan's vision

jmprieur: That's my recommendation. See my comment above

jmprieur: We already have one (see above). Just not yet implemented.

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderBoundProvider.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        private readonly ITokenAcquisition _tokenAcquisition;
+
+        /// <summary>
+        /// Initializes a new instance of the DefaultAuthorizationHeaderBoundProvider class using the specified
+        /// authorization header provider.
+        /// </summary>
+        /// <param name="authorizationHeaderProvider">The provider used to supply authorization headers for HTTP requests.</param>
+        /// <param name="tokenAcquisition">The token acquisition service.</param>
+        public DefaultAuthorizationHeaderBoundProvider(
```
jmprieur: Let's modify DefaultAuthorizationHeaderProvider rather to implement the `IAuthorizationHeaderProvider<OperationResult<AuthorizationHeaderInformation, AuthorizationHeaderError>>`

src/Microsoft.Identity.Web.TokenAcquisition/MsalMtlsHttpClientFactory.cs
```diff
@@ -0,0 +1,120 @@
...
+    /// </summary>
+    public sealed class MsalMtlsHttpClientFactory : IMsalMtlsHttpClientFactory
```
jmprieur, line 18: Shouldn't this be in MSAL.NET?
there seem to be 0 dependency on Abstractions or IdWeb.

tests/Microsoft.Identity.Web.Test.Common/Mocks/MockMtlsHttpClientFactory.cs
jmprieur: yes, exactly. There is a need for an end to end sample, and even tests. Something like this would be useful (let's make sure it's right

```csharp
// Registration
services.AddTokenAcquisition()
        .AddTokenBindingScenario(); // extension provided by PR

// Acquire header with \mTLS-Pop scheme
var headerInformation = await authorizationHeaderProvider.CreateAuthorizationHeaderAsync(
    "https://api.contoso.com/.default",
    new AuthorizationHeaderProviderOptions {
        ProtocolScheme = "mtls-pop" 
    });

// HttpClient configured (name/policy registered in AddTokenBindingScenario)
var client = httpClientFactory.CreateClient("ContosoBoundApi");
client.DefaultRequestHeaders.Add("Authorization", headerInformation.value);
var resp = await client.GetAsync("/data");
```

jmprieur (commented):
Sorry I forgot to press "Submit review"

src/Microsoft.Identity.Web.TokenAcquisition/ConfidentialClientApplicationBuilderExtension.cs
cpp11nullptr: Sound good, will add.

Directory.Build.props
cpp11nullptr: It's aligned with https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/8e0a65ae718168677c8e939b8f390a165e8fae69/build/platform_and_feature_flags.props#L9.

cpp11nullptr: @bgavrilMS, @gladjohn, would be it safe to define SUPPORTS_MTLS for any !net462 platform among targeted by Microsoft.Identity.Web?

src/Microsoft.Identity.Web.TokenAcquisition/ConfidentialClientApplicationBuilderExtension.cs
cpp11nullptr: @jmprieur , the certificate is required for mTLS PoP, for this particular case it's supplied by credential loader. I guess, for other scenarios (such as managed identity) there should be added corresponding extension method, but we haven't designed those scenarios yet.

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderBoundProvider.cs
cpp11nullptr: Thanks, I'll make implement IAuthorizationHeaderBoundProvider within DefaultAuthorizationHeaderProvider with removing DefaultAuthorizationHeaderBoundProvider.

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
cpp11nullptr: Good suggestion, will do.

src/Microsoft.Identity.Web.TokenAcquisition/MsalMtlsHttpClientFactory.cs
cpp11nullptr: It was initial approach but we decided that IdWeb would be a better place for this (particularly, it's needed by IdWeb for calling downstream API). FYI, @bgavrilMS

src/Microsoft.Identity.Web.TokenAcquisition/PublicAPI/net462/PublicAPI.Shipped.txt
cpp11nullptr: It's because of adding defaulted `isTokenAcquisitionSingleton` parameter (so corresponding API goes into PublicAPI.Unshipped.txt). However, after refactoring, this change will be reverted.

src/Microsoft.Identity.Web.TokenAcquisition/ServiceCollectionExtensions.cs
cpp11nullptr: It'll be refactored.

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderBoundProvider.cs
cpp11nullptr: As it was agreed, it was removed in favor of DefaultAuthorizationHeaderProvider.

Directory.Build.props
MZOLN: Yeah I was trying to understand what we would use it for and why it's needed and whether we could use something else instead, like #if Windows

jmprieur: Oh I see. this is to make sure we don't have the mTLS code in net462.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisitionOptionsHelper.cs
cpp11nullptr: This part was refactored (no helper is needed).

src/Microsoft.Identity.Web.TokenAcquisition/MsalMtlsHttpClientFactory.cs
cpp11nullptr: There is no business ask for it yet, so let's it keep as is without introducing additional dependencies there.

src/Microsoft.Identity.Web.TokenAcquisition/MsalMtlsHttpClientFactory.cs
cpp11nullptr: I'd keep it simple as for now because there shouldn't be many unique certificates and rotation of existing ones won't have enough frequency to cause issues with hot path.

src/Microsoft.Identity.Web.TokenAcquisition/MsalMtlsHttpClientFactory.cs
cpp11nullptr: It follows mTLS HTTP client implementation from MSAL.NET so I presume, yes (at least for now).

src/Microsoft.Identity.Web.TokenAcquisition/MsalMtlsHttpClientFactory.cs
cpp11nullptr: As it was replied in other comment, there is no business ask for it yet, so let's it keep as is for now.

src/Microsoft.Identity.Web.TokenAcquisition/MsalMtlsHttpClientFactory.cs
cpp11nullptr: It follows common C# guidelines and used in other places within the SDK.

src/Microsoft.Identity.Web.TokenAcquisition/MsalMtlsHttpClientFactory.cs
cpp11nullptr: It seems that not all .NET versions supports it.

tests/Microsoft.Identity.Web.Test.Common/Mocks/MockMtlsHttpClientFactory.cs
cpp11nullptr: Added mTLS PoP client and web API samples.

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderProvider.cs
```diff
@@ -97,6 +97,33 @@ public async Task<string> CreateAuthorizationHeaderAsync(
             return result.CreateAuthorizationHeader();
         }
 
+        /// <inheritdoc/>
+        public async Task<OperationResult<AuthorizationHeaderInformation, AuthorizationHeaderError>> CreateAuthorizationHeaderAsync(
```
bgavrilMS, line 101: Can the existing methods that return Task of string be refactored to call this one?

Smth like: 

- for bearer tokens, old methods should continue to work
- for pop tokens, it's ok if old methods throw exception

src/Microsoft.Identity.Web.TokenAcquisition/MsalMtlsHttpClientFactory.cs
```diff
@@ -0,0 +1,120 @@
...
+using Microsoft.Identity.Client;
+
+namespace Microsoft.Identity.Web
+{
+    /// <summary>
+    /// Provides a factory for creating HTTP clients configured for mTLS authentication with using binding certificate.
+    /// It uses a hybrid approach with leveraging IHttpClientFactory for non-mTLS HTTP clients and maintaining
+    /// a pool of mTLS clients with using certificate thumbprint as a key.
+    /// </summary>
+    public sealed class MsalMtlsHttpClientFactory : IMsalMtlsHttpClientFactory
```
bgavrilMS, line 18: Does this need to be public?

src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
```diff
@@ -50,6 +50,7 @@ public ConfidentialClientApplicationOptions ConfidentialClientApplicationOptions
         public LogLevel LogLevel { get; set; }
         public string? RedirectUri { get; set; }
         public bool EnableCacheSynchronization { get; set; }
+        public bool IsTokenBinding { get; set; }
```
bgavrilMS, line 53: Is this a convenience API? I thought we'd use the ProtocolScheme property to denote that app developer wants to use mtls POP ?

tests/DevApps/MtlsPop/MtlsPopClient/Program.cs
```diff
@@ -0,0 +1,55 @@
...
+
+            var sp = tokenAcquirerFactory.Build();
+
+            Console.WriteLine("Scenario 1: calling web API with mTLS PoP token...");
+            var webApi = sp.GetRequiredService<IDownstreamApi>();
+            var result = await webApi.GetForAppAsync<IEnumerable<WeatherForecast>>("WebApi",
+                options =>
+                {
+                    options.AcquireTokenOptions.ExtraParameters ??= new Dictionary<string, object>();
+                    options.AcquireTokenOptions.ExtraParameters["RequestBoundToken"] = true; // mTLS PoP
```
bgavrilMS, line 34: I thought we settled on using `ProtocolScheme` for allowing app developer to opt-in to this type of token. https://github.com/AzureAD/microsoft-identity-abstractions-for-dotnet/blob/35845ee8d662be389802d538802535a6d40607c0/src/Microsoft.Identity.Abstractions/DownstreamApi/AuthorizationHeaderProviderOptions.cs#L97

bgavrilMS: ExtraParamters is for extensibility, but mTLS POP is a core feature.

src/Microsoft.Identity.Web.TokenAcquisition/MsalMtlsHttpClientFactory.cs
cpp11nullptr: It's used in Microsoft.Identity.Web.DownstreamApi as well so yes.

tests/DevApps/MtlsPop/MtlsPopClient/Program.cs
cpp11nullptr: Thanks, changed to use `AuthenticationOptionsName` (which is equivalent for `ProtocolSchema` and used by token acquirer).

src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
cpp11nullptr: Per latest change it's set based on `AuthenticationOptionsName` and used only within token acquisition scenario.

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderProvider.cs
cpp11nullptr: It's intended to be used for token binding flow (although, it's expressed only by result type, not function name), which also supports only app tokens so I'd leave it as is to separate concerns.

PR #3620, opened 2025-11-23
Jmprieur/fix authority parsing

### LINKED ISSUE #3612: Fix merging logic: Parse Authority into Instance and TenantId when provided alone (AAD & B2C)
## Background
In @AzureAD/microsoft-identity-web#3609, we've identified that configuring MicrosoftIdentityApplicationOptions with only Authority fails for AAD scenarios. However, we also need to ensure B2C scenarios (which are being deprecated, but remain relevant to some customers) work as expected.

## Root Cause
- `PrepareAuthorityInstanceForMsal()` returns early if Instance is null, resulting in a malformed authority string being passed to MSAL.
- `ParseAuthorityIfNecessary()` only triggers if both Instance and TenantId are unset, but is invoked too late.
- In B2C, authority parsing is more complex due to user-flow policies.

## Fix Spec
- Ensure that when only Authority is set (and Instance/TenantId are unset), Authority is parsed early enough into Instance and TenantId before MSAL validation.
- For B2C, ensure policies in authorities are handled, and parsing logic gracefully supports B2C paths (including /tfp/ etc.).

## Sample Failure (AAD)
```csharp
[Fact]
public async Task AcquireToken_WithMicrosoftIdentityApplicationOptions_ClientCredentialsAsync_WithAuthority()
{
    services.Configure<MicrosoftIdentityApplicationOptions>(s_optionName, option =>
    {
        option.Authority = "https://login.microsoftonline.com/msidlab4.onmicrosoft.com/v2.0";
        option.ClientId = "f6b698c0-140c-448f-8155-4aa9bf77ceba";
        option.ClientCredentials = s_clientCredentials;
    });
    await CreateGraphClientAndAssertAsync(tokenAcquirerFactory, services);
}
```
(E2E: [TokenAcquirer.cs#L229](https://github.com/AzureAD/microsoft-identity-web/blob/233f2e0db92cbfe6d3cc8b76b95223fb3d74f4ea/tests/E2E%20Tests/TokenAcquirerTests/TokenAcquirer.cs#L229))

## Sample B2C config
```csharp
services.Configure<MicrosoftIdentityApplicationOptions>(s_optionName, option =>
{
    option.Authority = "https://<b2c domain>.b2clogin.com/tfp/<tenant>/<policy>/v2.0";
    option.ClientId = "...";
    option.ClientCredentials = s_clientCredentials;
});
```

## Acceptance Criteria
- [ ] Authority-only works for AAD, CIAM, and B2C scenarios
- [ ] Authority is parsed into Instance/TenantId/Policy before MSAL validation
- [ ] Sample E2E test(s) pass
- [ ] New B2C test verifies parsing and success path
- [ ] Code blocks updated in relevant areas, with permalinks
- [ ] Update code comments to clarify B2C path handling
- [ ] Document changes for migration

## References
- [MergedOptions.cs:ParseAuthorityIfNecessary](https://github.com/AzureAD/microsoft-identity-web/blob/main/src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs)
- [TokenAcquisition.cs:BuildConfidentialClientApplicationAsync](https://github.com/AzureAD/microsoft-identity-web/blob/main/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs)

---
Bridge: Propose implementation diff and test samples when Jean-Marc replies "Go ahead".

jmprieur:
# fixes authority parsing bug (#3612)

Fixes #3612. Done with Copilot in stages (see issue and it's 4 sub issues)

### DIFF
--- a/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
@@ -569,8 +569,9 @@ namespace Microsoft.Identity.Web
                 }
             }
 
-            // For non-managed identity flows, resolve the tenant
-            tenant = ResolveTenant(tenant, mergedOptions);
+            // For non-managed identity flows we only resolve tenant if the caller explicitly provided an override.
+            // This preserves the ability to use an authority-only configuration with meta-tenants like 'common'.
+            string? resolvedOverrideTenant = tenant != null ? ResolveTenant(tenant, mergedOptions) : null;
 
             if (tokenAcquisitionOptions is not null)
             {
@@ -592,15 +593,14 @@ namespace Microsoft.Identity.Web
                 addInOptions.InvokeOnBeforeTokenAcquisitionForApp(builder, tokenAcquisitionOptions);
             }
 
-            // MSAL.net only allows .WithTenantId for AAD authorities. This makes sense as there should
-            // not be cross tenant operations with such an authority.
+            // Apply tenant override only for AAD authorities and only if non-empty
             if (!mergedOptions.Instance.Contains(Constants.CiamAuthoritySuffix
 #if NET6_0_OR_GREATER
                 , StringComparison.OrdinalIgnoreCase
 #endif
-                ))
+                ) && !string.IsNullOrEmpty(resolvedOverrideTenant))
             {
-                builder.WithTenantId(tenant);
+                builder.WithTenantId(resolvedOverrideTenant);
             }
 
             if (tokenAcquisitionOptions != null)
@@ -976,7 +976,41 @@ namespace Microsoft.Identity.Web
                 }
                 else if (mergedOptions.IsB2C)
                 {
-                    authority = $"{mergedOptions.PreparedInstance}{ClaimConstants.Tfp}/{mergedOptions.Domain}/{mergedOptions.DefaultUserFlow}";
+                    // B2C authority construction requires the tenant segment. If Domain was not configured
+                    // (scenario: authority-only configuration providing Instance + SignUpSignInPolicyId), derive it.
+                    string? domain = mergedOptions.Domain;
+                    if (string.IsNullOrEmpty(domain))
+                    {
+                        // Try tenantId first if provided
+                        if (!string.IsNullOrEmpty(mergedOptions.TenantId))
+                        {
+                            domain = mergedOptions.TenantId;
+                        }
+                        else if (!string.IsNullOrEmpty(mergedOptions.Instance))
+                        {
+                            try
+                            {
+                                // Extract first label from host (e.g. fabrikamb2c from fabrikamb2c.b2clogin.com)
+                                var host = new Uri(mergedOptions.Instance).Host;
+                                var firstLabel = host.Split('.').FirstOrDefault();
+                                if (!string.IsNullOrEmpty(firstLabel))
+                                {
+                                    domain = firstLabel + ".onmicrosoft.com";
+                                }
+                            }
+                            catch
+                            {
+                                // Ignore derivation failures; will throw below if still null.
+                            }
+                        }
+                    }
+
+                    if (string.IsNullOrEmpty(domain))
+                    {
+                        throw new ArgumentException("B2C Domain could not be determined. Provide Domain or TenantId when using B2C authority-only configuration.");
+                    }
+
+                    authority = $"{mergedOptions.PreparedInstance}{ClaimConstants.Tfp}/{domain}/{mergedOptions.DefaultUserFlow}";
                     builder.WithB2CAuthority(authority);
                 }
                 else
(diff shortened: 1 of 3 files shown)


src/Microsoft.Identity.Web.TokenAcquisition/MergedOptionsStore.cs
```diff
@@ -2,21 +2,32 @@
...
 
 namespace Microsoft.Identity.Web
 {
     internal class MergedOptionsStore : IMergedOptionsStore
     {
         private readonly ConcurrentDictionary<string, MergedOptions> _options;
+        private readonly ILoggerFactory? _loggerFactory;
 
-        public MergedOptionsStore()
+        public MergedOptionsStore(ILoggerFactory? loggerFactory)
```
jmprieur: @copilot : This is a behavioral breaking change, as before this, the logging service was not needed.

please change this method so that it takes a IServiceProvider, and then get the ILoggerFactory from the service provider if possible (and otherwise just initialize _loggerFactory to null. Please also update the InternalApi.Unshipped files accordingly.

src/Microsoft.Identity.Web.TokenAcquisition/MergedOptionsStore.cs
```diff
@@ -1,22 +1,35 @@
...
 
 namespace Microsoft.Identity.Web
 {
     internal class MergedOptionsStore : IMergedOptionsStore
     {
         private readonly ConcurrentDictionary<string, MergedOptions> _options;
+        private readonly ILoggerFactory? _loggerFactory;
 
-        public MergedOptionsStore()
+        public MergedOptionsStore(IServiceProvider serviceProvider)
```
cpp11nullptr, line 16: Is it possible to get injected ILoggerFactory instead?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -976,7 +976,41 @@ private async Task<IConfidentialClientApplication> BuildConfidentialClientApplic
...
+                            domain = mergedOptions.TenantId;
+                        }
+                        else if (!string.IsNullOrEmpty(mergedOptions.Instance))
+                        {
+                            try
+                            {
+                                // Extract first label from host (e.g. fabrikamb2c from fabrikamb2c.b2clogin.com)
+                                var host = new Uri(mergedOptions.Instance).Host;
+                                var firstLabel = host.Split('.').FirstOrDefault();
+                                if (!string.IsNullOrEmpty(firstLabel))
```
cpp11nullptr, line 996: Can be there situation when B2C DNS name has subdomains, like `api.fabrikamb2c.b2clogin.com`? Maybe, 3rd part from the end would work better in this case?

tests/E2E Tests/TokenAcquirerTests/TokenAcquirer.cs
```diff
@@ -25,7 +25,7 @@ namespace TokenAcquirerTests
 {
     [Collection(nameof(TokenAcquirerFactorySingletonProtection))]
 #if !FROM_GITHUB_ACTION
-    public class TokenAcquirer
+    public partial class TokenAcquirer
```
cpp11nullptr, line 28: nit: perhaps, TokenAcquirerTests would be better name (but I realize that it was before).

src/Microsoft.Identity.Web.TokenAcquisition/MergedOptionsStore.cs
jmprieur: I used the locator pattern on purpose to avoid a behavioral breaking change. Just AddTokenAcquisition didn't require a logger abstraction so far.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
jmprieur: No. Not with a b2clogin URL. They are owned by msft

src/Microsoft.Identity.Web.TokenAcquisition/LoggingEventId.cs
```diff
@@ -29,6 +29,9 @@ internal static class LoggingEventId
         public static readonly EventId CredentialLoadAttemptFailed = new EventId(406, "CredentialLoadAttemptFailed");
         public static readonly EventId UsingSignedAssertionFromCustomProvider = new EventId(407, "UsingSignedAssertionFromCustomProvider");
 
+        // MergedOptions EventIds 408+
+        public static readonly EventId AuthorityConflict = new EventId(408, "AuthorityConflict");
```
pmaytak, line 33: 500+? 400+ is CCA stuff.


```suggestion
        // MergedOptions EventIds 500+
        public static readonly EventId AuthorityConflict = new EventId(500, "AuthorityConflict");
```

PR #3622, opened 2025-11-24
Add token binding

cpp11nullptr:
## Purpose

Introduce a “token binding scenario” for confidential client (app-only) token acquisition and downstream API calls:

- Bind access tokens to a client certificate (expected `cnf` claim – thumbprint / key reference).
- Use mutual TLS (mTLS) when calling downstream APIs.
- Support Bearer / PoP / mTLS protocol selection via `AuthorizationHeaderProviderOptions.ProtocolScheme`.
- Provide a service collection extension (one-line enablement) wiring token acquisition + binding + downstream calls.

## High-level scenario

1. Developer enables token binding scenario via new service collection extension (PR states: “add service collection extension to enable token binding”).
2. A binding certificate is sourced from ClientCredentials (direct cert or via Managed Identity).
3. Developer specifies a protocol scheme (e.g. `ProtocolScheme = "Pop"` or future `ProtocolScheme = "mtls-pop"` / `"mtls"`) in `AuthorizationHeaderProviderOptions`.
4. Token acquisition (client credentials flow) is configured accordingly; Entra ID returns a token containing a binding claim (`cnf`) when certificate binding is active.
5. The authorization header provider produces the header (currently `Bearer <token>`; protocol-specific transformations for schemes like Pop are applied internally).
6. An mTLS-capable HTTP client presents the same certificate during TLS handshake.
7. Downstream API validates that presented client certificate matches token’s `cnf` claim (x5t thumbprint).
8. Fallback path: if no certificate is available (or protocol does not require it), a regular unbound token is retrieved.


src/Microsoft.Identity.Web.TokenAcquisition/MsalMtlsHttpClientFactory.cs
```diff
@@ -0,0 +1,176 @@
...
+using Microsoft.Identity.Client;
+
+namespace Microsoft.Identity.Web
+{
+    /// <summary>
+    /// Provides a factory for creating HTTP clients configured for mTLS authentication with using binding certificate.
+    /// It uses a hybrid approach with leveraging IHttpClientFactory for non-mTLS HTTP clients and maintaining
+    /// a pool of mTLS clients with using certificate thumbprint as a key.
+    /// </summary>
+    public sealed class MsalMtlsHttpClientFactory : IMsalMtlsHttpClientFactory, IDisposable
```
bgavrilMS, line 19: Does this class need to be public?

src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
```diff
@@ -56,6 +56,7 @@ public ConfidentialClientApplicationOptions ConfidentialClientApplicationOptions
         public LogLevel LogLevel { get; set; }
         public string? RedirectUri { get; set; }
         public bool EnableCacheSynchronization { get; set; }
+        public bool IsTokenBinding { get; set; }
```
bgavrilMS: If this is convenience API, can it not just check the Protocol string property?

tests/DevApps/MtlsPop/MtlsPopClient/appsettings.json
```diff
@@ -0,0 +1,25 @@
...
+            }
+        ],
+        "SendX5c": true
+    },
+    "WebApi": {
+        "BaseUrl": "https://localhost:7060/",
+        "RelativePath": "WeatherForecast",
+        "RequestAppToken": true,
+        "AcquireTokenOptions": {
+            "AuthenticationOptionsName": "MTLS_POP"
```
bgavrilMS: I am still confused about this, I though the property name would be "ProtocolScheme" .

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderProvider.cs
```diff
@@ -97,6 +97,33 @@ public async Task<string> CreateAuthorizationHeaderAsync(
...
+        /// <inheritdoc/>
+        public async Task<OperationResult<AuthorizationHeaderInformation, AuthorizationHeaderError>> CreateAuthorizationHeaderAsync(
+            DownstreamApiOptions downstreamApiOptions,
+            ClaimsPrincipal? claimsPrincipal = null,
+            CancellationToken cancellationToken = default)
+        {
+            var newTokenAcquisitionOptions = CreateTokenAcquisitionOptionsFromApiOptions(downstreamApiOptions, cancellationToken);
+
+            // Token binding flow currently supports only app tokens.
+            var tokenAcquisitionResult = await _tokenAcquisition.GetAuthenticationResultForAppAsync(
```
jmprieur, line 136: what happens if developers intend to use a user token?

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderProvider.cs
```diff
@@ -97,6 +97,33 @@ public async Task<string> CreateAuthorizationHeaderAsync(
...
+                downstreamApiOptions.Scopes?.FirstOrDefault() ?? string.Empty,
```
jmprieur: Do empty scopes work?

src/Microsoft.Identity.Web.TokenAcquisition/DefaultTokenAcquisitionHost.cs
```diff
@@ -42,6 +43,12 @@ public DefaultTokenAcquisitionHost(
 
         public string GetEffectiveAuthenticationScheme(string? authenticationScheme)
         {
+            if (string.Equals(authenticationScheme, "MTLS_POP", StringComparison.OrdinalIgnoreCase))
```
jmprieur: Not sure what you expect to do here

the ASP.NET Core authenticationScheme is the name of the options. You should not suppose that there will be named options or a config section named MTLS_POP.

src/Microsoft.Identity.Web.TokenAcquisition/IDWebErrorMessage.cs
```diff
@@ -21,6 +21,7 @@ internal static class IDWebErrorMessage
         public const string MissingRequiredScopesForAuthorizationFilter = "IDW10108: RequiredScope Attribute does not contain a value. The scopes need to be set on the controller, the page or action. See https://aka.ms/ms-id-web/required-scope-attribute. ";
         public const string ClientCertificatesHaveExpiredOrCannotBeLoaded = "IDW10109: No credential could be loaded. This can happen when certificates passed to the configuration have expired or can't be loaded and the code isn't running on Azure to be able to use Managed Identity, Pod Identity etc. Details: ";
         public const string ClientSecretAndCredentialsCannotBeCombined = "IDW10110: ClientSecret top level configuration cannot be combined with ClientCredentials. Instead, add a new entry in the ClientCredentials array describing the secret.";
+        public const string MissingTokenBindingCertificate = "IDW10111: A certificate, which is required for token binding, is missing in loaded credentials.";
```
jmprieur: which kind of certificate? an encrypt certificate? a sign certificate?

src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
jmprieur: Let's not have this. This is not surfaced in the MicrosoftIdentityApplicationOptions, so shouldn't be here.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -588,6 +590,11 @@ public async Task<AuthenticationResult> GetAuthenticationResultForAppAsync(
                    .AcquireTokenForClient(new[] { scope }.Except(_scopesRequestedByMsal))
                    .WithSendX5C(mergedOptions.SendX5C);
 
+            if (mergedOptions.IsTokenBinding)
```
jmprieur: Where does this come from?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -747,6 +754,8 @@ private MergedOptions GetMergedOptions(string? authenticationScheme, TokenAcquis
                 mergedOptions = _tokenAcquisitionHost.GetOptions(authenticationScheme ?? tokenAcquisitionOptions?.AuthenticationOptionsName, out _);
             }
 
+            mergedOptions.IsTokenBinding = string.Equals(authenticationScheme, "MTLS_POP", StringComparison.OrdinalIgnoreCase);
```
jmprieur: The authenticationScheme is the name of the configuration. And null is used by IdWeb for the default authentication scheme. See https://learn.microsoft.com/en-us/aspnet/core/security/authorization/limitingidentitybyscheme?view=aspnetcore-10.0

You can't assume that there is an authentication scheme for mTLS Pop (especially for outbound calls). The way to say it's mTLS Pop is the protocolScheme property of the AuthorizationHeaderOptions / DownstreamApiOptions

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -1021,11 +1030,23 @@ private async Task<IConfidentialClientApplication> BuildConfidentialClientApplic
 
                 try
                 {
-                    await builder.WithClientCredentialsAsync(
-                        mergedOptions.ClientCredentials!,
-                        _logger,
-                        _credentialsLoader,
-                        new CredentialSourceLoaderParameters(mergedOptions.ClientId!, authority));
+                    if (mergedOptions.IsTokenBinding)
```
jmprieur: Given that the mTLS Pop signal (the protocolScheme property) is per request, I think we need to change the MSAL.NET design to have the binding certificate per request (that is on the AcquireTokenForClientBuilder). Or otherwise, use the client cert, and have a bool on the AcquireTokenForClientBuilder (WithTokenBinding), which would use this cert as a binding cert.

@bgavrilMS I would like to have your perspective.

tests/DevApps/MtlsPop/MtlsPopClient/appsettings.json
jmprieur: Same for me (hence the comments above). AuthenticationOptionsName would point to a section named "MTSL_POP" (at the same level of "AzureAd") and there is none.

As @bgavrilMS wrote, you need to use the "AuthenticationScheme" property in the Web Api to specify that this is MTLS Pop ... and this becomes a per-request directive (not on the CCA)

tests/DevApps/MtlsPop/MtlsPopWebApi/MtlsPopAuthorizationHandler.cs
```diff
@@ -0,0 +1,120 @@
...
+
+            if (string.IsNullOrEmpty(authToken))
+            {
+                _logger.LogWarning("No auth token found");
+                return Task.CompletedTask;
+            }
+
+            try
+            {
+                var handler = new JwtSecurityTokenHandler();
```
jmprieur: this should be JsonWebTokenHandler.

tests/DevApps/MtlsPop/MtlsPopWebApi/MtlsPopAuthorizationHandler.cs
```diff
@@ -0,0 +1,120 @@
...
+using Microsoft.AspNetCore.Authorization;
+using Microsoft.IdentityModel.Tokens;
+using System.IdentityModel.Tokens.Jwt;
+using System.Security.Cryptography;
+using System.Security.Cryptography.X509Certificates;
+using System.Text.Json;
+
+namespace MtlsPopSample
+{
+    public class MtlsPopRequirement : IAuthorizationRequirement
```
jmprieur: mTLS Pop is AuthN, not AuthZ
Here you used an AuthZ requirement. you should rather create an authentication handler if you want the web API to accept mTLS Pop in 3P cases. But this might be better to bring down MISE in the future?

tests/DevApps/MtlsPop/MtlsPopWebApi/Program.cs
```diff
@@ -0,0 +1,172 @@
...
+            builder.Services.AddControllers();
+
+            // Learn more about configuring OpenAPI at https://learn.microsoft.com/aspnet/core/fundamentals/openapi/aspnetcore-openapi
+            builder.Services.AddEndpointsApiExplorer();
+
+            builder.Services.AddMicrosoftIdentityWebApiAuthentication(builder.Configuration);
+
+            builder.Services.AddAuthorization(options =>
+            {
+                options.AddPolicy("MtlsPop", policy =>
```
jmprieur: ⚠️
This is not how it should be done. mTLS Pop is a protocol (hence a new ASP.NET Core authentication handler), not an Authorization policy)

tests/E2E Tests/TokenAcquirerTests/TokenAcquirer.cs
```diff
@@ -424,6 +427,58 @@ public async Task AcquireTokenWithMs10AtPop_ClientCredentialsAsync()
...
+            });
+
+            services.AddInMemoryTokenCaches();
+
+            var serviceProvider = tokenAcquirerFactory.Build();
+            ITokenAcquirer tokenAcquirer = tokenAcquirerFactory.GetTokenAcquirer(s_optionName);
+
+            var tokenAcquisitionOptions = new TokenAcquisitionOptions
+            {
+                AuthenticationOptionsName = "MTLS_POP"
```
jmprieur: that's not how this should be. The authenticationOptionsName is the name of a configuration section. You have only one which is AzureAd (The default one)

jmprieur (changes requested):
I've left comments. there are changes to make

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderProvider.cs
cpp11nullptr: I'll refactor it to fallback to existing CreateAuthorizationHeaderAsync() if it's not token binding scenario.

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderProvider.cs
cpp11nullptr: Good point, I guess, defaulting to empty scope shouldn't be there, will change it.

src/Microsoft.Identity.Web.TokenAcquisition/IDWebErrorMessage.cs
cpp11nullptr: It's signing certificate, will change a message.

src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs
cpp11nullptr: Since some token acquisition code doesn't have access to acquire token options, which merged option are being created from, having this property can help with better maintenance - I'll make it `internal` with adding a comment of its purpose. @jmprieur , let me know if it's a valid justification or not, thanks.

src/Microsoft.Identity.Web.TokenAcquisition/MsalMtlsHttpClientFactory.cs
cpp11nullptr: It's used by **DownstreamApi** project so should be visible (`public`) outside of **TokenAcquisition** project.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
cpp11nullptr: It's set based on acquire token options (also, replied in other comment).

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
cpp11nullptr: Thanks! Based on our discussion, a better solution will be using **ExtraParameters** instead, will make this change.

tests/DevApps/MtlsPop/MtlsPopClient/appsettings.json
cpp11nullptr: I'll change it to use **ProtocolScheme**.

tests/E2E Tests/TokenAcquirerTests/TokenAcquirer.cs
cpp11nullptr: It'll be refactored.

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderProvider.cs
bgavrilMS: I think we should throw exception. Be explicit about what is supported and what isn't ?

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderProvider.cs
bgavrilMS: Empty scopes don't work. Just like Bearer token requests for App Tokens.

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderProvider.cs
cpp11nullptr: It makes sense.

tests/DevApps/MtlsPop/MtlsPopWebApi/Program.cs
cpp11nullptr: Thanks for clarifying, I'll update the sample.

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderProvider.cs
```diff
@@ -97,10 +100,66 @@ public async Task<string> CreateAuthorizationHeaderAsync(
...
+            };
+
+            return new(authorizationHeaderInformation);
+        }
+
         private static TokenAcquisitionOptions CreateTokenAcquisitionOptionsFromApiOptions(
             AuthorizationHeaderProviderOptions? downstreamApiOptions,
             CancellationToken cancellationToken)
         {
+            var extraParameters = downstreamApiOptions?.AcquireTokenOptions.ExtraParameters;
```
MZOLN: nit: Not following this, was this a necessary change? Why would we always want to allocate a dictionary? Maybe extract those 156-161 into a private static method and add remarks? We might get additional binding protocol.

src/Microsoft.Identity.Web.TokenAcquisition/MsalMtlsHttpClientFactory.cs
```diff
@@ -0,0 +1,176 @@
...
+    {
+        private const long MaxMtlsHttpClientCountInPool = 1000;
+        private const long MaxResponseContentBufferSizeInBytes = 1024 * 1024;
+
+        // Please see (https://aka.ms/msal-httpclient-info) for important information regarding the HttpClient.
+        private static readonly Dictionary<string, HttpClient> s_mtlsHttpClientPool = new Dictionary<string, HttpClient>();
```
MZOLN, line 25: nit: the "s_" prefix reminds me of the olden times and hungarian notation that was prevalent in old APIS. The C# coding guidelines once C# was release clearly stated that it's a code smell and to avoid that. Im sure that we tried to in all our MISE code bases. I don't think that we need to additionally indicate that this field is static.

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderProvider.cs
cpp11nullptr: As it was discussed offline, extra parameters are good fit for this purpose (particularly, for SDK extensibility, as in case of MISE).

src/Microsoft.Identity.Web.TokenAcquisition/MsalMtlsHttpClientFactory.cs
```diff
@@ -0,0 +1,176 @@
...
+        public HttpClient GetHttpClient(X509Certificate2 x509Certificate2)
+        {
+            if (x509Certificate2 == null)
+            {
+                return GetHttpClient();
+            }
+
+            string key = x509Certificate2.Thumbprint;
+
+            s_cacheLock.EnterReadLock();
```
MZOLN: Consider using the EnterUpgradeableReadLock with the pattern from the doc. It should be available on all the supported frameworks here. https://learn.microsoft.com/en-us/dotnet/api/system.threading.readerwriterlockslim.enterupgradeablereadlock?view=net-10.0

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderProvider.cs
MZOLN: Yeah Im not questioning the logic but readability and extensibility of the chosen implementation.

src/Microsoft.Identity.Web.TokenAcquisition/MsalMtlsHttpClientFactory.cs
cpp11nullptr: This convention is used in other places around as well, also mentioned in https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/identifier-names#naming-conventions (with some remarks, though). In case if it should be changed, it's better to change it across whole codebase.

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderProvider.cs
cpp11nullptr: I'll extracted it to separated method.

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
```diff
@@ -558,15 +596,38 @@ public Task<HttpResponseMessage> CallApiForAppAsync(
...
+                        effectiveOptions,
+                        user,
+                        cancellationToken).ConfigureAwait(false);
+
+                    if (authorizationHeaderResult.Succeeded)
+                    {
+                        authorizationHeaderInformation = authorizationHeaderResult.Result;
+                        authorizationHeader = authorizationHeaderInformation?.AuthorizationHeaderValue ?? string.Empty;
+                    }
+                }
```
jmprieur, line 626: What happens if not succeeded?

src/Microsoft.Identity.Web.TokenAcquisition/ConfidentialClientApplicationBuilderExtension.cs
```diff
@@ -58,6 +58,34 @@ public static async Task<ConfidentialClientApplicationBuilder> WithClientCredent
             }
         }
 
+        public static async Task<ConfidentialClientApplicationBuilder> WithBindingCertificateAsync(
```
jmprieur: how is this different from a normal client certificate? why is this necessary? Maybe a comment to explain?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
jmprieur: this is a case where we could get it from the tokenAcquisitionOptions?
`tokenAcquisitionOptions?.ExtraParameters?.TryGetValue(TokenBindingParameterName, out var isTokenBindingValue) == true`

 Set higher when the ProtocolScheme is MTLS_Pop? why would we need to have the boolean in mergedOptions?

tests/DevApps/MtlsPop/MtlsPopWebApi/MtlsPopAuthenticationHandler.cs
```diff
@@ -0,0 +1,117 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
jmprieur, line 1: I'm not a fan of redoing less well what is done in MISE.
Unfortunately, customers will copy/paste that ...

jmprieur: fortunately this is for testing.

tests/E2E Tests/TokenAcquirerTests/TokenAcquirer.cs
```diff
@@ -424,6 +427,61 @@ public async Task AcquireTokenWithMs10AtPop_ClientCredentialsAsync()
...
+            };
+
+            // Act
+            var result = await tokenAcquirer.GetTokenForAppAsync("https://graph.microsoft.com/.default", tokenAcquisitionOptions);
+
+            // Assert
+            Assert.NotNull(result.AccessToken);
+            Assert.StartsWith("eyJ0e", result.AccessToken, StringComparison.OrdinalIgnoreCase);
+
+            var tokenParts = result.AccessToken.Split('.');
```
jmprieur: Better use JSonWebToken

tests/Microsoft.Identity.Web.Test/TokenAcquisitionTests.cs
```diff
@@ -126,6 +126,41 @@ public async Task ExtraBodyParametersAreSentToEndpointTest()
...
+            var tokenAcquirerFactory = InitTokenAcquirerFactory();
+            IServiceProvider serviceProvider = tokenAcquirerFactory.Build();
+            var tokenAcquisition = serviceProvider.GetRequiredService<ITokenAcquisitionInternal>() as TokenAcquisition;
+
+            var tokenAcquisitionOptions = new TokenAcquisitionOptions();
+            if (isTokenBindingParameter.HasValue)
+            {
+                tokenAcquisitionOptions.ExtraParameters = new Dictionary<string, object>
+                {
+                    { "IsTokenBinding", isTokenBindingParameter.Value }
```
jmprieur: This is per request

tests/Microsoft.Identity.Web.Test/TokenAcquisitionTests.cs
```diff
@@ -126,6 +126,41 @@ public async Task ExtraBodyParametersAreSentToEndpointTest()
...
+                System.Reflection.BindingFlags.NonPublic | System.Reflection.BindingFlags.Instance);
+            Assert.NotNull(method);
+
+#pragma warning disable CS8601 // Possible null reference assignment.
+            var mergedOptions = method.Invoke(tokenAcquisition, new object?[] { null, tokenAcquisitionOptions }) as MergedOptions;
+#pragma warning restore CS8601 // Possible null reference assignment.
+
+            // Assert
+            Assert.NotNull(mergedOptions);
+            Assert.Equal(expectedIsTokenBinding, mergedOptions.IsTokenBinding);
```
jmprieur: This is per app.

I don't understand why we need to change the merged options?

src/Microsoft.Identity.Web.TokenAcquisition/ConfidentialClientApplicationBuilderExtension.cs
cpp11nullptr: As it was agreed, this method will be removed with reusing existing `WithClientCredentialsAsync` API.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
cpp11nullptr: Good point, it'll be refactored, thanks!

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
cpp11nullptr: Looking into upstream code, it shouldn't happen because in case of any issues there will be thrown an exception but I'll add safe-guard there.

src/Microsoft.Identity.Web.TokenAcquisition/ConfidentialClientApplicationBuilderExtension.cs
```diff
@@ -21,16 +21,22 @@ public static ConfidentialClientApplicationBuilder WithClientCredentials(
...
         }
 
         public static async Task<ConfidentialClientApplicationBuilder> WithClientCredentialsAsync(
             this ConfidentialClientApplicationBuilder builder,
             IEnumerable<CredentialDescription> clientCredentials,
             ILogger logger,
             ICredentialsLoader credentialsLoader,
-            CredentialSourceLoaderParameters? credentialSourceLoaderParameters)
+            CredentialSourceLoaderParameters? credentialSourceLoaderParameters,
+            bool isTokenBinding)
```
bgavrilMS, line 39: I don't think this is needed. You should let MSAL throw exceptions if unsupported credential is used.

src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderProvider.cs
```diff
@@ -11,10 +11,15 @@
 
 namespace Microsoft.Identity.Web
 {
-    internal sealed class DefaultAuthorizationHeaderProvider : IAuthorizationHeaderProvider
+    internal sealed class DefaultAuthorizationHeaderProvider : IAuthorizationHeaderProvider, IBoundAuthorizationHeaderProvider
     {
+        private static readonly object _boxedTrue = true;
```
bgavrilMS: namiong: s_boxedTrue

src/Microsoft.Identity.Web.TokenAcquisition/ConfidentialClientApplicationBuilderExtension.cs
cpp11nullptr: This is needed for custom certificate loaders which have `CredentialType.SignedAssertion` credential type but provide both certificate and client assertion - in this case we need to set only certificate in the builder.

bgavrilMS (approved):

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
cpp11nullptr: (resolved as part of rewritten logic)

PR #3647, opened 2025-12-10
Add CAE claims support for FIC + Managed Identity

gladjohn:
# Add CAE claims support for FIC + Managed Identity

## This change wires CAE claims all the way through Federated Identity Credentials (FIC) and Managed Identity, and adds coverage to ensure that:

- when CAE claims are present, app tokens bypass cache and are obtained from the identity provider, and
- FIC (both MSI‑based and OIDC‑based) correctly propagates FMI path / tenant and cooperates with those CAE claims.
- It also introduces a small test‑only hook so tests can fully mock the managed identity pipeline without affecting production code.

### DIFF
--- a/src/Microsoft.Identity.Web.Certificateless/ManagedIdentityClientAssertion.cs
+++ b/src/Microsoft.Identity.Web.Certificateless/ManagedIdentityClientAssertion.cs
@@ -58,7 +58,7 @@ namespace Microsoft.Identity.Web
                 managedIdentityClientId,
                 tokenExchangeUrl,
                 logger,
-                ManagedIdentityClientAssertionTestHook.HttpClientFactory)
+                ManagedIdentityClientAssertionTestHook.HttpClientFactoryForTests)
         {
         }
 
--- a/src/Microsoft.Identity.Web.Certificateless/ManagedIdentityClientAssertionTestHook.cs
+++ b/src/Microsoft.Identity.Web.Certificateless/ManagedIdentityClientAssertionTestHook.cs
@@ -9,11 +9,11 @@ namespace Microsoft.Identity.Web.TestOnly
     /// TEST-ONLY hook so unit tests can override the HttpClient factory used by
     /// ManagedIdentityClientAssertion.
     /// </summary>
-    public static class ManagedIdentityClientAssertionTestHook
+    internal static class ManagedIdentityClientAssertionTestHook
     {
         /// <summary>
         /// Gets or sets the <see cref="IMsalHttpClientFactory"/> used by <c>ManagedIdentityClientAssertion</c> for unit testing purposes.
         /// </summary>
-        public static IMsalHttpClientFactory? HttpClientFactory { get; set; }
+        internal static IMsalHttpClientFactory? HttpClientFactoryForTests { get; set; }
     }
 }
--- a/src/Microsoft.Identity.Web.Certificateless/InternalAPI.Unshipped.txt
+++ b/src/Microsoft.Identity.Web.Certificateless/InternalAPI.Unshipped.txt
@@ -1,2 +1,5 @@
 #nullable enable
 Microsoft.Identity.Web.ManagedIdentityClientAssertion.ManagedIdentityClientAssertion(string? managedIdentityClientId, string? tokenExchangeUrl, Microsoft.Extensions.Logging.ILogger? logger, Microsoft.Identity.Client.IMsalHttpClientFactory? testHttpClientFactory) -> void
+Microsoft.Identity.Web.TestOnly.ManagedIdentityClientAssertionTestHook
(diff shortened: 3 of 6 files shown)


src/Microsoft.Identity.Web.Certificateless/ManagedIdentityClientAssertion.cs
```diff
@@ -49,7 +50,33 @@ public ManagedIdentityClientAssertion(string? managedIdentityClientId, string? t
...
+        /// <param name="managedIdentityClientId">Optional ClientId of the Managed Identity</param>
+        /// <param name="tokenExchangeUrl">Optional audience of the token to be requested from Managed Identity. Default value is "api://AzureADTokenExchange". 
+        /// This value is different on clouds other than Azure Public</param>
+        /// <param name="logger">A logger.</param>
+        /// <param name="httpClientFactory">Optional MSAL HttpClient factory.</param>
+        public ManagedIdentityClientAssertion(
+            string? managedIdentityClientId,
+            string? tokenExchangeUrl,
+            ILogger? logger,
+            IMsalHttpClientFactory? httpClientFactory)
```
bgavrilMS: Is this for test only? Maybe rename it as "test only". We should not inject HTTP in MSI

src/Microsoft.Identity.Web.Certificateless/ManagedIdentityClientAssertionTestHook.cs
```diff
@@ -0,0 +1,19 @@
...
+    /// <summary>
+    /// TEST-ONLY hook so unit tests can override the HttpClient factory used by
+    /// ManagedIdentityClientAssertion.
+    /// </summary>
+    public static class ManagedIdentityClientAssertionTestHook
+    {
+        /// <summary>
+        /// Gets or sets the <see cref="IMsalHttpClientFactory"/> used by <c>ManagedIdentityClientAssertion</c> for unit testing purposes.
+        /// </summary>
+        public static IMsalHttpClientFactory? HttpClientFactory { get; set; }
```
bgavrilMS: Isn't there a IManagedIdentityTestHttpClientFactory that can be reused?

src/Microsoft.Identity.Web.Certificateless/ManagedIdentityClientAssertion.cs
gladjohn: renamed

src/Microsoft.Identity.Web.Certificateless/ManagedIdentityClientAssertionTestHook.cs
gladjohn: I looked at reusing `IManagedIdentityTestHttpClientFactory`, but it lives in the `TokenAcquisition` project while `ManagedIdentityClientAssertion` is in `Certificateless`. Using it directly would add a new dependency between those projects. I am not sure if that is Ok to do. 

The tests does reuse `TestManagedIdentityHttpFactory` to create the` IMsalHttpClientFactory` and pass that in via the internal ctor + ManagedIdentityClientAssertionTestHook, so we get the same behavior without changing the layering or public API.

bgavrilMS: Can't you move it from TokenAcquisition to Certificateless ? Moving a type to a new assembly is not a breaking change.

bgavrilMS: In fact, IManagedIdentityTestHttpClientFactory is not even public. So there is a way to have a test hook that does not expose public API.

bgavrilMS: Can be a different type, to keep it internal.

gladjohn: made it internal per offline conversation

src/Microsoft.Identity.Web.Certificateless/ManagedIdentityClientAssertion.cs
```diff
@@ -49,7 +50,33 @@ public ManagedIdentityClientAssertion(string? managedIdentityClientId, string? t
...
-        public ManagedIdentityClientAssertion(string? managedIdentityClientId, string? tokenExchangeUrl, ILogger? logger)
+        public ManagedIdentityClientAssertion(
+            string? managedIdentityClientId,
+            string? tokenExchangeUrl,
+            ILogger? logger)
+            : this(
+                managedIdentityClientId,
+                tokenExchangeUrl,
+                logger,
+                ManagedIdentityClientAssertionTestHook.HttpClientFactoryForTests)
```
jmprieur, line 61: Maybe add a comment that this is static, and null by default unless set in tests.

PR #3661, opened 2026-01-06
Add token binding documentation

cpp11nullptr:


docs/calling-downstream-apis/calling-downstream-apis-README.md
```diff
@@ -60,6 +61,14 @@ graph LR
 - **Token type**: New access token via OBO flow
 - **Methods**: `CreateAuthorizationHeaderForUserAsync()` from web API context
 
+### Token Binding (mTLS PoP)
+- **Scenario**: Enhanced security where tokens are cryptographically bound to certificates
```
bgavrilMS:
```suggestion
- **Scenario**: Enhanced security where tokens are cryptographically bound to certificates as per [RFC 8705](https://datatracker.ietf.org/doc/html/rfc8705)
```

docs/calling-downstream-apis/token-binding.md
```diff
@@ -0,0 +1,609 @@
+# Token Binding with mTLS Proof-of-Possession (mTLS PoP)
+
+## Overview
+
+Token binding (also known as mTLS PoP - Mutual TLS Proof-of-Possession) is an advanced security feature that cryptographically binds access tokens to a specific X.509 certificate. This binding ensures that even if a token is intercepted, it cannot be used by an attacker without possession of the corresponding private key.
```
bgavrilMS:
```suggestion
Certificate token binding (also known as mTLS PoP - Mutual TLS Proof-of-Possession) is an advanced security feature that cryptographically binds access tokens to a specific X.509 certificate. It is described in [RFC 8705](https://datatracker.ietf.org/doc/html/rfc8705). The binding ensures that even if a token is intercepted, it cannot be used by an attacker without possession of the corresponding private key. 
```

docs/calling-downstream-apis/token-binding.md
```diff
@@ -0,0 +1,609 @@
+# Token Binding with mTLS Proof-of-Possession (mTLS PoP)
+
+## Overview
+
```
bgavrilMS, line 4: Pls add a note that not all clients are allowed to obtain mtls POP certs, as the feature is currently in private preview.

docs/calling-downstream-apis/token-binding.md
```diff
@@ -0,0 +1,609 @@
...
+- **Validate certificates**: Implement proper certificate validation on the server
+- **Use strong ciphers**: Configure secure cipher suites
+
+### Token Handling
+- **Short lifetimes**: Use short-lived tokens (recommended: 1 hour)
+- **Proper storage**: Never log or expose tokens
+- **Validate thoroughly**: Check all claims, expiration, and binding
+
+## Best Practices
+
```
bgavrilMS, line 583: I would add smth along the lines of "Use a certificate that stores the private key material in hardware, e.g. in TPM, for best protection"

docs/calling-downstream-apis/token-binding.md
cpp11nullptr: Added.

gladjohn (approved):

docs/calling-downstream-apis/calling-downstream-apis-README.md
```diff
@@ -10,6 +10,7 @@ Use this decision tree to select the best method for your scenario:
 |--------------------------------------|---------------------------------------|---------------------------------------------------------|
 | Microsoft Graph                      | You need to call Microsoft Graph APIS | GraphServiceClient                                      |
 | Azure SDK (Storage, KeyVault, etc.)  | You need to call Azure APIs (Azure SDK) | MicrosoftIdentityTokenCredential with Azure SDK clients  |
+| Custom API with Token Binding        | Enhanced security with certificate binding (mTLS PoP) | IDownstreamApi with `ProtocolScheme: "MTLS_POP"`        |
```
bgavrilMS, line 13: comment

Issue #3654, opened 2025-12-15
Infinite retry loop in Agent Identities flow when client credential config is invalid (regression from PR #3430)


jmprieur:
### Microsoft.Identity.Web Library

Microsoft.Identity.Web.TokenAcquisition

### Microsoft.Identity.Web version

3.12.0 (main branch, post-PR #3430)

### Web app

Not Applicable

### Web API

Not Applicable

### Token cache serialization

Not Applicable

## Summary:

After PR #3430 ("Reload certificates for all client credential based issues"), the token acquisition retry handler now attempts a certificate reload and retry for any InvalidClient error during client credentials token acquisition, not just specific certificate problems. This change introduces the risk of infinite retry loops when using `.WithAgentIdentities()` and a misconfigured client credential (e.g., wrong ClientID, expired ClientSecret, etc).

## Root cause analysis:

- The PR broadened the retry condition in `IsInvalidClientCertificateOrSignedAssertionError` to all `InvalidClient` errors.
- The Agent Identities flow (see `WithAgentIdentities`/`WithAgentUserIdentity`/`AgentUserIdentityMsalAddIn`) chains token requests: it first performs a client credential token request for the agent application, then another for the agent identity, and finally for the target resource. Each layer uses signed assertions and can propagate errors upward.
- If the first request fails with InvalidClient (wrong ClientID/Secret or expired keys), the retry machinery resets certs and recurses, but the same error reoccurs in every loop, as the core problem is not a certificate rotation but a configuration error (e.g., client not found, secret invalid, etc). This leads to an infinite loop and the call never returns.

## Confirmed symptoms:
- Observed with `.WithAgentIdentities()` token acquisitions.
- Easy to repro by setting a bad client id/secret/key for the agent application.
- Application hangs indefinitely (no exception escapes, call never returns).
- High CPU may be observed due to excessive retries, and throttling from ESTS

See full code analysis and stack below for more details.

### Minimum reproduction: 
1. Configure downstream web API to use `.WithAgentIdentities()` and valid FIC trust, but set the agent application's `ClientID` or `ClientSecret` incorrect or expired.
2. Attempt to acquire a token via `IAuthorizationHeaderProvider.CreateAuthorizationHeaderForAppAsync()` or `...ForUserAsync()` with the options from `.WithAgentIdentities()`.
3. The call will never return (infinite loop), and no exception escapes. 

_Example code:_
```csharp
[...]

PR #3673, opened 2026-01-15
Fix Certificate Reload Infinite Recursion Bug

4gust:
# Fix Certificate Reload Infinite Recursion Bug
## Description
Certificate reload logic was triggering for all invalid_client errors, not just certificate-related ones. This caused unnecessary retries for unrelated authentication failures like wrong passwords or missing app registrations.
Additionally, the shared _retryClientCertificate boolean flag had thread-safety issues in concurrent scenarios.
Solution
Restored specific error checking - Only reload certificates for actual cert errors:
•	AADSTS700027 - Invalid key
•	AADSTS700024 - Invalid time range
•	AADSTS7000214 - Certificate revoked
•	AADSTS1000502 - Certificate expired Replaced shared flag with per-call counter - Each call tracks its own retry count (max 1 retry), 
preventing infinite loops and other conditions.
Fixes issues : 
#3653 and #3654

## Solution
•	Removed: `private bool _retryClientCertificate` (shared state)
•	Added: `private const int MaxCertificateRetryAttempts = 1`
•	Added `int retryCount` parameter to private overloads of:	`AddAccountToCacheFromAuthorizationCodeAsync(AuthCodeRedemptionParameters)` and `GetAuthenticationResultForUserAsync(IEnumerable<string>, string?, string?, string?, ClaimsPrincipal?, TokenAcquisitionOptions?)`
•	GetAuthenticationResultForAppAsync(string, string?, string?, TokenAcquisitionOptions?)
•	Updated `IsInvalidClientCertificateOrSignedAssertionError(MsalServiceException, int)` to check retryCount < MaxCertificateRetryAttempts
### Others

### DIFF
--- a/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
@@ -53,8 +53,8 @@ namespace Microsoft.Identity.Web
         private readonly ConcurrentDictionary<string, SemaphoreSlim> _appSemaphores = new();
 
         private const string TokenBindingParameterName = "IsTokenBinding";
+        private const int MaxCertificateRetryAttempts = 1;
 
-        private bool _retryClientCertificate;
         protected readonly IMsalHttpClientFactory _httpClientFactory;
         protected readonly ILogger _logger;
         protected readonly IServiceProvider _serviceProvider;
@@ -125,6 +125,13 @@ namespace Microsoft.Identity.Web
         public async Task<AcquireTokenResult> AddAccountToCacheFromAuthorizationCodeAsync(
             AuthCodeRedemptionParameters authCodeRedemptionParameters)
         {
+            return await AddAccountToCacheFromAuthorizationCodeAsync(authCodeRedemptionParameters, retryCount: 0).ConfigureAwait(false);
+        }
+
+        private async Task<AcquireTokenResult> AddAccountToCacheFromAuthorizationCodeAsync(
+            AuthCodeRedemptionParameters authCodeRedemptionParameters,
+            int retryCount)
+        {
             _ = Throws.IfNull(authCodeRedemptionParameters.Scopes);
             MergedOptions mergedOptions = _tokenAcquisitionHost.GetOptions(authCodeRedemptionParameters.AuthenticationScheme, out string effectiveAuthenticationScheme);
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -899,15 +925,20 @@ public async Task RemoveAccountAsync(
...
+        private bool IsInvalidClientCertificateOrSignedAssertionError(MsalServiceException exMsal, int retryCount)
         {
-            return !_retryClientCertificate &&
-                string.Equals(exMsal.ErrorCode, Constants.InvalidClient, StringComparison.OrdinalIgnoreCase) &&
-                !exMsal.ResponseBody.Contains("AADSTS7000215" // No retry when wrong client secret.
-#if NET6_0_OR_GREATER
-                , StringComparison.OrdinalIgnoreCase
+            return retryCount < MaxCertificateRetryAttempts &&
+#if !NETSTANDARD2_0 && !NET462 && !NET472
+                (exMsal.Message.Contains(Constants.InvalidKeyError, StringComparison.OrdinalIgnoreCase)
```
bgavrilMS, line 932: Can we have a dictionary of these error codes to avoid the if / else and duplication?

tests/Microsoft.Identity.Web.Test/IsInvalidClientCertificateErrorTests.cs
```diff
@@ -0,0 +1,584 @@
...
+            // Assert
+            Assert.False(result, "Should return false for AADSTS50012 (invalid client credentials) - too generic");
+        }
+
+        /// <summary>
+        /// Test that the method returns false when a retry has already been attempted.
+        /// This prevents infinite retry loops.
+        /// </summary>
+        [Fact]
+        public void IsInvalidClientCertificateError_RetryCountAtMax_ReturnsFalse()
```
bgavrilMS, line 200: Seems like a lot of these tests can consolidated into 1

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
jmprieur: You should merge this one first: https://github.com/AzureAD/microsoft-identity-web/pull/3653, @bgavrilMS , @4gust

tests/Microsoft.Identity.Web.Test/IsInvalidClientCertificateErrorTests.cs
jmprieur: The only test test really matters would be the e2e with agentid and wrong cred, same with agent user identity

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
jmprieur: Please

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -124,6 +124,13 @@ public TokenAcquisition(
 #endif
         public async Task<AcquireTokenResult> AddAccountToCacheFromAuthorizationCodeAsync(
             AuthCodeRedemptionParameters authCodeRedemptionParameters)
+        {
+            return await AddAccountToCacheFromAuthorizationCodeAsync(authCodeRedemptionParameters, retryCount: 0).ConfigureAwait(false);
```
jmprieur, line 128: The auth code flow is not used by agent identities ...

bgavrilMS: But the retry logic applies to all flows. `GetAuthenticationResultForUserAsync` is also covered and that calls ROPC / OBO or AuthCode. ROPC is used.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
bgavrilMS: Fair enough @4gust - these fixes can be separated.

tests/Microsoft.Identity.Web.Test/IsInvalidClientCertificateErrorTests.cs
bgavrilMS: Just to update, @4gust and I looked at the Agentic tests, and when a bad credential is used, the tests indeed have infinite recursion.

PR #3683, opened 2026-01-27
Add trim AOT-safe APIs for AddMicrosoftIdentityWebApi and EnableTokenAcquisitionToCallDownstreamApi

anuchandy:
## Summary

This PR adds two trim-AOT-compatible APIs (to support Azure MCP Server trimming, which is required for azmcp 2.0 GA targeting mid-April).

These new APIs avoid reflection-based `ConfigurationBinder.Bind()` by using custom configuration binders, enabling Native AOT compilation and trimming.

These custom binders binds the properties defined in the repository's JSON schema files `microsoft-identity-web.json` and `Credentials.json`, which serve as the source of truth for the configuration contract.

### API Changes

| Original API (unchanged) | Trim-Safe API (new) |
|-------------|---------------|
| `AddMicrosoftIdentityWebApi(configuration, configSectionName, jwtBearerScheme, subscribeToEvents)` | `AddMicrosoftIdentityWebApiTrimSafe(configuration, configSectionName, jwtBearerScheme, subscribeToEvents)` |
| `EnableTokenAcquisitionToCallDownstreamApi()` | `EnableTokenAcquisitionToCallDownstreamApiTrimSafe()` |

### How It Works

The `TrimSafe` APIs internally use hand-written configuration binders that avoid reflection, making them compatible with Native AOT and trimming scenarios. 

## Usage

```csharp
// Using in Azure MCP Server or any web api app adhering to AOT compilance
//
builder.Services.AddAuthentication(JwtBearerDefaults.AuthenticationScheme)
    .AddMicrosoftIdentityWebApiTrimSafe(builder.Configuration)
    .EnableTokenAcquisitionToCallDownstreamApiTrimSafe();
```


src/Microsoft.Identity.Web.TokenAcquisition/WebApiBuilders.cs
```diff
@@ -51,5 +51,34 @@ public static MicrosoftIdentityAppCallsWebApiAuthenticationBuilder EnableTokenAc
...
+        /// Allows a higher level abstraction of security token (i.e. System.IdentityModel.Tokens.Jwt and more modern, Microsoft.IdentityModel.JsonWebTokens)
+        /// to be used with Microsoft Identity Web.
+        /// Developers should continue to use `EnableTokenAcquisitionToCallDownstreamApi`.
+        /// This API is not considered part of the public API and may change.
+        /// </summary>
+        /// <param name="authenticationScheme">Authentication scheme.</param>
+        /// <param name="services">The services being configured.</param>
+        /// <param name="configuration">IConfigurationSection.</param>
+        /// <returns>The authentication builder to chain.</returns>
+        public static MicrosoftIdentityAppCallsWebApiAuthenticationBuilder EnableTokenAcquisition(
```
jmprieur, line 65: ❌ Despite the XML doc saying "This API is not considered part of the public API and may change", it's marked public static and added to PublicAPI.Unshipped.txt

src/Microsoft.Identity.Web.TokenAcquisition/Internal/ConfidentialClientApplicationOptionsBinder.cs
```diff
@@ -0,0 +1,96 @@
...
+    /// and its base class <see cref="ApplicationOptions"/>.
+    /// </summary>
+    internal static class ConfidentialClientApplicationOptionsBinder
+    {
+        /// <summary>
+        /// Binds the <see cref="ConfidentialClientApplicationOptions"/> from the specified configuration section.
+        /// </summary>
+        /// <param name="options">The options instance to bind to.</param>
+        /// <param name="configurationSection">The configuration section containing the values.</param>
+        public static void Bind(ConfidentialClientApplicationOptions options, IConfigurationSection? configurationSection)
```
jmprieur, line 22: I don't think we need this binder. Isn't ConfidentialClientApplicationOptions provided from the MergedOptions?

src/Microsoft.Identity.Web.TokenAcquisition/Internal/CredentialDescriptionBinder.cs
```diff
@@ -0,0 +1,158 @@
...
+using Microsoft.Extensions.Configuration;
+using Microsoft.Identity.Abstractions;
+
+namespace Microsoft.Identity.Web.Internal
+{
+    /// <summary>
+    /// AOT-safe binder for <see cref="CredentialDescription"/>.
+    /// Binds configuration values based on the JSON schema defined in Credentials.json.
+    /// </summary>
+    internal static class CredentialDescriptionBinder
```
jmprieur, line 15: Some properties are not bound (for instance the AutoDecrypt credentials)
If we want to go this route (which is fragile), we'd need to have serious tests to check that all the properties we want are correctly bound.

Also this should go in Microsoft.Identity.Abstractions if we want to do that

src/Microsoft.Identity.Web.TokenAcquisition/Internal/MicrosoftIdentityApplicationOptionsBinder.cs
```diff
@@ -0,0 +1,193 @@
...
+
+namespace Microsoft.Identity.Web.Internal
+{
+    /// <summary>
+    /// AOT-safe binder for <see cref="MicrosoftIdentityApplicationOptions"/>.
+    /// Binds properties from the JSON schema AzureAdSection that are defined
+    /// on <see cref="MicrosoftIdentityApplicationOptions"/> and its base classes
+    /// (<see cref="MicrosoftEntraApplicationOptions"/> and <see cref="IdentityApplicationOptions"/>).
+    /// </summary>
+    internal static class MicrosoftIdentityApplicationOptionsBinder
```
jmprieur, line 17: @anuchandy what is not trimmable in MicrosoftIdentityApplicationOptions?

src/Microsoft.Identity.Web.TokenAcquisition/Internal/MicrosoftIdentityOptionsBinder.cs
```diff
@@ -0,0 +1,273 @@
...
+    /// on <see cref="MicrosoftIdentityOptions"/> and its base class (OpenIdConnectOptions).
+    /// </summary>
+    internal static class MicrosoftIdentityOptionsBinder
+    {
+        /// <summary>
+        /// Binds the <see cref="MicrosoftIdentityOptions"/> from the specified configuration section.
+        /// </summary>
+        /// <param name="options">The options instance to bind to.</param>
+        /// <param name="configurationSection">The configuration section containing the values.</param>
+        public static void Bind(MicrosoftIdentityOptions options, IConfigurationSection? configurationSection)
```
jmprieur, line 22: ❌ This binder has its own BindCertificateDescription and BindCertificateDescriptionCollection methods, while CredentialDescriptionBinder exists separately. There's partial overlap but not full reuse (CertificatelDescription inherits from CredentialDescription)

src/Microsoft.Identity.Web.TokenAcquisition/WebApiBuilders.cs
anuchandy: Thanks - I’ll double-check PublicAPI.Unshipped.txt and fix it. Right, this class is already under a .Internal namespace

anuchandy: I Checked this. The [`WebApiBuilders`](https://github.com/AzureAD/microsoft-identity-web/blob/5c87b5cc8bac14576c13810c32445db3bf060b2e/src/Microsoft.Identity.Web.TokenAcquisition/WebApiBuilders.cs#L16) class is declared as `public `static, which means its public methods are technically part of the public API surface from the analyzer's perspective. The `Microsoft.Identity.Web.Internal` namespace here is a naming convention to signal to devs that this API should not be used, but it doesn't change the actual C# accessibility level. 

The code change here follows the existing pattern - the original `EnableTokenAcquisition` method (the AOT-unsafe one) with the same XML comment is already in the [PublicAPI.Shipped.txt](https://github.com/AzureAD/microsoft-identity-web/blob/5c87b5cc8bac14576c13810c32445db3bf060b2e/src/Microsoft.Identity.Web.TokenAcquisition/PublicAPI/net10.0/PublicAPI.Shipped.txt#L157).

src/Microsoft.Identity.Web.TokenAcquisition/Internal/ConfidentialClientApplicationOptionsBinder.cs
anuchandy: thanks, to enable `EnableTokenAcquisitionToCallDownstreamApiTrimSafe()`, I was following the same steps as the existing AOT-unsafe `EnableTokenAcquisitionToCallDownstreamApi()`. As we can see in the current (AOT-unsafe) implementation, it does bind into `ConfidentialClientApplicationOptions`. I’ve captured that AOT-unsafe flow below.

To support the same logic in the new AOT-safe equivalent, we introduced `ConfidentialClientApplicationOptionsBinder` to perform the binding.

With my limited familiarity with this package, I’m not sure whether the existing code was doing this unnecessarily- but I was aiming to stay consistent with the current behavior.

```
┌─────────────────────────────────────────────────────────────────────────────────┐
│  MicrosoftIdentityWebApiAuthenticationBuilderWithConfiguration                  │
│                                                                                 │
│  EnableTokenAcquisitionToCallDownstreamApi()                                    │
│                                                                                 │
│    return EnableTokenAcquisitionToCallDownstreamApi(                            │
│               options => ConfigurationSection?.Bind(options));                  │
└─────────────────────────────────────────┬───────────────────────────────────────┘
                                          │
                                          ▼
┌─────────────────────────────────────────────────────────────────────────────────┐
│  MicrosoftIdentityWebApiAuthenticationBuilder                                   │
│                                                                                 │
│  EnableTokenAcquisitionToCallDownstreamApi(                                     │
│      Action<ConfidentialClientApplicationOptions>                               │
│          configureConfidentialClientApplicationOptions)                         │
│                                                                                 │
│    CallsWebApiImplementation(                                                   │
│        Services,                                                                │
│        JwtBearerAuthenticationScheme,                                           │
│        configureConfidentialClientApplicationOptions,                           │
│        ConfigurationSection);                                                   │
└─────────────────────────────────────────┬───────────────────────────────────────┘
                                          │
                                          ▼
┌─────────────────────────────────────────────────────────────────────────────────┐
│  MicrosoftIdentityWebApiAuthenticationBuilder                                   │
│                                                                                 │
│  CallsWebApiImplementation(...)                                                 │
│                                                                                 │
│    services.Configure(jwtBearerAuthenticationScheme,                            │
│                       configureConfidentialClientApplicationOptions);           │
│                       ▲                                                         │
│                       │                                                         │
│         Registers: options => ConfigurationSection?.Bind(options)               │
│                    (binds ConfidentialClientApplicationOptions)                 │
└─────────────────────────────────────────────────────────────────────────────────┘

```

src/Microsoft.Identity.Web.TokenAcquisition/Internal/CredentialDescriptionBinder.cs
anuchandy: Sure - I’ll work on adding tests for all the binders. I’ll also take a look at the other repo; I wasn’t aware the work spans two repos/packages, so thanks for the pointer.

src/Microsoft.Identity.Web.TokenAcquisition/Internal/MicrosoftIdentityApplicationOptionsBinder.cs
anuchandy: The `MicrosoftIdentityApplicationOptions` type itself doesn't have any inherently untrimmable properties - all its properties (including inherited ones from `MicrosoftEntraApplicationOptions` and `IdentityApplicationOptions`) are bound by this hand-written binder.

What is trim/AOT-unfriendly is the original use of `ConfigurationBinder.Bind()` to populate this type.

src/Microsoft.Identity.Web.TokenAcquisition/Internal/CredentialDescriptionBinder.cs
```diff
@@ -0,0 +1,158 @@
...
+
+            // CustomSignedAssertion - for SourceType = CustomSignedAssertion
+            var customSignedAssertionProviderName = configurationSection[nameof(CredentialDescription.CustomSignedAssertionProviderName)];
+            if (!string.IsNullOrEmpty(customSignedAssertionProviderName))
+            {
+                options.CustomSignedAssertionProviderName = customSignedAssertionProviderName;
+            }
+
+            // Note: DecryptKeysAuthenticationOptions is a complex nested type used for AutoDecryptKeys.
+            // Skipping for now as it requires a separate binder and is less commonly used.
```
cpp11nullptr, line 123: Can we create a work item with providing a link there to do not forget about it?

src/Microsoft.Identity.Web.TokenAcquisition/Internal/CredentialDescriptionBinder.cs
```diff
@@ -0,0 +1,158 @@
...
+            // SourceType (enum) - required property that determines which other properties are relevant
+            var sourceTypeValue = configurationSection[nameof(CredentialDescription.SourceType)];
+            if (!string.IsNullOrEmpty(sourceTypeValue) && Enum.TryParse<CredentialSource>(sourceTypeValue, ignoreCase: true, out var sourceType))
+            {
+                options.SourceType = sourceType;
+            }
+
+            // ClientSecret - for SourceType = ClientSecret
+            var clientSecret = configurationSection[nameof(CredentialDescription.ClientSecret)];
+            if (!string.IsNullOrEmpty(clientSecret))
```
cpp11nullptr, line 38: In case if `clientSecret` is empty string, `options.ClientSecret` will contain old value - is it expected? It applies thorough.

src/Microsoft.Identity.Web.TokenAcquisition/Internal/CredentialDescriptionBinder.cs
cpp11nullptr: Perhaps, unit tests with reflection (so it can iterate over all properties of option class to see whether its bound to configuration) can alleviate it in case if this path to bind is chosen.

src/Microsoft.Identity.Web.TokenAcquisition/Internal/ConfidentialClientApplicationOptionsBinder.cs
```diff
@@ -0,0 +1,96 @@
...
+            {
+                options.TenantId = tenantId;
+            }
+
+            if (configurationSection[nameof(options.Instance)] is string instance)
+            {
+                options.Instance = instance;
+            }
+
+            if (configurationSection[nameof(options.AadAuthorityAudience)] is string aadAuthorityAudience &&
```
cpp11nullptr, line 45: There are some repetitive blocks of code which can be made more declarative, can we consider extracting it to helper functions with calling it something like below? As benefit, there will be a bit clearer bind code and each bind util method can be granularly unit tested.

```csharp
..
BinderUtils.BindEnum<AadAuthorityAudience>(options, configurationSection, nameof(options.AadAuthorityAudience));
BinderUtils.BindEnum<AzureCloudInstance>(options, configurationSection, nameof(options.AzureCloudInstance));
BinderUtils.BindString(options, configurationSection, nameof(options.RedirectUri));
..
```

PR #3684, opened 2026-01-28
Add validation baseline exclusions

JoshLozensky:
Replacing validation exclusions removed in [PR #3659](https://github.com/AzureAD/microsoft-identity-web/pull/3659), these are needed to avoid packing errors when packing the GraphServiceClient project which has a different baseline validation version number from the rest of Id Web

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -56,6 +56,12 @@
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
   </ItemGroup>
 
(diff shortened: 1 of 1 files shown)


bgavrilMS (approved):

JoshLozensky:
@jmprieur I misspoke in the description and have edited to clarify that the different version is for baseline validation of the package, not a different version for the package itself. I'm seeing that the clientsemver used to be directly set in the csproj for the GraphServiceClient package and that is when the package-specific baseline validation was added as well. I'll try removing the unique baseline validation version from the package and test in another branch, merging this one so the build is fixed asap.

PR #3700, opened 2026-02-03
Remove unnecessary flag switch that causes official build to break.

MZOLN:
This seem unnecessary and causes overrides of binaries while signing.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -4,7 +4,7 @@
     <!-- This needs to be greater than or equal to the validation baseline version. The conditional logic around TargetNetNext is there
     to avoid NU5104 for packing a release version library with prerelease deps. By adding preview to it, that warning is avoided.
     -->
-    <MicrosoftIdentityWebVersion Condition="'$(MicrosoftIdentityWebVersion)' == ''">4.3.1-preview</MicrosoftIdentityWebVersion>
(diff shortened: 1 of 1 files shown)


JoshLozensky (approved):

Directory.Build.props
```diff
@@ -4,7 +4,7 @@
     <!-- This needs to be greater than or equal to the validation baseline version. The conditional logic around TargetNetNext is there
     to avoid NU5104 for packing a release version library with prerelease deps. By adding preview to it, that warning is avoided.
     -->
-    <MicrosoftIdentityWebVersion Condition="'$(MicrosoftIdentityWebVersion)' == ''">4.3.1-preview</MicrosoftIdentityWebVersion>
```
jmprieur, line 7: It should be 4.3.1
IT should never have -preview1 in this file.

@copilot please update the agent.md to check that

src/Microsoft.Identity.Web.TokenAcquisition/Microsoft.Identity.Web.TokenAcquisition.csproj
```diff
@@ -6,7 +6,6 @@
     <ProjectGuid>{74737C0A-25E6-4E2C-B3AA-F7EECB7A07C7}</ProjectGuid>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <PackageReadmeFile>README.md</PackageReadmeFile>
-    <GeneratePackageOnBuild>True</GeneratePackageOnBuild>
```
jmprieur, line 9: What's the issue with this one?

PR #3711, opened 2026-02-05
Fix losing certificates for credentials with `CredentialSource.Certificate`

pentp:
`DefaultCertificateLoader.ResetCertificates` incorrectly resets certificates for `CredentialSource.Certificate` which causes extremely confusing and hard to debug errors about null certificates later.
Also cleaned up a bit of redundant code.

### DIFF
--- a/src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
+++ b/src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
@@ -79,7 +79,8 @@ namespace Microsoft.Identity.Web
         {
             _ = Throws.IfNull(credentialDescription);
 
-            if (credentialDescription.CachedValue == null)
+            if (credentialDescription.CachedValue == null
+                && (credentialDescription.SourceType == CredentialSource.CustomSignedAssertion || CredentialSourceLoaders.ContainsKey(credentialDescription.SourceType)))
             {
                 // Get or create a semaphore for this credentialDescription
                 var semaphore = _loadingSemaphores.GetOrAdd(credentialDescription.Id, (v) => new SemaphoreSlim(1));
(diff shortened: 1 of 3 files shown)


src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
```diff
@@ -79,7 +79,8 @@ public async Task LoadCredentialsIfNeededAsync(CredentialDescription credentialD
         {
             _ = Throws.IfNull(credentialDescription);
 
-            if (credentialDescription.CachedValue == null)
+            if (credentialDescription.CachedValue == null
+                && (credentialDescription.SourceType == CredentialSource.CustomSignedAssertion || CredentialSourceLoaders.ContainsKey(credentialDescription.SourceType)))
```
bgavrilMS, line 83: @pentp - can you please explain why this was done?

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
```diff
@@ -140,11 +141,11 @@ public void ResetCredentials(IEnumerable<CredentialDescription> credentialDescri
         {
             foreach (var credentialDescription in credentialDescriptions)
             {
-                credentialDescription.CachedValue = null;
```
bgavrilMS, line 143: I think we would need unit tests.

src/Microsoft.Identity.Web.Certificate/DefaultCredentialsLoader.cs
pentp: This is principally for `CredentialSource.Certificate` - it doesn't have an associated loader and its `CachedValue` will always be null, so this just skips the semaphore setup and locking entirely.

PR #3714, opened 2026-02-09
Update OnBeforeTokenAcquisitionForOnBehalfOf event to construct ClaimsPricipal from token

rayluo:
This pull request introduces a new `OnBehalfOfEventArgs` class to provide richer context for on-behalf-of token acquisition events and updates the related public API signatures to use this new class instead of passing a `ClaimsPrincipal` directly. This change improves extensibility and clarity for event handlers involved in token acquisition flows.

Key changes:

**API Enhancements:**

* Introduced new `OnBehalfOfEventArgs` class, which encapsulates the user (`ClaimsPrincipal`) and the user assertion token, providing additional context for on-behalf-of token acquisition operations. (`src/Microsoft.Identity.Web.TokenAcquisition/OnBehalfOfEventArgs.cs`, [src/Microsoft.Identity.Web.TokenAcquisition/OnBehalfOfEventArgs.csR1-R27](diffhunk://#diff-10b376eb108fb8e3f8fc7c268ddd07817c0010b05153fb7fa637551ebee3a9e0R1-R27))
* Updated the public API surface for all supported frameworks to include the new `OnBehalfOfEventArgs` class and its properties. (`src/Microsoft.Identity.Web.TokenAcquisition/PublicAPI/net8.0/PublicAPI.Unshipped.txt`, `src/Microsoft.Identity.Web.TokenAcquisition/PublicAPI/net9.0/PublicAPI.Unshipped.txt`, `src/Microsoft.Identity.Web.TokenAcquisition/PublicAPI/net10.0/PublicAPI.Unshipped.txt`, `src/Microsoft.Identity.Web.TokenAcquisition/PublicAPI/net462/PublicAPI.Unshipped.txt`, `src/Microsoft.Identity.Web.TokenAcquisition/PublicAPI/net472/PublicAPI.Unshipped.txt`, [src/Microsoft.Identity.Web.TokenAcquisition/PublicAPI/net10.0/PublicAPI.Unshipped.txtR4-R13](diffhunk://#diff-3fa7a4c9e11b9133c8fa013832f87021bd224b9156e7af0ac943241175d58bb9R4-R13))

**Breaking Changes to Event Handler Signatures:**

* Changed the signatures of `BeforeTokenAcquisitionForOnBehalfOf.Invoke` and its async counterpart to accept an `OnBehalfOfEventArgs` instance instead of a `ClaimsPrincipal`, across all relevant API files. This affects both synchronous and asynchronous event hooks.
* Updated internal API declarations for `TokenAcquisitionExtensionOptions.InvokeOnBeforeTokenAcquisitionForOnBehalfOf` and its async version to use `OnBehalfOfEventArgs` instead of `ClaimsPrincipal`.

These changes collectively make the token acquisition event handling more robust and future-proof by encapsulating relevant data in a dedicated event args class.PR #3712 changed the usage of user to userHint, but based on [the logic at the beginning of the affected method](https://github.com/AzureAD/microsoft-identity-web/compare/rayluo/UpdateOboEvent?expand=1#diff-3600735102cf2582ea4dc8277f799690b8a8135f4971e0be919fbecd053ae2caL1186-L1196), the single source of truth in terms of OBO shall be derived from tokenUsedToCallTheWebApi. This change constructs a ClaimsPrincipal from the token string.

### DIFF
--- a/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
@@ -463,9 +463,7 @@ namespace Microsoft.Identity.Web
                 var dict = MergeExtraQueryParameters(mergedOptions, tokenAcquisitionOptions);
                 if (dict != null)
                 {
-#pragma warning disable CS0618 // Type or member is obsolete
                     builder.WithExtraQueryParameters(dict);
-#pragma warning restore CS0618 // Type or member is obsolete
                 }
 
                 if (tokenAcquisitionOptions.ExtraHeadersParameters != null)
@@ -687,9 +685,7 @@ namespace Microsoft.Identity.Web
 
                 if (dict != null)
                 {
-#pragma warning disable CS0618 // Type or member is obsolete
                     builder.WithExtraQueryParameters(dict);
-#pragma warning restore CS0618 // Type or member is obsolete
                 }
                 if (tokenAcquisitionOptions.ExtraHeadersParameters != null)
                 {
@@ -1337,9 +1333,7 @@ namespace Microsoft.Identity.Web
                                 dict.Remove(subAssertionConstant);
                             }
 
-#pragma warning disable CS0618 // Type or member is obsolete
                             builder.WithExtraQueryParameters(dict);
-#pragma warning restore CS0618 // Type or member is obsolete
                         }
                         if (tokenAcquisitionOptions.ExtraHeadersParameters != null)
                         {
@@ -1501,9 +1495,7 @@ namespace Microsoft.Identity.Web
 
                 if (dict != null)
                 {
-#pragma warning disable CS0618 // Type or member is obsolete
                     builder.WithExtraQueryParameters(dict);
-#pragma warning restore CS0618 // Type or member is obsolete
                 }
                 if (tokenAcquisitionOptions.ExtraHeadersParameters != null)
                 {


src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -1255,7 +1255,23 @@ private void NotifyCertificateSelection(
                     {
                         if (addInOptions != null)
                         {
-                            await addInOptions.InvokeOnBeforeTokenAcquisitionForOnBehalfOfAsync(builder, tokenAcquisitionOptions, userHint!).ConfigureAwait(false);
+                            ClaimsPrincipal? principalToUse;
+
+                            // Create ClaimsPrincipal from the token if available, otherwise pass null
+                            if (!string.IsNullOrEmpty(tokenUsedToCallTheWebApi))
```
trwalke: Oh yes, I see, the token can come from multiple sources so it is better to use this value. However, do we need to recreate the ClaimsPrincipal here? It doesnt seem like any additional data is being added outside of the original token so I guess we can just send the token itself (unless that is a security concern)?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -1255,7 +1255,23 @@ private void NotifyCertificateSelection(
...
+                            {
+                                var handler = new JwtSecurityTokenHandler();
+                                var jwtToken = handler.ReadJwtToken(tokenUsedToCallTheWebApi);
```
bgavrilMS: It seems pretty heavy to re-parse the token? Can we rationalize all the ClaimsPrincipals here?

- from the original token `userHint?.GetBootstrapToken() ?? _tokenAcquisitionHost.GetTokenUsedToCallWebAPI();` 
- the                     `ClaimsPrincipal? user = _tokenAcquisitionHost.GetUserFromRequest();`
- a ClaimsPrincipal made of `tokenUsedToCallTheWebApi`

How are these different in the 4 cases of OBO / LR OBO (session Key is / isn't null, token is / isn't null).

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
rayluo: Indeed. At least for the current CDT OBO project, the event callback does not really need a `ClaimsPrincipal`. In fact, it does not even currently need the real string value of the token. (Because in this specific case, a cdt token will always be available via a different mechanism `CreateAuthorizationHeaderForUserAsync(new AcquireTokenOptions().WithCdt("the cdt"))`. It just needs the presence or absence of the string token to know whether it is in the initial OBO or in the subsequent long-running state.)

The implementation in this PR currently builds a ClaimsPrincipal only because the callback was already implemented that way (and shipped?).

So, from CDT OBO perspective, it would be totally OK to change this callback to relay the string token only. But I'll leave it to @trwalke and @bgavrilMS to comment whether changing the callback signature at this point would be a breaking change.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
rayluo: Please refer to the conversation in the other thread above.

The handling of those 4 cases is in the other PR in MISE code base. From CDT OBO perspective, there is only 2 cases relevant.

* token is not null - this is the initial OBO, it will be converted into an SP OBO with the app token inside cdt. The presence or absence of session key will simply be passed along.
* token is null - this is the refresh of long-running obo. The result will be packed into a cdt.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
bgavrilMS: I don't think there's anybody using this extensiblity point. We haven't even released ID.Web with it. It's ok to change it. My main concern with JwtToken(string) is that it will introduce parsing. 

From a future proofing perspective, maybe it's worth having both `string` and `ClaimsPrincipal`

trwalke: Update this to use a DTO. this will allow us to send what is needed and add to it in the future

src/Microsoft.Identity.Web/AzureFunctionsAuthenticationHttpContextExtension.cs
```diff
@@ -20,7 +20,7 @@ public static class AzureFunctionsAuthenticationHttpContextExtension
         /// Calling this method from your Azure function validates the token and exposes the identity of the user or app on behalf of which your function is called,
         /// in the HttpContext.User member, where your function can make use of it.
         /// </summary>
-        /// <param name="httpContext">The current HTTP Context, such as req.HttpContext.</param>
+        /// <param name="httpContext">The current HTTP UserAssertionToken, such as req.HttpContext.</param>
```
bgavrilMS: @trwalke - I think your "search and replace" went a bit too far.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -1255,23 +1297,13 @@ private void NotifyCertificateSelection(
...
-                                var jwtToken = handler.ReadJwtToken(tokenUsedToCallTheWebApi);
-                                var identity = new CaseSensitiveClaimsIdentity(jwtToken.Claims);
-                                identity.BootstrapContext = tokenUsedToCallTheWebApi;
-                                principalToUse = new ClaimsPrincipal(identity);
-                            }
-                            else
-                            {
-                                principalToUse = null;
-                            }
+                                User = userHint,
```
rayluo, line 1298: Non-blocking suggestions:

Do we consider rename the `user` property here to `userHint`? Reason being, a same name variable `user` was defined in the closest scope here (just 20+ lines above), yet we are populating this property with the value `userHint` defined in outer scope (80+ lines above). The reason we have this PR in the first place was because those differences weren't clear. If we don't improve the naming here, the same trap could bite us in the future. Just my 2 cents.

trwalke: Makes sense, I updated the name of the variable above

src/Microsoft.Identity.Web/AzureFunctionsAuthenticationHttpContextExtension.cs
trwalke: Yeah, I used the rename feature in the right click menu and it looks like it will now randomly change text all over the solution now. Seems like a VS Bug

PR #3728, opened 2026-02-18
Propagate long running web API session key back to caller in user token acquisition flow

cpp11nullptr:
When long running web API session is created with "AllocateForMe" session key, a [new](https://github.com/AzureAD/microsoft-identity-web/blob/dca1fc0404c1bfd5ae1349b77a36c4864ea9918d/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs#L1262) session key is auto-generated and set in `tokenAcquisitionOptions`. This change propagates a new session key to caller of `TokenAcquirer::GetTokenForUserAsync()` so consumer side can use this key to fetch cached token later.

Note that *CreateAuthorizationHeaderForUserAsync* scenario already [does](https://github.com/AzureAD/microsoft-identity-web/blob/dca1fc0404c1bfd5ae1349b77a36c4864ea9918d/src/Microsoft.Identity.Web.TokenAcquisition/DefaultAuthorizationHeaderProvider.cs#L185) it so no changes are needed there.

### DIFF
--- a/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquirer.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquirer.cs
@@ -39,7 +39,8 @@ namespace Microsoft.Identity.Web
                 ).ConfigureAwait(false);
 
             // Propagate LongRunningWebApiSessionKey (possibly auto-generated) back to the caller
-            if (tokenAcquisitionOptions is not null && effectiveOptions is not null)
+            if (tokenAcquisitionOptions is not null && effectiveOptions is not null
+                && !string.IsNullOrEmpty(effectiveOptions.LongRunningWebApiSessionKey))
             {
                 tokenAcquisitionOptions.LongRunningWebApiSessionKey = effectiveOptions.LongRunningWebApiSessionKey;
             }
--- a/tests/Microsoft.Identity.Web.Test/TokenAcquirerTests.cs
+++ b/tests/Microsoft.Identity.Web.Test/TokenAcquirerTests.cs
@@ -216,6 +216,51 @@ namespace Microsoft.Identity.Web.Test
             Assert.Null(callerOptions.LongRunningWebApiSessionKey);
         }
 
+        [Theory]
+        [InlineData(null)]
+        [InlineData("")]
+        public async Task GetTokenForUserAsync_WhenEffectiveKeyIsNullOrEmpty_DoesNotOverwriteCallerKey(string? effectiveKeyValue)
+        {
+            // Arrange
+            const string originalKey = "test-key";
+            var authResult = CreateMockAuthenticationResult();
+            var callerOptions = new AcquireTokenOptions
+            {
+                LongRunningWebApiSessionKey = originalKey,
+            };
+
+            _tokenAcquisition.GetAuthenticationResultForUserAsync(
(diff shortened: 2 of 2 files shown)


src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquirer.cs
```diff
@@ -28,15 +28,22 @@ async Task<AcquireTokenResult> ITokenAcquirer.GetTokenForUserAsync(
...
                 tokenAcquisitionOptions?.UserFlow,
                 user,
-                GetEffectiveTokenAcquisitionOptions(tokenAcquisitionOptions, authenticationScheme, cancellationToken)
+                effectiveOptions
                 ).ConfigureAwait(false);
 
+            // Propagate LongRunningWebApiSessionKey (possibly auto-generated) back to the caller
+            if (tokenAcquisitionOptions is not null && effectiveOptions is not null)
+            {
+                tokenAcquisitionOptions.LongRunningWebApiSessionKey = effectiveOptions.LongRunningWebApiSessionKey;
```
trwalke, line 45: so this is already happening here: https://github.com/AzureAD/microsoft-identity-web/blob/d371fc6f1d6b3069c6372d91c5bf7d819ce7b53e/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs#L1272

Are you sure none of the user flows will set this twice?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquirer.cs
```diff
@@ -28,15 +28,22 @@ async Task<AcquireTokenResult> ITokenAcquirer.GetTokenForUserAsync(
...
                 tokenAcquisitionOptions?.Tenant,
                 tokenAcquisitionOptions?.UserFlow,
                 user,
-                GetEffectiveTokenAcquisitionOptions(tokenAcquisitionOptions, authenticationScheme, cancellationToken)
+                effectiveOptions
                 ).ConfigureAwait(false);
 
+            // Propagate LongRunningWebApiSessionKey (possibly auto-generated) back to the caller
+            if (tokenAcquisitionOptions is not null && effectiveOptions is not null)
+            {
```
trwalke, line 44: Should probably check to see if effectiveOptions.LongRunningWebApiSessionKey is no null or empty

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquirer.cs
cpp11nullptr: Before this change, tokenAcquisitionOptions, which are passed to GetAuthenticationResultForUserAsync() in GetTokenForUserAsync() method, are created as temporary variable (result of GetEffectiveTokenAcquisitionOptions() call), so `tokenAcquisitionOptions.LongRunningWebApiSessionKey = sessionKey` is lost when GetAuthenticationResultForUserAsync() is completed.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquirer.cs
cpp11nullptr: Done.

trwalke (approved):

PR #3747, opened 2026-03-11
MTLS Without Tokens Support - DownstreamApi

tlupes:
# Support for mTLS authentication only mode for DownstreamApi

Adds support for mTLS only mode for DownstreamApi. This allows you to configure a protocol of "MTLS" and the configured certificate will be attached to the request without and authentication header.

This is set up to work with all existing certificate infrastructure as Token Acquisition. This means that the Managed Certificate logic will fit right in.

## Description

This essentially does the following:

* Refactors all the certificate handling logic (including the observers) into a reusable "CredentialProvider"
* Modifies TokenAquesition to use this CredentialProvider
* Configures DownstreamApi to use this CredentialProvider as well.
    * If Protocol is MTLS, tokens are skipped
    * If Protocol is MTLS, it takes the first certificate credential and adds it as a certificate.
    * If Protocol is MTLS, the result of the call is checked right there to see if it's a certificate error. If so, it retries once.


src/Microsoft.Identity.Web.TokenAcquisition/ConfidentialClientApplicationBuilderExtension.cs
```diff
@@ -75,100 +99,14 @@ public static async Task<ConfidentialClientApplicationBuilder> WithClientCredent
             }
         }
 
-        internal /* for test */ async static Task<CredentialDescription?> LoadCredentialForMsalOrFailAsync(
-            IEnumerable<CredentialDescription> clientCredentials,
-            ILogger logger,
-            ICredentialsLoader credentialsLoader,
-            CredentialSourceLoaderParameters? credentialSourceLoaderParameters)
+        // Used for backcomapt support.
```
cpp11nullptr: nit: `backcompat`.

src/Microsoft.Identity.Web.TokenAcquisition/CredentialsProvider.cs
```diff
@@ -0,0 +1,364 @@
...
+
+            // If deselected, clear the values so they can be reloaded.
+            if (action == CerticateObserverAction.Deselected)
+            {
+                certificateDescription.Certificate = null;
+                certificateDescription.CachedValue = null;
+            }
+        }
+
+        private class LogMessages
```
cpp11nullptr: Can we consider extracting it to separate file?

cpp11nullptr (approved):
Can we also add sample app (see tests/DevApps/MtlsPop as an example) which shows a developer experience?

cpp11nullptr (approved):

src/Microsoft.Identity.Web.TokenAcquisition/CredentialsProvider.cs
```diff
@@ -0,0 +1,193 @@
...
+                    {
+
+                        await _credentialsLoader.LoadCredentialsIfNeededAsync(credential, credentialSourceLoaderParameters);
+                    }
+                    catch (Exception ex)
+                    {
+                        LogMessages.AttemptToLoadCredentialsFailed(_logger, credential, ex);
+                        errorMessage += $"Credential {credential.Id} failed because: {ex} \n";
+                    }
+
```
neha-bhargava, line 76: nit: extra space

src/Microsoft.Identity.Web.TokenAcquisition/CredentialsProvider.LogMessages.cs
```diff
@@ -7,9 +7,9 @@
 
 namespace Microsoft.Identity.Web
 {
-    internal partial class ConfidentialClientApplicationBuilderExtension
+    internal partial class CredentialsProvider
```
neha-bhargava, line 10: Why is this renamed?

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
```diff
@@ -25,18 +28,28 @@ namespace Microsoft.Identity.Web
...
         // This MSAL HTTP client factory is used to create HTTP clients with mTLS binding certificate.
         // Note, that it doesn't replace _httpClientFactory to keep backward compatibility and ability
         // to create named HTTP clients for non-mTLS scenarios.
-        private readonly IMsalHttpClientFactory? _msalHttpClientFactory;
+        private readonly IMsalHttpClientFactory _msalHttpClientFactory;
 
         private readonly IOptionsMonitor<DownstreamApiOptions> _namedDownstreamApiOptions;
 
         private const string Authorization = "Authorization";
+        private const string MtlsProtocolScheme = "MTLS";
```
neha-bhargava: is MTLS scheme different than MTLS POP below?

src/Microsoft.Identity.Web.TokenAcquisition/CredentialsProvider.LogMessages.cs
tlupes: Originally a lot of the loading logic, and thus log messages, were in ConfidentialClientApplicationBuilderExtension. All of the credential loading logic has been moved to CredentialsProvider, so that it can be reused. These log messages used to only be used in the Extension class, now they are only used in the CredentialsProvider class, so it was renamed for consistency.

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
tlupes: Added a comment! They are very different, MTLS uses the certificate as the authentication and authorization. MTLS_POP uses the certificate (essentially) to validate the token, the token itself for authentication and authorization.

neha-bhargava (approved):
Looks good. Thanks!

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
```diff
@@ -47,6 +75,7 @@ internal partial class DownstreamApi : IDownstreamApi
         /// <param name="namedDownstreamApiOptions">Named options provider.</param>
         /// <param name="httpClientFactory">HTTP client factory.</param>
         /// <param name="logger">Logger.</param>
+        [Obsolete("Use the overload which takes msalHttpClientFactory and credentialsProvider for mTLS support")]
```
bgavrilMS: Most ppl will not use mtls or mtls_pop . Isn't this constructor the best suited for them?

tlupes: This is internal, but I will mark it as not obsolete in case somebody does not have credentialsProvider set in their service collection.

src/Microsoft.Identity.Web.TokenAcquisition/ConfidentialClientApplicationBuilderExtension.cs
```diff
@@ -4,16 +4,18 @@
...
 using Microsoft.Identity.Abstractions;
 using Microsoft.Identity.Client;
+using static Microsoft.Identity.Web.TokenAcquisition;
 
 namespace Microsoft.Identity.Web
 {
     internal static partial class ConfidentialClientApplicationBuilderExtension
     {
-        [Obsolete(IDWebErrorMessage.WithClientCredentialsIsObsolete, false)]
+        [Obsolete(IDWebErrorMessage.WithClientCredentialsIsObsolete, true)]
```
bgavrilMS: Why is this fully deprecated ? Should we just delete it? (note that it's part of internal class, it's not public API)

tlupes: This was done to satisfy the Internal API analyzer. I will take a look for APIs that aren't used outside of tests and remove them.

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
```diff
@@ -585,10 +701,37 @@ public Task<HttpResponseMessage> CallApiForAppAsync(
...
+            // Obtention of the authorization header (except when calling an anonymous endpoint)
+            // which is done by not specifying any scopes or mTLS scheme.
+            if (string.Equals(effectiveOptions.ProtocolScheme, MtlsProtocolScheme, StringComparison.OrdinalIgnoreCase))
+            {
+                if (_credentialsProvider == null)
+                {
+                    throw new InvalidOperationException("mTLS authentication requires a Credentials Provider object to be registered, but no such service was found.");
+                }
+
+                credential = await _credentialsProvider.GetCredentialAsync(
```
gladjohn, line 715: Should this pass `mergedOptions` to preserve the selected authentication scheme?

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
```diff
@@ -528,52 +583,113 @@ public Task<HttpResponseMessage> CallApiForAppAsync(
...
+                        {
+                            retry = true;
+                            hasRetried = true;
+                        }
+                    }
                 }
             }
+            while (retry);
 
             return downstreamApiResult;
```
gladjohn, line 671: Should DownstreamApi also be calling some certificate observer notification logic after the response comes back (to track successful or failed mTLS certificate usage), similar to how TokenAcquisition notifies observers at lines 755 and 766?

gladjohn (approved):

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
tlupes: We could adjust it to have `Protocol = mergedOptions.ProtocolScheme`, however I elected to just pass in the constant to ensure that there's consistent casing there (as the options could be like mtls, MTLS,mTlS, but it's good to expose something consistent downstream).

tlupes: Passing the whole MergedOptions object would be pretty heavy. Perhaps that should be added in the future, but for now what's provided is more than enough information.

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
tlupes: This is actually handled here in line 641 and 653. It only needs to handle mTLS as everything related to tokens is handled in TokenAcquisition.

tlupes: _perhaps_ we could do a bit more tracking for mTLS_POP scenarios? But that can be a separate change, and I don't know if it's really needed. Those are already captured when the token is obtained, might be valuable to also track when the mTLS part is used, but I don't know if that's needed.

src/Microsoft.Identity.Web.TokenAcquisition/Microsoft.Identity.Web.TokenAcquisition.csproj
```diff
@@ -28,6 +28,10 @@
     <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="3.1.0" />
   </ItemGroup>
 
+  <ItemGroup>
+    <None Remove="PublicAPI\net9.0\InternalAPI.Unshipped.txt" />
```
bgavrilMS: Pls remove this.

src/Microsoft.Identity.Web.TokenAcquisition/Microsoft.Identity.Web.TokenAcquisition.csproj
```diff
@@ -60,6 +64,7 @@
     <AdditionalFiles Include="PublicAPI/$(TargetFramework)/PublicAPI.Unshipped.txt" />
     <AdditionalFiles Include="PublicAPI/$(TargetFramework)/InternalAPI.Shipped.txt" />
     <AdditionalFiles Include="PublicAPI/$(TargetFramework)/InternalAPI.Unshipped.txt" />
+    <AdditionalFiles Include="PublicAPI\net9.0\InternalAPI.Unshipped.txt" />
```
bgavrilMS: ?

tlupes: VS must have auto-added this. Will push a change removing the changes here once the merge validation completes

bgavrilMS (approved):

gladjohn (approved):

gladjohn (approved):

PR #3780, opened 2026-04-15
Fix flaky certificate test failures on CI

gladjohn:
- CertificatesObserverTests.ObserverSendsCorrectEvents: Increase cert store verification timeout from 5s to 15s. On slow CI runners the certificate store propagation can exceed the previous limit.

- CertificateRetryCounterTests.GetAuthenticationResultForAppAsync_ WithBadClientId_CompletesQuickly: Mock the MSAL HTTP layer with InvalidClientMsalHttpClientFactory to avoid real network calls that time out or get blocked by CI firewalls. Increase xUnit timeout from 5s to 30s as a safety net. The mocked test now completes in ~640ms.

### DIFF
--- a/tests/Microsoft.Identity.Web.Test/CertificateRetryCounterTests.cs
+++ b/tests/Microsoft.Identity.Web.Test/CertificateRetryCounterTests.cs
@@ -143,7 +143,7 @@ namespace Microsoft.Identity.Web.Test
         /// Regression test: Simulates the Agent Identities scenario where nested token acquisitions
         /// with bad config should fail quickly, not hang.
         /// </summary>
-        [Fact(Timeout = 5000)] // 5 second timeout - if it takes longer, it's likely hanging
+        [Fact(Timeout = 30000)] // 30 second timeout - generous to accommodate slow CI, but still validates no infinite hang
         public async Task GetAuthenticationResultForAppAsync_WithBadClientId_CompletesQuickly()
         {
             // Arrange
--- a/tests/Microsoft.Identity.Web.Test/CertificatesObserverTests.cs
+++ b/tests/Microsoft.Identity.Web.Test/CertificatesObserverTests.cs
@@ -2,14 +2,10 @@
 // Licensed under the MIT License.
 
 using System;
-using System.Collections.Concurrent;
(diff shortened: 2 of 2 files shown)


tests/Microsoft.Identity.Web.Test/CertificateRetryCounterTests.cs
```diff
@@ -143,7 +146,7 @@ await Task.Run(async () =>
         /// Regression test: Simulates the Agent Identities scenario where nested token acquisitions
         /// with bad config should fail quickly, not hang.
         /// </summary>
-        [Fact(Timeout = 5000)] // 5 second timeout - if it takes longer, it's likely hanging
+        [Fact(Timeout = 30000)] // 30 second timeout - generous to accommodate slow CI, but still validates no infinite hang
```
bgavrilMS, line 146: Won't this be enough? The instance discovery should be very fast.

gladjohn: Fixed two flaky test failures on CI:
 
 
  - CertificatesObserverTests: Replaced async polling loop (arbitrary timeouts) with synchronous verification — X509Store.Add is synchronous, no polling needed. Also backdated cert NotBefore to avoid time-boundary races.
  - CertificateRetryCounterTests: Bumped xUnit timeout from 5s → 30s to accommodate slow CI runners

tests/Microsoft.Identity.Web.Test/CertificateRetryCounterTests.cs
```diff
@@ -143,7 +143,7 @@ await Task.Run(async () =>
...
         public async Task GetAuthenticationResultForAppAsync_WithBadClientId_CompletesQuickly()
         {
             // Arrange
```
bgavrilMS, line 149: Won't this be enough? The instance discovery should be very fast.

gladjohn: You're right. Instance discovery should be fast. But the test is cosnsitently timing out at 5s.

bgavrilMS (approved):

PR #3781, opened 2026-04-16
Add CDT validation for OBO and long-running web API flows

trwalke:
Removed


bgavrilMS:
Let's not add any CDT logic in this SDK pls

PR #3793, opened 2026-04-24
DownstreamApi: ignore caller-supplied reserved or duplicate headers in ExtraHeaderParameters

iNinja:
### Description
Hardens header handling in `DownstreamApi.cs` so that headers supplied via `DownstreamApiOptions.ExtraHeaderParameters` cannot override headers the library is responsible for setting itself.

### Behaviour change

When iterating `ExtraHeaderParameters` to apply caller-supplied headers to the outbound `HttpRequestMessage`:

- Headers whose names match a reserved set (e.g. `Authorization`, `Proxy-Authorization`, `Host`, `Content-Length`, `Transfer-Encoding`, `Connection`, `Upgrade`, `Expect`, plus the `Sec-` and `Proxy-` prefixes) are skipped. Comparison is case-insensitive.
- Headers that duplicate a name the library has already set on the request are skipped.

Skipped entries are logged via existing `ILogger` plumbing (`ReservedHeaderIgnored`, `DuplicateHeaderIgnored`) at `Warning` level so operators can spot misconfigurations. The request is otherwise sent normally — no exception is thrown.

### What's included

- `src/Microsoft.Identity.Web.DownstreamApi/ReservedHeaderNames.cs` — new internal helper (`IsReserved(string)`) with the reserved name list and prefix checks.
- `src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs` — skip + log in the `ExtraHeaderParameters` loop.
- `src/Microsoft.Identity.Web.DownstreamApi/Logging/Logging.cs` — two new log event definitions.
- `tests/Microsoft.Identity.Web.Test/DownstreamWebApiSupport/DownstreamApiTests.cs` — five new tests covering reserved-name skip (case variants), prefix skip, duplicate skip, and the happy path that benign headers still flow through.
- `InternalAPI.Unshipped.txt` updates for the affected packages.

### Notes for reviewers

- No public API change; reserved-name handling is internal and additive.
- Behaviour is on by default — no opt-in flag. Callers relying on overriding `Authorization` (or other reserved names) via `ExtraHeaderParameters` will see those entries silently dropped with a warning log; the recommended path remains the existing `AcquireTokenOptions` / `AuthenticationScheme` plumbing.
- Defaults are conservative; the reserved set deliberately mirrors what `HttpClient` itself treats as restricted.


bgavrilMS (approved):

bgavrilMS (approved):

pmaytak (approved):

pmaytak:
@copilot resolve the merge conflicts in this pull request

pmaytak (approved):

bgavrilMS (approved):

PR #3798, opened 2026-04-29
API File Cleanup

tlupes:
# API File Cleanup

Reduce developer toil by removing internal API files, documenting methods which are used cross-project, and reduce API file duplication.

## Description

This removes the InternalAPI tracking files from nearly all projects and consolidates per-target-framework PublicAPI folder layouts into a smaller, shared structure. There is no functional change to any code files, this is purely to the tracking files (and a few comments).

### Motivation

The Microsoft.CodeAnalysis.PublicApiAnalyzers "internal API" tracking (RS0051–RS0058) was producing significant
maintenance overhead:

 - Every internal member required updates to InternalAPI.Unshipped.txt per target framework.
 - Tracking churn was disproportionate to the value, since the vast majority of Internal APIs are not used between projects.

In addition, the Per-TFM folders for Public API duplicated nearly identical content and required careful editing of each with each new member.

#### .editorconfig

 - Disabled the internal-API analyzer rules (RS0051, RS0052, RS0053, RS0054, RS0055, RS0057, RS0058).
 - Kept public-API analyzer rules.

#### Internal API Files

Removed all Internal API files with the sole exception of Microsoft.Identity.Web.Diagnostics, as that project is entirely internal and intended to be used by other Microsoft.Identity.Web.* projects.

#### Public API Files

- Moved all public API files into PublicAPI folder for consistency.
- Audited all Public API files to find duplication and move them into folders that relate to what they support, not their TFM. This consolidated (nearly) all projects into two buckets: ones that share all members with all TFM, and ones that have different members for NetCore and NetFramework. A single special case, Web, had to have special handling.

Projects that can use a single set of Public API files:

* AgentIdentifies
* Azure
* Certificate
* Certificateless
* Diagnostics (this also applies to the Internal API files)
* GraphServiceClientBeta
* Owin
* OidcFIC
* TokenCache

Projects that need different files for NetCore and NetFramework:

* DownstreamApi
* GraphServiceClient
* MicrosoftGraph
* MicrosoftGraphBeta
* TokenAcquisition

The only special case is Microsoft.Identity.Web, which has some members which are only available for .net8 and .net9. That is handled with a special "NetCore_Pre10" folder. This assembly does not support NetFramework, so it still only has two folders (NetCore and NetCore_Pre10).

#### Internal API Usage Comments

Added comments to each internal member which has legitimate usage in other Microsoft.Identity.Web Non-Test assemblies, with a listing of which assemblies use them and a warning to not modify the members.

The majority of these comments are on Microsoft.Identity.Web.TokenAcquisition.

## Impact

 - No public API changes. Public API tracking files (PublicAPI.Shipped.txt / PublicAPI.Unshipped.txt) remain in 
place, and the RS0016/RS0017 analyzers continue to enforce them.
 - No runtime/behavioral changes. Source edits are comments only.


bgavrilMS (approved):

gladjohn (approved):

src/Microsoft.Identity.Web.TokenAcquisition/AuthCodeRedemptionParameters.cs
```diff
@@ -5,6 +5,11 @@
 
 namespace Microsoft.Identity.Web
 {
+    /*
```
neha-bhargava, line 8: If we have to have these kinds of comments, was it not better to just have the tool check it?

tlupes: There are 53 of these comments and 590 internal members :). There's no way to tell the tool to check only a subset of APIs.

If we track ALL internal apis, then it makes it really hard to understand what actually needs to be not changed and what can be safely changed (a bit of a signal overload where too much noise makes the signal useless). We either need tooling that is a lot more granular (I've written it for other repos, but nothing general purpose enough to reuse here), or just accept that we'll have comments on 10% of the internal members and be careful to not muck with signatures.

bgavrilMS: The problem with the tool is that it controls everything. The internal API surface is massive and completely ignored anyway. comments like this are more useful.

 I don't think internal APIs exposed to "friend" assemblies are a big problem (I mean they are absolutely a bad pattern, friend assemblies are justifiable only for test assemblies, but now it's done). We require ppl to use the same versions of the "friend assemblies" so version mismatches are unlikely. 

If we really start to have problems here, it'd better to make the API public.

PR #3815, opened 2026-05-13
MTLS Without Tokens Support - MicrosoftIdentityMessageHandler Support

tlupes:
# MTLS Without Tokens Support - MicrosoftIdentityMessageHandler Support

Continuation of the PR #3747 , this adds the same level of support to MicrosoftIdentityMessageHandler.

## Description

* Moved shared constants from DownstreamAPI to `Microsoft.Identity.Web.TokenAcquisition/Constants.cs`
* Moved UnauthorizedHttpRequestException to `Microsoft.Identity.Web.TokenAcquisition` (Internal class)
* Add InternalsVisibleTo between `Microsoft.Identity.Web.TokenAcquisition` and `Microsoft.Identity.Web.DownstreamApi` so that the above changes are accessible.
* Minor changes to `MicrosoftIdentityHttpClientBuilderExtensions`:
    * Previously this would always pass a null logger to the `MicrosoftIdentityMessageHandler`. Now, it will pass one if configured.
    * Previously it would always make a new `IMsalMtlsHttpFactory` even if one were configured. Now, it will use a configured one if available.
    * Support to pass a ICredentialsProvider to the handler.
* Support within MicrosoftIdentityMessageHandler for mTLS without tokens


src/Microsoft.Identity.Web.TokenAcquisition/MicrosoftIdentityHttpClientBuilderExtensions.cs
```diff
@@ -421,11 +431,62 @@ public static IHttpClientBuilder AddMicrosoftIdentityMessageHandler(
...
+                return credentialsProvider;
+            }
+
+            var loader = services.GetService<ICredentialsLoader>();
+            if (loader is not null)
+            {
+                return new CredentialsProvider(
+                    GetLogger<CredentialsProvider>(services),
+                    loader,
+                    services.GetService<ICertificatesObserver[]>() ?? [],
```
neha-bhargava: **Likely bug.** `IServiceProvider.GetService<ICertificatesObserver[]>()` only returns a value when an `ICertificatesObserver[]` is *explicitly* registered as that exact type. The standard DI pattern (used in this repo's own tests — `AddSingleton<ICertificatesObserver>(observer1)`) registers individual instances, so this call resolves to `null`, the `?? []` fires, and the inline-constructed `CredentialsProvider` receives an empty observer list — silently dropping cert-rotation telemetry whenever this fallback path is taken.

Should be:
```csharp
services.GetServices<ICertificatesObserver>().ToArray()
```

src/Microsoft.Identity.Web.TokenAcquisition/MicrosoftIdentityHttpClientBuilderExtensions.cs
```diff
@@ -421,11 +431,62 @@ public static IHttpClientBuilder AddMicrosoftIdentityMessageHandler(
...
+            var credentialsProvider = services.GetService<ICredentialsProvider>();
+            if (credentialsProvider is not null)
+            {
+                return credentialsProvider;
+            }
+
+            var loader = services.GetService<ICredentialsLoader>();
+            if (loader is not null)
+            {
+                return new CredentialsProvider(
```
neha-bhargava, line 471: Constructing a fresh `CredentialsProvider` inside the `AddHttpMessageHandler` factory delegate means every handler instance gets its own provider, bypassing any internal caching it may rely on (and creating fresh observer plumbing each time). Prefer registering a default `ICredentialsProvider` once in DI (e.g. `services.TryAddSingleton<ICredentialsProvider>(sp => new CredentialsProvider(...))` from the same module that wires `AddMicrosoftIdentityMessageHandler`) and just resolving it here. That also fixes the observer-array bug below by going through canonical DI resolution.

src/Microsoft.Identity.Web.TokenAcquisition/MicrosoftIdentityMessageHandler.cs
```diff
@@ -435,12 +530,13 @@ private async Task<HttpResponseMessage> SendWithAuthenticationAsync(
...
 
-            // If a binding certificate is present (mTLS PoP), send through the mTLS HTTP client.
+            // If a binding certificate is present (mTLS PoP / mTLS), send through the mTLS HTTP client.
             // This bypasses the normal handler pipeline because the underlying HttpClientHandler
             // must be configured with the client certificate for mutual TLS authentication.
             // We must clone the request because the original HttpRequestMessage has already been
             // marked as "sent" by the outer HttpClient pipeline, and HttpRequestMessage cannot
             // be sent twice.
+            HttpResponseMessage result;
             if (bindingCertificate is not null && _mtlsHttpClientFactory is not null)
```
neha-bhargava: **Likely bug — silent insecure fallback.** If `_credentialsProvider` is set but `_mtlsHttpClientFactory` is `null` (legal via the new public ctor, and reachable when no `IHttpClientFactory`/`IMsalMtlsHttpClientFactory` is registered), an MTLS-protocol request reaches this point with `bindingCertificate != null && _mtlsHttpClientFactory == null` and silently falls through to `base.SendAsync(request, ...)` — i.e. the request is sent through the **non-mTLS** pipeline with **no client certificate** and **no Authorization header** (the latter was deliberately not set in the MTLS branch).

The caller asked for mTLS authentication; we send anonymously. The server will reject, but the failure mode is opaque.

Suggest mirroring the existing guard (line ~478) and throwing `InvalidOperationException` (or `MicrosoftIdentityAuthenticationException`) when `bindingCertificate is not null && _mtlsHttpClientFactory is null` in the MTLS-only path.

src/Microsoft.Identity.Web.TokenAcquisition/MicrosoftIdentityMessageHandler.cs
```diff
@@ -383,17 +441,19 @@ private async Task<HttpResponseMessage> SendWithAuthenticationAsync(
             HttpRequestMessage request,
             MicrosoftIdentityMessageHandlerOptions options,
             IList<string> scopes,
+            bool isRetry,
```
neha-bhargava: **Design.** A `bool isRetry` parameter that gates a recursive call from inside the same method is a classic flag-driven control-flow smell. `SendWithAuthenticationAsync` now mixes four responsibilities: credential acquisition, send dispatch (mTLS client vs base pipeline), credential-usage notification, and retry orchestration. It will be painful to extend the next time a new retry condition appears (e.g. transient network, token refresh).

Consider extracting an artifact record and splitting:
```csharp
private readonly record struct AuthArtifacts(
    string? AuthHeader,
    X509Certificate2? BindingCertificate,
    CredentialSourceLoaderParameters? LoaderParams,
    CredentialDescription? CredentialDescription);

Task<AuthArtifacts> AcquireAsync(options, scopes, ct);
Task<HttpResponseMessage> SendOnceAsync(request, artifacts, ct);
Task<HttpResponseMessage> SendWithCertRetryAsync(request, options, scopes, ct); // owns the bounded loop
```
This localizes the cert-failure retry policy in one place and removes the recursion-by-flag. Each method then has one reason to change.

src/Microsoft.Identity.Web.TokenAcquisition/MicrosoftIdentityMessageHandler.cs
```diff
@@ -453,11 +549,56 @@ private async Task<HttpResponseMessage> SendWithAuthenticationAsync(
...
+                        bindingCertificate,
+                        false,
+                        new UnauthorizedHttpRequestException($"Response has status {result.StatusCode} - {result.ReasonPhrase}"));
+
+                    // Retry again in case notifying that the certificate failed notification caused a rotation which will now succeed.
+                    // The original HttpRequestMessage has been sent and cannot be re-used, so clone it for the retry.
+                    if (!isRetry)
+                    {
+                        using var retryRequest = await CloneHttpRequestMessageAsync(request).ConfigureAwait(false);
+                        return await SendWithAuthenticationAsync(
```
neha-bhargava: **Retry interaction.** This new internal cert-failure retry composes with the WWW-Authenticate claims-challenge retry in `SendAsync` (line ~398). On a persistent 401 in MTLS-only mode, three sends can fire:

1. Initial send → 401 → this internal retry runs, `_credentialsProvider.GetCredentialAsync` re-invoked.
2. Internal retry → 401 → returns.
3. Outer `SendAsync` sees 401, parses claims, retries via `SendWithAuthenticationAsync(..., isRetry: true)`.

The third attempt is **semantically meaningless** for MTLS-only: there's no token to refresh with the challenge claims, so it just re-runs `GetCredentialAsync` for no benefit (and burns another credential-loader call which may hit disk/Key Vault).

Suggest either gating the outer claims-challenge retry on `!string.Equals(options.ProtocolScheme, Constants.MtlsProtocolScheme, ...)`, or having the inner method signal "already retried; outer shouldn't retry again".

src/Microsoft.Identity.Web.TokenAcquisition/MicrosoftIdentityMessageHandler.cs
```diff
@@ -453,11 +549,56 @@ private async Task<HttpResponseMessage> SendWithAuthenticationAsync(
...
+                        loaderParameters,
+                        credentialDescription,
+                        bindingCertificate,
+                        true,
+                        null);
+                }
+                else if (Constants.AuthFailureHttpStatusCodes.Contains(result.StatusCode))
+                {
+                    // Only alert if the failure is potentially due to the certificate.
+                    // This to to avoid needlessly refreshing the certificate on non-certificate related failures.
```
neha-bhargava:
```suggestion
                    // This is to avoid needlessly refreshing the certificate on non-certificate related failures.
```

src/Microsoft.Identity.Web.TokenAcquisition/Constants.cs
```diff
@@ -250,6 +251,23 @@ public static class Constants
...
 
+        /// <summary>
+        /// HTTP status codes which may indicate that a certificate based authnetication failure occurred.
+        /// </summary>
+        /*
+         * Used by Microsoft.Identity.Web.DownstreamApi
+         * Any changes to this member (including removal) can cause runtime failures.
+         * Treat as a public member.
+         */
+        internal static readonly HashSet<HttpStatusCode> AuthFailureHttpStatusCodes =
```
neha-bhargava: Could consider using IReadOnlySet to make sure it is immutable?

```csharp
internal static readonly IReadOnlySet<HttpStatusCode> AuthFailureHttpStatusCodes =
    new HashSet<HttpStatusCode>
    {
        HttpStatusCode.BadRequest,
        HttpStatusCode.Unauthorized,
        HttpStatusCode.Forbidden,
        (HttpStatusCode)495,
        (HttpStatusCode)496,
    };
```

src/Microsoft.Identity.Web.TokenAcquisition/Constants.cs
```diff
@@ -250,6 +251,23 @@ public static class Constants
             CertificateWasRevoked,              // AADSTS7000277 - Certificate was revoked
         };
 
+        /// <summary>
+        /// HTTP status codes which may indicate that a certificate based authnetication failure occurred.
```
neha-bhargava:
```suggestion
        /// HTTP status codes which may indicate that a certificate based authentication failure occurred.
```

src/Microsoft.Identity.Web.TokenAcquisition/GlobalSuppressions.cs
```diff
@@ -19,3 +19,4 @@
...
 [assembly: SuppressMessage("ApiDesign", "RS0026:Do not add multiple public overloads with optional parameters", Justification = "New overload adds mTLS PoP support alongside existing shipped constructor", Scope = "member", Target = "~M:Microsoft.Identity.Web.MicrosoftIdentityMessageHandler.#ctor(Microsoft.Identity.Abstractions.IAuthorizationHeaderProvider,Microsoft.Identity.Web.MicrosoftIdentityMessageHandlerOptions,Microsoft.Identity.Client.IMsalMtlsHttpClientFactory,Microsoft.Extensions.Logging.ILogger{Microsoft.Identity.Web.MicrosoftIdentityMessageHandler})")]
+[assembly: SuppressMessage("ApiDesign", "RS0026:Do not add multiple public overloads with optional parameters", Justification = "<Pending>", Scope = "member", Target = "~M:Microsoft.Identity.Web.MicrosoftIdentityMessageHandler.#ctor(Microsoft.Identity.Abstractions.IAuthorizationHeaderProvider,Microsoft.Identity.Web.MicrosoftIdentityMessageHandlerOptions,Microsoft.Identity.Client.IMsalMtlsHttpClientFactory,Microsoft.Identity.Web.ICredentialsProvider,Microsoft.Extensions.Logging.ILogger{Microsoft.Identity.Web.MicrosoftIdentityMessageHandler})")]
```
neha-bhargava: `Justification = "<Pending>"` should be filled in before merge — every other suppression in this file has a real one. Suggested wording:

> `"New overload adds mTLS-without-tokens support (ICredentialsProvider parameter) alongside existing shipped constructors."`

src/Microsoft.Identity.Web.Diagnostics/UnauthorizedHttpRequestException.cs
```diff
@@ -0,0 +1,32 @@
...
+{
+    /// <summary>
+    /// Exception for a failed HTTP call. This is exclusively used by reporting and never thrown.
+    /// </summary>
+    /*
+     * Used by Microsoft.Identity.Web.DownstreamApi
+     * Any changes to this member (including removal) can cause runtime failures.
+     * Treat as a public member.
+     */
+    internal class UnauthorizedHttpRequestException : Exception
```
neha-bhargava, line 11: Sharing this type across assemblies via `InternalsVisibleTo` is the worst of both worlds: technically `internal` (so no Public API tracking, no semver protection) but in practice a cross-assembly contract that `Microsoft.Identity.Web.DownstreamApi` depends on. The comment above explicitly says *"Treat as a public member"*.

Suggest making the type `public` and adding it to `PublicAPI.Unshipped.txt` here. That removes the IVT coupling for this symbol, surfaces future changes through the Public API analyzer, and matches the de-facto reality.

src/Microsoft.Identity.Web.DownstreamApi/Microsoft.Identity.Web.DownstreamApi.csproj
```diff
@@ -10,10 +10,6 @@
     <EnableConfigurationBindingGenerator>true</EnableConfigurationBindingGenerator>
   <IsAotCompatible>true</IsAotCompatible>
 </PropertyGroup>
-  <!-- Include AOT attribute polyfills for older TFMs -->
-  <ItemGroup Condition="'$(TargetFramework)' == 'net462' or '$(TargetFramework)' == 'net472' or '$(TargetFramework)' == 'netstandard2.0'">
-    <Compile Include="..\Shared\CodeAnalysisAttributes.cs" Link="Polyfills\CodeAnalysisAttributes.cs" />
-  </ItemGroup>
   <ItemGroup>
```
neha-bhargava, line 13: Why is this removed?

tests/Microsoft.Identity.Web.Test/CertificatesObserverTests.cs
```diff
@@ -240,18 +251,31 @@ static void RemoveCertificate(X509Certificate2? certificate)
                 mockHttpFactory.ConfigureSuccessfulTokenResponse(authority);
                 mockHttpFactory.ValidCertificates.Add(cert1);
 
-                IDownstreamApi downstreamApi = provider.GetRequiredService<IDownstreamApi>();
+                string apiUrl = authority + relativeUrl;
 
-                DownstreamApiOptions options = new DownstreamApiOptions()
+                Func<Task<HttpResponseMessage>> Invoke = async () =>
```
neha-bhargava: Nit: 
```suggestion
                Func<Task<HttpResponseMessage>> invoke = async () =>
```

neha-bhargava (commented):
Left some comments

src/Microsoft.Identity.Web.Diagnostics/UnauthorizedHttpRequestException.cs
tlupes: I'm really hesitant to do this. This exception is never thrown; it is used only for telemetry; there is no reason for it to be part of the public contract.

Perhaps a good middle ground is to add it to Microsoft.Identity.Web.Diagnostics, which in purely internal and tracks internal members?

src/Microsoft.Identity.Web.DownstreamApi/Microsoft.Identity.Web.DownstreamApi.csproj
tlupes: With the added InternalsVisibleTo it's no longer needed (and in fact results in errors).

src/Microsoft.Identity.Web.TokenAcquisition/MicrosoftIdentityHttpClientBuilderExtensions.cs
tlupes: This is already in place, this code exists for customers who are manually constructing the object

src/Microsoft.Identity.Web.TokenAcquisition/Constants.cs
```diff
@@ -240,7 +241,7 @@ public static class Constants
         /// <summary>
         /// Error codes indicating certificate or signed assertion issues that warrant retry with a new certificate.
         /// </summary>
-        internal static readonly HashSet<string> s_certificateRelatedErrorCodes = new (StringComparer.OrdinalIgnoreCase)
+        internal static readonly IReadOnlyCollection<string> s_certificateRelatedErrorCodes = new HashSet<string>(StringComparer.OrdinalIgnoreCase)
```
bgavrilMS: naming inconsistent between this and AuthFailureHttpStatusCodes . Should this be a property + comment that this is "internal" for test purposes only?

src/Microsoft.Identity.Web.TokenAcquisition/MicrosoftIdentityHttpClientBuilderExtensions.cs
```diff
@@ -421,11 +431,62 @@ public static IHttpClientBuilder AddMicrosoftIdentityMessageHandler(
...
+            var msalMtlsHttpClientFactory = services.GetService<IMsalMtlsHttpClientFactory>();
+            if (msalMtlsHttpClientFactory is not null)
+            {
+                return msalMtlsHttpClientFactory;
+            }
+
+            var httpClientFactory = services.GetService<IHttpClientFactory>();
+            if (httpClientFactory is not null)
+            {
+                return new MsalMtlsHttpClientFactory(httpClientFactory);
```
bgavrilMS, line 454: I don't know how this helps. All mTLS traffic will bypass the factory. I guess it's fine, I don't have an alternative for this. Maybe log a warning?

src/Microsoft.Identity.Web.TokenAcquisition/MicrosoftIdentityMessageHandler.cs
```diff
@@ -408,25 +468,60 @@ private async Task<HttpResponseMessage> SendWithAuthenticationAsync(
...
                     authHeader = boundResult.Result?.AuthorizationHeaderValue!;
                     bindingCertificate = boundResult.Result?.BindingCertificate;
+                    loaderParameters = null;
+                    credentialDescription = null;
+                }
+                else if (string.Equals(options.ProtocolScheme, Constants.MtlsProtocolScheme, StringComparison.OrdinalIgnoreCase))
+                {
+                    if (_credentialsProvider is null)
+                    {
+                        throw new InvalidOperationException("mTLS authentication requires a Credentials Provider object to be registered, but no such service was found.");
```
bgavrilMS: Consider adding an aka.ms/idweb/mtls page where you describe the sceanrios in more detail. As an app developer, I wouldn't know how to action this exception.

src/Microsoft.Identity.Web.TokenAcquisition/MicrosoftIdentityMessageHandler.cs
```diff
@@ -408,25 +468,60 @@ private async Task<HttpResponseMessage> SendWithAuthenticationAsync(
...
+                    }
+
+                    loaderParameters = new CredentialSourceLoaderParameters(string.Empty, string.Empty)
```
bgavrilMS: nit: add a comment that there is no Entra app backing this up, so no client ID / authoirty is expected

src/Microsoft.Identity.Web.TokenAcquisition/MicrosoftIdentityMessageHandler.cs
```diff
@@ -435,14 +530,20 @@ private async Task<HttpResponseMessage> SendWithAuthenticationAsync(
...
             // We must clone the request because the original HttpRequestMessage has already been
             // marked as "sent" by the outer HttpClient pipeline, and HttpRequestMessage cannot
             // be sent twice.
-            if (bindingCertificate is not null && _mtlsHttpClientFactory is not null)
+            HttpResponseMessage result;
+            if (bindingCertificate is not null)
             {
+                if (_mtlsHttpClientFactory is null)
+                {
+                    throw new InvalidOperationException("Authentication using mTLS requires a MtlsHttpClientFactory object to be registered, but no such service was found.");
```
bgavrilMS: A help page would be good. 

BUt at the same time, can the app developer get into this state? I thought we provide an HTTP pipeline in all cases.

bgavrilMS (approved):

src/Microsoft.Identity.Web.TokenAcquisition/MicrosoftIdentityHttpClientBuilderExtensions.cs
tlupes: It's trying to handle scenarios where not all services are available, but some are. In this, if `IMsalMtlsHttpClientFactory` is available, then it will use that. Failing that, then it will check if `IHttpClientFactory` is available and use that. Failing that, it returns null.

src/Microsoft.Identity.Web.TokenAcquisition/MicrosoftIdentityMessageHandler.cs
tlupes: This turned into a more major refactor, but it is a lot easier to read

src/Microsoft.Identity.Web.TokenAcquisition/MicrosoftIdentityMessageHandler.cs
tlupes: Added a link, I will make the link work in the near future. Once this is in I'll write up a help page for it.

src/Microsoft.Identity.Web.TokenAcquisition/MicrosoftIdentityMessageHandler.cs
tlupes: HTTP pipeline will not be present if the consumer did not register an http factory (e.g. `services.AddHttp();`)

src/Microsoft.Identity.Web.Diagnostics/UnauthorizedHttpRequestException.cs
```diff
@@ -0,0 +1,27 @@
...
+// Licensed under the MIT License.
+
+using System;
+
+namespace Microsoft.Identity.Web.Diagnostics
+{
+    /// <summary>
+    /// Exception for a failed HTTP call. This is exclusively used by reporting and never thrown.
+    /// </summary>
+    internal class UnauthorizedHttpRequestException : Exception
```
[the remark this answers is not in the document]
tlupes: Transitive dependency

src/Microsoft.Identity.Web.TokenAcquisition/MicrosoftIdentityMessageHandler.cs
```diff
@@ -12,6 +12,7 @@
 using Microsoft.Extensions.Logging;
 using Microsoft.Identity.Abstractions;
 using Microsoft.Identity.Client;
+using Microsoft.Identity.Web.Diagnostics;
```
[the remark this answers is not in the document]
tlupes: Transitive dependency

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
```diff
@@ -21,6 +21,7 @@
 using Microsoft.Extensions.Options;
 using Microsoft.Identity.Abstractions;
 using Microsoft.Identity.Client;
+using Microsoft.Identity.Web.Diagnostics;
```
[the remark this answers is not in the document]
tlupes: Transitive dependency

src/Microsoft.Identity.Web.TokenAcquisition/MicrosoftIdentityHttpClientBuilderExtensions.cs
neha-bhargava: I am curious how the flow works if the msalMtlsHttpClientFactory is null. If you change create one from httpClientFactory then it will not use the certificate when creating the httpClient. Is that an expected flow?

tests/Microsoft.Identity.Web.Test/MicrosoftIdentityMessageHandlerTokenBindingTests.cs
```diff
@@ -360,7 +360,7 @@ public async Task SendAsync_WithMtlsPop_NullBindingCertificate_UsesBaseSendAsync
         }
 
         [Fact]
-        public async Task SendAsync_WithMtlsPop_NullMtlsFactory_UsesBaseSendAsync()
+        public async Task SendAsync_WithMtlsPop_NullMtlsFactory_Throws()
```
neha-bhargava: Maybe add a test with IHttpClientFactory mock if that is a valid flow?

neha-bhargava (approved):

src/Microsoft.Identity.Web.TokenAcquisition/MicrosoftIdentityHttpClientBuilderExtensions.cs
tlupes: Now, if it is null, then a "mtls only" version will be created which actually has all the functionality for the message handler.

src/Microsoft.Identity.Web.TokenAcquisition/MsalMtlsHttpClientFactory.cs
```diff
@@ -127,6 +132,19 @@ public void Dispose()
...
 
+        /// <summary>
+        /// Creates a new instance of <see cref="MsalMtlsHttpClientFactory"/> configured for mTLS-only operations.
+        /// </summary>
+        /// <remarks>
+        /// Intended for use where the HttpClientFactory is not available or not desired, and all HTTP operations require mTLS authentication.
+        /// This factory will manage a pool of mTLS-configured HttpClient instances based on certificate thumbprints.
+        /// </remarks>
+        /// <returns>A new instance of <see cref="MsalMtlsHttpClientFactory"/> for mTLS-only operations.</returns>
+        internal static MsalMtlsHttpClientFactory CreateMtlsOnly()
```
neha-bhargava, line 143: Maybe a better and maintainable approach is to use a constructor will null?

neha-bhargava (approved):

src/Microsoft.Identity.Web.TokenAcquisition/MsalMtlsHttpClientFactory.cs
neha-bhargava: Minor non blocking comment

bgavrilMS (approved):

bgavrilMS (approved):
It would be good to add some docs around the sceanrio and usage.

src/Microsoft.Identity.Web.TokenAcquisition/MsalMtlsHttpClientFactory.cs
tlupes: I wanted to avoid that as using the null is potentially dangerous (runtime failures). This is why I made the internal static method to construct one with null so it's really explicit when you are using the null form. I can't make an overload that's nullable (C# does not allow that) so we end up with this wonkiness.

tlupes:
> It would be good to add some docs around the sceanrio and usage.

Will do, I will write up docs for this once it is merged.

neha-bhargava (approved):

PR #3825, opened 2026-05-21
Add sanitazing of redirect uri in authorization scenarios

cpp11nullptr:


src/Microsoft.Identity.Web/AuthorizeForScopesAttribute.cs
```diff
@@ -134,7 +134,9 @@ public override void OnException(ExceptionContext context)
                             : httpRequest.Query[Constants.XReturnUrl];
 
                         UrlHelper urlHelper = new UrlHelper(context);
-                        if (urlHelper.IsLocalUrl(redirectUri))
+                        if (urlHelper.IsLocalUrl(redirectUri)
```
iNinja, line 137: Could we move this check into a shared helper rather than inlining it here? `LoginLogoutEndpointRouteBuilderExtensions.IsLocalUrl` already has the same check inside the method. Duplicating the logic means both need to be updated together if we add more encodings. A shared helper would keep it in one place.

tests/Microsoft.Identity.Web.Test/Blazor/LoginLogoutEndpointRouteBuilderExtensionsTests.cs
```diff
@@ -43,6 +43,13 @@ public class LoginLogoutEndpointRouteBuilderExtensionsTests
         // Bare hostnames / non-slash-prefixed — blocked.
         [InlineData("evil.example", "/")]
         [InlineData("home", "/")]
+        // Percent-encoded slash/backslash — blocked (reverse proxies may decode these).
```
iNinja, line 46: The `AuthorizeForScopesAttribute` path has the same percent-encoded check but isn't covered by tests. It would be worth having them there as well.

src/Microsoft.Identity.Web/AuthorizeForScopesAttribute.cs
cpp11nullptr: Done.

tests/Microsoft.Identity.Web.Test/Blazor/LoginLogoutEndpointRouteBuilderExtensionsTests.cs
cpp11nullptr: Done.

iNinja (approved):

pmaytak (approved):

XiaoxinMS2 (approved):

4gust (approved):

Issue #3840, opened 2026-06-03
Agentic (User FIC) token flow bypasses MSAL cache when ClaimsPrincipal is null


rido-min:
## Summary

When using the agentic token flow (User FIC via `AcquireTokenByUsernamePassword`) from a bot/service context where there is no authenticated HTTP user, `CreateAuthorizationHeaderAsync()` is called with a `null` `ClaimsPrincipal`. This causes MSAL to skip the silent token flow and make a network round-trip to Entra on **every single call**, even when a valid cached token exists.

## Reproduction

1. Register an Agent Identity (FIC) app
2. Call `IAuthorizationHeaderProvider.CreateAuthorizationHeaderAsync()` with `WithAgentUserIdentity()` options and `claimsPrincipal: null`
3. Observe MSAL logs: every call shows `source: IdentityProvider` (network), never `source: Cache`
4. Call the same method again within token lifetime — still hits the network

## Root Cause

In `TokenAcquisition.TryGetAuthenticationResultForConfidentialClientUsingRopcAsync()`, the silent flow guard at [line ~440](https://github.com/AzureAD/microsoft-identity-web/blob/master/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs) requires:

```csharp
if (!forceRefresh && user != null && user.GetMsalAccountId() != null)
{
    // AcquireTokenSilent — cache hit path
}
```

When `user` is null (common in bot/service scenarios where the agent identity comes from request parameters, not HTTP context), this check always fails. The code falls through to `AcquireTokenByUsernamePassword`, which goes to the network.

After the ROPC call succeeds, the account ID is written back to the ClaimsPrincipal (line ~520):

```csharp
if (user != null && user.GetMsalAccountId() == null)
{
    user.AddIdentity(...);
}
```

But since `user` is null, this is also skipped. Even if the caller passes a non-null empty ClaimsPrincipal (our current workaround), the account information is only persisted on that specific object instance. If the caller creates a new ClaimsPrincipal per request, caching still doesn't work.

## Impact

In our Teams bot SDK, a single incoming message triggers 2-4 outbound API calls, each needing an agentic token. This means **2-4 unnecessary HTTP round-trips to login.microsoftonline.com per message** (~250ms each), adding 500ms-1s of latency. At scale this also increases Entra throttling risk.

## Current Workaround

[...]

PR #3842, opened 2026-06-04
Use MSAL's  recent UserFIC API for agentic flows

Avery-Dunn:
Replaces the internal ROPC piggybacking mechanism for agentic User FIC token acquisition with MSAL .NET's native `AcquireTokenByUserFederatedIdentityCredential` API. This resolves a customer-reported caching bug (#3840), modernizes the agentic flow to use purpose-built MSAL APIs, and simplifies the implementation.

## Background

ID Web's agentic User FIC flow previously hijacked the ROPC (`AcquireTokenByUsernamePassword`) path via an internal add-in (`AgentUserIdentityMsalAddIn`) that rewrote HTTP request bodies at the last moment. This had several drawbacks:

- **No cache support:** ROPC's silent-flow guard requires a non-null `ClaimsPrincipal` with oid/tid claims. In agentic scenarios, `ClaimsPrincipal` is typically null, so the cache was always bypassed — causing 2–4 unnecessary network calls per request ([#3840](https://github.com/AzureAD/microsoft-identity-web/issues/3840)).
- **Fragile HTTP body manipulation:** Opaque, hard to debug, bypasses MSAL's validation and telemetry.
- **No type safety:** String manipulation of grant types rather than MSAL's typed API surface.

MSAL .NET's `AcquireTokenByUserFederatedIdentityCredential` is a first-class API for this scenario with built-in cache support. Version 4.84.2 added the `Guid userObjectId` overload for OID-based flows.

### Multi-CCA Pattern

| Component | Role |
|---|---|
| **Blueprint CCA** | ID Web's existing CCA. Handles **Leg 1**: acquires FMI token (T1) via `AcquireTokenForClient` + `WithFmiPath(agentAppId)`. |
| **Agent CCA** | Per-agent CCA whose client assertion callback chains to the blueprint for Leg 1. Handles **Leg 2** (`AcquireTokenForClient` → T2) and **Leg 3** (`AcquireTokenByUserFederatedIdentityCredential` → user token). |

Both use MSAL's shared static cache (`EnableSharedCacheOptions`), providing natural cache key isolation via distinct `ClientId` values while ensuring tokens survive CCA re-creation.

### Three-Leg Flow

```
1. Caller → CreateAuthorizationHeaderForUserAsync(scopes, options, claimsPrincipal: null)
2. TryGetAuthenticationResultForAgentUserFicAsync detects agentic flow (UPN or OID)
3. Silent lookup via stored account identifier → cache HIT? Return cached token.
4. Cache MISS:
   Leg 1: Blueprint CCA → AcquireTokenForClient + WithFmiPath → T1 (FMI token)
   Leg 2: Agent CCA → AcquireTokenForClient (T1 as client assertion) → T2
   Leg 3: Agent CCA → AcquireTokenByUserFederatedIdentityCredential(scopes, UPN/OID, T2) → user token
5. Store account identifier for future silent lookups
6. Return user token
```

On subsequent calls for the same (agent, user, tenant) tuple, step 3 returns the cached token with zero network calls.

### Account Identifier Storage

A `ConcurrentDictionary<string, string>` maps `"{agentAppId}:{USER_IDENTIFIER}:{TENANTID}"` → MSAL account identifier. This replaces the role that `ClaimsPrincipal` oid/tid claims serve in other ID Web flows. Entries are cleaned up when the CCA dictionary is cleared at the size threshold.

### Agent CCA Eviction

As DOS protection, the agent CCA dictionary is cleared entirely when it exceeds a configurable threshold (default 10,000). Since all agent CCAs use MSAL's shared static cache, clearing the dictionary only discards lightweight CCA objects — tokens remain accessible to newly-built CCAs via `AcquireTokenSilent`.

### Tenant Propagation

The assertion callback extracts the tenant from `AssertionRequestOptions.TokenEndpoint` (when the host matches the configured instance) and applies `WithTenantId` to Leg 1. This ensures multi-tenant scenarios work correctly, matching the pattern used by `OidcIdpSignedAssertionProvider`.

### `Directory.Build.props`
- Bump MSAL .NET from 4.84.1 → 4.84.2 (adds `Guid userObjectId` overload)

### `TokenAcquisition.cs`
- **`TryGetAuthenticationResultForAgentUserFicAsync` (new):** Detects UPN/OID agentic flows, performs silent retrieval or the 3-leg flow via native MSAL APIs
- **`GetOrBuildAgentUserFicCcaAsync` (new):** Builds and caches agent CCAs with assertion callbacks that chain to the blueprint CCA; applies shared cache and size-threshold eviction
- **`ExtractTenantFromTokenEndpointIfSameInstance` (new):** Extracts tenant from token endpoint URL when host matches configured instance
- **Early return in `TryGetAuthenticationResultForConfidentialClientUsingRopcAsync`:** Intercepts agentic flows before the ROPC path
- **Silent catch broadened:** `MsalException` instead of `MsalUiRequiredException`
- **Dead code removal:** Agent identity extraction blocks from the ROPC method

### `TokenAcquisition.Logger.cs`
- **6 structured log messages** (`LoggerMessage.Define`):
  - `AgentUserFicFlowDetected`, `AgentUserFicSilentSuccess`, `AgentUserFicSilentFailure`
  - `AgentUserFicAcquisitionComplete`, `AgentCcaCreated`, `AgentCcaEviction`

### `LoggingEventId.cs`
- 6 new event IDs (600–605)

### `AgentIdentitiesExtension.cs`
- Removed add-in callback registration from `AddAgentIdentities()` (`AddOidcFic()` preserved)

### `AgentUserIdentityMsalAddIn.cs`
- **Deleted.** Fully superseded by native MSAL API usage.

### `TokenAcquisitionTests.cs`
- **4 core agent User FIC tests:** UPN caching, OID caching, ClaimsPrincipal independence, UPN/OID isolation
- **3 shared cache isolation tests:** Multi-agent/multi-user correctness, per-instance eviction behavior, shared cache token survival across CCA re-creation
- **1 size-threshold eviction test:** Verifies dictionary clear at threshold
- **4 `ExtractTenantFromTokenEndpointIfSameInstance` tests:** Same/different instance, null inputs, invalid URI

## No Breaking Changes

All public APIs are unchanged:
- `WithAgentUserIdentity(options, agentAppId, username)` — now uses native UPN path internally
- `WithAgentUserIdentity(options, agentAppId, userId)` — now uses native OID path internally
- `AddAgentIdentities()` — still registers OidcFic; no longer registers the (internal) add-in callback

The deleted `AgentUserIdentityMsalAddIn` was `internal static` with no external consumers.

## Known Limitations

- **National cloud FIC scope:** Hardcoded to `api://AzureADTokenExchange/.default` (public cloud). National cloud support requires cloud-aware inference in MSAL itself — to be addressed in a follow-up.

## Resolves

- #3840 — Agentic User FIC cache bypass when `ClaimsPrincipal` is null


src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -59,6 +59,28 @@ class OAuthConstants
...
+        /// which can only be obtained from GetAccountAsync(identifier) using an identifier that
+        /// comes back from a prior token acquisition. In all other ID Web flows, this identifier
+        /// is stored in the ClaimsPrincipal (via GetMsalAccountId / oid+tid claims). In the
+        /// agentic scenario, however, ClaimsPrincipal is typically null or freshly created per
+        /// request (bot/service pattern), so there is no persistent object to write back to.
+        /// This dictionary fills that role, keyed by "{agentAppId}:{USERNAME}:{TENANTID}".
+        /// Entries are cleaned up when MSAL evicts the corresponding account from its cache.
+        /// </summary>
+        private readonly ConcurrentDictionary<string, string> _agentUserFicAccountIds = new();
+        private static readonly string[] s_ficScopes = new[] { "api://AzureADTokenExchange/.default" };
```
neha-bhargava, line 93: These are always the public token exchange?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -59,6 +59,28 @@ class OAuthConstants
         private readonly ConcurrentDictionary<string, IConfidentialClientApplication?> _applicationsByAuthorityClientId = new();
         private readonly ConcurrentDictionary<string, SemaphoreSlim> _appSemaphores = new();
 
+        /// <summary>
+        /// Caches agent CCAs for the native User FIC flow. Each agent CCA uses an assertion
+        /// callback that chains to the blueprint CCA for Leg 1 (FMI token acquisition).
+        /// Key format: "{agentAppId}:{authenticationScheme}".
```
neha-bhargava: What is the reason to have the authenticationScheme included in the key for CCA?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -526,6 +558,194 @@ private async Task<AuthenticationResult> GetAuthenticationResultForUserInternalA
...
+                            .WithFmiPath(agentAppId)
+                            .WithSendX5C(blueprintOptions.SendX5C)
+                            .ExecuteAsync(options.CancellationToken)
+                            .ConfigureAwait(false);
+
+                        return leg1.AccessToken;
+                    })
+                    .WithAuthority(authority)
+                    .WithHttpClientFactory(_httpClientFactory)
+                    .WithInstanceDiscovery(false) // Blueprint already validated the authority.
```
neha-bhargava: If it is already validated then this is cached. Should not explicitly disable the instance discovery

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -59,6 +59,28 @@ class OAuthConstants
         private readonly ConcurrentDictionary<string, IConfidentialClientApplication?> _applicationsByAuthorityClientId = new();
         private readonly ConcurrentDictionary<string, SemaphoreSlim> _appSemaphores = new();
 
+        /// <summary>
+        /// Caches agent CCAs for the native User FIC flow. Each agent CCA uses an assertion
+        /// callback that chains to the blueprint CCA for Leg 1 (FMI token acquisition).
+        /// Key format: "{agentAppId}:{authenticationScheme}".
+        /// </summary>
+        private readonly ConcurrentDictionary<string, IConfidentialClientApplication> _agentUserFicCcas = new();
```
neha-bhargava: How many agentic CCA apps can exists do you know? Should there be an upper limit to this?

Avery-Dunn: The basic structure is that 1 "blueprint" Entra apps can some number of "agent" Entra apps, and each agent app has some number of users it's acting on behalf of, but I'm not sure what a realistic number of agent apps would actually be.

There are a few places in ID Web with similar unbound list of CCA instances:
-https://github.com/AzureAD/microsoft-identity-web/blob/94b18604fe3c2b5c58de8e4ac8b31a7ace750df2/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs#L59
-https://github.com/AzureAD/microsoft-identity-web/blob/94b18604fe3c2b5c58de8e4ac8b31a7ace750df2/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.ManagedIdentity.cs#L27

So on the one hand an unbounded dictionary matches the existing style, but on the other hand there are a lot more reasons to create "agent" apps than normal CCA apps so this list will likely be much larger than the others.

However, we haven't really discussed this issue, and any solution we come up with would also be helpful guidance for our customers, so that might be out of scope for this PR.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
Avery-Dunn: Removed `WithInstanceDiscovery(false)` in the latest commit.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
Avery-Dunn: Expanded comment explaining why `authenticationScheme` is part of the key: the authenticationScheme is included because different schemes may resolve to different blueprint credentials (e.g. certificates), and the agent CCA's assertion callback captures the scheme to chain back to the correct blueprint CCA.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
Avery-Dunn: The existing behavior in ID Web for these agent scenarios hardcoded the public cloud endpoint, so my PR originally tried to match that.

However, in the latest commit I refactored it to be customizable: now the public cloud endpoint is only the default, and can be manually overridden by setting `ExtraParameters[Constants.TokenExchangeUrlKey]` (this is how national cloud versions of AzureADTokenExchange are handled in other parts of ID Web today, such as in `OidcIdpSignedAssertionProvider` and `GetFicTokenAsync`)

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -59,6 +59,43 @@ class OAuthConstants
...
 
+        /// <summary>
+        /// Caches agent CCAs for the native User FIC flow. Each agent CCA uses an assertion
+        /// callback that chains to the blueprint CCA for Leg 1 (FMI token acquisition).
+        /// Key format: "{agentAppId}:{authenticationScheme}".
+        /// The authenticationScheme is included because different schemes may resolve to
+        /// different blueprint credentials (e.g. certificates), and the agent CCA's assertion
+        /// callback captures the scheme to chain to the correct blueprint CCA.
+        /// </summary>
+        private readonly ConcurrentDictionary<string, IConfidentialClientApplication> _agentUserFicCcas = new();
```
bgavrilMS: Can we try to reuse the existing CCAs? I am concerned because there is a lot of logic / hooks around the creation of these CCA objects, and we'd need to replicate them? For example, translating from the Identity.Web object model for configuration to the MSAL one (force refresh, auth scheme etc. etc.); It would be good to reuse as much as possible.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
bgavrilMS: We need to fix this holistically. I assigned you some other bugs related to this.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
Avery-Dunn: To clarify, what do you mean by "reuse the existing CCAs"?

We cannot re-use these CCA instances for different agents, as shown in this POC there is a fundamental incompatibility with MSAL's existing cache key design and the "1+N" client IDs that the agent ID flow requires: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/6008

That POC also shows that a relatively small code change could reduce it down to a single CCA instance, however in our previous discussions about this you said we should not need to go down that route and we should instead focus on providing guidance on how to use existing FIC APIs.

Now that we've started seeing real-world usage like this, should we start exploring our options to resolve this again?

bgavrilMS: I see. I'm ok with some extensibility API in MSAL.NET to address this. This code duplication will cause problem further on.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -526,6 +555,256 @@ private async Task<AuthenticationResult> GetAuthenticationResultForUserInternalA
...
+        private async Task<IConfidentialClientApplication> GetOrBuildAgentUserFicCcaAsync(
+            string agentAppId,
+            string authority,
+            string? authenticationScheme,
+            string[] ficScopes)
+        {
+            // Include authenticationScheme in the CCA cache key so different schemes
+            // (pointing to different blueprint credentials) get separate CCAs.
+            string ccaCacheKey = $"{agentAppId}:{authenticationScheme ?? string.Empty}";
+
```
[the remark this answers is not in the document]
Avery-Dunn: Within a single TokenAcquisition instance, the authority is derived from the one configured MergedOptions and does not vary per call: multi-authority scenarios require separate TokenAcquisition instances. Added a comment to the code explaining this.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -526,6 +555,256 @@ private async Task<AuthenticationResult> GetAuthenticationResultForUserInternalA
...
+            var agentCca = await GetOrBuildAgentUserFicCcaAsync(
+                agentAppId!, application.Authority, authScheme, ficScopes).ConfigureAwait(false);
+
+            bool forceRefresh = tokenAcquisitionOptions?.ForceRefresh ?? false;
+
+            // Try silent retrieval first using a stored account identifier from a prior call.
+            // Include tenantId in the key so cross-tenant calls don't collide.
+            string normalizedTenant = tenantId?.ToUpperInvariant() ?? string.Empty;
+            string accountLookupKey = $"{agentAppId}:{userIdentifierForCacheKey}:{normalizedTenant}";
+            if (!forceRefresh
```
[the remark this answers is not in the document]
Avery-Dunn: A given (agent, user, tenant) tuple maps to a single MSAL account identity regardless of which auth scheme was used. The CCA selected for the call is already scheme-specific, and GetAccountAsync returns the same account from any CCA that shares the user's cache partition. Added a comment explaining the rationale.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -526,6 +555,256 @@ private async Task<AuthenticationResult> GetAuthenticationResultForUserInternalA
...
+                        MergedOptions blueprintOptions = _tokenAcquisitionHost.GetOptions(capturedAuthScheme, out _);
+                        var blueprintCca = await GetOrBuildConfidentialClientApplicationAsync(
+                            blueprintOptions, isTokenBinding: false).ConfigureAwait(false);
+
+                        var leg1 = await blueprintCca
+                            .AcquireTokenForClient(capturedFicScopes)
+                            .WithFmiPath(agentAppId)
+                            .WithSendX5C(blueprintOptions.SendX5C)
+                            .ExecuteAsync(options.CancellationToken)
+                            .ConfigureAwait(false);
```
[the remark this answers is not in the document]
Avery-Dunn: The assertion callback now extracts the tenant from AssertionRequestOptions.TokenEndpoint (when the host matches the blueprint's configured instance) and applies WithTenantId to Leg 1. This matches the pattern used by OidcIdpSignedAssertionProvider.ExtractTenantFromTokenEndpointIfSameInstance. Added unit tests for the tenant extraction logic.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -526,6 +598,331 @@ private async Task<AuthenticationResult> GetAuthenticationResultForUserInternalA
...
```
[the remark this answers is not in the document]
Avery-Dunn: The assertion callback now extracts the tenant from AssertionRequestOptions.TokenEndpoint (when the host matches the blueprint's configured instance) and applies WithTenantId to Leg 1. This matches the pattern used by OidcIdpSignedAssertionProvider.ExtractTenantFromTokenEndpointIfSameInstance. Added unit tests for the tenant extraction logic.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -526,6 +598,331 @@ private async Task<AuthenticationResult> GetAuthenticationResultForUserInternalA
...
+            if (_agentCcaSweepTimer is not null)
+            {
+                return;
+            }
+
+            var newTimer = new Timer(
+                _ => SweepExpiredAgentCcas(),
+                null,
+                Timeout.InfiniteTimeSpan, // Don't start immediately
+                Timeout.InfiniteTimeSpan);
```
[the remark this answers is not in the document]
Avery-Dunn: TokenAcquisition is registered as a singleton with process lifetime, so the timer lives for the duration of the host and is cleaned up on process exit. Adding IDisposable to TokenAcquisition would be a larger change (it's not disposable today). Added an inline comment documenting this design choice.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -526,6 +598,331 @@ private async Task<AuthenticationResult> GetAuthenticationResultForUserInternalA
...
+                        {
+                            silentBuilder.WithTenantId(tenantId);
+                        }
+
+                        return await silentBuilder.ExecuteAsync().ConfigureAwait(false);
+                    }
+                    catch (MsalUiRequiredException ex)
+                    {
+                        Logger.TokenAcquisitionError(_logger, ex.Message, ex);
+                    }
```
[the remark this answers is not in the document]
Avery-Dunn: Broadened from MsalUiRequiredException to MsalException so any silent-flow failure falls back to the full 3-leg acquisition instead of propagating up.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -526,6 +598,331 @@ private async Task<AuthenticationResult> GetAuthenticationResultForUserInternalA
...
+                if (kvp.Value.IsExpired(AgentCcaMaxIdleMilliseconds))
+                {
+                    if (_agentUserFicCcas.TryRemove(kvp.Key, out _))
+                    {
+                        _agentCcaSemaphores.TryRemove(kvp.Key, out _);
+
+                        // Clean up companion account ID entries for this agent CCA.
+                        // Account IDs are keyed as "{agentAppId}:{USER_IDENTIFIER}:{TENANTID}",
+                        // and the CCA key is "{agentAppId}:{authenticationScheme}".
+                        // Extract the agentAppId prefix to match related account entries.
```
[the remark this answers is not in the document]
Avery-Dunn: Added semaphore.Dispose() after TryRemove in the latest commit.

tests/Microsoft.Identity.Web.Test/TokenAcquisitionTests.cs
```diff
@@ -417,5 +417,377 @@ private static System.Security.Cryptography.X509Certificates.X509Certificate2 Cr
...
+            var mockHttpClient = serviceProvider.GetRequiredService<IMsalHttpClientFactory>() as MockHttpClientFactory;
+
+            // UPN flow handlers (3 legs)
+            AddAgentUserFicMockHandlers(mockHttpClient!, userAccessToken: "upn-user-token");
+            // OID flow handlers (3 legs — Leg 1 may be cached from UPN flow, but Leg 2 + Leg 3 are needed)
+            // Note: Leg 1 (blueprint FMI) is cached in the blueprint CCA, but Leg 2 uses the agent CCA
+            // which also caches T2. Since OID and UPN go to the same agent CCA, Leg 2 may be cached.
+            // We still need a Leg 3 handler for the OID grant type.
+            mockHttpClient!.AddMockHandler(CreateUserFicTokenHandler(accessToken: "oid-user-token"));
+
```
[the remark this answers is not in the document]
Avery-Dunn: The UpnAndOidCachesAreIsolated test intentionally reuses cached Legs 1/2 from the UPN flow for the OID flow, which reflects realistic behavior (the blueprint CCA and agent CCA cache app tokens across grant types). Only Leg 3 differs because UPN vs OID are distinct user identifiers.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
Avery-Dunn: @neha-bhargava: In the latest commit I've added a simple eviction strategy: every CCA instance gets a last-access timestamp, and every 30 minutes we clean up any CCA instance that hasn't been accessed in 8 hours (presumably, most tokens in the instance's cache will have expired anyway).

This may not be the best strategy, but it should be good enough for ID Web's internal behavior while we investigate potential changes to MSAL and more robust guidance for our other customers.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -527,6 +603,408 @@ private async Task<AuthenticationResult> GetAuthenticationResultForUserInternalA
...
+
+            return evictedCount;
+        }
+
+        /// <summary>
+        /// Resolves the FIC token exchange scope from ExtraParameters, falling back to the
+        /// public cloud default. Matches the override pattern used by
+        /// <see cref="TokenAcquirerExtensions.GetFicTokenAsync"/> and OidcIdpSignedAssertionProvider.
+        /// </summary>
+        private static string[] ResolveFicScopes(IDictionary<string, object>? extraParameters)
```
bgavrilMS: I don't think this is how we should solve the multi-cloud problem.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -60,6 +61,89 @@ class OAuthConstants
         private readonly ConcurrentDictionary<string, IConfidentialClientApplication?> _applicationsByAuthorityClientId = new();
         private readonly ConcurrentDictionary<string, SemaphoreSlim> _appSemaphores = new();
 
+        /// <summary>
+        /// Wraps an agent CCA with a last-access timestamp for idle-based eviction.
+        /// The sweep timer removes entries that haven't been accessed within
+        /// <see cref="AgentCcaMaxIdleMilliseconds"/>.
+        /// </summary>
+        internal sealed class AgentCcaEntry
```
bgavrilMS: Are we over-engineering this? If there are too many apps in the dictionary, say over 10000, just clear up the dictionary.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -60,6 +61,89 @@ class OAuthConstants
...
+        /// request (bot/service pattern), so there is no persistent object to write back to.
+        /// This dictionary fills that role, keyed by "{agentAppId}:{USER_IDENTIFIER}:{TENANTID}"
+        /// where USER_IDENTIFIER is either the normalized UPN or OID.
+        /// Entries are cleaned up when MSAL evicts the corresponding account from its cache.
+        /// </summary>
+        internal readonly ConcurrentDictionary<string, string> _agentUserFicAccountIds = new();
+
+        /// <summary>
+        /// Default FIC token exchange URL for the public cloud. For other clouds, callers can
+        /// override via ExtraParameters[Constants.TokenExchangeUrlKey]:
```
bgavrilMS: I don't think this is a good developer experience. Identity.Web knows the cloud, so it should infer the credential audience. 

I recommend you do not fix this issue in this PR. Let's discuss a better design.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -527,6 +603,408 @@ private async Task<AuthenticationResult> GetAuthenticationResultForUserInternalA
...
+            }
+            else if (extraParameters.TryGetValue(Constants.UserIdKey, out object? userIdObj)
+                     && userIdObj is string oidStr && Guid.TryParse(oidStr, out Guid parsedOid))
+            {
+                userObjectId = parsedOid;
+                userIdentifierForCacheKey = parsedOid.ToString("D").ToUpperInvariant();
+            }
+            else
+            {
+                // Neither UPN nor valid OID — not a user FIC flow we can handle.
```
bgavrilMS, line 630: We should have some logging for all this orchestration.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
Avery-Dunn: Discussed separately, I've removed the ExtraParameters override option and restored the existing behavior (default public cloud only)

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
Avery-Dunn: Discussed separately, I've removed the ExtraParameters override option and restored the existing behavior (default public cloud only)

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
Avery-Dunn: Added more logging in the latest commit.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
Avery-Dunn: Discussing separately for a better management style, this just seemed like a "good enough" solution while we either made changes to MSAL .NET or figured out some better guidance that would also work for our other customers.

bgavrilMS: The apps themselves are not large. The tokens are cached in memory caches or distributed caches and we can manage eviction there.

I can see merit to this as a DOS protection - unbounded growth in that dictionary can in theory occur if app has no validation on their side. But all in all, if we get to a ridicoulus number of apps, we can just wipe the dictionary. The important thing is to not lose the tokens, and those should be fine.

Avery-Dunn: > But all in all, if we get to a ridicoulus number of apps, we can just wipe the dictionary. The important thing is to not lose the tokens, and those should be fine.

Could you clarify this a bit more? The in-memory token caches are tied to CCA instances, so if we wipe the dictionary of CCA instances wouldn't we lose the tokens?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
Avery-Dunn: Discussed separately, this will be handled by follow up items about extensibility APIs in MSAL.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
bgavrilMS: No. 

Tokens are cached statically in MSAL (the instance cache doesn't go away). Or in a global distributedcache. The token caches do not go away. Each CCA instance loads the cache.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -527,6 +593,391 @@ private async Task<AuthenticationResult> GetAuthenticationResultForUserInternalA
...
+                        var silentResult = await silentBuilder.ExecuteAsync().ConfigureAwait(false);
+                        Logger.AgentUserFicSilentSuccess(_logger, accountLookupKey);
+                        return silentResult;
+                    }
+                    catch (MsalException ex)
+                    {
+                        // Catch all MSAL exceptions (not just MsalUiRequiredException) so that
+                        // any silent failure (cache errors, client errors, etc.) falls back to
+                        // the full 3-leg acquisition rather than propagating up as a hard failure.
+                        Logger.AgentUserFicSilentFailure(_logger, accountLookupKey, ex.ErrorCode ?? ex.GetType().Name, ex);
```
Robbie-Microsoft: `accountLookupKey` contains the raw UPN (`upn.ToUpperInvariant()`) or OID. It's logged here at Information, and at line 697 at Debug - both outside `EnablePiiLogging`. UPN/OID are EUII; hash or gate the user segment before logging.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -527,6 +593,391 @@ private async Task<AuthenticationResult> GetAuthenticationResultForUserInternalA
...
+            if (_agentCcaSweepTimer is not null)
+            {
+                return;
+            }
+
+            var newTimer = new Timer(
+                _ => SweepExpiredAgentCcas(),
+                null,
+                Timeout.InfiniteTimeSpan, // Don't start immediately
+                Timeout.InfiniteTimeSpan);
```
Robbie-Microsoft: The "singleton with process lifetime" assumption holds only via `AddOidcFic`/`AddAgentIdentities`, which force `AddTokenAcquisition(true)`. The agentic code is in base `TokenAcquisition`, registered scoped by default. On a scoped instance this `Timer` leaks per request (no `IDisposable`) and the CCA/account caches don't persist across requests. Implementing `IDisposable` to dispose the timer covers both lifetimes.

Robbie-Microsoft:
PR Description is out of sync with the head commit - can you update the description to reflect current code?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
Avery-Dunn: This is similar to the concern raised in: https://github.com/AzureAD/microsoft-identity-web/pull/3842#discussion_r3423920906

Although it's true the singleton assumption only holds when using `AddAgentIdentities`, that is fine: it's the only way to reach this code, and the scoped-instance scenario can't actually reach this code path.

Implementing IDisposable on TokenAcquisition would eliminate the risk of the timer leak if this behavior ever changes, but that is existing behavior and a larger change out of scope for this PR.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
Avery-Dunn: In the latest commit I've replaced accountLookupKey (contains UPN/OID) with non-PII parameters (agentAppId, normalizedTenant) in both AgentUserFicSilentSuccess and AgentUserFicSilentFailure log messages. The user identifier is no longer logged at any level.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
Avery-Dunn: I think I understand now: we still need to create 1 CCA instance per agent ID, however if they all share the cache they will still maintain isolation in the cache (due to the client IDs in the cache key being set by each of those CCA instances).

In the latest commits I've added the shared cache option to the CCA instances, and removed the complicated eviction with a simpler "clear dictionary when it reaches 10000" strategy.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -527,6 +563,326 @@ private async Task<AuthenticationResult> GetAuthenticationResultForUserInternalA
...
+                     && userIdObj is string oidStr && Guid.TryParse(oidStr, out Guid parsedOid))
+            {
+                userObjectId = parsedOid;
+                userIdentifierForCacheKey = parsedOid.ToString("D").ToUpperInvariant();
+            }
+            else
+            {
+                // Neither UPN nor valid OID — not a user FIC flow we can handle.
+                return null;
+            }
```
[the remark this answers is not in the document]
Avery-Dunn: Fixed in latest commit.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -527,6 +563,326 @@ private async Task<AuthenticationResult> GetAuthenticationResultForUserInternalA
...
+
+                // DOS protection: if the dictionary grows beyond the threshold, clear it.
+                // Tokens survive in MSAL's shared static cache and will be found by new CCAs.
+                if (_agentUserFicCcas.Count > AgentCcaMaxCount)
+                {
+                    int cleared = _agentUserFicCcas.Count;
+                    _agentUserFicCcas.Clear();
+                    _agentUserFicAccountIds.Clear();
+                    Logger.AgentCcaEviction(_logger, cleared, 0);
+                }
```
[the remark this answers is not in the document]
Avery-Dunn: Fixed in latest commit.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -527,6 +563,326 @@ private async Task<AuthenticationResult> GetAuthenticationResultForUserInternalA
...
+            return result;
+        }
+
+        /// <summary>
+        /// Gets or builds an agent CCA for the native User FIC flow. Each agent CCA uses an
+        /// assertion callback that chains back to the blueprint CCA for Leg 1 (FMI token).
+        /// Each agent CCA has a unique ClientId (the agent app ID), providing natural cache
+        /// key isolation in the shared static cache. Entries are tracked with last-access
+        /// timestamps for idle-based eviction.
+        /// </summary>
```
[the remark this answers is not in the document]
Avery-Dunn: Fixed in latest commit.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -61,6 +61,50 @@ class OAuthConstants
         private readonly ConcurrentDictionary<string, IConfidentialClientApplication?> _applicationsByAuthorityClientId = new();
         private readonly ConcurrentDictionary<string, SemaphoreSlim> _appSemaphores = new();
 
+        /// <summary>
+        /// Maximum number of agent CCA instances to keep in the dictionary before
+        /// clearing it as a DOS protection measure. Since tokens are stored in MSAL's
+        /// shared static cache (not per-CCA), clearing the dictionary only discards
+        /// lightweight CCA objects — tokens remain accessible to newly-built CCAs.
+        /// </summary>
```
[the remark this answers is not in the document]
Avery-Dunn: Fixed in latest commit.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -61,6 +61,50 @@ class OAuthConstants
...
+        /// native User FIC flow. This is needed because AcquireTokenSilent requires an IAccount,
+        /// which can only be obtained from GetAccountAsync(identifier) using an identifier that
+        /// comes back from a prior token acquisition. In all other ID Web flows, this identifier
+        /// is stored in the ClaimsPrincipal (via GetMsalAccountId / oid+tid claims). In the
+        /// agentic scenario, however, ClaimsPrincipal is typically null or freshly created per
+        /// request (bot/service pattern), so there is no persistent object to write back to.
+        /// This dictionary fills that role, keyed by "{agentAppId}:{USER_IDENTIFIER}:{TENANTID}"
+        /// where USER_IDENTIFIER is either the normalized UPN or OID.
+        /// Entries are cleaned up when MSAL evicts the corresponding account from its cache.
+        /// </summary>
```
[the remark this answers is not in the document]
Avery-Dunn: Fixed in latest commit.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.Logger.cs
```diff
@@ -81,6 +81,62 @@ public static void TokenAcquisitionMsalAuthenticationResultTime(
...
+                LoggerMessage.Define<string>(
+                    LogLevel.Information,
+                    LoggingEventId.AgentCcaCreated,
+                    "[MsIdWeb] Created new agent CCA for cache key '{CcaCacheKey}'.");
+
+            private static readonly Action<ILogger, int, int, Exception?> s_agentCcaEviction =
+                LoggerMessage.Define<int, int>(
+                    LogLevel.Information,
+                    LoggingEventId.AgentCcaEviction,
+                    "[MsIdWeb] Agent CCA sweep evicted {EvictedCount} entries. Remaining: {RemainingCount}.");
```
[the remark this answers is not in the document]
Avery-Dunn: Fixed in latest commit.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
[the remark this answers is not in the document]
Avery-Dunn: Fixed in latest commit.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -527,6 +563,326 @@ private async Task<AuthenticationResult> GetAuthenticationResultForUserInternalA
...
+                || !extraParameters.TryGetValue(Constants.AgentIdentityKey, out object? agentObj))
+            {
+                return null;
+            }
+
+            string? agentAppId = agentObj as string;
+            if (string.IsNullOrEmpty(agentAppId))
+            {
+                return null;
+            }
```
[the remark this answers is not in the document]
Avery-Dunn: Not fixing, this normalization is not done anywhere else so adding it here would be inconsistent.

src/Microsoft.Identity.Web.AgentIdentities/AgentUserIdentityMsalAddIn.cs
```diff
@@ -1,102 +0,0 @@
...
-using System.Threading.Tasks;
-using Microsoft.Extensions.DependencyInjection;
-using Microsoft.Extensions.Options;
-using Microsoft.Identity.Abstractions;
-using Microsoft.Identity.Client;
-using Microsoft.Identity.Client.Extensibility;
-
-namespace Microsoft.Identity.Web.AgentIdentities
-{
-    internal static class AgentUserIdentityMsalAddIn
```
bgavrilMS, line 15: yay

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -527,6 +566,328 @@ private async Task<AuthenticationResult> GetAuthenticationResultForUserInternalA
...
+                            silentBuilder.WithTenantId(tenantId);
+                        }
+
+                        var silentResult = await silentBuilder.ExecuteAsync().ConfigureAwait(false);
+                        Logger.AgentUserFicSilentSuccess(_logger, agentAppId!, normalizedTenant);
+                        return silentResult;
+                    }
+                    catch (MsalException ex)
+                    {
+                        // Catch all MSAL exceptions (not just MsalUiRequiredException) so that
```
bgavrilMS: This seems a bit dangerous. MsalUiRequiredException is a signal that there is not silent auth artefact (an RT). Or that RT may not be able to be refreshed (e.g. MFA introduced)

What other scenarios are you thinking of when cathing all exceptions?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -61,6 +61,53 @@ class OAuthConstants
...
+        /// a silent attempt) or when the CCA dictionary is cleared due to size-threshold eviction.
+        /// </summary>
+        internal readonly ConcurrentDictionary<string, string> _agentUserFicAccountIds = new();
+
+        /// <summary>
+        /// When true, agent User FIC CCAs use MSAL's shared (process-level static) cache.
+        /// This allows tokens to survive CCA re-creation after eviction.
+        /// Defaults to true; set to false in tests that need per-instance cache isolation.
+        /// </summary>
+        internal bool UseSharedCacheForAgentCcas { get; set; } = true;
```
bgavrilMS: Who sets this to false? 

We should not bypass Id.Web token caching. This is https://learn.microsoft.com/en-us/entra/msal/dotnet/how-to/token-cache-serialization?tabs=aspnetcore

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -527,6 +566,328 @@ private async Task<AuthenticationResult> GetAuthenticationResultForUserInternalA
...
+
+                        return leg1.AccessToken;
+                    })
+                    .WithAuthority(authority)
+                    .WithHttpClientFactory(_httpClientFactory)
+                    .WithExperimentalFeatures();
+
+                if (UseSharedCacheForAgentCcas)
+                {
+                    builder.WithCacheOptions(CacheOptions.EnableSharedCacheOptions);
```
bgavrilMS: As stated above, we already have a way to set token caching, which is described here https://learn.microsoft.com/en-us/entra/msal/dotnet/how-to/token-cache-serialization?tabs=aspnetcore

I don't think we should bypass this.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -527,6 +566,328 @@ private async Task<AuthenticationResult> GetAuthenticationResultForUserInternalA
...
+                        // Propagate tenant override to Leg 1 when the caller specifies a tenant
+                        // (e.g., via WithTenantId on Leg 2/3). Extract tenant from the token
+                        // endpoint provided by MSAL, matching the pattern used by
+                        // OidcIdpSignedAssertionProvider.ExtractTenantFromTokenEndpointIfSameInstance.
+                        string? leg1Tenant = ExtractTenantFromTokenEndpointIfSameInstance(
+                                options.TokenEndpoint,
+                                blueprintOptions.Instance);
+                        if (!string.IsNullOrEmpty(leg1Tenant))
+                        {
+                            leg1Builder.WithTenantId(leg1Tenant);
```
bgavrilMS: There is API in MSAL called `WithTenantIdFromAuthority` - can we use that?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -527,6 +566,328 @@ private async Task<AuthenticationResult> GetAuthenticationResultForUserInternalA
...
+
+                        var leg1 = await leg1Builder
+                            .ExecuteAsync(options.CancellationToken)
+                            .ConfigureAwait(false);
+
+                        return leg1.AccessToken;
+                    })
+                    .WithAuthority(authority)
+                    .WithHttpClientFactory(_httpClientFactory)
+                    .WithExperimentalFeatures();
```
bgavrilMS: What is the experiemental feature?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -527,6 +566,328 @@ private async Task<AuthenticationResult> GetAuthenticationResultForUserInternalA
...
+        /// Extracts the tenant ID from an OAuth2 token endpoint URL when the endpoint belongs
+        /// to the same cloud instance as the configured authority. Returns null if the hosts
+        /// don't match (cross-cloud) or the URL format is unrecognized.
+        /// Token endpoint format: https://login.microsoftonline.com/{tenant}/oauth2/v2.0/token
+        /// </summary>
+        /// <remarks>
+        /// This is the same logic as OidcIdpSignedAssertionProvider.ExtractTenantFromTokenEndpointIfSameInstance,
+        /// duplicated here because that method is internal to the OidcFIC project.
+        /// </remarks>
+        internal static string? ExtractTenantFromTokenEndpointIfSameInstance(
```
bgavrilMS: We should not have this kind of functionality in Id.Web; MSAL should carry this.

bgavrilMS:
I share the concern about introducing a second CCA cache for the agent User FIC path.

I don't think the blueprint CCA and agent CCA can be the exact same `IConfidentialClientApplication` instance, because the agent CCA has a different `ClientId` and needs a client assertion callback that chains back to the blueprint app for the FMI token. But I do think we should avoid maintaining `_agentUserFicCcas` as a separate CCA lifecycle/configuration path.

A safer shape would be to reuse the existing `GetOrBuildConfidentialClientApplicationAsync` / `_applicationsByAuthorityClientId` infrastructure and only parameterize the pieces that differ:

1. Add an optional cache-key suffix / discriminator, e.g. `agent-user-fic:{agentAppId}:{authenticationScheme}:{blueprintKey}`, so agent CCAs do not collide with normal CCAs.
2. Add an optional credential customization callback to `BuildConfidentialClientApplicationAsync`, so the common builder still applies `CreateWithApplicationOptions`, authority handling, logging/PII settings, Azure region, client capabilities, HTTP factory, cache options, and `_tokenCacheProvider.Initialize(...)`.
3. For agent User FIC, clone/derive the blueprint options, set `ClientId = agentAppId`, and provide only the specialized `WithClientAssertion(...)` callback that acquires the Leg 1 FMI token from the blueprint CCA.
4. Detect the agent User FIC flow before building the normal user-flow CCA. Right now `GetAuthenticationResultForUserInternalAsync` builds `application` first, then detects User FIC, which can create an unused/wrong CCA before creating the real agent CCA.

Pseudo-shape:

```csharp
private async Task<IConfidentialClientApplication> GetOrBuildConfidentialClientApplicationAsync(
    MergedOptions mergedOptions,
    bool isTokenBinding,
    string? cacheKeySuffix = null,
    Func<ConfidentialClientApplicationBuilder, string, Task>? configureCredentialsAsync = null)
```

Then inside `BuildConfidentialClientApplicationAsync`, keep all common configuration centralized and replace only this part conditionally:

```csharp
if (configureCredentialsAsync is not null)
{
    await configureCredentialsAsync(builder, authority).ConfigureAwait(false);
}
else
{
    await builder.WithClientCredentialsAsync(...).ConfigureAwait(false);
}
```

This keeps the agent CCA as a distinct MSAL app where needed, but avoids a parallel cache/configuration path that can drift from the rest of Microsoft.Identity.Web. In particular, the current PR's agent CCA builder appears to bypass the normal token cache provider initialization and several normal MSAL app settings, which could make behavior inconsistent for distributed/session cache configurations and future CCA configuration changes.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
bgavrilMS: This is not done.

bgavrilMS:
> I share the concern about introducing a second CCA cache for the agent User FIC path.
> 
> I don't think the blueprint CCA and agent CCA can be the exact same `IConfidentialClientApplication` instance, because the agent CCA has a different `ClientId` and needs a client assertion callback that chains back to the blueprint app for the FMI token. But I do think we should avoid maintaining `_agentUserFicCcas` as a separate CCA lifecycle/configuration path.
> 
> A safer shape would be to reuse the existing `GetOrBuildConfidentialClientApplicationAsync` / `_applicationsByAuthorityClientId` infrastructure and only parameterize the pieces that differ:
> 
> 1. Add an optional cache-key suffix / discriminator, e.g. `agent-user-fic:{agentAppId}:{authenticationScheme}:{blueprintKey}`, so agent CCAs do not collide with normal CCAs.
> 2. Add an optional credential customization callback to `BuildConfidentialClientApplicationAsync`, so the common builder still applies `CreateWithApplicationOptions`, authority handling, logging/PII settings, Azure region, client capabilities, HTTP factory, cache options, and `_tokenCacheProvider.Initialize(...)`.
> 3. For agent User FIC, clone/derive the blueprint options, set `ClientId = agentAppId`, and provide only the specialized `WithClientAssertion(...)` callback that acquires the Leg 1 FMI token from the blueprint CCA.
> 4. Detect the agent User FIC flow before building the normal user-flow CCA. Right now `GetAuthenticationResultForUserInternalAsync` builds `application` first, then detects User FIC, which can create an unused/wrong CCA before creating the real agent CCA.
> 
> Pseudo-shape:
> 
> ```cs
> private async Task<IConfidentialClientApplication> GetOrBuildConfidentialClientApplicationAsync(
>     MergedOptions mergedOptions,
>     bool isTokenBinding,
>     string? cacheKeySuffix = null,
>     Func<ConfidentialClientApplicationBuilder, string, Task>? configureCredentialsAsync = null)
> ```
> 
> Then inside `BuildConfidentialClientApplicationAsync`, keep all common configuration centralized and replace only this part conditionally:
> 
> ```cs
> if (configureCredentialsAsync is not null)
> {
>     await configureCredentialsAsync(builder, authority).ConfigureAwait(false);
> }
> else
> {
>     await builder.WithClientCredentialsAsync(...).ConfigureAwait(false);
> }
> ```
> 
> This keeps the agent CCA as a distinct MSAL app where needed, but avoids a parallel cache/configuration path that can drift from the rest of Microsoft.Identity.Web. In particular, the current PR's agent CCA builder appears to bypass the normal token cache provider initialization and several normal MSAL app settings, which could make behavior inconsistent for distributed/session cache configurations and future CCA configuration changes.

This is my main comment @Avery-Dunn - the AI helped me formulate it, but it captures my concern.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
Avery-Dunn: This might've been a copy/paste issue from previous code, "WithFmiPathForClientAssertion" was an experimental feature but the "WithFmiPath" we're using now is not. 

Fix in the latest commit.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -527,6 +566,330 @@ private async Task<AuthenticationResult> GetAuthenticationResultForUserInternalA
...
+
+            if (extraParameters.TryGetValue(Constants.UsernameKey, out object? usernameObj)
+                && usernameObj is string upn && !string.IsNullOrEmpty(upn))
+            {
+                username = upn;
+                userIdentifierForCacheKey = upn.ToUpperInvariant();
+            }
+            else if (extraParameters.TryGetValue(Constants.UserIdKey, out object? userIdObj)
+                     && (userIdObj is string oidStr || (oidStr = userIdObj?.ToString()!) is not null)
+                     && Guid.TryParse(oidStr, out Guid parsedOid))
```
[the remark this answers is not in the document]
Avery-Dunn: Fixed in latest commit.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
Avery-Dunn: In the latest commit this now only catches MsalUiRequiredException.

Some of the existing ROPC behavior in ID Web just caught MsalException, and since the old FIC behavior used ROPC I think that style just transferred to this new flow.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
Avery-Dunn: This was kind of a leftover from testing shared caching behavior after simplifying the eviction strategy, but in the latest code it was actually only used by one remaining test that proved not having a shared cache meant tokens disappeared when the CCA instance cache was cleared.

Since that's fairly obvious and not something we need to test, I just removed it in the latest commit.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
Avery-Dunn: Yep, looks like that works much simpler and I adjusted it in the latest commit.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
Avery-Dunn: Removed in the latest commit, as you can see it was copied from some unrelated OidcFIC behavior but for the agentic flow that this PR adjusts it can be much simpler via WithTenantIdFromAuthority

Avery-Dunn:
@bgavrilMS Since the CCA instance management has been discussed and refactored several times I created a spinoff PR to handle it: https://github.com/AzureAD/microsoft-identity-web/pull/3930

It should address all of the issues and recommendations you brought up, and once it's reviewed and approved can be merged into this main PR.

bgavrilMS (approved):

PR #3847, opened 2026-06-08
Prevent OpenIdConnectMiddlewareDiagnostics from logging sensitive values

iarekk:
## Description

`OpenIdConnectMiddlewareDiagnostics.DisplayProtocolMessage` reflected every `OpenIdConnectMessage` property to the Debug log, leaking bearer credentials (tokens, codes, secrets, client assertions, passwords) and PII (login hints, usernames, sids, etc.) in cleartext to anyone running with `LogLevel.Debug`.

Each property is now gated by the same two flags `Microsoft.IdentityModel.Logging` uses for its own output:

| Class | Gate | Fields |
|---|---|---|
| Security artifact | `IdentityModelEventSource.LogCompleteSecurityArtifact` | `Code`, `IdToken`, `IdTokenHint`, `AccessToken`, `RefreshToken`, `ClientSecret`, `ClientAssertion`, `Password` |
| PII / correlation | `IdentityModelEventSource.ShowPII` | `LoginHint`, `Username`, `UserId`, `Nonce`, `State`, `Sid` |

`ShowPII` is the master switch — when it is `false`, `LogCompleteSecurityArtifact` has no effect, matching `LogHelper.FormatInvariant`. Redacted values reuse Wilson's `HiddenPIIString` / `HiddenSecurityArtifactString` constants so the wording and `aka.ms` links are consistent with the rest of the IdentityModel stack.

No public API changes. New `[Theory]` covers all four flag combinations.

### DIFF
--- a/src/Microsoft.Identity.Web.TokenAcquisition/IDWebErrorMessage.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/IDWebErrorMessage.cs
@@ -28,6 +28,9 @@ namespace Microsoft.Identity.Web
         public const string ClientSecretAndCredentialsCannotBeCombined = "IDW10110: ClientSecret top level configuration cannot be combined with ClientCredentials. Instead, add a new entry in the ClientCredentials array describing the secret.";
         public const string MissingTokenBindingCertificate = "IDW10115: A signing certificate, which is required for token binding, is missing in loaded credentials.";
         public const string TokenBindingRequiresEnabledAppTokenAcquisition = "IDW10116: Token binding requires enabled app token acquisition.";
+        public const string OpenIdConnectMiddlewareDiagnosticsRequiresDevelopmentEnvironment = "IDW10117: 'subscribeToOpenIdConnectMiddlewareDiagnosticsEvents' was set to true outside of the Development environment. " +
+            "OpenIdConnectMiddlewareDiagnostics logs full protocol messages, including bearer tokens and PII, and must only be enabled when IHostEnvironment.IsDevelopment() returns true. " +
+            "Set the parameter to false (or only set it to true inside an `if (builder.Environment.IsDevelopment()) { ... }` block) and redeploy.";
 
         // Authorization IDW10200 = "IDW10200:"
         public const string NeitherScopeOrRolesClaimFoundInToken = "IDW10201: Neither scope nor roles claim was found in the bearer token. Authentication scheme used: '{0}'. ";
--- a/src/Microsoft.Identity.Web/Resource/OpenIdConnectMiddlewareDiagnosticsEnvironmentValidator.cs
+++ b/src/Microsoft.Identity.Web/Resource/OpenIdConnectMiddlewareDiagnosticsEnvironmentValidator.cs
@@ -1,0 +1,47 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Hosting;
+
+namespace Microsoft.Identity.Web.Resource
+{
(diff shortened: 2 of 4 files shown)


iarekk:
@microsoft-github-policy-service agree company="Microsoft"

src/Microsoft.Identity.Web/Resource/OpenIdConnectMiddlewareDiagnostics.cs
```diff
@@ -138,16 +140,56 @@ private async Task OnRedirectToIdentityProviderAsync(RedirectContext context)
...
                 }
             }
         }
 
+        // OpenIdConnectMessage properties that carry bearer credentials or other secret material.
+        // Their values are redacted unless IdentityModelEventSource.LogCompleteSecurityArtifact is true,
+        // matching the stricter gate Microsoft.IdentityModel.Logging applies to security artifacts.
+        private static readonly HashSet<string> s_securityArtifactPropertyNames = new HashSet<string>(StringComparer.Ordinal)
+        {
+            nameof(OpenIdConnectMessage.Code),
```
bgavrilMS: This is interesting idea, but it fails open. OpenIdConnectMessage can always gain a new sensitive property.

iarekk: Good point! 

I'd like to give the user as much useful diagnostics as possible before they flip any flags, so I'll flip the approach:

- all properties are considered security artifacts by default
- some properties are safe to LogDebug `whitelist_1`
- some properties are known to be PII but not security aritfacts `whitelist_2`.

And before logging a property, we'd check against the two whitelists, and if it's not on them, it will be logged as a security artifact.

So the logic will be:

```
if on whitelist_1, log the value
else if on whitelist_2, log unmaksed iff LogPII==true
else log unmasked iff LogSecurityArtifacts==true
```

iarekk:
Good catch — pushed a fail-closed redesign in 065ac255.

Instead of denying a fixed list of sensitive properties, the diagnostic now allow-lists what is safe or PII and treats **everything else as a security artifact** (hidden unless both `ShowPII` and `LogCompleteSecurityArtifact` are set). Non-string properties not in the safe list are skipped entirely, so future additions cannot leak via `ToString()` either.

To make the classification deliberate rather than implicit, `AllOpenIdConnectMessageProperties_AreExplicitlyClassifiedOrKnownSecurityArtifacts` reflects over `OpenIdConnectMessage` and fails CI if a new property is unclassified — forcing a maintainer to opt it into the safe or PII bucket on every Wilson bump.

src/Microsoft.Identity.Web/Resource/OpenIdConnectMiddlewareDiagnostics.cs
```diff
@@ -138,16 +140,56 @@ private async Task OnRedirectToIdentityProviderAsync(RedirectContext context)
 
         private void DisplayProtocolMessage(OpenIdConnectMessage message)
         {
+            bool showPii = IdentityModelEventSource.ShowPII;
+            bool logCompleteSecurityArtifact = showPii && IdentityModelEventSource.LogCompleteSecurityArtifact;
```
bgavrilMS: LogCompleteSecurityArtifact is a settable property in https://github.com/AzureAD/azure-activedirectory-identitymodel-extensions-for-dotnet (Wilson). Wilson is used for basic token validation.

I don't see any codepath in Identity.Web where this is being set, so you are probably relying on the default value. 

Is this intentional?

How about we block it at the entry point - https://github.com/AzureAD/microsoft-identity-web/blob/d575152cc581193ec2e7ade1a72ea4956d031598/src/Microsoft.Identity.Web/WebAppExtensions/MicrosoftIdentityWebAppAuthenticationBuilderExtensions.cs?plain=1#L42C112-L42C112 i.e. throw if it's `!builder.Environment.IsDevelopment()`

PR #3864, opened 2026-06-16
Test + doc: x-ms-tokenboundauth header for AKV mTLS PoP via ExtraHeaderParameters

gladjohn:
## Summary

Azure Key Vault requires `x-ms-tokenboundauth: true` on mTLS PoP requests to trigger TLS renegotiation for client certificate binding. Without it, AKV returns 401: *Client certificate required for using MTLS_POP token.*

This PR validates and documents that IdWeb's existing `ExtraHeaderParameters` config surface handles this correctly — no production code changes needed.

## Changes

- **16 unit tests** covering:
  - Happy path (commercial + sovereign clouds)
  - MSI system-assigned and user-assigned full config shape
  - FIC full config shape
  - Duplicate prevention (case-insensitive)
  - Reserved header filtering
  - `CustomizeHttpRequestMessage` dynamic approach
  - Non-AKV resources correctly omit the header
- **Updated `daemon-app-msi-mtls` sample** with `ExtraHeaderParameters` in `appsettings.json`
- **Documented** the AKV requirement in `Program.cs` header comment

## Developer configuration

```json
{
  "AzureKeyVault": {
    "BaseUrl": "https://myvault.vault.azure.net/",
    "RequestAppToken": true,
    "ProtocolScheme": "MTLS_POP",
    "Scopes": ["https://vault.azure.net/.default"],
    "AcquireTokenOptions": {
      "ManagedIdentity": { "UserAssignedClientId": "<your-client-id>" }
    },
    "ExtraHeaderParameters": { "x-ms-tokenboundauth": "true" }
  }
}
```

## Why not a DelegatingHandler?

Per reviewer feedback: hardcoding vault host strings is brittle. `ExtraHeaderParameters` is the IdWeb-native extensibility point — it's declarative, JSON-configurable, and lets each developer control which services get the header.

### DIFF
--- a/tests/DevApps/daemon-app/daemon-app-msi-mtls/Program.cs
+++ b/tests/DevApps/daemon-app/daemon-app-msi-mtls/Program.cs
@@ -13,6 +13,10 @@ using System.Net;
 //
 // Trigger: AuthorizationHeaderProviderOptions.ProtocolScheme = "MTLS_POP" in appsettings.json
 //          (under AcquireTokenOptions, with RequestAppToken = true).
+//
+// Important: Azure Key Vault requires "x-ms-tokenboundauth": "true" header on mTLS PoP
+// requests to trigger TLS renegotiation for client cert binding. This is configured via
+// ExtraHeaderParameters in appsettings.json. Without it, AKV returns 401.
 
 var factory = TokenAcquirerFactory.GetDefaultInstance();
 
--- a/tests/DevApps/daemon-app/daemon-app-msi-mtls/appsettings.json
+++ b/tests/DevApps/daemon-app/daemon-app-msi-mtls/appsettings.json
@@ -13,6 +13,9 @@
             "ManagedIdentity": {
                 "UserAssignedClientId": "4b7a4b0b-ecb2-409e-879a-1e21a15ddaf6"
             }
+        },
+        "ExtraHeaderParameters": {
+            "x-ms-tokenboundauth": "true"
         }
     },
 
--- a/tests/Microsoft.Identity.Web.Test/DownstreamWebApiSupport/MtlsPopTokenBoundAuthHeaderTests.cs
+++ b/tests/Microsoft.Identity.Web.Test/DownstreamWebApiSupport/MtlsPopTokenBoundAuthHeaderTests.cs
@@ -1,0 +1,560 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net.Http;
+using System.Security.Claims;
+using System.Threading;
+using System.Threading.Tasks;
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Microsoft.Identity.Abstractions;
+using Microsoft.Identity.Web.Test.Resource;
+using Xunit;
+
+namespace Microsoft.Identity.Web.Tests
+{
+    /// <summary>
+    /// Validates that developers can configure the x-ms-tokenboundauth header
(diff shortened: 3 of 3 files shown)


src/Microsoft.Identity.Web.TokenAcquisition/TokenBoundAuthHeaderHandler.cs
```diff
@@ -0,0 +1,74 @@
...
+        {
+            if (string.IsNullOrWhiteSpace(host))
+            {
+                return false;
+            }
+
+            host = host!.TrimEnd('.');
+
+            // Azure Key Vault (commercial, China, US Gov).
+            return host.EndsWith(".vault.azure.net", StringComparison.OrdinalIgnoreCase)
```
bgavrilMS: How do you know this is a complete list?

gladjohn: I’m using the Azure Key Vault [documented](https://learn.microsoft.com/en-us/azure/key-vault/general/about-keys-secrets-certificates) DNS suffix table as the source of truth for currently supported vault data-plane endpoints.

The documented vault suffixes are:
- .vault.azure.net
- .vault.azure.cn
- .vault.usgovcloudapi.net

gladjohn: @bgavrilMS, following our offline discussion, I reviewed the configuration options currently available in Id Web.

### Existing developer configuration options

Today developers can already add the header explicitly in two ways:

**1. `ExtraHeaderParameters` (configuration-based)**

```json
"AzureKeyVault": {
  "ProtocolScheme": "MTLS_POP",
  "ExtraHeaderParameters": {
    "x-ms-tokenboundauth": "true"
  }
}
```

**2. `CustomizeHttpRequestMessage` (code-based)**

```csharp
options.CustomizeHttpRequestMessage = msg =>
{
    msg.Headers.TryAddWithoutValidation("x-ms-tokenboundauth", "true");
};
```

### Why I still believe auto-injection is the right default

While manual configuration is possible, there are a few reasons why I think the library should handle this automatically for Azure Key Vault:

* **Poor discoverability** – Without the header, AKV returns a 401 with *"Client certificate required for using MTLS_POP token"*. The error does not indicate that `x-ms-tokenboundauth: true` is missing. I validated this end-to-end on a real VM.
* **Not all callers use DownstreamApi** – `MsalMtlsHttpClientFactory` can be used directly with MSAL + HttpClient. Those callers do not necessarily have access to `DownstreamApiOptions` or `ExtraHeaderParameters`.
* **The library has sufficient context** – When `ProtocolScheme == MTLS_POP` and the destination host is `*.vault.azure.net`, we know exactly when the header is required.
* **No impact to other resources** – The logic is scoped to Key Vault hosts only. Graph, ARM, Storage, and other resources would never receive the header.

### Both approaches can coexist

The handler already checks whether the header exists before adding it. If a developer specifies the header through `ExtraHeaderParameters` or `CustomizeHttpRequestMessage`, their configuration takes precedence and no duplicate header is added.

In that sense, auto-injection acts as a safety net rather than replacing existing extensibility points.

### If we rely solely on developer configuration

The result is that every developer using mTLS PoP with Azure Key Vault must independently discover and configure this requirement:

* The AKV error message does not point to the missing header.
* The requirement is not obvious from the API surface.
* Direct MSAL + HttpClient callers may not have a convenient configuration hook available.

For those reasons, I still think automatic injection for Key Vault-specific mTLS PoP requests provides the best out-of-box experience while remaining fully compatible with existing customization mechanisms.

bgavrilMS (changes requested):
No hardcoding of resource strings. Brittle solution. Ppl should use extra header API

bgavrilMS (approved):

PR #3839, opened 2026-06-02
Add MSI mTLS PoP support: pure MI + FIC-with-MI (impl for devex #3832)

gladjohn:
## Summary

Implements the devex proposal in #3832: **mTLS Proof-of-Possession for Managed Identity**
(pure MSI) and **FIC backed by Managed Identity** (federated identity credentials with
MI as the assertion source).

Adds a small public extension point on the existing public `ClientAssertionProviderBase`
so any future binding-capable assertion provider (e.g. OIDC FIC, tracked separately in
#3851) can opt in without further plumbing changes in `Microsoft.Identity.Web.TokenAcquisition`.

Triggered by setting `AuthorizationHeaderProviderOptions.ProtocolScheme = "MTLS_POP"` on
the downstream API options (together with `RequestAppToken = true`). No new public API
on the IdWeb consumer surface.

## Devex (the entire dev-facing surface)

```json
"AzureKeyVault": {
  "RequestAppToken": true,
  "ProtocolScheme": "MTLS_POP",
  "AcquireTokenOptions": {
    "ManagedIdentity": { "UserAssignedClientId": "..." }
  }
}
```

For FIC-with-MI, add the `SignedAssertionFromManagedIdentity` credential under
`AzureAd.ClientCredentials` — the binding cert minted by MI flows through to the outer
CCA automatically.

### Source

| File | Change |
|------|--------|
| `ClientAssertionProviderBase.cs` | Two new `public virtual` extension points: `SupportsTokenBinding` (default `false`) and `GetSignedAssertionWithBindingAsync` (default returns `null`). |
| `ManagedIdentityClientAssertion.cs` | Overrides both: `SupportsTokenBinding => true`, `GetSignedAssertionWithBindingAsync` returns a `ClientSignedAssertion` (assertion + binding cert pair). Bearer and bound paths share a private `AcquireManagedIdentityTokenAsync` helper. |
| `ConfidentialClientApplicationBuilderExtension.cs` | Generic dispatch: checks `credential.CachedValue is ClientAssertionProviderBase { SupportsTokenBinding: true }` — no narrowing to a concrete type. Other signed-assertion sources still throw `IDW10115`. |
| `TokenAcquisition.cs` | Chains `WithMtlsProofOfPossession().WithAttestationSupport()` on the pure-MI builder when `IsTokenBinding` is set. |
| `IDWebErrorMessage.cs` | Reworded `IDW10115` to reflect the new opt-in model. |
[...]

PR #3835, opened 2026-06-01
Add bound-credential support for Bearer tokens (cert + mTLS)

gladjohn:
## What this PR adds

Honors the new `CredentialDescription.UseBoundCredential` property (from microsoft-identity-abstractions-for-dotnet#252) on the **Certificate** credential branch.

When `UseBoundCredential = true` and a certificate is configured, IdWeb now wires the cert via MSAL's `WithCertificate(cert, new CertificateOptions { SendCertificateOverMtls = true })`. This causes MSAL to send the cert over mTLS to Entra's mTLS endpoint (`mtlsauth.microsoft.com`) and return a **Bearer** access token bound to that certificate.

### Changes
| File | Purpose |
|---|---|
| `ConfidentialClientApplicationBuilderExtension.cs` | Cert branch in `WithClientCredentialsAsync` now branches on `UseBoundCredential` |
| `WithClientCredentialsTests.cs` | Two new dispatch tests (`UseBoundCredential = true` / `false`) |
| `tests/DevApps/daemon-app/daemon-app-cert-bound/` | DevApp harness verified end-to-end against `mtlsauth.microsoft.com` |

## What this PR does **not** add (deferred to follow-up PRs)

- ❌ **FIC (signed-assertion) bound flow** — no new `GetSignedClientAssertionAsync` virtual on `ClientAssertionProviderBase`, no override on `ManagedIdentityClientAssertion`. The `SignedAssertion` branch in the dispatch is untouched.
- ❌ **Managed Identity → mTLS-PoP token exchange** — MI path does not call `.WithMtlsProofOfPossession()` here.
- ❌ **KeyGuard attestation** — `Microsoft.Identity.Client.KeyAttestation.WithAttestationSupport()` is not added.

These are intentionally split out so reviewers can land the smallest possible cert-path change first.

## Dependencies

- Depends on `CredentialDescription.UseBoundCredential` from microsoft-identity-abstractions-for-dotnet#252.
- **CI will fail to build** until abstractions ships the new property. Opening as **draft** until that lands.

## How it was validated

- Full IdWeb unit suite green locally (934 pass / 0 fail / 11 skip) with `-p:UseLocalAbstractions=true` swapping in a local ProjectReference to the abstractions PR branch.
- Bearer-bound DevApp harness (`daemon-app-cert-bound`) successfully obtains an access token (AT length 517, 1h expiry) from `mtlsauth.microsoft.com`.

## Reviewer notes

[...]

PR #3866, opened 2026-06-16
feat: OIDC FIC + mTLS Proof-of-Possession support 

### LINKED ISSUE #3851: [Feature Request] Add token-binding (mTLS PoP) support to OidcIdpSignedAssertionProvider
## Context
 
 PR #3839 added token-binding (mTLS PoP) support to `ManagedIdentityClientAssertion`
 via two new public virtuals on `ClientAssertionProviderBase`:
 
 - `virtual bool SupportsTokenBinding => false;`
 - `virtual Task<ClientSignedAssertion?> GetSignedAssertionWithBindingAsync(...)`
 
 `OidcIdpSignedAssertionProvider` (in `Microsoft.Identity.Web.OidcFIC`) inherits
 the defaults, so it currently produces bearer-only JWT assertions. Per @bgavrilMS's
 PR review, OIDC FIC should also support binding.
 
 This issue tracks that follow-up work, split out of #3839 to keep the MSI scope
 landable.
 
 ## Problem statement
 
 `OidcIdpSignedAssertionProvider.GetClientAssertionAsync` calls
 `_tokenAcquirer.GetTokenForAppAsync(...)` against an external OIDC IdP and returns
 the resulting JWT. To produce a `ClientSignedAssertion` with a non-null
 `TokenBindingCertificate`, we need a binding cert *paired with* the OIDC-issued
 assertion — but the external OIDC IdP cannot itself issue that cert.
 
 ## Open design question
 
 Where does the binding certificate come from when the assertion is OIDC-issued?
 Three options to evaluate:
 
 1. **Paired credential** — declarative config lists an OIDC FIC credential
    alongside a cert/KeyGuard credential; provider stitches the two together.
 2. **OidcFIC-owned cert source** — provider owns a key (KeyGuard / cert store /
    ephemeral) and produces the cert internally.
 3. **New abstraction** — `IBindingCertificateSource` injectable into the
    provider, decoupling assertion source from cert source.
 
 @bgavrilMS — would appreciate your steer on the preferred shape before we cut
 code.
 
 ## Acceptance criteria
 
 - [ ] `OidcIdpSignedAssertionProvider` overrides `SupportsTokenBinding` and
       `GetSignedAssertionWithBindingAsync`.
 - [ ] Returned `ClientSignedAssertion.TokenBindingCertificate` is non-null and
       usable by MSAL's `WithMtlsProofOfPossession()`.
 - [ ] Cert-source design documented in the OIDC FIC config schema and devex doc.
 - [ ] Unit tests covering: bound success path, missing-cert config error,
       assertion/cert mismatch.
 - [ ] E2E test exercising end-to-end mTLS PoP flow with OIDC-issued FIC.
 
 ## References
 
 - PR #3839 — introduces the `SupportsTokenBinding` extension point
 - `src/Microsoft.Identity.Web.OidcFIC/OidcIdpSignedAssertionProvider.cs` — target
 - `src/Microsoft.Identity.Web.Certificateless/ManagedIdentityClientAssertion.cs` —
   reference impl pattern

gladjohn:
## Summary

Spike implementation for **OIDC Federated Identity Credentials + mTLS Proof-of-Possession** (issue #3851, follow-up to PRs #3839 and #3835).

Adds token-binding capability to `OidcIdpSignedAssertionProvider` / `OidcIdpSignedAssertionLoader` so that OIDC FIC credentials can produce **mTLS PoP bound tokens** using a paired-credential design.

## Design: Paired Credential

The inner OIDC IdP configuration section carries two credentials:
1. **Auth credential** - authenticates the inner leg to the external OIDC IdP (e.g., a certificate or secret).
2. **Binding credential** (`UseBoundCredential = true`) - resolved by the loader as the mTLS binding certificate. Marked `Skip = true` so the inner-leg token acquisition does not accidentally use it as its own authentication credential.

The provider returns a `ClientSignedAssertion` bundle (OIDC assertion + binding cert) via `GetSignedAssertionWithBindingAsync`, enabling MSAL's `WithClientAssertion(bundle)` + `WithMtlsProofOfPossession()` flow.

## What's included

| Area | Change |
|------|--------|
| **Loader** (`OidcIdpSignedAssertionLoader.cs`) | Bound-credential resolution; bearer-to-PoP reconstruction on cache miss; `MtlsPopProtocol` constant; `ConfigureAwait(false)` on warm-up |
| **Provider** (`OidcIdpSignedAssertionProvider.cs`) | `SupportsTokenBinding` override; `GetSignedAssertionWithBindingAsync`; tenant propagation decoupled from `ClientAssertionFmiPath`; inner-leg PoP routing via `IsTokenBinding` ExtraParameter |
| **E2E test** | `OidcFic_WithMtlsPop_ReturnsBoundToken` - hits real ESTS with MSI team's allowlisted self-FIC app; asserts `mtls_pop` scheme + `cnf` x5t#S256 thumbprint + `xms_cc` propagation |
| **Unit tests** | 23 tests covering loader + provider binding paths (bearer-to-PoP upgrade, null cert throw, Skip independence, rotation limitation) |
| **Daemon sample** | End-to-end self-FIC demo producing real bound `mtls_pop` tokens against `westus3.mtlsauth.microsoft.com` |

## Key technical decisions

- **Inner-leg PoP fix**: when the outer call requests `MTLS_POP`, the inner token exchange must also request `mtls_pop` via `ExtraParameters[IsTokenBinding] = true`. Without this, ESTS rejects with `AADSTS392199`.
- **Tenant propagation**: decoupled from `ClientAssertionFmiPath` - `AcquireTokenOptions.Tenant` is now set whenever `ExtractTenantFromTokenEndpointIfSameInstance` returns a same-instance match (not only when an FMI path exists).
- **Throw, don't return null**: if the binding certificate disappears between `SupportsTokenBinding` and `GetSignedAssertionWithBindingAsync`, the provider throws `InvalidOperationException` instead of returning null.

## Known limitation (out of scope for this spike)

IdWeb's outer-CCA cache key (`GetApplicationKey`) does not include the `isTokenBinding` flag. A prior bearer call can cache a string-assertion CCA that a subsequent PoP call incorrectly reuses. The E2E test sidesteps this with a one-shot retry on the transient 307. A separate fix to the cache key is needed.

## Test results

- 969/969 `Microsoft.Identity.Web.Test` (net9.0, 4 pre-existing skips)
- 4/4 E2E tests (including new `OidcFic_WithMtlsPop_ReturnsBoundToken`)
- Daemon sample produces real `mtls_pop` tokens (verified 3/3 consecutive runs)

Closes #3851


src/Microsoft.Identity.Web.OidcFIC/OidcIdpSignedAssertionLoader.cs
```diff
@@ -75,6 +79,15 @@ internal static partial class Logger
...
             [LoggerMessageAttribute(EventId = 504, Level = LogLevel.Error, Message = "[MsIdWeb] Failed to get signed assertion from {ProviderName}. Exception occurred: {Message}. Setting skip to true.")]
             public static partial void SignedAssertionProviderFailed(ILogger logger, string providerName, string message);
+
+            /// <summary>
+            /// Logger for when more than one bound credential is configured for OIDC FIC mTLS PoP.
+            /// First-valid wins (matches IdWeb's standard fallback pattern); this is informational only.
+            /// </summary>
+            /// <param name="logger">ILogger.</param>
+            /// <param name="count">Number of bound credentials detected.</param>
+            [LoggerMessageAttribute(EventId = 505, Level = LogLevel.Debug, Message = "[MsIdWeb] OIDC FIC mTLS PoP: {Count} bound credentials configured. First valid binding cert wins (KeyVault-fallback semantics). To pin a specific binding cert, leave only one UseBoundCredential = true entry in the inner ClientCredentials collection.")]
```
bgavrilMS: Are you saying we don't support multiple bound credentials? Can we enforce this i.e. throw?

gladjohn: Correct, we don't support multiple bound credentials. The binding cert is the PoP key identity, so having two marked `UseBoundCredential = true` would be ambiguous. Changed this to throw `InvalidOperationException` with a clear message instead of silently picking the first. Multiple regular creds (without UseBoundCredential) should still work.

src/Microsoft.Identity.Web.OidcFIC/OidcIdpSignedAssertionLoader.cs
```diff
@@ -8,12 +8,16 @@
...
-using Microsoft.Identity.Web;
 
 namespace Microsoft.Identity.Web.OidcFic
 {
     internal partial class OidcIdpSignedAssertionLoader : ICustomSignedAssertionProvider
     {
+        // Centralized to avoid drift with TokenAcquisition.ProtocolNames.MtlsPop (which is internal to
+        // that assembly). Value matches what TokenAcquisition writes into CredentialSourceLoaderParameters.Protocol
+        // when AuthorizationHeaderProviderOptions.ProtocolScheme = "MTLS_POP" is set.
+        private const string MtlsPopProtocol = "MTLS_POP";
```
bgavrilMS: Such a long comment for a self explanator constants.

bgavrilMS: Should we not keep it lowercase, as per Auth headre?

gladjohn: Fixed - now shortened to one line

gladjohn: Changed to "mtls_pop" to match the auth header token type.

src/Microsoft.Identity.Web.OidcFIC/OidcIdpSignedAssertionProvider.cs
```diff
@@ -21,17 +19,130 @@ internal class OidcIdpSignedAssertionProvider : ClientAssertionProviderBase
...
+
+        /// <summary>
+        /// Returns <c>true</c> when a binding certificate is currently available on the configured bound
+        /// inner-leg <see cref="CredentialDescription"/> (its <see cref="CredentialDescription.Certificate"/>
+        /// is non-null). The cert is read live, so if the underlying credential is later reset/cleared this
+        /// property flips back to <c>false</c> and the dispatch in
+        /// <c>ConfidentialClientApplicationBuilderExtension</c> will throw a clear
+        /// <c>MissingTokenBindingCertificate</c> rather than silently use a stale cert.
+        /// </summary>
+        public override bool SupportsTokenBinding => _boundCredentialDescription?.Certificate is not null;
```
bgavrilMS, line 56: What about the case when the credential is an MSI-FIC? This is typica: 

- app in cloud 1 uses OIDC FIC from cloud 2
- app in cloud 2 uses MSI FIC 

So this would a be 3 leg FIC

1. Get credential from MSI for cloud 2
2. Get credential from Entra for cloud 1, using credential from step 1
3. Get token from Entra for cloud 1, using credential from step 2

src/Microsoft.Identity.Web.OidcFIC/OidcIdpSignedAssertionLoader.cs
```diff
@@ -190,9 +201,37 @@ public async Task LoadIfNeededAsync(CredentialDescription credentialDescription,
                     configuration.GetSection(sectionName).Bind(microsoftIdentityApplicationOptions);
                 }
 
-                // Special case for Signed assertions with an FmiPath.
-                // The provider needs to postpone getting the signed assertion until the first call, when ClientAssertionFmiPath will be provided.
-                signedAssertion = new OidcIdpSignedAssertionProvider(_tokenAcquirerFactory, microsoftIdentityApplicationOptions, credentialDescription.TokenExchangeUrl, _logger);
+                // Resolve the bound inner-leg credential only when the caller actually needs MTLS_POP.
```
bgavrilMS, line 204: What about when caller asks for Bearer token with bound credentials ?

src/Microsoft.Identity.Web.OidcFIC/OidcIdpSignedAssertionLoader.cs
```diff
@@ -212,5 +251,81 @@ public async Task LoadIfNeededAsync(CredentialDescription credentialDescription,
...
+                    // The credential was flagged bound but materialised to null (e.g. unsupported source type,
+                    // missing thumbprint, KeyVault returned no cert). Don't pin to a null cert — fall through
+                    // and try the next bound entry, if any.
+                    continue;
+                }
+
+                // Take this cred off the inner-leg auth menu so GetTokenForAppAsync below doesn't pick it.
+                // The cert itself is read live from credential.Certificate by the provider, so the outer leg
+                // doesn't need the CredentialDescription entry to remain selectable as an auth credential.
+                credential.Skip = true;
```
bgavrilMS, line 324: This is an architectural red flag. This credential should not be in the business of modifying other credentilals. Only orchestrator objects are allowed to do that, i.e. set the SKIP flag.

bgavrilMS (changes requested):
Incorrect design.

gladjohn:
@bgavrilMS I am going to work with ESTS to build a 3 leg FIC infra first and then come back to this

PR #3867, opened 2026-06-16
fix: include isTokenBinding in CCA cache key to prevent bearer/PoP collision

gladjohn:
## Problem

`GetApplicationKey` did not include the `isTokenBinding` flag in the CCA cache key. When the same app made a bearer call followed by an mTLS PoP call (or vice versa), both resolved to the same cache key — so the second call reused the CCA built for the first.

This caused MSAL to throw at request time:
> *A string-returning client assertion callback cannot be used over mTLS. Use a certificate credential or a ClientSignedAssertion callback that can return a token-binding certificate.*

The root cause is that bearer builds the CCA with `WithClientAssertion(Func<string>)` while PoP builds it with `WithClientAssertion(Func<ClientSignedAssertion>)` or `WithCertificate()` — these are fundamentally incompatible credential wirings that must not share a cache entry.

## Fix

Append `-tokenBinding` to the cache key when `isTokenBinding = true`. This is a minimal, backwards-compatible change:
- Existing bearer callers pass `false` (default) → their keys are unchanged.
- PoP callers get a distinct key → build a separate CCA with the correct credential wiring.

## Test

Added `GetOrBuildCca_BearerThenTokenBinding_DoesNotReturnCachedBearerApp` which:
1. Builds a bearer CCA (`isTokenBinding: false`) — succeeds and caches.
2. Requests a PoP CCA (`isTokenBinding: true`) — with the fix, this is a cache miss and attempts a fresh build. Since the test uses a secret credential (which can't do mTLS binding), it correctly throws IDW10115.
3. Before the fix, step 2 silently returned the bearer CCA (no throw, wrong behavior).

## Results

- 949/949 `Microsoft.Identity.Web.Test` pass (net9.0, 4 pre-existing skips)
- No API surface changes — `GetApplicationKey` is private, parameter has a default value

### DIFF
--- a/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
@@ -220,7 +220,7 @@ namespace Microsoft.Identity.Web
                     $"Certificate error detected. Retrying with next certificate (attempt {retryCount + 1}/{MaxCertificateRetries}). {exMsal.Message}",
                     exMsal);
 
-                string applicationKey = GetApplicationKey(mergedOptions);
+                string applicationKey = GetApplicationKey(mergedOptions, isTokenBinding: false);
                 NotifyCertificateSelection(loaderParameters, mergedOptions, application!, false, exMsal);
                 _applicationsByAuthorityClientId[applicationKey] = null;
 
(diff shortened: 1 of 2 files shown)


tests/Microsoft.Identity.Web.Test/TokenAcquisitionAuthorityTests.cs
```diff
@@ -217,6 +217,57 @@ public async Task VerifyDifferentRegionsDifferentAppAsync()
...
+        /// correctly throws IDW10115 (secret can't do mTLS binding). The important thing this
+        /// test verifies is that the PoP call does NOT silently return the cached bearer CCA —
+        /// it attempts a fresh build (and fails for a valid reason: wrong credential type).
+        ///
+        /// Before the cache-key fix, the second call would return the bearer CCA unchanged
+        /// (no throw), and MSAL would later throw at request time with the confusing error:
+        /// "A string-returning client assertion callback cannot be used over mTLS."
+        /// </summary>
+        [Fact]
+        public async Task GetOrBuildCca_BearerThenTokenBinding_DoesNotReturnCachedBearerApp()
```
cpp11nullptr, line 234: Can we add `GetOrBuildCca_TokenBindingThenBearer_..` test just to be more safe?

cpp11nullptr (approved):

tests/Microsoft.Identity.Web.Test/TokenAcquisitionAuthorityTests.cs
gladjohn: added now

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -239,12 +239,14 @@ private async Task<AcquireTokenResult> AddAccountToCacheFromAuthorizationCodeInt
         /// when supporting managed identities.
         /// </summary>
         /// <param name="mergedOptions">Merged configuration options.</param>
-        /// <returns>Concatenated string of authority, cliend id and azure region</returns>
-        private static string GetApplicationKey(MergedOptions mergedOptions)
+        /// <param name="isTokenBinding">Whether mTLS token binding is requested.</param>
+        /// <returns>Concatenated string of authority, client id, azure region, and token-binding flag.</returns>
+        private static string GetApplicationKey(MergedOptions mergedOptions, bool isTokenBinding = false)
```
bgavrilMS: Not a fan of default params for private / internal methods. When the method is reused by others in the team, it makes them ignore the default parameter. Let it required to force ppl to think.

bgavrilMS (approved):

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
gladjohn: removed default params

bgavrilMS (approved):

PR #3871, opened 2026-06-17
Sidecar: gate agent identity parameters behind AllowOverrides

iNinja:
## Summary

Hardens the Entra Identity sidecar by ensuring agent identity query parameters (`AgentIdentity`, `AgentUsername`, `AgentUserId`) are only processed when the per-route `AllowOverrides` flag is enabled. Previously these parameters were applied unconditionally, bypassing the override gating that was already in place for other option overrides.

### Security hardening (commit 1)
- Moved `AgentOverrides.SetOverrides()` inside the `if (allowOverrides)` block in both `AuthorizationHeaderEndpoint` and `DownstreamApiEndpoint`.
- Added 6 new tests in `AgentOverrideGatingTests` covering all four sidecar routes, verifying that agent identity parameters are ignored when `AllowOverrides` is `false` and applied when `true`.

### Documentation (commit 2)
- Updated `WithDescription` strings on all four endpoints to clarify that agent identity parameters are subject to the `AllowOverrides` setting.
- Updated the sidecar README to document agent identity parameter behaviour under `AllowOverrides`.
- Regenerated the OpenAPI spec.

### Diagnostics (commit 3)
- Added a warning log (`AgentIdentityOverridesIgnored`) that fires when agent identity parameters are provided on a route where `AllowOverrides` is disabled. The warning names the route and the configuration key to set, so operators can diagnose and remediate without guessing.

## Testing

All 12 gating tests pass (6 existing `SidecarOverrideGatingTests` + 6 new `AgentOverrideGatingTests`).


bgavrilMS (approved):

soodt:
LGTM!

PR #3880, opened 2026-06-19
Use IIdentityLogger for MSAL logging in TokenAcquisition and ManagedIdentityClientAssertion (#3820)

### LINKED ISSUE #3820: TokenAcquisition and  exposes MSAL logging as unstructured logs
TokenAcquisition still using the old LogCallback `.WithLogging` API to expose the internal logging of MSAL, instead of the newer `IIdentityLogger` based overload that supports somewhat more structured logging.

https://github.com/AzureAD/microsoft-identity-web/blob/de4fd497c652c78be43d70c63e736a858c4706a3/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs#L1098-L1105

https://github.com/AzureAD/microsoft-identity-web/blob/de4fd497c652c78be43d70c63e736a858c4706a3/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.ManagedIdentity.cs#L101-L106

The same is also true of Microsoft.Identity.Web.Certificateless's ManagedIdentityClientAssertion class (which amusingly ends up with a logger Source context of `Microsoft.Identity.Web.DefaultCertificateLoader`, despite the assembly name):

https://github.com/AzureAD/microsoft-identity-web/blob/de4fd497c652c78be43d70c63e736a858c4706a3/src/Microsoft.Identity.Web.Certificateless/ManagedIdentityClientAssertion.cs#L99

This results in log lines whose message portions look like the following: 
```
False MSAL 4.83.1.0 MSAL.NetCore .NET 10.0.7 Microsoft Windows 10.0.26200 [2026-05-18 16:29:25Z - OMITTING_CORRELATION _GUID] === Token Acquisition (ConfidentialAuthCodeRequest) started: Scopes: Authority Host: login.microsoftonline.com
```

Elsewhere Microsoft.Identity.Web has been using `IIdentityLogger` via `IdentityLoggerAdapter` from `Microsoft.IdentityModel.LoggingExtensions`. If it is feasible to switch `TokenAcquisition` and `ManagedIdentityClientAssertion` over to using IdentityLoggerAdapter, I would appreciate it.

neha-bhargava:
## Summary

Fixes #3820.

`TokenAcquisition` and `ManagedIdentityClientAssertion` configured MSAL's internal logging via the legacy `WithLogging(LogCallback, …)` overload. That callback hands back a single pre-formatted string, so the entire MSAL log line (MSAL version, runtime, OS, correlation id, message) was flattened into one **unstructured** `ILogger` message, and the log-level threshold was computed up front by a hand-rolled `ConvertMicrosoftExtensionsLogLevelToMsal` helper.

This PR switches all MSAL logging configuration to the **`IIdentityLogger`** overload, passing `IdentityLoggerAdapter` from `Microsoft.IdentityModel.LoggingExtensions` — the same adapter IdentityWeb already uses for Wilson logging in `MicrosoftIdentityBaseAuthenticationBuilder`. MSAL now emits each log entry at its proper level and `ILogger` owns the filtering.

## Changes

- `TokenAcquisition.cs` (confidential client build) — use `new IdentityLoggerAdapter(_logger)`.
- `TokenAcquisition.ManagedIdentity.cs` (managed identity build) — use `new IdentityLoggerAdapter(_logger)`.
- `ManagedIdentityClientAssertion.cs` (Certificateless) — use `new IdentityLoggerAdapter(_logger)`; added the `Microsoft.IdentityModel.LoggingExtensions` package reference to the project.
- Removed the now-unused `Log` callbacks and `ConvertMicrosoftExtensionsLogLevelToMsal` helpers from both types.
- Changelog entry under **Behavior changes**.

## Notes / out of scope

- The issue also observes that `ManagedIdentityClientAssertion`'s log entries carry the source context `Microsoft.Identity.Web.DefaultCertificateLoader`. That comes from the `ILogger` category supplied by the **caller** that constructs the assertion, not from the logging API used here, so it is intentionally left out of this change. Happy to follow up separately if a dedicated category is desired.

## Testing

- `Microsoft.Identity.Web.TokenAcquisition` and `Microsoft.Identity.Web.Certificateless` build clean (0 warnings) across all target frameworks.
- `Microsoft.Identity.Web.Test` builds clean; no existing tests referenced the removed private helpers.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -11,6 +11,7 @@
 
 ### Behavior changes
 - **B2C OpenID Connect event handler: LRU cache for issuer address.** Issuer address lookups in the B2C OIDC event handler are now cached with an LRU cache, improving performance for repeated lookups. See [#3821](https://github.com/AzureAD/microsoft-identity-web/pull/3821).
+- **MSAL logs are now emitted as structured logs.** `TokenAcquisition` and `ManagedIdentityClientAssertion` now route MSAL's internal logging through `IdentityLoggerAdapter` (the `IIdentityLogger` overload) instead of the legacy `LogCallback`, so MSAL log entries flow through `ILogger` with their original log level instead of being flattened into a single unstructured message. See [#3820](https://github.com/AzureAD/microsoft-identity-web/issues/3820).
 
 ### Dependencies updates
 - Update MSAL.NET to 4.84.1. See [#3822](https://github.com/AzureAD/microsoft-identity-web/pull/3822).
--- a/src/Microsoft.Identity.Web.Certificateless/ManagedIdentityClientAssertion.cs
+++ b/src/Microsoft.Identity.Web.Certificateless/ManagedIdentityClientAssertion.cs
@@ -12,6 +12,7 @@ using Microsoft.Identity.Client.Extensibility;
 using Microsoft.Identity.Client.KeyAttestation;
 using Microsoft.Identity.Web.Certificateless;
 using Microsoft.Identity.Web.TestOnly;
+using Microsoft.IdentityModel.LoggingExtensions;
 
 namespace Microsoft.Identity.Web
 {
@@ -98,7 +99,7 @@ namespace Microsoft.Identity.Web
 
             if (_logger != null)
             {
-                builder = builder.WithLogging(Log, ConvertMicrosoftExtensionsLogLevelToMsal(_logger), enablePiiLogging: false);
+                builder = builder.WithLogging(new IdentityLoggerAdapter(_logger), enablePiiLogging: false);
                 _logger.LogInformation($"ManagedIdentityClientAssertion with tokenExchangeUrl={_tokenExchangeUrl}");
             }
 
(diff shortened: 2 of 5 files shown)


trwalke (commented):
The log level we use in MSAL is slightly different from the one used in IdWeb since IdWeb used the standard .Net log levels. will this translation still happen correctly?

https://learn.microsoft.com/en-us/aspnet/core/fundamentals/logging/?view=aspnetcore-10.0#log-level

bgavrilMS (approved):

neha-bhargava:
Good question — there are two hops now (MSAL `LogLevel` → `EventLogLevel` → `ILogger` level). Traced both upstream (MSAL `LoggerHelper.GetEventLogLevel` + Wilson `IdentityLoggerAdapter`):

| MSAL `LogLevel` | new | old callback |
|---|---|---|
| Error | `LogError` | `LogError` |
| Warning | `LogWarning` | `LogWarning` |
| Info | `LogInformation` | `LogInformation` |
| Verbose | `LogDebug` | `LogDebug` |
| Always | `LogTrace` | `LogInformation` |

4/5 are identical. Only `Always` shifts (Info → Trace), and that's just MSAL's "health metric" lines — not the Info/Error/Warning logs this PR is about. It's the standard MSAL → `IIdentityLogger` → `IdentityLoggerAdapter` behavior, i.e. the same adapter we already use for Wilson logging. Filtering is also now per-entry via `ILogger.IsEnabled` instead of a one-time threshold computed at construction. I can drop in a small custom `IIdentityLogger` if we want `Always` to stay at Information, but I'd lean toward matching the rest of the stack.

PR #3899, opened 2026-06-29
Adopt IAuthorizationHeaderProvider2 from Abstractions 12.3.0

4gust:
- Bump Microsoft.Identity.Abstractions to 12.3.0
- Implement IAuthorizationHeaderProvider2 on DefaultAuthorizationHeaderProvider and the public BaseAuthorizationHeaderProvider; consolidate all header-creation paths through a single BuildHeaderInformationAsync engine
- Propagate TokenAcquisitionMetadata and AdditionalResponseParameters from MSAL onto AuthorizationHeaderInformation via the existing AcquireTokenResultFactory
- DownstreamApi and MicrosoftIdentityMessageHandler prefer IAuthorizationHeaderProvider2 for mTLS PoP, with IBoundAuthorizationHeaderProvider kept as a back-compat fallback
- Add PublicAPI.Unshipped entries for the new virtuals on BaseAuthorizationHeaderProvider
- Add DefaultAuthorizationHeaderProviderV2Tests covering bearer/app/user flows, metadata + additional-response-parameter propagation, and token-binding guard

# {PR title}

## Description

{Detail}


src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
```diff
@@ -713,7 +713,28 @@ public Task<HttpResponseMessage> CallApiForAppAsync(
...
+                    {
+                        // in theory it shouldn't happen because in case of error during token acquisition
+                        // there will be thrown corresponding exception, so it's more a safeguard
+                        throw new InvalidOperationException("Cannot acquire bound authorization header.");
+                    }
+
+                    authorizationHeaderInformation = authorizationHeaderResult.Result;
+                    authorizationHeader = authorizationHeaderInformation?.AuthorizationHeaderValue!;
+                }
+                else if (_authorizationHeaderProvider is IBoundAuthorizationHeaderProvider boundAuthorizationHeaderBoundProvider
```
bgavrilMS: add a comment // for backwards compatibility.

bgavrilMS (approved):

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
gladjohn: Minor (readability / keep-in-sync): this branch and the `IAuthorizationHeaderProvider2` branch above both re-evaluate the same `string.Equals(effectiveOptions.ProtocolScheme, Constants.TokenBindingProtocolScheme, StringComparison.OrdinalIgnoreCase)`, and the two `&&` operands are in opposite order between the two branches, which makes the duplicated comparison easy to miss. No behavior change (it is already an `else if`), but hoisting it once keeps the branches in sync and matches the `isTokenBinding` hoist already used in `DefaultAuthorizationHeaderProvider.BuildHeaderInformationAsync`:

    bool isTokenBinding = string.Equals(effectiveOptions.ProtocolScheme, Constants.TokenBindingProtocolScheme, StringComparison.OrdinalIgnoreCase);

    if (isTokenBinding && _authorizationHeaderProvider is IAuthorizationHeaderProvider2 providerV2) { ... }
    else if (isTokenBinding && _authorizationHeaderProvider is IBoundAuthorizationHeaderProvider boundProvider) { ... }

src/Microsoft.Identity.Web.TokenAcquisition/MicrosoftIdentityMessageHandler.cs
```diff
@@ -484,6 +484,31 @@ private async Task<AuthArtifacts> AcquireAuthArtifactsAsync(
...
+                        claimsPrincipal: null,
+                        cancellationToken).ConfigureAwait(false);
+
+                    if (!boundResult.Succeeded)
+                    {
+                        throw new MicrosoftIdentityAuthenticationException(
+                            "Failed to acquire bound authorization header for mTLS PoP.");
+                    }
+
+                    return new AuthArtifacts(
```
gladjohn, line 506: Minor (maintainability): correctness here depends on this early `return`. The token-binding block just below (the `IBoundAuthorizationHeaderProvider` path) is a separate `if`, not an `else if`, so the two are mutually exclusive only because this V2 block returns first. It is correct today, but if this `return` were ever refactored into a fall-through, both blocks could execute when a provider implements both interfaces and the scheme is token-binding. Promoting the block below to `else if` (and hoisting the shared `bool isTokenBinding`) would make the mutual exclusivity structural rather than dependent on this return.

gladjohn (commented):
Two minor maintainability nits on the new mTLS-PoP token-binding dispatch (not blocking).

gladjohn (approved):

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
4gust: done

4gust: updated

src/Microsoft.Identity.Web.TokenAcquisition/MicrosoftIdentityMessageHandler.cs
4gust: updated

bgavrilMS (approved):

PR #3902, opened 2026-06-30
Flow finalized DownstreamApi request before authorization header creation

bgavrilMS:
## Summary
- finalize DownstreamApi request headers, query parameters, content, and customizations before creating the authorization header
- keep Authorization added after signing so request-binding providers do not include it in their signed material
- add coverage that the authorization provider sees the final request state

## Context
This supports MISE outbound SHR q/h/b signing by ensuring request-binding authorization providers receive the same HTTP request state that will be sent on the wire.

### DIFF
--- a/tests/Microsoft.Identity.Web.Test/DownstreamWebApiSupport/DownstreamApiTests.cs
+++ b/tests/Microsoft.Identity.Web.Test/DownstreamWebApiSupport/DownstreamApiTests.cs
@@ -145,6 +145,51 @@ namespace Microsoft.Identity.Web.Tests
             Assert.Equal(options.AcquireTokenOptions.ExtraQueryParameters, DownstreamApi.CallerSDKDetails);
         }
 
+        [Fact]
+        public async Task UpdateRequestAsync_WithScopes_FlowsFinalRequestToAuthorizationHeaderProviderAsync()
+        {
+            // Arrange
+            var authorizationHeaderProvider = new CapturingAuthorizationHeaderProvider();
+            var downstreamApi = new DownstreamApi(
+                authorizationHeaderProvider,
+                _namedDownstreamApiOptions,
+                _httpClientFactory,
+                _logger,
+                msalHttpClientFactory: null,
+                credentialsProvider: _provider);
+
+            var httpRequestMessage = new HttpRequestMessage(HttpMethod.Post, "https://example.com/path");
+            var content = new StringContent("test content");
+            var options = new DownstreamApiOptions
+            {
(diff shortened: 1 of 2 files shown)


4gust (approved):

tests/Microsoft.Identity.Web.Test/DownstreamWebApiSupport/DownstreamApiTests.cs
```diff
@@ -145,6 +145,51 @@ public async Task UpdateRequestAsync_WithScopes_AddsAuthorizationHeaderToRequest
...
+            // Act
+            await downstreamApi.UpdateRequestWithCertificateAsync(httpRequestMessage, content, options, false, new ClaimsPrincipal(), CancellationToken.None);
+
+            // Assert
+            Assert.Same(httpRequestMessage, authorizationHeaderProvider.CapturedRequest);
+            Assert.Same(content, authorizationHeaderProvider.CapturedRequest!.Content);
+            Assert.Contains("fromOptions=query-value", authorizationHeaderProvider.CapturedRequest.RequestUri!.Query, StringComparison.Ordinal);
+            Assert.True(authorizationHeaderProvider.CapturedRequest.Headers.Contains("X-From-Options"));
+            Assert.True(authorizationHeaderProvider.CapturedRequest.Headers.Contains("X-From-Customizer"));
+            Assert.False(authorizationHeaderProvider.AuthorizationHeaderPresentWhenCaptured);
```
gladjohn, line 189: can we also have a negative test ExtraHeaderParameter using a reserved name is still skipped even in the new position, ensuring the reorder didn't change the reserved/duplicate-skip semantics.

gladjohn (approved):

tests/Microsoft.Identity.Web.Test/DownstreamWebApiSupport/DownstreamApiTests.cs
4gust: Added UpdateRequestAsync_ReservedExtraHeaderParameters_AreSkippedInFlowedRequestAsync as a [Theory] over Authorization, Cookie, Host, X-Original-URL, and X-MS-CLIENT-PRINCIPAL. It asserts the reserved header from ExtraHeaderParameters is not flowed on the request captured by the authorization provider (new pre-signing position), that a non-reserved sibling still flows, and that the final request contains no caller-supplied value for the reserved name — confirming the reorder preserved reserved/duplicate-skip semantics.

gladjohn (approved):

PR #3908, opened 2026-07-01
Use Microsoft-hosted windows-2022 pool for PR pipeline

iarekk:
## Summary
Align the PR build agent pool with **MSAL.NET (Azure Pipelines)**.

The root `azure-pipelines.yml` PR build job previously ran on the custom `MwWilson1EsHostedPool`. This changes it to the Microsoft-hosted **`windows-2022`** vmImage, matching the pool MSAL.NET uses for its build/test jobs.

## Changes
- `azure-pipelines.yml`: replace `pool.name: MwWilson1EsHostedPool` with `pool.vmImage: 'windows-2022'`.
- Add the `visualstudio` demand alongside `msbuild` to mirror MSAL.NET's Windows build jobs.

## Reference
MSAL.NET `build/template-build-and-run-all-tests.yaml` uses:
```yaml
pool:
    vmImage: 'windows-2022'
    demands:
    - msbuild
    - visualstudio
```

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -10,9 +10,10 @@ pr:
 jobs:
 - job: "PR_build"
   pool:
-    name: MwWilson1EsHostedPool
+    vmImage: 'windows-2022'
     demands:
     - msbuild
+    - visualstudio
   steps:
     - template: build/template-install-dependencies.yaml
     - template: build/template-run-unit-tests.yaml


azure-pipelines.yml
```diff
@@ -10,9 +10,10 @@ pr:
 jobs:
 - job: "PR_build"
   pool:
-    name: MwWilson1EsHostedPool
```
gladjohn, line 13: think this VM has 'Managed Identity` enabled for some testing with MSI

Avery-Dunn: It's something I was able to remove from most jobs in the MISE PR pipeline too: https://dev.azure.com/IdentityDivision/DevEx/_git/MISE/pullrequest/24625

It might've been needed for something in the past, but most tasks seem to be able to run on the standard agents these days (not sure about ID Web though, but we'll know from actually pipeline runs).

Avery-Dunn (approved):

azure-pipelines.yml
iarekk: Thanks both! I'll merge it to see how the new pipeline behaves on ADO. Worst case we go back to this pool :)

PR #3917, opened 2026-07-02
Run PR pipeline on hosted agents with LabAuth cert from store

iarekk:
Combines the changes previously split across #3916 and #3917.

## Summary
Move the GitHub PR pipeline back to Microsoft-hosted agents (`windows-2022`) and make the integration tests work there.

## Changes
1. **Revert to hosted pool** — `azure-pipelines.yml` runs on `windows-2022` instead of `MwWilson1EsHostedPool` (reverts #3913).
2. **LabAuth cert from store** — the integration-test host (`IntegrationTestService`) previously loaded its `LabAuth` client certificate from the `msidlabs` KeyVault (`SourceType: KeyVault`), which Microsoft.Identity.Web resolves via `DefaultAzureCredential`. Hosted agents have no managed identity with KeyVault access, so tests failed with `IDW10109: ... ManagedIdentityCredential ... Identity not found`.

   A new opt-in switch (env var `UseLabAuthCertFromStore`, read in `IntegrationTestService/Program.cs`) overrides the client-certificate source for both schemes (`AzureAd`, `AzureAd2`) to load the `LabAuth` cert from the local store (`StoreWithDistinguishedName`, `LocalMachine/My`, `CN=LabAuth.MSIDLab.com`) — which `build/template-install-dependencies.yaml` already installs. The in-process tests use `WebApplicationFactory<Startup>`, which invokes `Program.CreateHostBuilder`, so the override applies.

## Safety
- **Default (unset) = KeyVault**, so the official OneBranch/Wilson pipeline and local dev are unchanged.
- The switch is enabled **only** in the hosted GitHub PR pipeline (`azure-pipelines.yml`); the OneBranch `.pipelines/*.yml` are untouched.

This mirrors how MSAL.NET's integration tests authenticate with the installed cert rather than a managed identity.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -9,6 +9,14 @@ pr:
 
 jobs:
 - job: "PR_build"
+  variables:
+    # Load the LabAuth client certificate from the local certificate store
+    # (installed into LocalMachine/My by template-install-dependencies.yaml)
+    # instead of the msidlabs KeyVault. Microsoft-hosted agents have no managed
+    # identity that can read the KeyVault, so DefaultAzureCredential would fail.
+    # The official OneBranch/Wilson pipeline does not set this and keeps using
+    # the KeyVault source.
+    UseLabAuthCertFromStore: true
   pool:
     vmImage: 'windows-2022'
     demands:
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -10,7 +10,7 @@ pr:
 jobs:
 - job: "PR_build"
   pool:
-    name: MwWilson1EsHostedPool
+    vmImage: 'windows-2022'
     demands:
     - msbuild
     - visualstudio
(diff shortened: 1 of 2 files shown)


azure-pipelines.yml
```diff
@@ -9,8 +9,16 @@ pr:
 
 jobs:
 - job: "PR_build"
+  variables:
+    # Load the LabAuth client certificate from the local certificate store
+    # (installed into LocalMachine/My by template-install-dependencies.yaml)
+    # instead of the msidlabs KeyVault. Microsoft-hosted agents have no managed
```
gladjohn, line 15: we need to first install the cert using the build steps

gladjohn: https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/blob/main/build/template-install-keyvault-secrets.yaml

iarekk: Thanks! I think it's done here already: https://github.com/AzureAD/microsoft-identity-web/blob/master/build/template-install-dependencies.yaml

gladjohn (approved):

4gust (approved):

PR #3910, opened 2026-07-02
Fix duplicate logging of MsalUiRequiredException

armin-azar:
# Fix duplicate logging of MsalUiRequiredException

- [x] You've read the [Contributor Guide](https://github.com/AzureAD/microsoft-identity-web/blob/master/CONTRIBUTING.md) and [Code of Conduct](https://github.com/AzureAD/microsoft-identity-web/blob/master/CODE_OF_CONDUCT.md).
- [x] You've included unit or integration tests for your change, where applicable.
- [x] You've included inline docs for your change, where applicable.
- [x] There's an open issue for the PR that you are making. If you'd like to propose a new feature or change, please open an issue to discuss the change or find an existing issue.

Summary of the changes (Less than 80 chars)

Stop double-logging MsalUiRequiredException

## Description

MsalUiRequiredException is already logged internally by MSAL.NET at error
level. Microsoft.Identity.Web additionally logged the same exception via
`Logger.TokenAcquisitionError` in two places in TokenAcquisition.cs, both
of which only rethrow/wrap the exception (no recovery logic depends on the
log call). This produced two log entries for a single MsalUiRequiredException,
as described in the issue.

This PR removes the redundant logging calls in those two catch blocks and
adds a comment explaining why nothing is logged there. All other
Logger.TokenAcquisitionError call sites (which handle exceptions MSAL does
not already log, e.g. certificate retry) are unchanged.

Added a unit test that reproduces the scenario from the issue (AcquireTokenSilent
called with no account/login hint, which MSAL throws MsalUiRequiredException for
synchronously) and asserts the log is not emitted.

Fixes #3528

PR #3925, opened 2026-07-04
Redeem authorization code with MSAL when complex client credentials are configured (#3631)

### LINKED ISSUE #3631: Authorization code should always be redeemed with MSAL

Microsoft.Identity.Web

4.1.0

### Web app

Sign-in users

### Web API

Not Applicable

### Token cache serialization

Not Applicable

### Description

I have configured my web application to use `SignedAssertionFromManagedIdentity` instead of `clientSecret` and `ResponseType=code`.
When redeeming the `authorization_code` on redirection, the app throws the exception

```
Microsoft.AspNetCore.Authentication.AuthenticationFailureException: An error was encountered while handling the remote login.
 ---> Microsoft.IdentityModel.Protocols.OpenIdConnect.OpenIdConnectProtocolException: Message contains error: 'invalid_client', error_description: 'AADSTS7000218: The request body must contain the following parameter: 'client_assertion' or 'client_secret'. Trace ID: 4fd981cf-1ae9-4f05-9c7b-f0608a8f0700 Correlation ID: eeb1c507-503c-4a69-84d1-42f2f39543cb Timestamp: 2025-11-28 14:53:40Z', error_uri: 'https://login.microsoftonline.com/error?code=7000218'.
   at Microsoft.AspNetCore.Authentication.OpenIdConnect.OpenIdConnectHandler.RedeemAuthorizationCodeAsync(OpenIdConnectMessage tokenEndpointRequest)
   at Microsoft.AspNetCore.Authentication.OpenIdConnect.OpenIdConnectHandler.HandleRemoteAuthenticateAsync()
   --- End of inner exception stack trace ---
   at Microsoft.AspNetCore.Authentication.RemoteAuthenticationHandler`1.HandleRequestAsync()
   at Microsoft.AspNetCore.Authentication.AuthenticationMiddleware.Invoke(HttpContext context)
   at Microsoft.AspNetCore.Diagnostics.DeveloperExceptionPageMiddlewareImpl.Invoke(HttpContext context)
```

I configured logging level to debug but I don't see relevant traces around the exception that would tell the credential source fails.

With client secret, it works.

My app runs as a docker container inside App Services. The managed identity works ok to retrieve manually an assertion with this endpoint

```c#
// add a minimal endpoint, that returns "Hello World!"
app.MapGet("/token", async (ILogger<Program> logger) =>
{
    var fic = new ManagedIdentityClientAssertion("zzz", null, logger);
    var a = await fic.GetSignedAssertionAsync(new Microsoft.Identity.Client.AssertionRequestOptions()
    {
        ClientID = "yyy",
        TokenEndpoint = "https://login.microsoftonline.com/xxx/oauth2/token"
    });
    return a;
});
```

### Reproduction steps

1. Program.cs
    ```c#
    var builder = WebApplication.CreateBuilder(args);
    builder.Services.AddMicrosoftIdentityWebAppAuthentication(builder.Configuration);
    var app = builder.Build();

    app.MapGet("/protected", () => "Hello").RequireAuthorization(p => p.RequireAuthenticatedUser());

    app.Run();
    ```
2. Configuration
```json
{
  "AzureAd": {
    "TenantId": "xxx",
    "ClientId": "yyy",
    "ResponseType": "code",
    "ClientCredentials": [ 
      {
        "SourceType": "SignedAssertionFromManagedIdentity",
        "ManagedIdentityClientId": "zzz"
      }
    ],
    "CallbackPath": "/signin-oidc"
  }
}
```

3. In Entra of tenant Id xxx, create an app registration, declare a FIC with the managed identity zzz. This app registration has a client id yyy.

### Error message

_No response_

### Id Web logs

_No response_

### Relevant code snippets

```csharp
provided above
```

### Regression

_No response_

### Expected behavior

I would expect a more meaningfull exception, or at least some traces. With `subscribeToOpenIdConnectMiddlewareDiagnosticsEvents` set to true, I do have traces from category `Microsoft.Identity.Web` but not useful. Otherwise, I don't see any traces from this category.

armin-azar:
# Redeem authorization code with MSAL when complex client credentials are configured (#3631)

Redeem auth code with MSAL for complex credentials in sign-in-only apps

### Fixes #3631

> See **Known limitations** below.
> Opening as a **interim** so the maintainers can weigh in on the design trade-offs before this is finalized.

## Problem

When a web app is configured with a **complex client credential**, meaning `SignedAssertionFromManagedIdentity`, a certificate, or anything other than a plain client secret, together with `ResponseType=code`, sign-in fails with **AADSTS7000218** (`client_assertion` or `client_secret` required).

Root cause: the default ASP.NET Core `OpenIdConnectHandler` performs the authorization-code redemption on the callback. It has no knowledge of the complex credential, so it sends the token request without a `client_assertion`, and Azure AD rejects it. Microsoft.Identity.Web only hooks `OnAuthorizationCodeReceived` (and therefore redeems the code via MSAL with the correct credential) when `EnableTokenAcquisitionToCallDownstreamApi()` is called. A sign-in-only app has no reason to call that method, so it hits the failure with no actionable error explaining why.

## Reproduction

I reproduced this end-to-end against my own Azure test subscription and Entra ID tenant **before** making any changes:

- Web app registered in Entra ID, configured with a federated credential (`SignedAssertionFromManagedIdentity`) and `ResponseType=code`, sign-in only (no `EnableTokenAcquisitionToCallDownstreamApi()`).
- Sign-in consistently failed on the callback with **AADSTS7000218**, confirming the report.

## Fix

Microsoft.Identity.Web now detects the scenario and redeems the code with MSAL automatically:

- **Detection**: a credential is considered "complex" when any `ClientCredentials` entry has a `SourceType` other than `ClientSecret`; combined with `ResponseType=code` this triggers the automatic path.
- **Wiring**: `OnAuthorizationCodeReceived` is wired via a `PostConfigure<OpenIdConnectOptions>` step so it runs after all `Configure` steps, including the one `EnableTokenAcquisitionToCallDownstreamApi()` registers.
- **Services**: the token acquisition services and a minimal **internal** no-op `IMsalTokenCacheProvider` are registered so a sign-in-only app does not have to register a token cache. The no-op type is kept internal and is not part of the public API.
- **Diagnostics**: instead of a bare AADSTS7000218, the library logs an informational message when auto-redemption is enabled, and a warning when it is needed but the token acquisition services are not available.

## Backward compatibility

- **`EnableTokenAcquisitionToCallDownstreamApi()` is unchanged.** A flag on `MergedOptions` (`AuthorizationCodeHandledByMicrosoftIdentityWeb`) records that the code is already being redeemed by Microsoft.Identity.Web, so the automatic path never wires a second, redundant handler.
- **Plain client secret with `ResponseType=code` is unaffected.** The OIDC handler already redeems those correctly, and the automatic path does not activate.
- **Existing workaround keeps working.** Apps that adopted the workaround from the issue thread, calling `EnableTokenAcquisitionToCallDownstreamApi()` and registering a dummy or no-op `IMsalTokenCacheProvider`, will **not** break after upgrading: that path still sets the flag above (so the automatic path stays off), and the service registrations are idempotent, so the dummy cache continues to be used exactly as before.
- **Apps that already redeem the code themselves** in a custom `OnAuthorizationCodeReceived` handler are **not** double-redeemed: the automatic handler invokes the application handler first and only redeems with MSAL if the code has not already been handled (`AuthorizationCodeReceivedContext.HandledCodeRedemption`), which avoids replaying the single-use code.

## Testing

- **Repro before the change**: confirmed AADSTS7000218 against a real Entra ID tenant (see Reproduction above).
- **Smoke test after the change**: with the same Entra ID app registration (federated credential plus `ResponseType=code`, sign-in only, no `EnableTokenAcquisitionToCallDownstreamApi()`), the interactive sign-in path now completes successfully; the code is redeemed with MSAL using the configured credential.
- **Unit tests added** (`WebAppExtensionsTests`):
  - Automatic redemption fires for `SignedAssertionFromManagedIdentity` plus `ResponseType=code` **without** `EnableTokenAcquisitionToCallDownstreamApi()`, and the token acquisition and cache services are registered automatically.
  - No double redemption: when the app already redeems the code in its own handler, MSAL redemption is skipped.
  - Plain client secret plus `ResponseType=code` does **not** auto-enable token acquisition (no added service footprint).
- Full existing unit test suite passes (`net8.0`): 991 passed, 11 pre-existing skips.

## Known limitations and open questions for maintainers

The automatic activation is convenient but has unconditional side effects, and I would like the team's guidance on how far to take this before it merges:

1. **Token cache slot.** The fallback no-op `IMsalTokenCacheProvider` is registered eagerly with `TryAddSingleton`. If an app sets `ResponseType=code` directly on its own options alongside a complex credential and later calls `.AddInMemoryTokenCaches()` or `.AddDistributedTokenCaches()` (also `TryAddSingleton`, first registration wins), the real cache is shadowed by the no-op. Options: register the real `MsalMemoryTokenCacheProvider` instead, defer cache registration, or let `TokenAcquisition` tolerate a missing provider.
2. **Sign-out.** The automatic path wires `OnRedirectToIdentityProviderForSignOut` to call `RemoveAccountAsync`, which builds the confidential client (loading the credential). A credential outage could turn a previously pure-redirect sign-out into a failure for a sign-in-only app.
3. **B2C.** The automatic path does not perform the `client_info` (uid/utid) claim injection that the `EnableTokenAcquisitionToCallDownstreamApi()` path does in `OnTokenValidated`, so account cleanup at sign-out may not locate the account in B2C or CIAM scenarios.
4. **Eager probe.** To decide whether to register services at `ConfigureServices` time, the options-configuration delegate is invoked once against a throwaway options instance, wrapped in a broad `catch`. This runs for every `AddMicrosoftIdentityWebApp` call, which means a delegate with side effects runs an extra time; it also cannot observe a `ResponseType` set directly on `OpenIdConnectOptions` (only what the delegate sets), so that sub-case is detected at `PostConfigure` time and logs a warning rather than redeeming. A design that always registers the (idempotent, cheap) token acquisition services and drops the probe would remove this, at the cost of resolving the cache-slot question in item 1.

Happy to iterate on any of these based on the direction you prefer.

Fixes #3631.


armin-azar:
Hi @gladjohn, following up on this one since it's been a couple of weeks. Thanks for reviewing my last PR (#3910) and hoping you might have bandwidth for this one too.
This fixes a real sign-in failure (`AADSTS7000218`) for apps using complex client credentials with `ResponseType=code`, reproduced end-to-end against a live Entra ID tenant, with unit tests and backward-compatibility checks included.
I flagged it as interim because I have a few open design questions (listed in the PR description) where I'd like maintainer input before finalizing but if it's easier, my recommendation is to go with the approach in item 1 (register the real token cache provider instead of the no-op) unless you'd prefer a different direction. Happy to adjust based on whatever guidance works best.
No rush if you're busy, just wanted to make sure this didn't slip through unnoticed. Thanks for taking a look!

PR #3930, opened 2026-07-07
Refactor CCA instance management in agentic flows

Avery-Dunn:
This PR is meant to address this comment in the main agentic refactoring PR: https://github.com/AzureAD/microsoft-identity-web/pull/3842#issuecomment-4897313513

Managing CCA instances is the main expected pain point of the new agentic behavior supported in recent versions of MSAL. This sort of per-agent CCA instance management will become irrelevant once MSAL extensibility APIs are improved, however until those improvements are made and released ID Web should manage the per-agent CCA instances in a proper way.

## Concern

The original PR introduced a dedicated `GetOrBuildAgentUserFicCcaAsync` that built agent CCAs via `ConfidentialClientApplicationBuilder.Create(agentAppId)` with manual `.WithAuthority()`, `.WithHttpClientFactory()`, and `.WithCacheOptions()` calls — bypassing the normal `BuildConfidentialClientApplicationAsync` path. This meant agent CCAs missed:

- `CreateWithApplicationOptions()` (AzureRegion, ClientCapabilities, ClientName/Version)
- `.WithLogging()` (structured logging)
- `_tokenCacheProvider.Initialize()` (**distributed cache support**)
- B2C/CIAM authority handling
- `.WithExperimentalFeatures()`

In addition, there are existing `GetOrBuildConfidentialClientApplicationAsync` / `_applicationsByAuthorityClientId` infrastructure that we could adjust to build the CCAs rather than keeping a parallel track for the per agent CCAs.

### 1. Cache key suffix

`GetOrBuildConfidentialClientApplicationAsync` now accepts an optional `agentAppId` parameter. When set, the cache key becomes `{blueprintKey}:agent:{agentAppId}`, storing agent CCAs in the shared `_applicationsByAuthorityClientId` dictionary alongside normal CCAs.

The `authenticationScheme` is not explicitly in the agent suffix because different schemes resolve to different `MergedOptions`, which produce different `GetApplicationKey()` base keys — so scheme isolation is implicit in the blueprint key portion.

### 2. Credential customization callback

`BuildConfidentialClientApplicationAsync` now accepts an optional `clientAssertionProvider` callback (`Func<AssertionRequestOptions, Task<string>>`). When provided, the builder calls `.WithClientAssertion(clientAssertionProvider)` instead of `.WithClientCredentialsAsync(...)`. All other configuration is shared:

- `CreateWithApplicationOptions(ccaOptions)` — authority, AzureRegion, ClientCapabilities, ClientName/Version
- `.WithHttpClientFactory()`, `.WithLogging()`, `.WithExperimentalFeatures()`
- Authority handling (OIDC/B2C/standard)
- Cache options (shared + distributed)
- `_tokenCacheProvider.Initialize()` on both AppTokenCache and UserTokenCache

**Deviation from suggestion:** The reviewer proposed `Func<ConfidentialClientApplicationBuilder, string, Task>` (exposing the builder + authority). We use `Func<AssertionRequestOptions, Task<string>>` (the MSAL assertion callback directly), which keeps the builder encapsulated — `GetOrBuildAgentUserFicCcaAsync` constructs the callback and passes it to the unified builder rather than receiving a raw builder to configure.

### 3. Clone/derive blueprint options

For agent CCAs, `BuildConfidentialClientApplicationAsync` creates a **fresh `ConfidentialClientApplicationOptions`**, populates it via `MergedOptions.UpdateConfidentialClientApplicationOptionsFromMergedOptions(mergedOptions, ccaOptions)`, then sets `ccaOptions.ClientId = agentAppId`. The blueprint's lazily-cached `ConfidentialClientApplicationOptions` instance is never mutated.

### 4. Earlier agent flow detection

`GetAuthenticationResultForUserInternalAsync` now calls `TryGetAuthenticationResultForAgentUserFicAsync` **before** `GetOrBuildConfidentialClientApplicationAsync`. For agent requests, the blueprint CCA is only built lazily inside the assertion callback when Leg 1 actually fires. The `IConfidentialClientApplication application` parameter was removed from `TryGetAuthenticationResultForAgentUserFicAsync` entirely since the agent CCA gets its authority from `MergedOptions`.

## Additional changes beyond the original comment

- **Selective agent eviction**: DOS eviction only removes entries with `:agent:` in the key — the blueprint CCA is preserved. The old separate-dictionary design cleared everything.
- **Removed `_agentUserFicCcas` and `_agentCcaSemaphores`**: Two entire `ConcurrentDictionary` fields eliminated. Agent CCAs reuse `_applicationsByAuthorityClientId` and `_appSemaphores`.
- **Agent redirect URI skip**: Agent CCAs skip `WithRedirectUri()` since redirect URIs aren't meaningful for client-credential + assertion flows.
- **`_applicationsByAuthorityClientId` visibility**: Changed from `private` to `internal` to support test assertions on shared dictionary contents.

## How this compares to the original (master) agent flow

The master branch already stored agent CCAs in `_applicationsByAuthorityClientId` via the normal builder path — but it did so through a fundamentally different mechanism:

| Aspect | Master (original) | This refactor |
|--------|-------------------|---------------|
| **CCA key differentiation** | `GetMergedOptions()` creates a new `MergedOptions` with the agent's `ClientId`, which naturally produces a different `GetApplicationKey()` | Explicit `:agent:{agentAppId}` suffix appended to the blueprint's key |
| **Configuration completeness** | `GetMergedOptions()` creates a bare `new MergedOptions()` with only ~7 fields copied (missing ClientName, ClientVersion, LogLevel, EnablePiiLogging, EnableCacheSynchronization, etc.) | `UpdateConfidentialClientApplicationOptionsFromMergedOptions` copies all fields from the fully-populated blueprint `MergedOptions` |
| **Token flow** | Piggybacked on ROPC (`AcquireTokenByUsernamePassword` with `password = "password"`) | Native MSAL UserFIC API (`AcquireTokenByUserFederatedIdentityCredential`) |
| **Leg 1 (FMI token)** | `OidcIdpSignedAssertionProvider` → re-enters `TokenAcquisition` via `ITokenAcquirerFactory` | Direct assertion callback → blueprint CCA's `AcquireTokenForClient().WithFmiPath()` |
| **Silent cache lookup** | Depended on `ClaimsPrincipal.GetMsalAccountId()` (typically null in agentic scenarios — **this was the reported bug**) | `_agentUserFicAccountIds` dictionary tracks MSAL account identifiers independently |
| **DOS protection** | None | Threshold eviction clears only `:agent:` entries, preserving the blueprint CCA |

In short, the master design used the shared dictionary but with **incomplete configuration cloning** and a **semantically wrong token flow** (ROPC). This refactor restores the shared-dictionary approach while fixing both the configuration completeness (full options cloning) and the token flow (native UserFIC API).

## Risks and unknowns

1. **Key format coupling**: The `:agent:` sentinel in cache keys is a string convention. If `GetApplicationKey()` ever produces keys containing `:agent:` for non-agent scenarios, the eviction logic would incorrectly clear those entries. This is unlikely given the current key format (`authority + clientId + azureRegion + credentialId`), but worth noting.

2. **`_agentUserFicAccountIds` remains separate**: Unlike CCA instances, account identifiers are not stored in the shared dictionary. This is intentional — they map `(agent, user, tenant)` tuples to MSAL account identifiers, which is a fundamentally different data structure from the CCA cache. There's no existing mechanism for this in the normal flow (which uses `ClaimsPrincipal` instead).

3. **Blueprint CCA built lazily**: For agent-only traffic, the blueprint CCA is not built until the assertion callback fires during Leg 1. If the assertion callback fails (e.g., credential loading error), the error surfaces inside the agent CCA's token acquisition rather than during CCA construction. This matches the old behavior where `OidcIdpSignedAssertionProvider` could also fail during callback execution.

## Test coverage

- **`AgentCca_StoredInSharedDictionary_WithAgentKeySegment`** — verifies agent and blueprint CCAs coexist in `_applicationsByAuthorityClientId` with correct key segments
- **`AgentCcaEviction_DoesNotAffect_BlueprintCca`** — verifies selective eviction preserves blueprint entries
- All existing agent tests (UPN/OID acquisition, silent cache, shared cache survival, eviction threshold) pass unchanged
- Full test suite (992-1020 tests across TFMs) passes with no regressions

### DIFF
--- a/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
+++ b/src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
@@ -58,29 +58,17 @@ namespace Microsoft.Identity.Web
         ///  Important: call GetOrBuildConfidentialClientApplication instead of accessing _applicationsByAuthorityClientId directly.
         ///  Write access to this dictionary is synchronized.
         /// </summary>
-        private readonly ConcurrentDictionary<string, IConfidentialClientApplication?> _applicationsByAuthorityClientId = new();
+        internal readonly ConcurrentDictionary<string, IConfidentialClientApplication?> _applicationsByAuthorityClientId = new();
         private readonly ConcurrentDictionary<string, SemaphoreSlim> _appSemaphores = new();
 
         /// <summary>
-        /// Maximum number of agent CCA instances to keep in the dictionary before
-        /// clearing it as a DOS protection measure. Tokens are stored in MSAL's
-        /// shared static cache, so clearing the dictionary only discards lightweight
+        /// Maximum number of agent CCA instances to keep in the shared application dictionary
+        /// before clearing agent entries as a DOS protection measure. Tokens are stored in
+        /// MSAL's shared static cache, so clearing agent entries only discards lightweight
         /// CCA objects — tokens remain accessible to newly-built CCAs.
+        /// Agent entries are identified by the ":agent:" segment in their cache key.
         /// </summary>
         internal int AgentCcaMaxCount { get; set; } = 10000;
-
-        /// <summary>
-        /// Caches agent CCAs for the native User FIC flow. Each agent CCA uses an assertion
-        /// callback that chains to the blueprint CCA for Leg 1 (FMI token acquisition).
-        /// Key format: "{agentAppId}:{authenticationScheme}".
-        /// The authenticationScheme is included because different schemes may resolve to
-        /// different blueprint credentials (e.g. certificates), and the agent CCA's assertion
-        /// callback captures the scheme to chain to the correct blueprint CCA.
-        /// When the dictionary exceeds <see cref="AgentCcaMaxCount"/>, it is cleared entirely;
-        /// tokens survive in MSAL's shared static cache and are found by new CCAs via silent calls.
-        /// </summary>
-        internal readonly ConcurrentDictionary<string, IConfidentialClientApplication> _agentUserFicCcas = new();
-        private readonly ConcurrentDictionary<string, SemaphoreSlim> _agentCcaSemaphores = new();
 
         /// <summary>
         /// Maps (agentAppId, user identifier, tenantId) tuples to MSAL account identifiers for the
@@ -341,13 +329,25 @@ namespace Microsoft.Identity.Web
             MergedOptions mergedOptions = GetMergedOptions(authenticationScheme, tokenAcquisitionOptions);
             user ??= await _tokenAcquisitionHost.GetAuthenticatedUserAsync(user).ConfigureAwait(false);
 
-            var application = await GetOrBuildConfidentialClientApplicationAsync(mergedOptions, isTokenBinding: false);
-
             if (tokenAcquisitionOptions is not null)
             {
                 tokenAcquisitionOptions.ExtraParameters ??= new Dictionary<string, object>();
                 tokenAcquisitionOptions.ExtraParameters[Constants.ExtensionOptionsServiceProviderKey] = _serviceProvider;
             }
+
+            // Detect agentic User FIC flow early — before building the blueprint CCA.
+            // Agent CCAs are built via the unified builder path with their own ClientId,
+            // so the blueprint CCA is only built lazily (inside the assertion callback)
+            // when actually needed for Leg 1 token acquisition.
+            var agentResult = await TryGetAuthenticationResultForAgentUserFicAsync(
+                tenantId, scopes, mergedOptions, tokenAcquisitionOptions).ConfigureAwait(false);
+            if (agentResult is not null)
+            {
+                LogAuthResult(agentResult);
+                return agentResult;
+            }
+
+            var application = await GetOrBuildConfidentialClientApplicationAsync(mergedOptions, isTokenBinding: false);
 
             CredentialSourceLoaderParameters loaderParameters = new CredentialSourceLoaderParameters(application.AppConfig.ClientId, application.Authority)
             {
@@ -442,15 +442,6 @@ namespace Microsoft.Identity.Web
             MergedOptions mergedOptions,
             TokenAcquisitionOptions? tokenAcquisitionOptions)
         {
-            // Handle agentic User FIC flow (both UPN and OID) using MSAL's native UserFIC API.
-            // This bypasses the ROPC piggybacking approach and provides proper cache behavior.
-            var agentUserFicResult = await TryGetAuthenticationResultForAgentUserFicAsync(
-                application, tenantId, scopes, mergedOptions, tokenAcquisitionOptions).ConfigureAwait(false);
-            if (agentUserFicResult is not null)
-            {
-                return agentUserFicResult;
-            }
-
             string? username = null;
             string? password = null;
 
@@ -579,7 +570,6 @@ namespace Microsoft.Identity.Web
         /// <returns>An <see cref="AuthenticationResult"/> if this is an agentic User FIC flow
         /// (UPN or OID); <c>null</c> if not an agentic flow (regular ROPC).</returns>
         private async Task<AuthenticationResult?> TryGetAuthenticationResultForAgentUserFicAsync(
-            IConfidentialClientApplication application,
             string? tenantId,
             IEnumerable<string> scopes,
             MergedOptions mergedOptions,
@@ -628,7 +618,7 @@ namespace Microsoft.Identity.Web
             Logger.AgentUserFicFlowDetected(_logger, agentAppId!, identifierType);
 
             var agentCca = await GetOrBuildAgentUserFicCcaAsync(
-                agentAppId!, application.Authority, authScheme).ConfigureAwait(false);
+                agentAppId!, authScheme, mergedOptions).ConfigureAwait(false);
 
             bool forceRefresh = tokenAcquisitionOptions?.ForceRefresh ?? false;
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -1350,9 +1302,17 @@ public async Task<IConfidentialClientApplication> GetConfidentialClientApplicati
...
             MergedOptions mergedOptions,
-            bool isTokenBinding)
+            bool isTokenBinding,
+            string? agentAppId = null,
+            Func<AssertionRequestOptions, Task<string>>? clientAssertionProvider = null)
         {
+            // For agent CCAs, incorporate the agent app ID into the cache key so each agent
+            // gets its own CCA instance while sharing the same dictionary and lifecycle.
             string key = GetApplicationKey(mergedOptions, isTokenBinding);
+            if (agentAppId is not null)
```
bgavrilMS: This should be part of GetApplicationKey

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -1370,11 +1330,37 @@ public async Task<IConfidentialClientApplication> GetConfidentialClientApplicati
...
+                    key = $"{key}:agent:{agentAppId}";
+                }
                 _applicationsByAuthorityClientId[key] = newApp;
+
+                // DOS protection for agent CCAs: if too many agent entries accumulate,
+                // clear them. Tokens survive in MSAL's shared static cache and will be
+                // found by new CCAs via AcquireTokenSilent.
+                if (agentAppId is not null)
+                {
+                    int agentCount = _applicationsByAuthorityClientId.Keys.Count(k => k.IndexOf(":agent:", StringComparison.Ordinal) >= 0);
```
bgavrilMS: This is O(n) operation, it can be quite expensive. How about just clearing the entire dictionary if it reaches 10000. Tokens are cached anyway.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -1350,9 +1302,17 @@ public async Task<IConfidentialClientApplication> GetConfidentialClientApplicati
 
         internal /* for testing */ async Task<IConfidentialClientApplication> GetOrBuildConfidentialClientApplicationAsync(
             MergedOptions mergedOptions,
-            bool isTokenBinding)
+            bool isTokenBinding,
+            string? agentAppId = null,
+            Func<AssertionRequestOptions, Task<string>>? clientAssertionProvider = null)
```
bgavrilMS: This `clientAssertionProvider ` is not part of the app key. I think it only makes sense for this callback to exist when an agentAppid is also defined (which implicitly defines the key). In other words, we expect the CCA for 1 agentAppId to map to a single `clientAssertionProvider`. In other words - agentAppId and clientAssertionProvider need to both be NULL or both be defined.

bgavrilMS: Another way of doing this would be have `GetOrBuildConfidentialClientApplicationAsync(mergedOptions, isTokenBiding, (agentAppId, blueprintOptions)` and for GetOrBuildConfidentialClientApplicationAsync to create the FUNC on the fly. This allows it you to have a key that includes details of `blueprintOptions` as well. 

But it can be overkill.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -724,97 +714,59 @@ private async Task<AuthenticationResult> GetAuthenticationResultForUserInternalA
...
-            var semaphore = _agentCcaSemaphores.GetOrAdd(ccaCacheKey, _ => new SemaphoreSlim(1, 1));
-            await semaphore.WaitAsync().ConfigureAwait(false);
-            try
-            {
-                if (_agentUserFicCcas.TryGetValue(ccaCacheKey, out var entry))
+            // Build the assertion callback that chains to the blueprint CCA for Leg 1.
+            // Capture authenticationScheme so the callback resolves the correct blueprint.
+            string? capturedAuthScheme = authenticationScheme;
+
+            Func<AssertionRequestOptions, Task<string>> assertionCallback = async (AssertionRequestOptions options) =>
```
bgavrilMS, line 760: You're calling this on every token acquisition. Isn't this expensive? The callback captures external variables etc.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
Avery-Dunn: `GetApplicationKey` now accepts an optional `agentAppId` parameter and appends the `:agent:{agentAppId}`suffix internally. The manual suffix concatenation in `GetOrBuildConfidentialClientApplicationAsync` has been removed

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
Avery-Dunn: Replaced the O(n) agent-only scan with a simpler `_applicationsByAuthorityClientId.Clear()`.

This should be safe because `_applicationsByAuthorityClientId` should only be populated by CCA instances that use the shared cache.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -1386,10 +1372,23 @@ public async Task<IConfidentialClientApplication> GetConfidentialClientApplicati
...
+        /// when <paramref name="agentAppId"/> is non-null.</param>
         private async Task<IConfidentialClientApplication> BuildConfidentialClientApplicationAsync(
             MergedOptions mergedOptions,
-            bool isTokenBinding)
+            bool isTokenBinding,
+            string? agentAppId = null,
+            Func<AssertionRequestOptions, Task<string>>? clientAssertionProvider = null)
         {
+            bool isAgentCca = agentAppId is not null;
+
```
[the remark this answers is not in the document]
Avery-Dunn: Added an `ArgumentException` guard at the top of `BuildConfidentialClientApplicationAsync` that throws if exactly one of `agentAppId`/`clientAssertionProvider` is null.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
Avery-Dunn: Added an `ArgumentException` guard at the top of `BuildConfidentialClientApplicationAsync` that throws if exactly one of `agentAppId`/`clientAssertionProvider` is null. 

The overload with a tuple shouldn't be necessary, the guard should adequately enforce them being paired.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
Avery-Dunn: Added an early cache-hit check in `GetOrBuildAgentUserFicCcaAsync` that returns the cached CCA before constructing the closure.

Some quick performance testing showed the closure allocation itself was pretty negligible compared to the total size of the call, but the early return avoids entering the async state machine in `GetOrBuildConfidentialClientApplicationAsync`, yielding ~15% time improvement on cache-hit calls compared to the original design (~105 µs to ~88 µs)

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -276,13 +266,24 @@ private async Task<AcquireTokenResult> AddAccountToCacheFromAuthorizationCodeInt
...
-        private static string GetApplicationKey(MergedOptions mergedOptions, bool isTokenBinding)
+        /// <param name="agentAppId">When non-null, appends an agent-specific segment so each
+        /// agent CCA gets its own entry in the shared dictionary.</param>
+        /// <returns>Concatenated string of authority, client id, azure region, credential id,
+        /// token-binding flag, and optional agent app id.</returns>
+        private static string GetApplicationKey(MergedOptions mergedOptions, bool isTokenBinding, string? agentAppId = null)
         {
             string credentialId = string.Join("-", mergedOptions.ClientCredentials?.Select(c => c.Id) ?? Enumerable.Empty<string>());
 
             string baseKey = DefaultTokenAcquirerFactoryImplementation.GetKey(mergedOptions.Authority, mergedOptions.ClientId, mergedOptions.AzureRegion) + credentialId;
```
bgavrilMS: You can further optimize allocations here by using a StringBuilder

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -1350,9 +1325,11 @@ public async Task<IConfidentialClientApplication> GetConfidentialClientApplicati
 
         internal /* for testing */ async Task<IConfidentialClientApplication> GetOrBuildConfidentialClientApplicationAsync(
             MergedOptions mergedOptions,
-            bool isTokenBinding)
+            bool isTokenBinding,
+            string? agentAppId = null,
+            Func<AssertionRequestOptions, Task<string>>? clientAssertionProvider = null)
```
bgavrilMS: rename to `agenticAssertionProvider` to make it clear this is not a general purpose assertion provider.

bgavrilMS (approved):
Approved with comments.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
Avery-Dunn: Renamed in latest commit

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
Avery-Dunn: Good point, adjusted in latest commit.

PR #3935, opened 2026-07-09
Restore independent PR pipeline + pool-aware MI identity + net462/472 unit tests

iarekk:
Re-lands the pipeline work reverted in #3936 (#3933 + #3934), plus fixes the managed-identity (MI) E2E identity so the test passes on **every** pipeline that runs it.

## 1. Restore reverted work
- 4-stage PR pipeline (Build / Integration / E2E / Managed Identity E2E), MI stage on the `MSALMSIV2` pool.
- Extra unit-test assemblies in the ADO unit glob + Sidecar E2E/integration exclusion filter.

## 2. Pool-aware MI identity (the reason for the revert)
The MI E2E test hardcoded a UAMI client id. Each build pool has a **different** identity assigned, so a single constant can't work everywhere — the DevEx `id4s-official` pipeline (Wilson pool) failed with *"Identity not found"* when master carried the `msiv2uami` id.

Fix: the test reads the id from `IDWEB_MI_UAMI_CLIENTID` and falls back to the Wilson-pool identity:
- **Default (fallback)** = `45344e7d` (`Msal_Integration_tests`, Wilson pool) → `id4s-official` needs **no changes**.
- **Override** = `6325cd32` (`msiv2uami`, `MSALMSIV2` pool), set in `template-run-managed-identity-e2e-tests.yaml` → used by the Id.Web PR MI stage and the IdWeb OneBranch official MI job (both consume that template).

## 3. .NET Framework unit tests
VSTest@2 can't discover the net462/net472 xUnit tests on hosted agents, so add `dotnet test -f net472` steps mirroring the GitHub Action. Not running net462

### DIFF
--- a/build/template-test-unit.yaml
+++ b/build/template-test-unit.yaml
@@ -1,0 +1,24 @@
+# template-test-unit.yaml
+# Run the Microsoft.Identity.Web unit tests.
+
+steps:
+
+- task: VSTest@2
+  displayName: 'Run unit tests'
+  inputs:
+    testSelector: 'testAssemblies'
+    testAssemblyVer2: |
+        tests\Microsoft.Identity.Web.Test\bin\**\Microsoft.Identity.Web.Test.dll
+        tests\Microsoft.Identity.Web.UI.Test\bin\**\Microsoft.Identity.Web.UI.Test.dll
+        tests\E2E Tests\Sidecar.Tests\bin\**\Sidecar.Tests.dll
+        tests\E2E Tests\OidcIdPSignedAssertionProviderTests\bin\**\OidcIdpSignedAssertionProviderTests.dll
+        tests\E2E Tests\CustomSignedAssertionProviderTests\bin\**\CustomSignedAssertionProviderTests.dll
+    searchFolder: '$(System.DefaultWorkingDirectory)'
+    # Match the GitHub Action: Sidecar.Tests ships E2E/integration classes that
+    # need a live sidecar and are excluded there. Skip the same ones here.
+    testFiltercriteria: 'FullyQualifiedName!~SidecarEndpointsE2ETests&FullyQualifiedName!~SidecarIntegrationTests'
+    runInParallel: true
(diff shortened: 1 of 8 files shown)


gladjohn (approved):

gladjohn:
BLOCKING MERGE

build/template-test-unit.yaml
```diff
@@ -22,3 +22,24 @@ steps:
     failOnMinTestsNotRun: true
     minimumExpectedTests: '1'
     runSettingsFile: 'build\CodeCoverage.runsettings'
+
+# VSTest@2 on the hosted agent cannot discover the .NET Framework xUnit tests
```
gladjohn, line 26: block comment

iarekk: should be good now, with the MI tests being aware of the pool

build/template-test-unit.yaml
```diff
@@ -0,0 +1,45 @@
...
+    minimumExpectedTests: '1'
+    runSettingsFile: 'build\CodeCoverage.runsettings'
+
+# VSTest@2 on the hosted agent cannot discover the .NET Framework xUnit tests
+# (net462/net472), so the glob above silently runs 0 of them. The GitHub Action
+# covers them via 'dotnet test -f net462/net472'; mirror that here. Only the main
+# unit test project multi-targets .NET Framework. --no-build reuses the solution
+# build output produced earlier in the stage.
+- task: DotNetCoreCLI@2
+  displayName: 'Run unit tests (.NET Framework 4.6.2)'
```
bgavrilMS: How long do these take? I am not sure about this. Does Identity.Web even have codepaths specific to net462 ? 

Maybe we should skip 462. This is legacy + limited support. In other words, if a net462 specific bug would be found, we would probably not fix

iarekk: @bgavrilMS extra 18 seconds for ne462:

image

gladjohn (approved):

PR #3942, opened 2026-07-13
Add OnBeforeAuthHeaderCreation / OnAfterAuthHeaderCreation hooks to DownstreamApi

neha-bhargava:
Targets **4.13.2**. Builds on the regression fix from #3943 (already merged).

## What

Adds explicit request-customization hooks to `DownstreamApi`, consuming the new `AuthorizationHeaderProviderOptions` members from `Microsoft.Identity.Abstractions` 12.5.0 (AzureAD/microsoft-identity-abstractions-for-dotnet#262):

- **`OnBeforeAuthHeaderCreation`** — runs before the header is created/signed. Use it to shape the request that a request-binding protocol signs (SignedHttpRequest `q`/`h`/`b`), so the signature covers the finalized request.
- **`OnAfterAuthHeaderCreation`** — runs after the header is attached. The pre-existing **`CustomizeHttpRequestMessage` is still invoked** at the same point for backwards compatibility (both run; distinct hooks).

Both hooks are propagated through `MicrosoftIdentityMessageHandler`.

## Also

`BaseAuthorizationHeaderProvider` and `DefaultAuthorizationHeaderProvider` declare `IAuthorizationHeaderProvider2` only (it already extends `IAuthorizationHeaderProvider`), and the Base provider's delegate field is typed `IAuthorizationHeaderProvider2` to avoid upcasting.

## Tests

- Flow test shapes the flowed request via `OnBeforeAuthHeaderCreation` (asserting it reaches the provider before signing).
- Lifecycle test: `OnBefore` sees no header (runs before); `OnAfter` and `CustomizeHttpRequestMessage` both observe it.
- **`with SHR` / `without SHR`** tests asserting the `Authorization` header is created and attached in both cases.
- Full suite: **1036 passed / 4 skipped / 0 failed**.

## Depends on

AzureAD/microsoft-identity-abstractions-for-dotnet#262 (Abstractions 12.5.0) — the Abstractions package floor bump is pending its release.

### DIFF
--- a/src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
+++ b/src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
@@ -734,6 +734,10 @@ namespace Microsoft.Identity.Web
                 httpRequestMessage.RequestUri = uriBuilder.Uri;
             }
 
+            // Opportunity to shape the request before the authorization header is created and (for request-binding
+            // protocols such as SignedHttpRequest q/h/b) signed, so the signature covers the finalized request.
+            effectiveOptions.OnBeforeAuthHeaderCreation?.Invoke(httpRequestMessage);
+
             // Obtention of the authorization header (except when calling an anonymous endpoint)
             // which is done by not specifying any scopes or mTLS scheme.
             if (string.Equals(effectiveOptions.ProtocolScheme, Constants.MtlsProtocolScheme, StringComparison.OrdinalIgnoreCase))
@@ -838,9 +842,10 @@ namespace Microsoft.Identity.Web
                 Logger.UnauthenticatedApiCall(_logger, null);
             }
 
-            // Opportunity to change the request message after the authorization header (including the Authorization
-            // header) has been set, just before the request is sent, per CustomizeHttpRequestMessage's documented
-            // contract. #3902 had moved this before header creation, which regressed callers that read the header.
+            // Observe or adjust the finalized request once the authorization header (including the Authorization
+            // header) has been set, just before it is sent. Fire the new OnAfterAuthHeaderCreation hook and the
+            // pre-existing CustomizeHttpRequestMessage (retained for backwards compatibility); both share this timing.
+            effectiveOptions.OnAfterAuthHeaderCreation?.Invoke(httpRequestMessage);
             effectiveOptions.CustomizeHttpRequestMessage?.Invoke(httpRequestMessage);
 
             return (authorizationHeaderInformation, credential);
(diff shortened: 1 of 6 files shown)


src/Microsoft.Identity.Web.TokenAcquisition/BaseAuthorizationHeaderProvider.cs
```diff
@@ -36,7 +36,7 @@ public BaseAuthorizationHeaderProvider(IServiceProvider serviceProvider)
             _headerProvider = new DefaultAuthorizationHeaderProvider(_tokenAcquisition);
         }
 
-        private readonly DefaultAuthorizationHeaderProvider _headerProvider;
+        private readonly IAuthorizationHeaderProvider _headerProvider;
```
bgavrilMS: Should this be `IAuthorizationHeaderProvider2` to avoid upcasting?

bgavrilMS (approved):

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
```diff
@@ -838,10 +842,10 @@ public Task<HttpResponseMessage> CallApiForAppAsync(
...
             }
 
-            // Opportunity to change the request message after the authorization header (including the Authorization
-            // header) has been set, just before the request is sent, per CustomizeHttpRequestMessage's documented
-            // contract. #3902 had moved this before header creation, which regressed callers that read the header.
-            effectiveOptions.CustomizeHttpRequestMessage?.Invoke(httpRequestMessage);
+            // Observe or adjust the finalized request once the authorization header (including the Authorization
+            // header) has been set, just before it is sent. OnAfterAuthHeaderCreation aliases the pre-existing
+            // CustomizeHttpRequestMessage (same underlying delegate), so this single invocation honors both.
+            effectiveOptions.OnAfterAuthHeaderCreation?.Invoke(httpRequestMessage);
```
bgavrilMS, line 848: Should you still fire `CustomizeHttpRequestMessage` for backwards compat reasons?

bgavrilMS (dismissed):

src/Microsoft.Identity.Web.TokenAcquisition/BaseAuthorizationHeaderProvider.cs
neha-bhargava: Good call - done. Typed _headerProvider as IAuthorizationHeaderProvider2 and dropped the casts at the three metadata delegation sites.

src/Microsoft.Identity.Web.DownstreamApi/DownstreamApi.cs
neha-bhargava: Yes - fixed.

bgavrilMS (approved):

gladjohn (approved):

trwalke (approved):

Issue #3961, opened 2026-07-20
AcquireTokenForClient becomes extremely slow with 164K tokens in one app-token cache partition


ararbi90:
### Microsoft.Identity.Web Library

Microsoft.Identity.Web

### Microsoft.Identity.Web version

Microsoft.Identity.Client: 4.74.1

### Web app

Sign-in users and call web APIs

### Web API

Protected web APIs (validating tokens)

### Token cache serialization

In-memory caches

### Description

Description

We operate a token broker that uses  Microsoft.Identity.Web  and MSAL to acquire application tokens for many downstream audiences. Requests commonly share the same broker client and tenant, creating a high-cardinality workload within one app-token cache partition.

During an outage, MSAL reported:

[FindAccessTokenAsync] Discovered 164663 access tokens in cache using partition key:
{clientId}_{tenantId}_AppTokenCache

For the same operation:

DurationInHttpInMs: 977
DurationTotalInMs: 1980434

The identity-provider request completed in approximately one second, but the overall token acquisition took approximately 33 minutes. Inbound token validation also completed quickly.

Our current hypothesis is that the delay occurred during local MSAL cache processing—such as partition lookup, filtering, overlap removal, or cache update—rather than during inbound validation or the token endpoint request.

Workload

The broker calls:

ITokenAcquirer.GetTokenForAppAsync(scope)

This delegates to MSAL  AcquireTokenForClient . The same broker client and tenant acquire tokens for many downstream scopes or audiences, allowing many access-token entries to accumulate under one app-token partition.

Versions observed during the incident:

Microsoft.Identity.Web: 4.7.0
Microsoft.Identity.Client: 4.74.1

### Reproduction steps

If you have  a large partition that will eventually happen

### Error message

There is not error message the call eventually is processed but it takes 30 minutes to request a token.

### Relevant code snippets

```csharp
Relevant code paths

ITokenAcquirer.GetTokenForAppAsync
→ MSAL AcquireTokenForClient
→ ClientCredentialRequest.GetCachedAccessTokenAsync
→ TokenCache.FindAccessTokenAsync
→ Accessor.GetAllAccessTokens(partitionKey)
→ filter tokens by scope and request properties

When refreshing a token:

TokenCache.SaveTokenResponseAsync
→ cache synchronization
→ DeleteAccessTokensWithIntersectingScopes
→ Accessor.SaveAccessToken

Key references:

• Microsoft.Identity.Web app-token acquisition
[...]

PR #3970, opened 2026-07-23
Don't short-circuit the in-memory token cache serialization provider

gladjohn:
## Summary

When an app registers `AddInMemoryTokenCache` / `AddInMemoryTokenCaches`, `TokenAcquisition` **bypasses the registered `IMemoryCache` serialization provider** and instead switches MSAL to its opaque **static (process-wide) cache**:

```csharp
if (_tokenCacheProvider is MsalMemoryTokenCacheProvider)
    builder.WithCacheOptions(CacheOptions.EnableSharedCacheOptions);   // static cache
...
if (!(_tokenCacheProvider is MsalMemoryTokenCacheProvider))           // Initialize() skipped
{ _tokenCacheProvider.Initialize(app.AppTokenCache); _tokenCacheProvider.Initialize(app.UserTokenCache); }
```

`Initialize()` is what wires MSAL's serialization callbacks, so skipping it for the in-memory provider means:

- the registered `IMemoryCache` is **never used** (tokens live in MSAL's static dictionary),
- `GetSuggestedCacheKey` (the cache-key partitioning hook) is a **silent no-op**,
- `MsalMemoryTokenCacheOptions.AbsoluteExpirationRelativeToNow` and the memory-cache **`SizeLimit` are silently ignored**,
- the cache is **unbounded with no eviction**.

This is the IdWeb-side driver behind very large single-partition app-token caches for high-cardinality / token-broker workloads (one `{clientId}_{tenantId}` app-token partition observed at **164,663** tokens; `AcquireTokenForClient` becomes O(N) over the partition, with saves serializing under MSAL's cache semaphore). Related: #3961.

## Change

Initialize the token cache provider for **all** cache types (so the in-memory provider's serialization callbacks — including `GetSuggestedCacheKey` and per-entry expiry — actually run), and enable MSAL's shared/static cache only when it is actually required:

```csharp
bool usesSharedInternalCache =
    isAgentCca ||
    (mergedOptions.UseFastUnboundedCache && _tokenCacheProvider is MsalMemoryTokenCacheProvider);
if (usesSharedInternalCache)
    builder.WithCacheOptions(CacheOptions.EnableSharedCacheOptions);
...
if (!usesSharedInternalCache)
{
    _tokenCacheProvider.Initialize(app.AppTokenCache);
    _tokenCacheProvider.Initialize(app.UserTokenCache);
}
```

Agent CCAs continue to use the shared/static cache (MSAL forbids combining the internal shared cache with external serialization). Now `GetSuggestedCacheKey`, entry TTL, and size limits actually work.

## Opt-in (config-driven)

A new **`MicrosoftIdentityOptions.UseFastUnboundedCache`** flag restores the previous fast, process-wide static cache. It is wired the same way as `EnableCacheSynchronization` — a config-bound option that flows into `MergedOptions` (`UpdateMergedOptionsFromMicrosoftIdentityOptions`) and is read in `TokenAcquisition`. So it is settable from **configuration**, not just code:

```jsonc
// appsettings.json
"AzureAd": {
  // ...
  "UseFastUnboundedCache": true
}
```

- Default is `false` (the new, bounded, serialized behavior).
- `true` = MSAL static/shared cache: fast, **unbounded** (no eviction), ignores expiry/size options and the partitioning hook.

## Behavior change (minor)

Consumers who set `AbsoluteExpirationRelativeToNow` / `SizeLimit` on the in-memory cache previously had them **ignored**; they now **take effect** (tokens may expire/evict where they previously persisted for the process lifetime). Opt out via `UseFastUnboundedCache` if the old semantics are required.

## Files

- `src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs` — un-short-circuit + `UseFastUnboundedCache` gate
- `src/Microsoft.Identity.Web.TokenAcquisition/MicrosoftIdentityOptions.cs` — new `UseFastUnboundedCache` (config-bindable)
- `src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs` — flow it into `MergedOptions`
- `src/Microsoft.Identity.Web.TokenAcquisition/PublicAPI/{NetCore,NetFramework}/PublicAPI.Unshipped.txt`
- Tests updated accordingly

## Testing

Unit suite green across TFMs (local):

| Suite / TFM | Result |
|---|---|
| Web.Test net462 / net472 | 44 / 44 |
| Web.Test net8.0 | 1031 passed, 0 failed, 11 skipped |
| Web.Test net9.0 | 1040 / 0 / 4 |
| Web.Test net10.0 | 1059 / 0 / 4 |
| UI.Test net8.0 / net9.0 | 46 / 46 |

Two behavior tests assert: (1) an app token acquired through the default in-memory cache is written to the backing `IMemoryCache` (serialization provider wired), and (2) with `UseFastUnboundedCache = true` nothing is written (static cache used). An initial naive "always initialize" version failed 9 `TokenAcquisitionTests.Agent*` tests with `MsalClientException: "You configured MSAL cache serialization at the same time with internal caching options. These are mutually exclusive."`, which drove the `usesSharedInternalCache` gate. Integration/E2E were not run (require lab credentials).

## Open item

`UseFastUnboundedCache` on `MicrosoftIdentityOptions` flows into `MergedOptions` in every host (`DefaultTokenAcquisitionHost.GetOptions` resolves the `MicrosoftIdentityOptions` merger), so it works whenever the value is set. Config-binding coverage depends on the entrypoint: `TokenAcquirerFactory.GetDefaultInstance("AzureAd")` binds the section only to `MicrosoftIdentityApplicationOptions`. Confirming the MISE-native config entrypoint is a follow-up (if it binds only `MicrosoftIdentityApplicationOptions`, the flag would also need to live in `Microsoft.Identity.Abstractions`).

---

🤖 Prototyped during an investigation with GitHub Copilot.

### DIFF
--- a/src/Microsoft.Identity.Web.TokenCache/InMemory/MsalMemoryTokenCacheOptions.cs
+++ b/src/Microsoft.Identity.Web.TokenCache/InMemory/MsalMemoryTokenCacheOptions.cs
@@ -32,5 +32,27 @@ namespace Microsoft.Identity.Web.TokenCacheProviders.InMemory
             get;
             set;
         }
+
+        /// <summary>
+        /// Gets or sets a value indicating whether the in-memory token cache should use MSAL's
+        /// internal shared (static) cache instead of the <see cref="Microsoft.Extensions.Caching.Memory.IMemoryCache"/>
+        /// serialization cache.
+        /// <para>
+        /// Defaults to <see langword="false"/>. When <see langword="false"/> (recommended), tokens are stored via
+        /// the <c>IMemoryCache</c> serialization provider, so the cache honors <see cref="AbsoluteExpirationRelativeToNow"/>,
+        /// the memory cache size limit, and per-entry expiry, and supports cache-key partitioning through
+        /// <c>MsalMemoryTokenCacheProvider.GetSuggestedCacheKey</c>.
+        /// </para>
+        /// <para>
+        /// Set to <see langword="true"/> only to restore the legacy behavior, where Microsoft.Identity.Web
+        /// enabled MSAL's opaque static cache and bypassed the serialization provider. In that mode the
+        /// expiry/size options above and the partitioning hook have no effect, and the cache is unbounded.
+        /// </para>
+        /// </summary>
+        public bool UseSharedCache
+        {
+            get;
+            set;
+        }
     }
 }
--- a/src/Microsoft.Identity.Web.TokenCache/InMemory/MsalMemoryTokenCacheOptions.cs
+++ b/src/Microsoft.Identity.Web.TokenCache/InMemory/MsalMemoryTokenCacheOptions.cs
@@ -32,27 +32,5 @@ namespace Microsoft.Identity.Web.TokenCacheProviders.InMemory
             get;
             set;
         }
-
-        /// <summary>
-        /// Gets or sets a value indicating whether the in-memory token cache should use MSAL's
-        /// internal shared (static) cache instead of the <see cref="Microsoft.Extensions.Caching.Memory.IMemoryCache"/>
-        /// serialization cache.
-        /// <para>
-        /// Defaults to <see langword="false"/>. When <see langword="false"/> (recommended), tokens are stored via
-        /// the <c>IMemoryCache</c> serialization provider, so the cache honors <see cref="AbsoluteExpirationRelativeToNow"/>,
-        /// the memory cache size limit, and per-entry expiry, and supports cache-key partitioning through
-        /// <c>MsalMemoryTokenCacheProvider.GetSuggestedCacheKey</c>.
-        /// </para>
-        /// <para>
-        /// Set to <see langword="true"/> only to restore the legacy behavior, where Microsoft.Identity.Web
-        /// enabled MSAL's opaque static cache and bypassed the serialization provider. In that mode the
-        /// expiry/size options above and the partitioning hook have no effect, and the cache is unbounded.
-        /// </para>
-        /// </summary>
-        public bool UseSharedCache
-        {
-            get;
-            set;
-        }
     }
 }
--- a/src/Microsoft.Identity.Web.TokenCache/InMemory/MsalMemoryTokenCacheProvider.cs
+++ b/src/Microsoft.Identity.Web.TokenCache/InMemory/MsalMemoryTokenCacheProvider.cs
@@ -40,6 +40,14 @@ namespace Microsoft.Identity.Web.TokenCacheProviders.InMemory
         }
 
         /// <summary>
+        /// Gets a value indicating whether this provider was configured to use MSAL's internal shared
+        /// (static) cache instead of the <see cref="IMemoryCache"/> serialization cache. Surfaces
+        /// <see cref="MsalMemoryTokenCacheOptions.UseSharedCache"/> so the token acquisition wiring can
+        /// decide whether to enable the shared cache and skip serialization initialization.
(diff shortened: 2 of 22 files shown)


src/Microsoft.Identity.Web.TokenCache/InMemory/MsalMemoryTokenCacheOptions.cs
```diff
@@ -32,5 +32,27 @@ public TimeSpan AbsoluteExpirationRelativeToNow
...
+        /// the memory cache size limit, and per-entry expiry, and supports cache-key partitioning through
+        /// <c>MsalMemoryTokenCacheProvider.GetSuggestedCacheKey</c>.
+        /// </para>
+        /// <para>
+        /// Set to <see langword="true"/> only to restore the legacy behavior, where Microsoft.Identity.Web
+        /// enabled MSAL's opaque static cache and bypassed the serialization provider. In that mode the
+        /// expiry/size options above and the partitioning hook have no effect, and the cache is unbounded.
+        /// </para>
+        /// </summary>
+        public bool UseSharedCache
```
bgavrilMS: Can ppl set this via config?

bgavrilMS (approved):

src/Microsoft.Identity.Web.TokenCache/InMemory/MsalMemoryTokenCacheOptions.cs
gladjohn: Not from appsettings.json out of the box.

bgavrilMS: 1. I would rename it to "UseFastUnboundedCache" or smth like that
2. Can we add it as config? I am concerned about MISE native.

src/Microsoft.Identity.Web.TokenCache/InMemory/MsalMemoryTokenCacheProvider.cs
```diff
@@ -39,6 +39,14 @@ public MsalMemoryTokenCacheProvider(
             _cacheOptions = cacheOptions.Value;
         }
 
+        /// <summary>
+        /// Gets a value indicating whether this provider was configured to use MSAL's internal shared
+        /// (static) cache instead of the <see cref="IMemoryCache"/> serialization cache. Surfaces
+        /// <see cref="MsalMemoryTokenCacheOptions.UseSharedCache"/> so the token acquisition wiring can
+        /// decide whether to enable the shared cache and skip serialization initialization.
+        /// </summary>
+        public bool UseSharedCache => _cacheOptions.UseSharedCache;
```
bgavrilMS: This is not a property of the MsalMemoryCacheProvider, it's more a general purpose property like https://github.com/AzureAD/microsoft-identity-web/blob/204830a0029fea0958b1b87c7a2925dbffee55b4/src/Microsoft.Identity.Web.TokenAcquisition/MergedOptions.cs?plain=1#L65C61-L65C61

bgavrilMS (changes requested):

src/Microsoft.Identity.Web.TokenCache/InMemory/MsalMemoryTokenCacheOptions.cs
gladjohn: added as a config

bgavrilMS (approved):

PR #3991, opened 2026-07-31
Source home account identifier (uid/utid) claims from the redeemed token

soodt:
# Source home account identifier (uid/utid) claims from the redeemed token

## Description

The web app sign-in adapters (ASP.NET Core and OWIN) now set the uid/utid claims — the
MSAL home account identifier and token-cache key — from the redeemed token's
Account.HomeAccountId instead of the client_info parameter on the sign-in response, so the
claims stay consistent with the account that was redeemed and cached.

- No public API change — internal only (the redeemed account is carried through to the
  sign-in events).
- B2C preserved; the existing guard for app-set uid/utid claims is retained; no claims are
  stamped when the redeemed token has no home account id.
- Standard sign-in flows are unaffected. Unit tests added for both adapters.


bgavrilMS:
@soodt - what bug are you trying to solve? This s a pretty tricky area of the code.

What testing has there been done. You need to manually test: 

- reopening from cached session
- case where access token is expired and Id.Web uses AcquireTokenSilent

src/Microsoft.Identity.Web.OWIN/AppBuilderExtension.cs
```diff
@@ -174,59 +175,8 @@ public static IAppBuilder AddMicrosoftIdentityWebApp(
...
-
-                                if (clientInfoFromServer.UniqueTenantIdentifier != null && clientInfoFromServer.UniqueObjectIdentifier != null)
-                                {
-                                    context.AuthenticationTicket.Identity.AddClaim(new Claim(ClaimConstants.UniqueTenantIdentifier, clientInfoFromServer.UniqueTenantIdentifier));
-                                    context.AuthenticationTicket.Identity.AddClaim(new Claim(ClaimConstants.UniqueObjectIdentifier, clientInfoFromServer.UniqueObjectIdentifier));
-                                }
-                            }
-                            httpContext.Session.Remove(ClaimConstants.ClientInfo);
-                        }
+                        (string? homeObjectId, string? homeTenantId) = GetAndRemoveHomeAccountIdentifier(context.OwinContext, httpContext);
```
bgavrilMS, line 178: Why do you need to do: Get & Remove and then Add ? Why not just GetOrAdd ?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -161,11 +161,21 @@ public TokenAcquisition(
 #endif
         public async Task<AcquireTokenResult> AddAccountToCacheFromAuthorizationCodeAsync(
```
bgavrilMS, line 162: Why do we need this overload?

bgavrilMS (changes requested):
It's not clear what problem is solved.

PR #3994, opened 2026-08-05
Resolve cloud-specific metadata by authority host

Avery-Dunn:
Makes Microsoft.Identity.Web resolve the correct cloud-specific Federated Identity Credential (FIC)
token-exchange audience/scope automatically from the request's authority host, instead of hardcoding the
public-cloud value. ID Web is the **translator** between MSAL's public cloud baseline and any caller- or
upstream-supplied metadata, so sovereign and private clouds work with no customer code in the common case,
while remaining fully overridable.

Related work in Abstractions and MSAL:
https://github.com/AzureAD/microsoft-identity-abstractions-for-dotnet/pull/266
https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/6104

## Motivation

FIC-based flows must present a cloud-specific token-exchange audience (e.g. `api://AzureADTokenExchangeUSGov`
for US Government) rather than the public-cloud `api://AzureADTokenExchange`. ID Web previously hardcoded the
public value across every FIC leg (agent identities, OIDC client-credentials, managed identity), so
applications in sovereign or private clouds hit opaque exchange failures with no override path.

This change centralizes the audience/scope decision in one place, keyed off the authority host, layering a
caller- or upstream-SDK metadata provider over MSAL's built-in baseline — so the right value is resolved
automatically for the clouds MSAL ships, and any cloud can be added or overridden.

### The resolver: `CloudMetadataResolution` (internal, static)

A single internal, **static** helper is the one place ID Web decides the FIC token-exchange value. Each FIC
leg holds an optional `ICloudMetadataProvider` resolved from DI and passes it in per call. Resolution layers,
highest precedence first:

1. an explicit per-call override supplied by the caller (e.g. a credential's `TokenExchangeUrl`);
2. an upstream `ICloudMetadataProvider` from DI — contributed by a caller or an upstream SDK; this is how
   internal-only sovereign clouds (which MSAL does not ship) become resolvable;
3. MSAL's built-in public baseline, read directly from `KnownCloudMetadata.Default`.

It exposes two forms so no call site hand-builds either:

- `ResolveTokenExchangeAudience(...)` — the **bare** audience, for managed-identity / resource contexts.
- `ResolveTokenExchangeScope(...)` — the **scope** (bare + `/.default`), for client-credentials / app-token
  contexts. The `/.default` computation is delegated to MSAL's `TokenExchangeScope.FromAudience`, the single
  cross-stack owner of that rule, so ID Web and MSAL can never diverge on the suffix.

Because the upstream provider and MSAL baseline expose the audience under the **same** key literal
(Abstractions' `CloudMetadataKeyNames.FederatedCredentialAudience` equals MSAL's
`CloudMetadataKeyNames.FederatedCredentialAudience`), ID Web "translates" between the two SDKs simply by
reading the same literal from whichever source resolves first — no runtime key remapping is needed, and
neither SDK depends on the other.

Being static, the resolver needs no DI registration or lifetime management: each consumer resolves
`ICloudMetadataProvider` from the container (so the winning `TryAdd` provider is honored) and hands it in.

### Fail-fast on an unknown cloud

When a **non-empty** authority host resolves to no cloud-specific value from any source, ID Web throws an
`InvalidOperationException` naming the host and pointing at the override APIs (`AddCloudMetadata(...)`, an
`ICloudMetadataProvider` registration, or an explicit `TokenExchangeUrl`) — rather than silently exchanging
against the wrong (public-cloud) audience. This "throw on an unknown cloud" behavior is deliberately chosen
over a fail-soft public-cloud fallback: switching to a fallback later would be a non-breaking relaxation,
whereas the reverse would break callers. The common **public-cloud** path is unaffected (it resolves from the
MSAL baseline), and the **managed-identity** leg — which has no AAD authority to key on — still uses the
documented public-cloud default rather than throwing.

### Configuration binding: `AddCloudMetadata(IConfiguration)` (new public API)

For non-upstream callers that need a cloud ID Web and MSAL do not ship, a new public extension registers a
provider purely from configuration — no code beyond binding an `appsettings.json` section:

```csharp
services.AddCloudMetadata(configuration.GetSection("CloudMetadata"));
```

```json
"CloudMetadata": {
  "login.microsoftonline.us": { "federated_credential_audience": "api://AzureADTokenExchangeUSGov" },
  "login.mynewcloud.example": { "federated_credential_audience": "api://AzureADTokenExchangeMyCloud" }
}
```

It binds each host→key/value section into an Abstractions `InMemoryCloudMetadataProvider` registered as
`ICloudMetadataProvider` via `TryAddSingleton` (so an explicitly registered provider — including an upstream
SDK's — wins and is left untouched). The section shape is identical to MISE's `AddMiseCloudMetadata(IConfiguration)`.

### Where resolution is applied

Every FIC leg that keys off an authority routes through the resolver:

- **OIDC client-credentials FIC** (`OidcIdpSignedAssertionProvider`) — resolves the exchange **scope** from
  the application's instance/authority host; an explicitly configured `TokenExchangeUrl` is passed as the
  per-call override. The provider is injected via an internal constructor overload (from DI).
- **Agent identity FIC** (`TokenAcquisition`, blueprint leg 1 + instance leg 2) — resolves the exchange
  **scope** from the agent authority host, replacing the former hardcoded `api://AzureADTokenExchange/.default`.
- **Managed-identity FIC** (`CredentialsProvider`) — resolves the **bare** audience from the request authority
  host, only when the caller did not set `TokenExchangeUrl` explicitly.
- **Managed-identity client assertion** (`ManagedIdentityClientAssertion`) — auto-resolves the audience
  per-request from the **calling confidential client's** authority host against MSAL's baseline
  (`KnownCloudMetadata.Default`), so a single shared instance still emits the correct per-cloud audience; an
  explicit `tokenExchangeUrl` still wins, and the public-cloud audience is the final fallback. (This leg lives
  in the Certificateless assembly, which has no Abstractions dependency, so it consults only the MSAL baseline,
  not a DI provider.)

## Key design decisions

1. **One decision point.** All FIC legs route through `CloudMetadataResolution`, so the audience-vs-scope
   choice and the `/.default` rule are defined exactly once; call sites never hand-build either form.
2. **A single override seam.** The Abstractions `ICloudMetadataProvider` (from DI) is the only override
   surface; the MSAL public baseline is read directly from `KnownCloudMetadata.Default` rather than exposed as
   a second DI seam. An upstream provider already outranks the baseline, so nothing is lost.
3. **No cross-SDK dependency.** MSAL owns public cloud strings; an upstream SDK owns internal-only strings; ID
   Web reads the shared key literal from whichever resolves first. MSAL and the upstream SDK never reference
   each other.
4. **Host-keyed; empty authority is not an error.** Resolution keys off the extracted host; a null/empty
   authority (e.g. the managed-identity leg) resolves to the documented public-cloud default rather than
   throwing, consistent with how ID Web's options can carry an empty instance string.

## New public API surface

Namespace `Microsoft.Identity.Web`:

| API | Kind |
|---|---|
| `CloudMetadataServiceCollectionExtensions.AddCloudMetadata(this IServiceCollection, IConfiguration) : IServiceCollection` | static ext |

This is the **only** new public API. `CloudMetadataResolution` and all consuming changes are `internal`. The
new surface consumes public types already shipped by MSAL (`KnownCloudMetadata`, `CloudMetadataKeyNames`,
`TokenExchangeScope`) and Microsoft.Identity.Abstractions (`ICloudMetadataProvider`,
`InMemoryCloudMetadataProvider`, `CloudMetadataKeyNames`).

## Other changes

| File | Description |
|------|-------------|
| `TokenAcquisition/CloudMetadataResolution.cs` | New internal **static** resolver — the single audience/scope decision point (per-call → upstream provider → MSAL baseline → throw); delegates `/.default` to `TokenExchangeScope.FromAudience` |
| `TokenAcquisition/CloudMetadataServiceCollectionExtensions.cs` | New public `AddCloudMetadata(IConfiguration)` — bind cloud metadata from configuration into an `InMemoryCloudMetadataProvider` |
| `TokenAcquisition/ServiceCollectionExtensions.cs` | Removed the old resolver-singleton factory registration (resolver is now static; providers resolved from DI directly) |
| `TokenAcquisition/TokenAcquisition.cs` | Agent FIC legs 1 & 2 resolve the exchange scope by authority host via the static resolver + a DI `ICloudMetadataProvider` |
| `TokenAcquisition/CredentialsProvider.cs` | Managed-identity FIC leg resolves the bare audience by authority host (DI provider) when none is set explicitly |
| `TokenAcquisition/MicrosoftIdentityHttpClientBuilderExtensions.cs` | Flows the DI `ICloudMetadataProvider` through to the credentials provider |
| `Certificateless/ManagedIdentityClientAssertion.cs` | Per-request audience auto-resolved from the calling client's authority host (MSAL baseline); explicit `tokenExchangeUrl` still wins; public-cloud default fallback |
| `OidcFIC/OidcIdpSignedAssertionProvider.cs` | Resolves the exchange scope via the static resolver (removes the local default + `/.default` literals); holds an injected `ICloudMetadataProvider` |
| `OidcFIC/OidcIdpSignedAssertionLoader.cs` | Passes the DI `ICloudMetadataProvider` into the provider |
| `TokenAcquisition/CloudMetadataResolver.cs` | **Deleted** — replaced by the static `CloudMetadataResolution` |
| `Test/CloudMetadataResolutionTests.cs` | New unit tests for the static resolver (precedence, per-call override, `/.default` idempotency, throw-on-unknown) |
| `Test/ManagedIdentityClientAssertionResolutionTests.cs` | New unit tests for the MI-leg per-request audience resolution |
| `Test/CloudMetadataResolverTests.cs` | **Deleted** — superseded |
| `Test/FederatedIdentityCaeTests.cs` | Cross-cloud FIC exchange tests asserting the credential-exchange `scope` on the wire |

## Test coverage

`CloudMetadataResolutionTests` + `ManagedIdentityClientAssertionResolutionTests` + `FederatedIdentityCaeTests`
(net8.0):
- Precedence/layering: per-call override > upstream `ICloudMetadataProvider` > MSAL baseline
  (`KnownCloudMetadata.Default`); a per-call override wins over provider and baseline.
- Known sovereign hosts resolve from the MSAL baseline; a non-empty host unknown to every source **throws**
  `InvalidOperationException`; an empty authority resolves to the public-cloud default without throwing.
- `TokenExchangeScope.FromAudience` appends `/.default` idempotently; the bare audience form stays bare.
- `AddCloudMetadata(IConfiguration)` registers a resolvable provider from a bound section.
- MI-leg (`ManagedIdentityClientAssertion`) auto-resolves the audience per-request from the calling client's
  authority host; explicit `tokenExchangeUrl` wins; public-cloud default fallback.
- `FederatedIdentityCaeTests` (cross-cloud, pseudo-E2E): build the real OIDC-CC FIC pipeline (ID Web → MSAL)
  over a mocked MSAL HTTP layer and assert the credential-exchange request's `scope` equals the expected
  cloud-specific audience + `/.default` — for public and US Gov, default and custom-override, plus an injected
  `ICloudMetadataProvider` overriding US Gov. The US-Gov-default case also guards the authority-vs-instance
  host-resolution fix.


src/Microsoft.Identity.Web.TokenAcquisition/CloudMetadataResolver.cs
```diff
@@ -0,0 +1,192 @@
...
+                    && !string.IsNullOrEmpty(upstreamAudience))
+                {
+                    return upstreamAudience;
+                }
+            }
+
+            // 3. MSAL public baseline, read directly from KnownCloudConfiguration.Default (same key literal as
+            //    the upstream provider — no remap needed). Reading .Default directly keeps the upstream
+            //    ICloudMetadataProvider as the single override seam rather than exposing a second DI seam.
+            return KnownCloudConfiguration.Default.GetSettingsByAuthorityHost(host).TokenExchangeAudience();
```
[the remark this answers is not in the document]
Avery-Dunn: ResolveAudience  now guards  if (string.IsNullOrEmpty(host)) return null;  before the  KnownCloudConfiguration.Default  lookup, matching the documented fail-soft "null/empty yields null" contract (pairs with the Abstractions short-circuit).

src/Microsoft.Identity.Web.OidcFIC/OidcIdpSignedAssertionLoader.cs
```diff
@@ -196,7 +196,12 @@ public async Task LoadIfNeededAsync(CredentialDescription credentialDescription,
 
                 // Special case for Signed assertions with an FmiPath.
                 // The provider needs to postpone getting the signed assertion until the first call, when ClientAssertionFmiPath will be provided.
-                signedAssertion = new OidcIdpSignedAssertionProvider(_tokenAcquirerFactory, microsoftIdentityApplicationOptions, credentialDescription.TokenExchangeUrl, _logger);
+                signedAssertion = new OidcIdpSignedAssertionProvider(
+                    _tokenAcquirerFactory,
+                    microsoftIdentityApplicationOptions,
+                    credentialDescription.TokenExchangeUrl,
+                    _logger,
+                    CloudMetadataResolver.FromServiceProvider(_serviceProvider));
```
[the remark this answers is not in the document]
Avery-Dunn: The loader now resolves the registered singleton ( _serviceProvider.GetService<CloudMetadataResolver>() ?? CloudMetadataResolver.FromServiceProvider(_serviceProvider) ), restoring the single-decision-point / shared warning-dedupe behavior and avoiding per-call allocations.

tests/Microsoft.Identity.Web.Test/FederatedIdentityCaeTests.cs
```diff
@@ -269,5 +272,207 @@ public async Task Fic_CustomSignedAssertion_ClaimsAndCapabilities_AreSent_OnSeco
...
+        [Fact]
+        public async Task Fic_CustomSignedAssertion_HonorsInjectedCloudMetadataProvider()
+        {
+            using var httpFactoryForTest = new MockHttpClientFactory();
+            // First request (credential exchange) — its "scope" is what we assert on.
+            var credentialRequestHttpHandler = httpFactoryForTest.AddMockHandler(
+                MockHttpCreator.CreateClientCredentialTokenHandler("token-exchange-1"));
+            // Second request (actual token acquisition).
+            var tokenRequestHttpHandler = httpFactoryForTest.AddMockHandler(
+                MockHttpCreator.CreateClientCredentialTokenHandler("final-access-token"));
```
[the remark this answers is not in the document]
Avery-Dunn: Dropped the unused variable (kept the  AddMockHandler(...)  call that queues the second response), so no unused-variable warnaserror risk.

src/Microsoft.Identity.Web.TokenAcquisition/CloudMetadataResolver.cs
```diff
@@ -0,0 +1,199 @@
...
+            }
+
+            // No source recognized the host, so fall back to the documented public-cloud default. If the
+            // caller supplied a (non-public) authority host, surface a one-time warning per host: a sovereign
+            // or private cloud that resolves only to the public default is almost always a missing
+            // ICloudMetadataProvider entry rather than intended behavior.
+            string? host = TryGetHost(authorityOrInstance);
+            if (_logger != null && !string.IsNullOrEmpty(host) && _warnedHosts.TryAdd(host!, 0))
+            {
+                _logger.LogWarning(
```
bgavrilMS, line 110: I think we should throw. ESTS will fail the call eventually, but it'll make everything hard to debug.

src/Microsoft.Identity.Web.TokenAcquisition/ServiceCollectionExtensions.cs
```diff
@@ -52,6 +52,12 @@ public static IServiceCollection AddTokenAcquisition(
                 services.TryAddSingleton<ICredentialsLoader, DefaultCertificateLoader>();
             }
 
+            // The resolver layers an optional upstream ICloudMetadataProvider (registered by a caller or by
+            // MISE — the single override seam) over MSAL's public baseline, which it reads directly from
+            // KnownCloudConfiguration.Default. It is the single audience-vs-scope decision point shared by all
+            // FIC legs. Registered via a factory so it observes whichever ICloudMetadataProvider wins TryAdd.
+            services.TryAddSingleton<CloudMetadataResolver>(sp => CloudMetadataResolver.FromServiceProvider(sp));
```
bgavrilMS, line 59: Why not follow the pattern of if (!HasImplemenetation) ?

src/Microsoft.Identity.Web.TokenAcquisition/CloudMetadataServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,79 @@
...
+    /// token-exchange audience, keyed by authority host) from configuration, so Microsoft.Identity.Web can
+    /// resolve the correct value for sovereign or private clouds it and MSAL do not ship — with no code beyond
+    /// binding an <c>appsettings.json</c> section.
+    /// </summary>
+    public static class CloudMetadataServiceCollectionExtensions
+    {
+        /// <summary>
+        /// Registers an <see cref="ICloudMetadataProvider"/> populated from <paramref name="cloudMetadataSection"/>
+        /// so Microsoft.Identity.Web resolves cloud-specific FIC token-exchange values (for clouds it and MSAL do
+        /// not know about) from configuration. This is the ID Web counterpart to MISE's
```
bgavrilMS, line 24: Don't mention MISE

src/Microsoft.Identity.Web.TokenAcquisition/CloudMetadataResolver.cs
```diff
@@ -0,0 +1,199 @@
...
+
+        /// <summary>
+        /// Resolves the FIC token-exchange <b>scope</b> (bare audience with <c>/.default</c> appended) for the
+        /// cloud that owns <paramref name="authorityOrInstance"/>. Suitable for client-credentials / app-token
+        /// contexts.
+        /// </summary>
+        /// <param name="authorityOrInstance">The request authority or instance (URL or bare host).</param>
+        /// <param name="perCallOverride">An optional explicit override that wins over all resolved sources. It
+        /// may be supplied bare or already suffixed with <c>/.default</c>; the suffix is applied idempotently.</param>
+        internal string ResolveTokenExchangeScope(string? authorityOrInstance, string? perCallOverride = null)
```
bgavrilMS, line 127: What's the difference between scope and audience?

src/Microsoft.Identity.Web.TokenAcquisition/CloudMetadataResolver.cs
```diff
@@ -0,0 +1,199 @@
...
+        {
+            string audience = ResolveTokenExchangeAudience(authorityOrInstance, perCallOverride);
+
+            // Delegate the audience→scope (/.default) computation to MSAL's CloudSettingsExtensions — the
```
bgavrilMS, line 131: THis is somewhat over-engineered. Just make sure the string has "./default" at the end - if possible without parsing.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -154,6 +159,8 @@ public TokenAcquisition(
             }
 
             _credentialsProvider = credentialsProvider;
+            _cloudMetadataResolver = serviceProvider.GetService<CloudMetadataResolver>()
```
bgavrilMS, line 162: `GetService` is a DI resolver. you're supposed to give it an interface and it'll return the object that was registered behind it, possibly by constructing it and its dependencies.

src/Microsoft.Identity.Web.TokenAcquisition/CloudMetadataResolver.cs
```diff
@@ -0,0 +1,199 @@
...
+    /// </para>
+    /// <para>
+    /// The upstream provider and MSAL baseline both expose the audience under the same key literal
+    /// (<see cref="AbstractionsCloudKeys.TokenExchangeAudience"/> equals
+    /// <c>Microsoft.Identity.Client.Instance.Discovery.MsalCloudKeys.TokenExchangeAudience</c>), so ID Web
+    /// "translates" between the two SDKs simply by reading the same literal from whichever source resolves
+    /// first — no runtime key remapping is needed.
+    /// </para>
+    /// </remarks>
+    internal sealed class CloudMetadataResolver
```
bgavrilMS, line 47: Can we not have a resolver and a provider? Can the design be simplified to use just a `ICloudMetadataProvider` and implemenattions? 

Generally the pattern is to use interfaces and implementations. But here you use an actual object, which is tricky to reason about (lifetime etc.). Can we just move the `ResolveAudience` methods somehwere else and rely on ICloudMetadataProvider ?

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
bgavrilMS: It would be better to relly only on serviceProvider.GetService<ICloudSettingsResolver>()

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
```diff
@@ -704,7 +711,10 @@ private async Task<AuthenticationResult> GetAuthenticationResultForUserInternalA
 
             // Leg 2: Get the agent's instance token (T2).
             // The assertion callback handles Leg 1 (blueprint → T1) transparently.
-            var leg2Builder = agentCca.AcquireTokenForClient(s_ficScopes);
+            // Resolve the cloud-specific FIC exchange scope from the agent authority host so sovereign
+            // clouds use the correct audience (public cloud is unchanged via the documented fallback).
+            var leg2Builder = agentCca.AcquireTokenForClient(
```
bgavrilMS, line 716: We should also inject the cloud settings into MSAL? So that MSAL knows about well-known clouds?

bgavrilMS: Maybe this should be done separately.

src/Microsoft.Identity.Web.TokenAcquisition/CloudMetadataResolver.cs
```diff
@@ -0,0 +1,199 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using System;
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using Microsoft.Identity.Abstractions;
+using Microsoft.Identity.Client.Instance.Discovery;
```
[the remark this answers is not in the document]
Avery-Dunn: No extension method now; CloudMetadataResolution calls the static TokenExchangeScope.FromAudience(...) with using Microsoft.Identity.Client.Instance.Discovery;

src/Microsoft.Identity.Web.TokenAcquisition/CloudMetadataResolver.cs
Avery-Dunn: As of the latest commit we now throws `InvalidOperationException` with an informative error message when the cloud is unknown.

src/Microsoft.Identity.Web.TokenAcquisition/ServiceCollectionExtensions.cs
Avery-Dunn: The old unconditional `TryAddSingleton<ICloudConfiguration>(Default)` is gone as of the latest commit. 

ID Web no longer registers a default provider (a null provider results in the MSAL baseline only), and `AddCloudMetadata` uses `TryAddSingleton` to avoid overwriting the caller's custom metadata.

src/Microsoft.Identity.Web.TokenAcquisition/CloudMetadataServiceCollectionExtensions.cs
Avery-Dunn: Fixed in the latest commit.

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
Avery-Dunn: As of the latest commit I removed the injected resolver object: `TokenAcquisition` now pulls the interface `ICloudMetadataProvider` from DI and passes it to the static `CloudMetadataResolution` helper.

src/Microsoft.Identity.Web.TokenAcquisition/CloudMetadataResolver.cs
Avery-Dunn: As of the latest commit resolve logic is in a static `CloudMetadataResolution` helper, and the DI sideis the `ICloudMetadataProvider` interface + implementations (`InMemoryCloudMetadataProvider`, `MiseCloudMetadata`, MSAL `KnownCloudMetadata` baseline).

src/Microsoft.Identity.Web.TokenAcquisition/TokenAcquisition.cs
Avery-Dunn: > Maybe this should be done separately.

Yeah, cross-SDK injection into MSAL is intentionally out of scope here: MISE/caller metadata flows into MSAL via ID Web, which injects metadata into MSAL APIs as needed for a given scenario. In the future, if MSAL has some metadata that it can auto-resolve then we can add an injection mechanism.

src/Microsoft.Identity.Web.TokenAcquisition/CloudMetadataResolver.cs
Avery-Dunn: `TokenExchangeScope.FromAudience` is now the only method for handling this, and just uses `EndsWith("/.default") ? audience : audience + "/.default"`

src/Microsoft.Identity.Web.TokenAcquisition/CloudMetadataResolver.cs
Avery-Dunn: There's not much of a difference: originally this split was done because in three places MSAL's `AcquireTokenForClient` used the endpoint as a scope with the `/.default` suffix, whereas MISE/ID Web managed identity paths treated the `TokenExchangeUrl` as the expected audience and without the `/.default` suffix.

After the other simplifications it became clear that this class was a bit over-engineered though, so in the latest commit this class was renamed (since it was FIC-focused, rather than a generic metadata helper), the `ResolveTokenExchangeScope` was dropped, and the other helper is now used more directly where ID Web actually calls into MSAL with the scope.

tests/Microsoft.Identity.Web.Test/FederatedCredentialAudienceResolverTests.cs
```diff
@@ -0,0 +1,167 @@
...
+                });
+
+            string audience = FederatedCredentialAudienceResolver.ResolveTokenExchangeAudience(
+                UsGovAuthority, perCallOverride: null, provider);
+
+            Assert.Equal("api://AzureADTokenExchangeCustomGov", audience);
+        }
+
+        [Fact]
+        public void ResolveAudience_UpstreamProvider_ResolvesNewCloudUnknownToMsal()
```
bgavrilMS, line 78: Good test. It made me understand the API better than the PR description.

bgavrilMS (approved):

trwalke (approved):

trwalke (approved):

PR #3995, opened 2026-08-05
Fix #3895: BlazorAuthenticationChallengeHandler sends a local returnUrl instead of the absolute NavigationManager.Uri

### LINKED ISSUE #3895: Blazor MapLoginAndLogout: incremental-consent returnUrl (absolute NavigationManager.Uri) coerced to '/' — user loses their page

4.11.0 (also present on `main` at time of writing)

### Where

Blazor Server — incremental / JIT consent via `BlazorAuthenticationChallengeHandler` together with the `MapLoginAndLogout()` Minimal-API endpoints.

### Summary

On the Blazor consent-challenge path the post-consent `returnUrl` is dropped to `/`, so step-up / incremental consent always returns the user to the application root instead of the page they were on.

### Root cause

The two halves of MIW disagree on the shape of the `returnUrl`:

1. `BlazorAuthenticationChallengeHandler.ChallengeUser` builds the challenge URL with
   `returnUrl = Uri.EscapeDataString(navigation.Uri)` — and `NavigationManager.Uri` is an **absolute** URL.
   (`src/Microsoft.Identity.Web/Blazor/BlazorAuthenticationChallengeHandler.cs`, `ChallengeUser`.)
2. The `/authentication/login` endpoint sets
   `RedirectUri = RedirectUriHelper.IsLocalUrl(returnUrl) ? returnUrl : "/"`
   (`src/Microsoft.Identity.Web/Blazor/LoginLogoutEndpointRouteBuilderExtensions.cs`, `GetAuthProperties`),
   and `RedirectUriHelper.IsLocalUrl` returns `false` for any absolute URL
   (`src/Microsoft.Identity.Web/Internal/RedirectUriHelper.cs`).

Net effect: the absolute `returnUrl` produced by `ChallengeUser` never satisfies `IsLocalUrl`, so the endpoint replaces it with `/`.

### Prior art — the MVC path already solves this

`Microsoft.Identity.Web.UI`'s `AccountController.Challenge` accepts same-origin absolute URLs and coerces them to `PathAndQuery`, with a regression test pinned to the exact Blazor `NavigationManager.Uri` case:

- `src/Microsoft.Identity.Web.UI/Areas/MicrosoftIdentity/Controllers/AccountController.cs` — the `Challenge` action's same-origin coercion.
- `tests/Microsoft.Identity.Web.UI.Test/.../AccountControllerTests.cs` — the regression test for an absolute `NavigationManager.Uri` return URL.

The newer Blazor `MapLoginAndLogout` helper did not inherit that coercion. Apps that use the Minimal-API path (and have therefore removed `Microsoft.Identity.Web.UI`) have no path that preserves the originating page.

### Expected behaviour

The Blazor consent challenge should return the user to the page they were on. Either:

- `BlazorAuthenticationChallengeHandler.ChallengeUser` should pass a **local** `returnUrl` (e.g. `new Uri(navigation.Uri).PathAndQuery`), or
- `GetAuthProperties` / `RedirectUriHelper` should accept same-origin absolute return URLs (mirroring `AccountController.Challenge`).

### Repro

1. Blazor Server app using `AddMicrosoftIdentityWebApp(...)` + `app.MapGroup("/authentication").MapLoginAndLogout()`, **without** `Microsoft.Identity.Web.UI`.
2. From a non-root page (e.g. `/admin/things/edit/5`), call a Graph API that requires a scope not yet consented, so `MicrosoftIdentityWebChallengeUserException` is thrown.
3. Handle it via `BlazorAuthenticationChallengeHandler.HandleExceptionAsync(ex)`.
4. Complete consent.
5. **Actual:** the browser lands on `/`. **Expected:** it returns to `/admin/things/edit/5`.

### Workaround

A small component-layer handler that mirrors `ChallengeUser` but passes `new Uri(NavigationManager.Uri).PathAndQuery` (with a defensive `IsLocalUrl` re-check, as `AccountController.Challenge` does) as the `returnUrl`, which `/authentication/login` then honours.

DevonEast:
Fixes #3895.

## Problem

`BlazorAuthenticationChallengeHandler.ChallengeUser` passes `NavigationManager.Uri` — always an absolute URL — as the `returnUrl` query parameter of the `/authentication/login` endpoint mapped by `MapLoginAndLogout()`. That endpoint validates `returnUrl` with `RedirectUriHelper.IsLocalUrl`, which (correctly) rejects absolute URLs and falls back to `/`. Net effect: every incremental-consent or Conditional Access round-trip in Blazor Server drops the user on the app root instead of returning them to the page they were on.

The two halves of the flow ship in the same assembly and are documented as a pair, but they disagree about the shape of `returnUrl`.

## Why fix the sender, not the validator

The strict local-only validation in `GetAuthProperties` is a deliberate hardening: its regression tests pin that absolute URLs must be coerced to `/` (the earlier behaviour of coercing absolutes via `PathAndQuery` at the endpoint was removed as part of the open-redirect hardening). Relaxing the validator would reopen that surface. The sender, by contrast, can trivially produce the local form — it *is* the current page's URI.

This also mirrors what the MVC path already does: `AccountController.Challenge` accepts same-origin absolute URLs by coercing them to `PathAndQuery` (with an `IsLocalUrl` re-check on the coerced value), explicitly because the consent handler passes `NavigationManager.Uri`. This PR gives the Blazor Minimal-API path the equivalent treatment — at the sending end, so the endpoint's validation stays strict.

## Change

`ChallengeUser` now sends `new Uri(navigation.Uri).PathAndQuery`:

- **Preserves the app path base** (`https://host/app/page?x=1` → `/app/page?x=1`) — which is why it uses `PathAndQuery` rather than `NavigationManager.ToBaseRelativePath`.
- **Drops the fragment**, matching the MVC coercion's documented behaviour.
- **Defensive re-check**: `PathAndQuery` of a same-origin URL can still begin with `//` or `/\` (e.g. a request path of `//evil.example/x`), which a downstream `Location` header would treat as protocol-relative. The coerced value is re-validated with `RedirectUriHelper.IsLocalUrl` and falls back to `/` — the same re-check `AccountController.Challenge` performs.

No public API change; no change to `LoginLogoutEndpointRouteBuilderExtensions`.

## Tests

`BlazorAuthenticationChallengeHandlerTests` gains a concrete `TestNavigationManager` (a subclass calling `Initialize()` and overriding `NavigateToCore` — the pattern the ASP.NET Core repo uses), which makes `NavigationManager.Uri`/`NavigateTo` unit-testable. With it:

- **New**: `ChallengeUser_SendsLocalReturnUrl_PreservingPathAndQuery` — returnUrl is the escaped local `/admin/reports?tab=2`, `forceLoad: true`.
- **New**: `ChallengeUser_LocalReturnUrl_PreservesPathBase` — an app under `/app/` keeps its base.
- **New**: `ChallengeUser_ProtocolRelativePathShape_CoercedToRoot` — a `//evil.example/x` path shape coerces to `/`.
- **Un-skipped**: both `HandleExceptionAsync_Detects…` tests previously skipped as "NavigationManager cannot be mocked" now run, asserting navigation occurred with `forceLoad` and the requested scope present.

The existing `GetAuthProperties_CoercesNonLocalReturnUrls` theory is untouched and still pins the endpoint's strict validation — the flow now composes: handler sends local, endpoint accepts local. All 37 tests in the Blazor test namespace pass on net10.0.

## Repro (before) / behaviour (after)

Blazor Server app without `Microsoft.Identity.Web.UI`, `MapLoginAndLogout()` mapped, incremental consent triggered from `/admin/reports?tab=2`:

- **Before**: `returnUrl=https%3A%2F%2Fapp…%2Fadmin%2Freports%3Ftab%3D2` → `IsLocalUrl` false → `RedirectUri = "/"` → user lands on the app root after consent.
- **After**: `returnUrl=%2Fadmin%2Freports%3Ftab%3D2` → `IsLocalUrl` true → user returns to `/admin/reports?tab=2`.

### DIFF
--- a/src/Microsoft.Identity.Web/Blazor/BlazorAuthenticationChallengeHandler.cs
+++ b/src/Microsoft.Identity.Web/Blazor/BlazorAuthenticationChallengeHandler.cs
@@ -72,7 +72,22 @@ public class BlazorAuthenticationChallengeHandler(
     /// </summary>
     public void ChallengeUser(ClaimsPrincipal user, string[]? scopes = null, string? claims = null)
     {
-        var currentUri = navigation.Uri;
+        // NavigationManager.Uri is always absolute, but the /login endpoint mapped by
+        // MapLoginAndLogout validates returnUrl with RedirectUriHelper.IsLocalUrl, which
+        // rejects absolute URLs and falls back to "/" — so passing the absolute URI loses
+        // the user's page after the consent round-trip. Send the app-local PathAndQuery
+        // instead (preserves any path base, drops the fragment — matching the MVC
+        // AccountController.Challenge coercion of same-origin absolute URLs).
+        //
+        // Defensive re-check: PathAndQuery can still begin with "//" or "/\" for a request
+        // path like "//evil.example/x", which a downstream Location header would treat as
+        // protocol-relative. Re-run IsLocalUrl on the coerced value and fall back to "/",
+        // mirroring the endpoint's own validation.
+        var returnUrl = new Uri(navigation.Uri).PathAndQuery;
+        if (!RedirectUriHelper.IsLocalUrl(returnUrl))
+        {
+            returnUrl = "/";
+        }
 
         // Build scopes string (add OIDC scopes)
         var allScopes = (scopes ?? [])
@@ -87,7 +102,7 @@ public class BlazorAuthenticationChallengeHandler(
         var domainHint = Uri.EscapeDataString(GetDomainHint(user));
 
         // Build the challenge URL
-        var challengeUrl = $"/authentication/login?returnUrl={Uri.EscapeDataString(currentUri)}" +
+        var challengeUrl = $"/authentication/login?returnUrl={Uri.EscapeDataString(returnUrl)}" +
                           $"&scope={scopeString}" +
                           $"&loginHint={loginHint}" +
                           $"&domainHint={domainHint}";
--- a/tests/Microsoft.Identity.Web.Test/Blazor/BlazorAuthenticationChallengeHandlerTests.cs
+++ b/tests/Microsoft.Identity.Web.Test/Blazor/BlazorAuthenticationChallengeHandlerTests.cs
@@ -17,6 +17,29 @@ namespace Microsoft.Identity.Web.Test.Blazor
 {
     public class BlazorAuthenticationChallengeHandlerTests
     {
+        /// <summary>
+        /// Concrete <see cref="NavigationManager"/> for unit tests. <c>Uri</c> is driven by
+        /// <see cref="NavigationManager.Initialize(string, string)"/> and navigation is captured
+        /// by overriding <see cref="NavigationManager.NavigateToCore(string, NavigationOptions)"/> —
+        /// the same pattern the ASP.NET Core repo uses to test NavigationManager consumers.
+        /// </summary>
+        private sealed class TestNavigationManager : NavigationManager
+        {
+            public string? LastNavigatedTo { get; private set; }
+            public bool LastForceLoad { get; private set; }
+
+            public TestNavigationManager(string baseUri, string uri)
+            {
+                Initialize(baseUri, uri);
+            }
+
+            protected override void NavigateToCore(string uri, NavigationOptions options)
+            {
+                LastNavigatedTo = uri;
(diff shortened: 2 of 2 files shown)


DevonEast:

bgavrilMS (approved):

iNinja (approved):

PR #3996, opened 2026-08-05
Add tests for bearer-over-mTLS user flows

Robbie-Microsoft:
## Summary

Adds test coverage proving that mTLS bearer transport (`SendCertificateOverMtls`, from [MSAL #6009](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/6009)) works for the **user** flows, not just app flows. idweb already wires this via `CredentialDescription.UseBoundCredential = true` (`WithCertificate(cert, new CertificateOptions { SendCertificateOverMtls = true })`), so **no production code is required** — this PR is test-only.

Earlier revisions also surfaced [MSAL #5947](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/5947) (`GetRefreshToken()` accessor + `CacheOptions.DisableInternalCache`). Per review feedback that has been **removed**: internal token caching should not be disablable, there is no feature request for it, and the SPO scenario that motivated it does not use Microsoft.Identity.Web. That code and its changelog entry are dropped; what remains is test coverage only.

## What's covered

New `MtlsBearerUserFlowTests.cs` proves bound-certificate user flows route the token request over the mTLS endpoint (`mtlsauth.microsoft.com`) carrying `client_assertion`, while unbound controls stay on `login.microsoftonline.com`.

**At the MSAL/CCA boundary** (`MtlsBearerUserFlowTests`, 4):
- On-Behalf-Of (bound)
- refresh_token (bound, via `IByRefreshToken.AcquireTokenByRefreshToken`)
- authorization_code (bound)
- On-Behalf-Of (unbound control)

**Through idweb's public header provider** (`MtlsBearerUserFlowHeaderProviderTests`, 2) — directly answers "it works for `GetHeaderForApp`, but is it done for `GetHeaderForUser`?":
- `IAuthorizationHeaderProvider.CreateAuthorizationHeaderForUserAsync` with a bound certificate routes the OBO request over `mtlsauth.microsoft.com` and returns a `Bearer` header
- unbound control uses the regular `login.microsoftonline.com` endpoint

## Testing

All 6 mTLS user-flow tests pass on net8.0. No production code changes; no changelog entry (test-only).

PR #3997, opened 2026-08-06
DevEx - Add documentation for bearer-over-mTLS user flows

gladjohn:
Document the use of bearer-over-mTLS client authentication for user flows, including configuration examples and flow coverage.

### DIFF
--- a/docs/design/User-flow-bound-credential-mtls-devex.md
+++ b/docs/design/User-flow-bound-credential-mtls-devex.md
@@ -119,7 +119,7 @@ Then `await api.CallApiForUserAsync("GraphAPI")`. No mTLS-specific C#.
    `WithCertificate(cert, new CertificateOptions { SendCertificateOverMtls = true })`
    (`ConfidentialClientApplicationBuilderExtension.WithClientCredentialsAsync`).
 2. Post-#6009, auth-code / OBO / silent route to `mtlsauth.microsoft.com`,
-   present the cert at TLS, and send `client_assertion` (x5c auto-enabled).
+   present the cert at TLS (x5c auto-enabled).
 3. MSAL returns a **bearer** token; IdWeb's default `MsalMtlsHttpClientFactory`
    supplies the mTLS transport — no extra wiring.
 4. `IDownstreamApi` calls downstream with `Authorization: Bearer` as usual.
--- a/docs/design/User-flow-bound-credential-mtls-devex.md
+++ b/docs/design/User-flow-bound-credential-mtls-devex.md
@@ -1,0 +1,145 @@
+# Bearer-over-mTLS client auth for user flows (`UseBoundCredential`)
+
+## Summary
+
+`UseBoundCredential: true` presents the app's certificate at the TLS layer
+(`mtlsauth.microsoft.com`) to authenticate to Entra ID, instead of a
+`client_assertion` JWT in the body. The access token is still a plain **bearer**
+token; only the app→ESTS leg changes. Maps to MSAL
+`CertificateOptions.SendCertificateOverMtls`.
+
+The knob **already ships** (abstractions 12.1.0) and already works for **app
+tokens** — see the merged sample
+[`daemon-app-cert-bound`](../../tests/DevApps/daemon-app/daemon-app-cert-bound).
+This spec **extends the same knob to user flows** (auth code, OBO, silent).
+
+**No new public API.** MSAL [#6009](https://github.com/AzureAD/microsoft-authentication-library-for-dotnet/pull/6009)
+makes the delegated flows honor the flag; IdWeb just needs the MSAL bump +
+tests + docs/samples.
+
+## Not to be confused with `ProtocolScheme = "MTLS_POP"`
+
+| | `UseBoundCredential: true` | `ProtocolScheme: "MTLS_POP"` ([#3832](./msi-fic-pure-mtls-pop-devex.md)) |
+|---|---|---|
+| Set on | the **credential** (`ClientCredentials`) | the **downstream API** options |
+| mTLS applies to | app → ESTS **client auth** | the **downstream API** call |
+| Access token | **bearer** | **sender-constrained** (PoP) |
+| Credential | certificate only | certificate, MI, FIC-with-MI |
+
+Independent; may both be set.
+
+## Flow coverage
+
+| Flow | Entry point | Bearer-over-mTLS |
+|---|---|---|
+| App token (daemon) | `RequestAppToken` / `CallApiForApp` | ✅ shipped |
+| Web-app sign-in | `AddMicrosoftIdentityWebApp` | ✅ via #6009 |
+| On-behalf-of | `EnableTokenAcquisitionToCallDownstreamApi` | ✅ via #6009 |
+| Silent / refresh-token | `AcquireTokenSilent` | ✅ via #6009 |
+
+## Config — web API (OBO)
+
+```json
+{
+  "AzureAd": {
+    "Instance": "https://login.microsoftonline.com/",
+    "TenantId": "<tenant>",
+    "ClientId": "<api-client-id>",
+    "Audience": "api://<api-client-id>",
+    "ClientCredentials": [
+      {
+        "SourceType": "StoreWithThumbprint",
+        "CertificateStorePath": "CurrentUser/My",
+        "CertificateThumbprint": "<thumbprint>",
+        "UseBoundCredential": true
+      }
+    ]
+  },
+  "DownstreamApis": {
+    "GraphAPI": {
+      "BaseUrl": "https://graph.microsoft.com/v1.0",
+      "Scopes": [ "https://graph.microsoft.com/.default" ]
+    }
+  }
+}
+```
+
+## Config — web app (sign-in + call Graph)
+
+```json
+{
+  "AzureAd": {
+    "Instance": "https://login.microsoftonline.com/",
+    "TenantId": "<tenant>",
+    "ClientId": "<client-id>",
+    "CallbackPath": "/signin-oidc",
+    "ClientCredentials": [
+      {
+        "SourceType": "StoreWithThumbprint",
+        "CertificateStorePath": "CurrentUser/My",
+        "CertificateThumbprint": "<thumbprint>",
+        "UseBoundCredential": true
+      }
+    ]
+  },
+  "DownstreamApi": {
+    "BaseUrl": "https://graph.microsoft.com/v1.0",
+    "Scopes": [ "user.read" ]
+  }
+}
+```
+
+Only delta vs. a plain-bearer config: the `"UseBoundCredential": true` line, and
+a **certificate** source (not `ClientSecret` — bearer-over-mTLS is cert-only).
+`DownstreamApi(s)` is unchanged — no `ProtocolScheme`, no extra headers.
+
+## Wiring (unchanged from any cert app)
+
+```csharp
+// Web app
+builder.Services
+    .AddMicrosoftIdentityWebAppAuthentication(builder.Configuration)
+    .EnableTokenAcquisitionToCallDownstreamApi()
+        .AddDownstreamApi("DownstreamApi", builder.Configuration.GetSection("DownstreamApi"))
+        .AddInMemoryTokenCaches();
+
+// Web API (OBO)
+builder.Services
+    .AddMicrosoftIdentityWebApiAuthentication(builder.Configuration)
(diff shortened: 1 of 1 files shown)


docs/design/User-flow-bound-credential-mtls-devex.md
```diff
@@ -0,0 +1,145 @@
+# Bearer-over-mTLS client auth for user flows (`UseBoundCredential`)
+
+## Summary
+
+`UseBoundCredential: true` presents the app's certificate at the TLS layer
+(`mtlsauth.microsoft.com`) to authenticate to Entra ID, instead of a
+`client_assertion` JWT in the body. The access token is still a plain **bearer**
```
Robbie-Microsoft, line 7: Nit: "instead of a `client_assertion` JWT" contradicts step 2 of *How it works*, which says `client_assertion` is still sent (x5c auto-enabled). Step 2 is correct — mTLS changes the channel, not the assertion. Verified on the wire in #3996: bound user-flow requests hit `mtlsauth.microsoft.com` **and** carry `client_assertion`.

```suggestion
(`mtlsauth.microsoft.com`) to authenticate to Entra ID. The `client_assertion`
JWT is still sent. The access token is still a plain **bearer**
```

gladjohn: modified

Robbie-Microsoft: This got fixed in the wrong direction — step 2 lost `client_assertion` but the Summary still says "instead of". The assertion is still sent; mTLS changes the channel, not the credential material. Wire dump from a bound OBO call on 4.87.0:

```
endpoint  = https://mtlsauth.microsoft.com/{tenant}/oauth2/v2.0/token
POST keys = assertion, client_assertion, client_assertion_type, client_id,
            client_info, grant_type, requested_token_use, scope
client_assertion_type = urn:ietf:params:oauth:client-assertion-type:jwt-bearer
client_assertion      = 2122 chars
```

Also makes "x5c auto-enabled" on line 122 dangling — x5c is a header of the assertion that line no longer mentions (and is why it's 2122 chars). Suggest restoring it there.

```suggestion
(`mtlsauth.microsoft.com`) to authenticate to Entra ID. The `client_assertion`
JWT is still sent. The access token is still a plain **bearer**
```

docs/design/User-flow-bound-credential-mtls-devex.md
```diff
@@ -0,0 +1,145 @@
...
+tests + docs/samples.
+
+## Not to be confused with `ProtocolScheme = "MTLS_POP"`
+
+| | `UseBoundCredential: true` | `ProtocolScheme: "MTLS_POP"` ([#3832](./msi-fic-pure-mtls-pop-devex.md)) |
+|---|---|---|
+| Set on | the **credential** (`ClientCredentials`) | the **downstream API** options |
+| mTLS applies to | app → ESTS **client auth** | the **downstream API** call |
+| Access token | **bearer** | **sender-constrained** (PoP) |
+| Credential | certificate only | certificate, MI, FIC-with-MI |
```
Robbie-Microsoft, line 27: Not cert-only. `ConfidentialClientApplicationBuilderExtension` honors the flag for `CredentialType.SignedAssertion` too (→ `WithBoundClientAssertion`, throws `MissingTokenBindingCertificate` if the provider can't bind). Only `Secret` ignores it. As written this turns away the FIC audience.

```suggestion
| Credential | certificate, or signed assertion (FIC / MI / OIDC IdP) | certificate, MI, FIC-with-MI |
```

Line 138 needs the same correction.

docs/design/User-flow-bound-credential-mtls-devex.md
```diff
@@ -0,0 +1,145 @@
...
+## How it works
+
+1. `UseBoundCredential: true` → IdWeb builds the CCA with
+   `WithCertificate(cert, new CertificateOptions { SendCertificateOverMtls = true })`
+   (`ConfidentialClientApplicationBuilderExtension.WithClientCredentialsAsync`).
+2. Post-#6009, auth-code / OBO / silent route to `mtlsauth.microsoft.com`,
+   present the cert at TLS (x5c auto-enabled).
+3. MSAL returns a **bearer** token; IdWeb's default `MsalMtlsHttpClientFactory`
+   supplies the mTLS transport — no extra wiring.
+4. `IDownstreamApi` calls downstream with `Authorization: Bearer` as usual.
```
Robbie-Microsoft, line 125: Missing the highest-value devex piece: how a dev confirms it's actually on. One line does it —

```suggestion
4. `IDownstreamApi` calls downstream with `Authorization: Bearer` as usual.

## Verifying it works

`AuthenticationResultMetadata.TokenEndpoint` contains `mtlsauth.` when the
credential is bound, and `login.microsoftonline.com` when it isn't. That is the
assertion the tests in #3996 make.
```

docs/design/User-flow-bound-credential-mtls-devex.md
```diff
@@ -0,0 +1,145 @@
...
+  },
+  "DownstreamApi": {
+    "BaseUrl": "https://graph.microsoft.com/v1.0",
+    "Scopes": [ "user.read" ]
+  }
+}
+```
+
+Only delta vs. a plain-bearer config: the `"UseBoundCredential": true` line, and
+a **certificate** source (not `ClientSecret` — bearer-over-mTLS is cert-only).
```
Robbie-Microsoft, line 93: Worth stating that this fails *silently*: `case CredentialType.Secret` just calls `WithClientSecret` and drops the flag — no log, no throw. Dev sets `UseBoundCredential: true` on a secret, sees no error, assumes mTLS. Should we log a warning in idweb?

docs/design/User-flow-bound-credential-mtls-devex.md
```diff
@@ -0,0 +1,145 @@
...
+        .AddInMemoryTokenCaches();
+```
+
+Then `await api.CallApiForUserAsync("GraphAPI")`. No mTLS-specific C#.
+
+## How it works
+
+1. `UseBoundCredential: true` → IdWeb builds the CCA with
+   `WithCertificate(cert, new CertificateOptions { SendCertificateOverMtls = true })`
+   (`ConfidentialClientApplicationBuilderExtension.WithClientCredentialsAsync`).
```
Robbie-Microsoft, line 120: Worth a sentence here: `GetApplicationKey` appends `+bound` per credential, so a bound config never reuses the unbound CCA/cache entry. That's the first question anyone flipping this flag on an existing deployment will ask.

docs/design/User-flow-bound-credential-mtls-devex.md
```diff
@@ -0,0 +1,145 @@
...
+* Delegated-flow tests (auth-code, OBO, silent).
+* Two samples: `web-app-bound-credential`, `web-api-obo-bound-credential`.
+* Docs: cert-credentials page + a note in `token-binding.md` contrasting with `MTLS_POP`.
+
+## Prerequisites
+
+* MSAL with #6009 (user-flow bearer-over-mTLS).
+* Abstractions ≥ 12.1.0 (already referenced — `UseBoundCredential` shipped in [PR #255](https://github.com/AzureAD/microsoft-identity-abstractions-for-dotnet/pull/255)).
+* A **certificate** credential (other credential types ignore the flag).
+* Cert registered on the app; tenant/app allow-listed for mTLS client auth (else `AADSTS700027`).
```
Robbie-Microsoft, line 139: `mtlsauth.microsoft.com` is commercial-only. Add Gov/China endpoints (or state explicitly they're out of scope) — the `MTLS_POP` devex has that table and readers will expect parity.

docs/design/User-flow-bound-credential-mtls-devex.md
```diff
@@ -0,0 +1,145 @@
...
+Independent; may both be set.
+
+## Flow coverage
+
+| Flow | Entry point | Bearer-over-mTLS |
+|---|---|---|
+| App token (daemon) | `RequestAppToken` / `CallApiForApp` | ✅ shipped |
+| Web-app sign-in | `AddMicrosoftIdentityWebApp` | ✅ via #6009 |
+| On-behalf-of | `EnableTokenAcquisitionToCallDownstreamApi` | ✅ via #6009 |
+| Silent / refresh-token | `AcquireTokenSilent` | ✅ via #6009 |
```
Robbie-Microsoft: Entry-point column is MSAL-level here — `AcquireTokenSilent` isn't something the IdWeb dev calls. The IdWeb-level equivalents are `IAuthorizationHeaderProvider.CreateAuthorizationHeaderForUserAsync` and `IDownstreamApi.CallApiForUserAsync`.

Rows 36-37 have the inverse problem: they name registration methods, not call sites. Worth making the column consistently one layer.

```suggestion
| Silent / refresh-token | `IAuthorizationHeaderProvider.CreateAuthorizationHeaderForUserAsync` | ✅ via #6009 |
```

docs/design/User-flow-bound-credential-mtls-devex.md
```diff
@@ -0,0 +1,145 @@
...
+   present the cert at TLS (x5c auto-enabled).
+3. MSAL returns a **bearer** token; IdWeb's default `MsalMtlsHttpClientFactory`
+   supplies the mTLS transport — no extra wiring.
+4. `IDownstreamApi` calls downstream with `Authorization: Bearer` as usual.
+
+## Work items
+
+* Bump `Microsoft.Identity.Client` to the build containing #6009.
+* Delegated-flow tests (auth-code, OBO, silent).
+* Two samples: `web-app-bound-credential`, `web-api-obo-bound-credential`.
```
Robbie-Microsoft, line 131: Stale: the MSAL bump is done (4.87.0 already referenced) and delegated-flow tests landed in #3996 — worth linking so readers see current state.

bgavrilMS (approved):

Robbie-Microsoft (changes requested):
above comments

gladjohn:
Thanks for the review folks - forgot I had this open. I will address these comments and respond back

docs/design/User-flow-bound-credential-mtls-devex.md
gladjohn: Thanks. But, I don’t think we should add this here. Endpoint selection and `AuthenticationResultMetadata.TokenEndpoint` are handled by MSAL internally and are not part of the IdWeb developer contract. This document should focus on the IdWeb configuration and the bearer token returned to the downstream API.

docs/design/User-flow-bound-credential-mtls-devex.md
gladjohn: Thanks. I agree the user-visible behavior should be clear, but the internal credential dispatch and logging behavior are outside this document’s scope. I’ll state that `UseBoundCredential` has no effect on client-secret credentials. Any warning or fail-fast behavior should be tracked as a separate implementation change.

docs/design/User-flow-bound-credential-mtls-devex.md
gladjohn: Thanks. CCA cache-key construction and the `+bound` discriminator are internal IdWeb implementation details, not a supported developer contract. We should validate bound/unbound isolation in tests, but I don’t think the internal cache-key format belongs in this developer-experience document.

docs/design/User-flow-bound-credential-mtls-devex.md
gladjohn: Fixed in `1e914a5`. I removed the incorrect claim that mTLS replaces the client assertion and rewrote the summary at the IdWeb contract level. I also removed the MSAL transport/assertion implementation details from **How it works**.

docs/design/User-flow-bound-credential-mtls-devex.md
gladjohn: Fixed in `1e914a5`. The column is now **IdWeb usage**, the app and OBO rows use `IDownstreamApi` / `IAuthorizationHeaderProvider`, and silent refresh is described as behavior handled internally rather than as a developer call to `AcquireTokenSilent`.

docs/design/User-flow-bound-credential-mtls-devex.md
gladjohn: Fixed in `1e914a5`. The section now records MSAL 4.87.0 and the authorization-code, OBO, and direct refresh-token tests from #3996 as complete. Actual silent-acquisition refresh coverage, samples, and public documentation remain outstanding.

PR #4004, opened 2026-08-24
Make managed identity key attestation optional

gladjohn:
## Problem

`Microsoft.Identity.Web.Certificateless` currently references `Microsoft.Identity.Client.KeyAttestation`. That dependency transitively brings the native-only `Microsoft.Azure.Security.KeyGuardAttestation` package into applications that never request key attestation, including consumers that arrive through `Microsoft.Identity.Web.TokenAcquisition` and `Microsoft.Identity.Web.Azure`.

NuGet dependencies cannot be conditioned on whether an API is used, so keeping the reference in `Certificateless` means every consumer receives the native DLL and PDB. It also creates compatibility problems for .NET Framework / `packages.config` consumers. See AzureAD/microsoft-authentication-library-for-dotnet#6133.

## Change

This PR separates managed identity mTLS PoP from optional Credential Guard key attestation:

- Keeps `Microsoft.Identity.Web.Certificateless` independent through the `IManagedIdentityAttestationProvider` extension point.
- Adds the optional `Microsoft.Identity.Web.KeyAttestation` package as the only IdWeb package that references `Microsoft.Identity.Client.KeyAttestation`.
- Adds `AddMicrosoftIdentityWebKeyAttestation()` for explicit opt-in.
- Uses the provider consistently for both managed identity mTLS PoP paths:
  - direct managed identity token acquisition;
  - `SignedAssertionFromManagedIdentity` / FIC token acquisition.
- Keeps the custom credential-loader override so the FIC path receives the provider without coupling `Certificateless` to the optional package.
- Updates the existing attested FIC E2E test to explicitly register the optional package.
- Updates packaging, API surface, changelog, and mTLS PoP documentation.

## Behavior

| Configuration | Managed identity mTLS PoP behavior |
| --- | --- |
| Default | `WithMtlsProofOfPossession()` |
| Install `Microsoft.Identity.Web.KeyAttestation` and call `AddMicrosoftIdentityWebKeyAttestation()` | `WithMtlsProofOfPossession()` + `WithAttestationSupport()` |

Only applications that opt in receive the native KeyGuard attestation assets. The unattested managed identity mTLS PoP flow remains supported by MSAL.

For `TokenAcquirerFactory` applications, opt-in is explicit before `Build()`:

```csharp
var tokenAcquirerFactory = TokenAcquirerFactory.GetDefaultInstance();
tokenAcquirerFactory.Services.AddMicrosoftIdentityWebKeyAttestation();

```

## Review focus

- `IManagedIdentityAttestationProvider` is the dependency-inversion seam; the core packages do not reference the optional implementation package.
- `KeyAttestedManagedIdentityCredentialLoader` intentionally overrides the built-in `SignedAssertionFromManagedIdentity` loader through the existing custom-loader mechanism.
- Absence of the provider intentionally selects the unattested flow and logs that choice once per component instance.
- No backward-compatibility shim, loader redesign, or broader API redesign is included.

## Validation

- Built the attested `TokenAcquirerTests` E2E project for `net8.0`: 0 warnings, 0 errors.
- Focused `KeyAttestationTests`: 2 passed.
- Previously validated the modified credential/attestation test set: 44 passed.
- Packed and inspected the affected NuGet dependency groups:
  - core `Certificateless`, `TokenAcquisition`, and `Azure` packages do not depend on KeyAttestation;
  - only `Microsoft.Identity.Web.KeyAttestation` depends on `Microsoft.Identity.Client.KeyAttestation`.

### DIFF
--- a/src/Microsoft.Identity.Web.Certificateless/IManagedIdentityAttestationProvider.cs
+++ b/src/Microsoft.Identity.Web.Certificateless/IManagedIdentityAttestationProvider.cs
@@ -1,0 +1,21 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+using Microsoft.Identity.Client;
+
+namespace Microsoft.Identity.Web
+{
+    /// <summary>
+    /// Enables optional key attestation on a managed identity mTLS proof-of-possession request.
+    /// </summary>
+    public interface IManagedIdentityAttestationProvider
+    {
+        /// <summary>
+        /// Configures key attestation on the managed identity request.
+        /// </summary>
+        /// <param name="builder">The managed identity request builder.</param>
+        /// <returns>The configured request builder.</returns>
+        AcquireTokenForManagedIdentityParameterBuilder EnableAttestation(
+            AcquireTokenForManagedIdentityParameterBuilder builder);
+    }
+}
--- a/Microsoft.Identity.Web.linux.slnf
+++ b/Microsoft.Identity.Web.linux.slnf
@@ -4,6 +4,7 @@
     "projects": [
       "src/Microsoft.Identity.Web.Diagnostics/Microsoft.Identity.Web.Diagnostics.csproj",
       "src/Microsoft.Identity.Web.Certificateless/Microsoft.Identity.Web.Certificateless.csproj",
+      "src/Microsoft.Identity.Web.KeyAttestation/Microsoft.Identity.Web.KeyAttestation.csproj",
       "src/Microsoft.Identity.Web.Certificate/Microsoft.Identity.Web.Certificate.csproj",
       "src/Microsoft.Identity.Web.TokenCache/Microsoft.Identity.Web.TokenCache.csproj",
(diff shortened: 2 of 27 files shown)


src/Microsoft.Identity.Web.Certificateless/IManagedIdentityAttestationProvider.cs
```diff
@@ -0,0 +1,21 @@
...
+// Licensed under the MIT License.
+
+using Microsoft.Identity.Client;
+
+namespace Microsoft.Identity.Web
+{
+    /// <summary>
+    /// Enables optional key attestation on a managed identity mTLS proof-of-possession request.
+    /// </summary>
+    public interface IManagedIdentityAttestationProvider
```
bgavrilMS: nit: Attestation is not related to MSI. Maybe this should be called `IHardwareKeyAttestationProvider` or `IKeyAttestationProvider`

gladjohn: agreed - Renamed to IKeyAttestationProvider. Also, renamed internal implementation accordingly

4gust (approved):

bgavrilMS (approved):

PR #4008, opened 2026-08-31
Add inbound PoP (SHR) token validation to the Entra sidecar

soodt:
# Add inbound PoP (SHR) token validation to the Entra sidecar

## Description

Adds inbound **Signed HTTP Request (SHR) proof-of-possession** validation to the `Microsoft.Identity.Web.Sidecar`. The `/Validate` endpoint now accepts either a `Bearer` token (unchanged) or, for **app-only (client-credentials)** tokens, the `PoP` scheme, returning `protocol: "PoP"` with the embedded token's claims.

## Scope
App-only tokens only — no OBO/actor, no user/delegated, no PFT/CDT-over-PoP, no mTLS-PoP.
**Timestamp-only** freshness; **no server nonce**.

## Tests
- Unit: tamper / expiry / bad-signature / wrong-issuer / wrong-audience / missing-`cnf`
  negatives, plus default options-mapping.
- Round-trip: mock-IdP JWKS end-to-end, and real-Entra E2E pipeline tests.

## Docs & samples
Sidecar `README`, `appsettings.json` sample block, `.http` sample, and PoP examples added
to the Python and TypeScript dev-adapter sample clients.


src/Microsoft.Identity.Web.Sidecar/Configuration/PopValidationOptions.cs
```diff
@@ -0,0 +1,99 @@
...
+/// </para>
+/// </summary>
+public class PopValidationOptions
+{
+    /// <summary>
+    /// Gets or sets a value indicating whether the <c>ts</c> (timestamp) claim is validated - the
+    /// replay/freshness guard. Default <c>true</c>; disabling it removes replay protection.
+    /// </summary>
+    [DefaultValue(true)]
+    public bool ValidateTs { get; set; } = true;
```
bgavrilMS: Is there a business reason to have this configurable? If not, I would not expose this. It needs to stay `true`

src/Microsoft.Identity.Web.Sidecar/Configuration/PopValidationOptions.cs
```diff
@@ -0,0 +1,99 @@
...
+    [DefaultValue(false)]
+    public bool ValidateQ { get; set; }
+
+    /// <summary>
+    /// Gets or sets a value indicating whether the <c>h</c> (headers) claim is validated - request
+    /// binding. Default <c>false</c>; enabling it hardens binding but the caller must sign <c>h</c> or
+    /// every request fails.
+    /// </summary>
+    [DefaultValue(false)]
+    public bool ValidateH { get; set; }
```
bgavrilMS: Headres are not sent to the container. So ValidateH dones't actually work

src/Microsoft.Identity.Web.Sidecar/appsettings.json
```diff
@@ -38,6 +38,19 @@
...
+    //    // Inbound Signed HTTP Request (PoP) validation for app-only tokens.
+    //    // All flags are optional; shown here with their secure defaults.
+    //    "PopValidation": {
+    //        "ValidateTs": true,
+    //        "ValidateM": true,
+    //        "ValidateU": true,
+    //        "ValidateP": true,
+    //        "ValidateQ": false,
+    //        "ValidateH": false,
+    //        "AcceptUnsignedHeaders": true,
```
bgavrilMS: As stated above, validating headers is not implemented. We should implement minimal options here - my recommendation is to remove validateH and AcceptUnsignedHeaders as options (but internally keep those default values - validateH - false, AcceptUnsignedHeaders - true).

bgavrilMS (approved):
Approved with comments.

src/Microsoft.Identity.Web.Sidecar/Pop/PopHttpRequestFactory.cs
```diff
@@ -0,0 +1,54 @@
...
+            return false;
+        }
+
+        if (!Uri.TryCreate(uri, UriKind.Absolute, out Uri? parsedUri))
+        {
+            error = $"The '{PopConstants.OriginalUriHeaderName}' header must be an absolute URI.";
+            return false;
+        }
+
+        requestData = new HttpRequestData
```
iNinja, line 46: `ValidateH` is operator-configurable, but this request data contains no headers. Enabling it therefore validates the SHR's `h` claim against an empty collection and rejects correctly signed headers. Please either populate headers through a defined trusted forwarding contract and add round-trip tests, or remove `ValidateH` and `AcceptUnsignedHeaders` from the supported configuration surface.

src/Microsoft.Identity.Web.Sidecar/Pop/ShrPopValidationService.cs
```diff
@@ -0,0 +1,161 @@
...
+        TokenValidationParameters accessTokenValidationParameters = bearerTvp.Clone();
+
+        // Ensure the cloned TVP can resolve the embedded token's signing keys via live JWKS. The stock
+        // JwtBearer post-configure copies ConfigurationManager onto the shared TVP lazily inside its
+        // OnMessageReceived event, which only fires when the Bearer scheme authenticates first. Copying
+        // it here removes that ordering dependency; if it were ever null, inner-token validation fails closed.
+        accessTokenValidationParameters.ConfigurationManager ??=
+            bearerOptions.ConfigurationManager as BaseConfigurationManager;
+
+        var validationContext = new SignedHttpRequestValidationContext(
```
iNinja: `CreateValidationParameters` runs before the guarded `try`, and `SignedHttpRequestLifetime` throws for negative values. A bad configuration can therefore let the sidecar start but make every PoP request return 500. Please validate and bound these options at startup, and keep parameter construction inside the controlled failure path as defence in depth.

src/Microsoft.Identity.Web.Sidecar/Endpoints/ValidateRequestEndpoints.cs
```diff
@@ -29,6 +30,15 @@ private static Results<Ok<ValidateAuthorizationHeaderResult>, ProblemHttpResult>
         HttpContext httpContext,
         [FromServices] IConfiguration configuration)
     {
+        // A validated inbound SHR PoP request: the PoP handler has already validated the outer signature
+        // and the embedded app-only access token (delegated/user tokens are rejected there) and stashed
+        // it here. Scope enforcement (AzureAd:Scopes -> VerifyUserHasAnyAcceptedScope) targets delegated
+        // 'scp' claims and so does not apply to app-only PoP; the app-only guard above is the equivalent gate.
+        if (httpContext.Items[PopConstants.ValidatedAccessTokenItemKey] is JsonWebToken popToken)
```
iNinja, line 41: The PoP branch returns before `AzureAd:Scopes` is evaluated, while the endpoint documentation says `/Validate` enforces configured scopes. Since this PoP path is restricted to app-only tokens and scopes apply to delegated tokens, please document that distinction and clarify that authorization of the returned app identity remains the caller's responsibility. The nearby comment should also avoid describing app-only classification as an equivalent authorization gate.

src/Microsoft.Identity.Web.Sidecar/Configuration/PopValidationOptions.cs
soodt: Fixed. Removed ValidateH from the config surface and hardcoded it to false

src/Microsoft.Identity.Web.Sidecar/appsettings.json
soodt: Fixed

src/Microsoft.Identity.Web.Sidecar/Configuration/PopValidationOptions.cs
soodt: Agreed.  Removed ValidateTs it's now hardcoded true

src/Microsoft.Identity.Web.Sidecar/Pop/PopHttpRequestFactory.cs
soodt: Fixed. Removed ValidateH from the config surface and hardcoded it to false

src/Microsoft.Identity.Web.Sidecar/Pop/ShrPopValidationService.cs
soodt: Both done. Added  IValidateOptions<SidecarOptions> with  ValidateOnStart() bounding  SignedHttpRequestLifetime > 0 (fails fast at boot) and moved parameter/context construction inside the guarded try  so bad config yields a 401, not a 500.

src/Microsoft.Identity.Web.Sidecar/Endpoints/ValidateRequestEndpoints.cs
soodt: Reworded the comment. Let me know if it looks alright.

iNinja (approved):

PR #4014, opened 2026-08-31
Isolate configured token cache providers

iNinja:
## Summary

- Use a distinct cache provider for each configured token-cache registration.
- Preserve existing sharing for parameterless in-memory token-cache registration.

## Release notes

Configured token-cache registrations no longer share an implicitly cached provider.
Parameterless in-memory token-cache registration continues to share its provider.

## Testing

- Focused tests passed on .NET 8 and .NET Framework 4.7.2.

### DIFF
--- a/src/Microsoft.Identity.Web.TokenCache/TokenCacheExtensions.cs
+++ b/src/Microsoft.Identity.Web.TokenCache/TokenCacheExtensions.cs
@@ -61,8 +61,10 @@ namespace Microsoft.Identity.Web
         /// </code>
         /// If using distributed token caches, use AddDistributedTokenCache.
         /// </example>
-        /// <remarks>Don't use this method in ASP.NET Core. Just add use the ConfigureServices method
-        /// instead.</remarks>
+        /// <remarks>
+        /// Don't use this method in ASP.NET Core. Use the ConfigureServices method instead.
+        /// Each invocation creates an isolated token cache provider.
+        /// </remarks>
         internal static IConfidentialClientApplication AddTokenCaches(
             this IConfidentialClientApplication confidentialClientApp,
             Action<IServiceCollection> initializeCaches)
@@ -90,8 +92,10 @@ namespace Microsoft.Identity.Web
         /// </code>
         ///
         /// </example>
-        /// <remarks>Don't use this method in ASP.NET Core. Just add use the ConfigureServices method
-        /// instead.</remarks>
+        /// <remarks>
+        /// Don't use this method in ASP.NET Core. Use the ConfigureServices method instead.
+        /// The token cache provider and its in-memory cache are shared process-wide across calls.
+        /// </remarks>
         public static IConfidentialClientApplication AddInMemoryTokenCache(
             this IConfidentialClientApplication confidentialClientApp)
         {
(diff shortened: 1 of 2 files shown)


4gust (approved):

bgavrilMS (approved):

bgavrilMS (approved):

bgavrilMS (commented):
Still loking at this.

bgavrilMS (changes requested):
As per my convo with GPT 6 Astra, this breaks the following scneario

```csharp
async Task<string> GetTokenAsync()
{
    var app = ConfidentialClientApplicationBuilder
        .Create(clientId)
        .WithAuthority(authority)
        .WithClientSecret(clientSecret)
        .Build();

    app.AddInMemoryTokenCache(services =>
    {
        services.Configure<MemoryCacheOptions>(options =>
        {
            options.SizeLimit = 100 * 1024 * 1024;
        });
    });

    var result = await app
        .AcquireTokenForClient(new[] { "https://graph.microsoft.com/.default" })
        .ExecuteAsync();

    return result.AccessToken;
}

await GetTokenAsync();
await GetTokenAsync();  // expected: from cache actual: from STS
```

bgavrilMS:
The configuration bug is real, but this change also removes cross-CCA cache reuse for the configured standalone extensions. This is more than provider isolation.

For example, an application can repeatedly build a CCA and attach the same configured cache:

```csharp
var app = ConfidentialClientApplicationBuilder
    .Create(clientId)
    .WithAuthority(authority)
    .WithClientSecret(clientSecret)
    .Build();

app.AddInMemoryTokenCache(services =>
{
    services.Configure<MemoryCacheOptions>(options =>
    {
        options.SizeLimit = 100 * 1024 * 1024;
    });
});

var result = await app.AcquireTokenForClient(scopes).ExecuteAsync();
```

Before this change, successive CCAs using that overload reused the backing memory cache. After this change, each registration builds a new service provider and therefore a new memory cache, even with identical configuration. Repeating this pattern loses cross-CCA token reuse and causes additional token-endpoint requests.

For `AddDistributedTokenCache(configure)`, each new adapter constructs its own L1 `MemoryCache`. A shared Redis backend still allows L2 token reuse, but repeated CCA construction now produces repeated cold-L1 reads and independent memory budgets. With `AddDistributedMemoryCache()`, the default L2 storage is also separate.

Scope clarification: normal Id.Web DI registrations and internally created CCAs remain shared, as does parameterless `AddInMemoryTokenCache()`. The regression is in repeated use of the configured standalone extensions.

The previous tests exercised distributed L1 and L2 reuse across recreated CCAs; replacing those checks with isolation checks removes coverage of the behavior we need to preserve.

Please preserve reuse for repeated registrations of the same logical cache while fixing configuration isolation. An explicit cache identity or reusable registration could distinguish genuinely different caches without making CCA creation the cache-lifetime boundary. Arbitrary `Action<IServiceCollection>` callbacks cannot reliably be compared for configuration equivalence, so this needs an explicit ownership/configuration contract rather than method-based deduplication or unconditional isolation.

Regression coverage should include two CCAs with identical client/authority/scopes and the same logical cache: a second-call cache hit for configured in-memory caching, an L1 hit for distributed caching with L1 enabled, and an L2 hit with L1 disabled. Separate logical caches should independently honor their configurations/backends.

PR #4015, opened 2026-08-31
Use app acquisition for EasyAuth app authentication results

iNinja:
## Summary

- Use the existing MSAL client credentials flow for both EasyAuth app-token methods.
- Return a genuine app-only authentication result instead of the provider token associated with the signed-in user.
- Keep EasyAuth user provider tokens opaque and preserve the existing user-token behavior.

## Release notes

EasyAuth `GetAuthenticationResultForAppAsync` now performs client-credential acquisition and returns an app-only authentication result instead of the provider token associated with the signed-in user. Applications using this method must have valid application credentials and the required application permissions.

EasyAuth user-token behavior is unchanged. User-token methods continue to return the provider token supplied by App Services authentication for the current request.

## Testing

- 6 focused `AppServicesAuthenticationTokenAcquisitionTests` passed on .NET 8.


tests/Microsoft.Identity.Web.Test/AppServicesAuthenticationTokenAcquisitionTests.cs
```diff
@@ -0,0 +1,710 @@
...
+            AppServicesAuthenticationTokenAcquisition acquisition = CreateAcquisition(
+                CreateContext(accessToken),
+                httpClientFactory);
+
+            // Act / Assert
+            await Assert.ThrowsAsync<NotSupportedException>(() =>
+                acquisition.GetAccessTokenForUserAsync(
+                    new[] { "User.Read" },
+                    authenticationScheme: null));
+        }
```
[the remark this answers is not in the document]
iNinja: Added a mismatch test where the token contains `Mail.Read` and the request asks for `User.Read`, covering the valid-but-mismatched branch.

4gust (approved):

src/Microsoft.Identity.Web/AppServicesAuth/AppServicesAuthenticationTokenAcquisition.cs
```diff
@@ -4,12 +4,15 @@
 using System;
```
bgavrilMS, line 4: This PR proposes significant changes. Were these manually tested ? I am not aware of having automated EasyAuth tests (we should!).

src/Microsoft.Identity.Web/AppServicesAuth/AppServicesAuthenticationTokenAcquisition.cs
```diff
@@ -247,14 +250,412 @@ public Task<AuthenticationResult> GetAuthenticationResultForAppAsync(
...
+
+            return directTenant ?? optionsTenant;
+        }
+
+        private static IReadOnlyCollection<string> ValidateUserToken(
+            string accessToken,
+            IReadOnlyCollection<string> requestedScopes,
+            string? explicitTenant)
+        {
+            using JsonDocument payload = ReadTokenPayload(accessToken);
```
bgavrilMS: Clients can't parse tokens. They could be JWE.

bgavrilMS (changes requested):
Clients MUST not parse tokens.

src/Microsoft.Identity.Web/AppServicesAuth/AppServicesAuthenticationTokenAcquisition.cs
iNinja: Reworked this to focus on a narrow set of changes, removing the token inspection.

4gust:
I tested the manually and the fix works.

bgavrilMS (approved):

PR #4023, opened 2026-09-02
Sidecar: validate Host header outside Development

soodt:
# Sidecar: validate Host header outside Development

## Description

Adds ASP.NET Core Host filtering for `Microsoft.Identity.Web.Sidecar` outside the
Development environment

## What

- Registers `AddHostFiltering` only when the environment is not Development.
- Applies Host filtering in the request pipeline (scoped with `UseWhen`) rather than globally, so the `/healthz` endpoint stays exempt.
- In Development, neither check is applied, so the inner-loop experience is unchanged.

### DIFF
--- a/src/Microsoft.Identity.Web.Sidecar/Program.cs
+++ b/src/Microsoft.Identity.Web.Sidecar/Program.cs
@@ -71,40 +71,12 @@ public class Program
             options.AddOperationTransformer(new OptionsOverrideOperationTransformer());
         });
 
-        if (!builder.Environment.IsDevelopment())
-        {
-            // Outside development, accept requests only for local host names.
-            builder.Services.AddHostFiltering(options =>
-            {
-                options.AllowedHosts = ["localhost", "127.0.0.1", "[::1]"];
-            });
-
-            // Remove the framework's global Host filtering so it can be applied in the
-            // pipeline below instead, scoped to exempt the health endpoint. The filter is
-            // matched by its framework-internal type name; if that changes, the health test fails.
-            ServiceDescriptor? hostFilteringStartupFilter = builder.Services.FirstOrDefault(
-                static descriptor =>
-                    descriptor.ServiceType == typeof(IStartupFilter) &&
-                    descriptor.ImplementationType?.Name == "HostFilteringStartupFilter");
-
-            if (hostFilteringStartupFilter is not null)
-            {
-                builder.Services.Remove(hostFilteringStartupFilter);
-            }
-        }
-
         var app = builder.Build();
 
         if (!app.Environment.IsDevelopment())
         {
-            // Validate the Host header outside development, but keep the health
-            // endpoint reachable for probes that target the pod's routable address.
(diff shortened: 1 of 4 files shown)


src/Microsoft.Identity.Web.Sidecar/Program.cs
```diff
@@ -71,10 +71,39 @@ public static void Main(string[] args)
...
+            }
+        }
+
         var app = builder.Build();
 
         if (!app.Environment.IsDevelopment())
         {
+            // Validate the Host header outside development, but keep the health
+            // endpoint reachable for probes that target the pod's routable address.
+            app.UseWhen(
```
iNinja: The exemption uses exact path equality, but endpoint routing also accepts `/healthz/` for the mapped health endpoint. With the current predicate, `/healthz/` still passes through Host Filtering and returns `400` when an orchestrator sends the pod or container address as the Host.`n`nCould we make the exemption cover the same path variants as the health endpoint, without exempting arbitrary subpaths, and add a test for `/healthz/` with an unexpected Host?

src/Microsoft.Identity.Web.Sidecar/Program.cs
```diff
@@ -71,10 +71,39 @@ public static void Main(string[] args)
...
 
+        if (!builder.Environment.IsDevelopment())
+        {
+            // Outside development, accept requests only for local host names.
+            builder.Services.AddHostFiltering(options =>
+            {
+                options.AllowedHosts = ["localhost", "127.0.0.1", "[::1]"];
+            });
+
+            // Remove the framework's global Host filtering so it can be applied in the
```
iNinja: Could we avoid depending on the framework-internal `HostFilteringStartupFilter` name and registration shape here?`n`nThe Host check and `LocalCallerRestriction` implement one Sidecar-specific boundary: outside Development, non-health requests must come from loopback and use a loopback Host. I think it would be clearer to enforce both in `LocalCallerRestriction`, while keeping the health exemption in one place. The Host check can use the public `HostString.MatchesAny` helper with `localhost`, `127.0.0.1`, and `[::1]`, preserving the framework's port, casing, and IPv6 matching behavior. Invalid callers can continue returning `403`, while invalid Hosts return `400`.`n`nThat would remove `AddHostFiltering`, the internal service-descriptor lookup, and the conditional `UseHostFiltering` branch, leaving a small fixed policy that is directly covered by the Sidecar tests.

src/Microsoft.Identity.Web.Sidecar/Program.cs
soodt: Fixed.

src/Microsoft.Identity.Web.Sidecar/Program.cs
soodt: Fixed.

iNinja (approved):

iarekk (approved):

